PR #108, opened 2026-07-15
revert: re-land the Genes, Genomes, and Human Genetics expansion as separate PRs

ewen-poch:
Revert the combined connector expansion (#101) so the Genes & Ontologies, Genomes, and Human Genetics enhancements can re-land as one PR per connector. Same code returns immediately in the follow-up PRs.


ewen-poch:
Abandoning this revert — keeping the combined #101 as merged; future connector work will be split into separate PRs.

PR #120, opened 2026-07-15
feat(ketcher): interactive Ketcher Chemistry sketcher app + molecule viewer

ewen-poch:
Adds **Ketcher Chemistry** — an interactive 2D molecule sketcher exposed as a Featured connector app, plus a read-only viewer for chemical file artifacts.

## What's included
- **Read-only viewer** for `.ket/.mol/.sdf/.smi/.rxn` artifacts (lazy-loaded Ketcher/Indigo, mirrors the PdbPreview/3Dmol pattern).
- **Interactive sketcher tile** in the preview workbench (`toolKind: 'ketcher'`), modeled on the live NotebookPreview tile — registers an imperative handle on mount, applies commands, throttles edits.
- **4 agent tools** (connector `ketcher`): `open_sketcher` (creates a `.ket` artifact, mounts+activates a tile, returns `artifact_id`), `set_structure`, `highlight_atoms`, `get_structure`.
- **Main↔renderer bridge**: `KetcherBroker` (modeled on `ApprovalBroker`) + `ketcher:command/reply/mounted/unmounted` IPC. Tools are hosted in the main process via `ConnectorService.callBundled`'s special-case (same enabled/blocked/approval gate as every connector; only dispatch is diverted to `KetcherService`).
- **Persistence**: tile edits round-trip back to the `.ket` artifact (throttled), resilient to the pending→message finalize move.
- **Settings**: Ketcher appears as a Featured connector with per-tool permissions + skip-approvals, surfaced automatically through the shared registry/catalog path.

## Dependencies
- Adds `ketcher-react`/`-core`/`-standalone@3.17.0` (React 19-compatible; bundled WASM Indigo, lazy-loaded into separate chunks).

## Testing
- ✅ `npm run typecheck`, `npm run lint`, `npx electron-vite build` (renderer bundles the Ketcher WASM/worker chunks) all pass.
- ✅ Full `npx vitest run` green (1555 passed / 43 skipped), incl. `KetcherBroker`, `KetcherService`, and `ConnectorService` routing unit tests.
- ⚠️ **Needs a `npm run dev` GUI hand-verify before merge**: the live round-trip (open_sketcher → tile mount → set/highlight/get → reply, and Ketcher WASM runtime load) and the highlight rendering path are unit-tested against mocks only. `open_sketcher` resolves the active turn's artifact run, so it's an in-agent-turn tool (not a terminal-cell call).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -326,7 +326,6 @@
       "version": "7.29.7",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.7",
         "@babel/generator": "^7.29.7",
@@ -832,7 +831,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
@@ -877,7 +875,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       }
@@ -1446,6 +1443,7 @@
       "dev": true,
       "license": "BSD-2-Clause",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "cross-dirname": "^0.1.0",
         "debug": "^4.3.4",
@@ -1465,6 +1463,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "graceful-fs": "^4.2.0",
         "jsonfile": "^6.0.1",
@@ -1479,6 +1478,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "universalify": "^2.0.0"
       },
@@ -1491,6 +1491,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "engines": {
         "node": ">= 10.0.0"
       }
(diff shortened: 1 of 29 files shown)


ewen-poch:
## ⚠️ DO NOT MERGE — blocked by a CSP / Web Worker issue found during `npm run dev` verification

Live dev-verify surfaced two renderer-runtime issues the build/CI can't catch (they only fail at runtime when the Ketcher chunk loads):

1. **Fixed (`a638974`):** Ketcher's `util`/`assert` deps reference the Node `process` global **unguarded at chunk load**; the sandboxed renderer has no `process`. Added a minimal `process`/`global` shim (`src/renderer/src/lib/node-globals-shim.ts`, imported first in `main.tsx`). Verified the bare `process.*` refs **survive into the production bundle** (`index.modern-*` = Ketcher), so this is needed for both dev and packaged builds — not a dev-only artifact.

2. **BLOCKER (unresolved):** `ketcher-standalone` spawns its Indigo backend as a **blob-URL Web Worker**, which the renderer CSP blocks:
   > Refused to create a worker from 'blob:…' — violates `script-src 'self'` (worker-src not set, falls back to script-src). → `Failed to fetch dynamically imported module … ketcher-react.js`

   The CSP lives in `src/renderer/index.html`:
   `default-src 'self'; script-src 'self'; …` (no `worker-src`).

   Unblocking requires a **security policy decision**: add `worker-src 'self' blob:` (and possibly `child-src blob:`) to the CSP — i.e. allow blob-URL workers app-wide. Both the read-only viewer and the interactive tile depend on the Ketcher chunk loading, so **both are blocked** until this is resolved and the live sketcher round-trip (open_sketcher → set/highlight/get) is hand-verified.

Everything else is ready: full app built (viewer + tile + broker + tools + settings), CI green (`npm ci`, lint, typecheck, coverage, `electron-vite build` incl. WASM bundling), unit tests green. Holding as **draft** until the CSP call is made.

ewen-poch:
Superseded and closing. During `npm run dev` verification, `ketcher-react`/`ketcher-standalone` proved too heavy for this Electron + Vite renderer: it needs (1) a CSP `worker-src blob:` exception for the Indigo WASM Web Worker, (2) a 4 GB build heap to bundle that WASM, and (3) a React-instance dedupe fight that still threw "Invalid hook call" from its prebundled `index.modern` chunk. That is a full molecule *editor*; our users only need to view structures that analyses produce. Replacing with a lightweight, read-only OpenChemLib SVG viewer (pure JS — no worker, no CSP change, no heap bump, no React conflict) in a fresh PR.

PR #133, opened 2026-07-15
fix(acp): drop the manual resume gate; interrupted sessions auto-resume on send

ewen-poch:
## Problem

Clicking **Resume** on an interrupted chat session could appear to do nothing — the button would spin on "Resuming…" (or fail quietly), leaving the conversation stuck. The manual Resume flow was a second, more fragile reattach path layered on top of one that already worked.

## Root cause

Reattaching an interrupted session is inherently reliable: on the agent side it is a local read of the persisted transcript (`getOrCreateSession`), not something that should hang. More importantly, the **send path already auto-resumes**: `sendWorkspaceMessage` detects a session that isn't attached and transparently calls `resumeSession(...)` before sending. The composer was never gated on `interrupted` either, so a user could already just type and continue.

The `SessionInterruptedBanner` + manual Resume button was therefore redundant — and the only path that could hang. Guarding it with a timeout would have been a symptom fix; the right change is to remove the gate.

## Change

- Delete `SessionInterruptedBanner` and its manual Resume path (button, `isResuming`, `onResumeSession`).
- Suppress the `INTERRUPTED_SESSION_ERROR` restore sentinel in the composer error region, so a freshly restored session looks clean (no banner, no red error box) and is immediately typeable. A real, non-sentinel error still surfaces.
- Clear the `interrupted` flag in `appendUserMessage` so a session that receives a new message is no longer flagged.
- Remove the now-dead `resumeInterruptedSession` / `resumeInterruptedWorkspaceSession` / `markResumed`.

The main-process `session/resume` path is unchanged, and no timeout is introduced.

## Behavior after

An interrupted session looks ready. Typing and sending a message transparently re-attaches the runtime via the existing send-path auto-resume, streams the reply, and clears the interrupted state — no separate button, no dead end.

## Testing

- `npm run typecheck` — passes (node + web).
- `npx vitest run` over the workspace, store, and acp renderer suites — 416 passing.
- `npx eslint --cache .` — clean.
- Manually verified in `npm run dev` with a crafted interrupted session: no banner, composer usable, send auto-resumes.

Net **+22 / −323** across 9 files.


ewen-poch:
Superseded by #140. That approach removed the manual Resume gate in favor of auto-resume-on-send; we instead kept and hardened Resume to reconnect and continue on an abnormal disconnect. Closing.

PR #204, opened 2026-07-19
feat(settings): always show agent-card uninstall with an inline explainer

ewen-poch:
## Summary

In Settings → Model → **Agent framework**, both runtime cards (`ClaudeStatusCard`, `OpencodeStatusCard`) previously showed the **Uninstall** button only for an app-managed install, so a runtime found on PATH / installed via npm gave the user no signal about why it couldn't be removed from the app, or how to remove it. Now every card with a detected runtime shows the Uninstall button. It's actionable only for a non-active, app-managed install; otherwise it's greyed out with a trailing `?` icon **inside** the button whose tooltip explains why:

- **Not app-managed** — the app doesn't own this binary; how to remove it manually (`npm uninstall -g …`, your package manager, or by deleting it from your PATH), then re-detect.
- **Active framework** — it's the active agent; switch to another framework first, then uninstall.

## Implementation

- New shared `RuntimeUninstallControl` replaces the near-duplicate uninstall block in both cards, plus a pure `uninstallDisabledHint()` helper (`runtime-uninstall-hint.ts`) so the exact English copy and its branching are unit-tested directly.
- The greyed-with-reason button is `aria-disabled` (not natively `disabled`, which swallows hover events and would keep the tooltip from ever opening) with its click neutralized, so the button itself stays hoverable as the tooltip trigger. A transient busy state (uninstalling/detecting) takes priority and uses the native `disabled` attribute with no `?`. This `aria-disabled`-for-hoverable-tooltip exception is documented in `docs/design.md` (Focus / Disabled).
- Onboarding omits `onUninstall`, so it stays button-free; a not-detected card still shows the install picker.
- No change to `SettingsPage` wiring or the existing `UninstallRuntimeDialog` confirmation flow.

## Testing

- `npm run typecheck`, `npm run lint`, `npm run test` (2957 passed / 43 skipped), and `prettier --check` all green.
- New `RuntimeUninstallControl.render.test.tsx` covers the enabled path, both greyed reasons (aria-disabled + inline `?`), the transient-busy path, and busy-takes-priority-over-a-standing-reason; `uninstallDisabledHint` copy is asserted directly. Updated the two card tests and `SettingsPage.render.test.tsx`'s active-card assertion.
- Not covered automatically: the live Radix tooltip won't open under jsdom's synthetic events, so hovering the greyed button (`npm run dev`) to confirm the copy renders is a manual check.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -328,6 +328,7 @@ Workspace-only tokens without a shadcn counterpart, plus shadow tokens. For shar
 - All focusable controls use `focus-visible:outline-none focus-visible:ring-[3px] focus-visible:ring-ring/50`.
 - Inputs may add `focus-visible:border-ring/50`; the light focus border target is `rgb(134 182 239)`.
 - Disabled controls use `disabled:pointer-events-none disabled:opacity-50`; shared buttons also use `touch-manipulation`.
+- **Exception — a disabled control that must explain _why_:** a natively `disabled` element receives no pointer events, so a `Tooltip` on it never opens. When the disabled state needs an on-hover explanation (e.g. the agent-framework Uninstall button when the runtime is not app-managed or is the active backend), render it as `aria-disabled` + `opacity-50 cursor-not-allowed` with a neutralized `onClick` instead of the native `disabled` attribute, so it keeps the greyed look yet stays hoverable as the tooltip trigger. Reserve this for standing (non-transient) reasons; transient busy states (in-flight install/detect) still use native `disabled` with no tooltip.
 - Sidebar buttons and icon triggers may use `cursor-pointer`; decorative row wrappers provide hover/active styling only and must not imply a click target outside the nested button.
 - Hover, focus, and active states must not change width, height, padding, or border width.
 
--- a/src/renderer/src/pages/settings/ClaudeStatusCard.render.test.tsx
+++ b/src/renderer/src/pages/settings/ClaudeStatusCard.render.test.tsx
@@ -75,10 +75,11 @@ describe('ClaudeStatusCard surface', () => {
       )
     })
 
-    // The button now always shows for a detected runtime, but a non-managed install can't be removed.
-    expect(findUninstallButton()?.disabled).toBe(true)
-    const help = container.querySelector('button[aria-label^="Why can\'t"]')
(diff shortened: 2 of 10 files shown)


ewen-poch:
Addressed the review findings in 42c7bf4:

- **[P2] Busy state now has priority.** `disabled={!hint && busy}` → the busy state (`isDetecting`/`isUninstalling`) is computed first and suppresses the standing hint, so during detect/uninstall the button is natively `disabled` with no `?`, even for a non-managed or active runtime. This matches the PR description and avoids showing a stale explanation mid-operation. Added a test locking in the priority (`managed: false` + `isDetecting`).

- **[P3] Documented the `aria-disabled` exception.** Added an explicit exception to the *Focus / Disabled* convention in `docs/design.md`: a disabled control that must explain *why* on hover uses `aria-disabled` + `opacity-50 cursor-not-allowed` with a neutralized click (kept hoverable so the `Tooltip` can open), reserved for standing reasons; transient busy states still use native `disabled`. So the tests now assert a documented behavior, not an off-spec one.

- **[P2] Spec updated.** `docs/internal/2026-07-18-uninstall-managed-framework-design.md` gets a "Revisions after PR #204 review" section: the Uninstall button is now always shown for a *detected* runtime (greyed + `?` for non-managed / active, actionable only when app-managed and inactive), superseding the earlier "shown only when app-managed / PATH-npm never offered uninstall" wording. (That doc lives in the gitignored `docs/internal/`, so it isn't part of this PR's diff.)

Full suite green (2957 passed / 43 skipped), lint/typecheck/format clean.

ewen-poch:
[P3] Fixed in 7cd7824. `isInstalling` (the global install flag) is now threaded through both status cards into `RuntimeUninstallControl` and folded into its busy state (`isUninstalling || isDetecting || isInstalling`). During an install the Uninstall button is natively `disabled` with no `?`, for both cards — matching `docs/design.md` (Focus / Disabled) and the existing card-selection lock. This also closes a latent bug where a managed-inactive Uninstall stayed clickable mid-install. Chose to wire install-busy through (rather than narrow the docs) since locking uninstall during an install is the consistent, safer behavior. Added a test; full suite green (2958 passed / 43 skipped).

PR #205, opened 2026-07-19
feat(web): add localhost web UI and headless backend mode

imjszhang:
## Summary

Adds an optional localhost web interface and headless backend mode so Open Science can be used from a browser while the Electron main process continues running agent sessions, tray, and Notebook processes — even after the desktop window is closed.

This merges the originally planned embedded web server + RPC bridge and headless mode work into a single PR.

## Motivation

- Allow browser access to the same agent sessions and project workspace as the desktop app
- Keep the agent runtime alive when the UI is closed (headless backend)
- Minimize coupling with the fast-moving upstream codebase

### Low-coupling approach

Instead of duplicating business logic or adding web-specific IPC handlers, this PR introduces a **generic RPC bridge**:

1. **`rpc-capture`** wraps `ipcMain.handle` and records registered channels
2. **HTTP `/rpc/:channel`** invokes existing handlers with a synthetic `event.sender`
3. **WebSocket `/events`** mirrors `webContents.send` broadcasts to browser clients
4. **`bootstrap.ts`** builds a `window.api` shim from a generated preload → channel map

Upstream IPC modules are largely untouched. Existing handlers are reused as-is.

### Event mirroring

Seven broadcast call sites (acp, settings, storage, reviewer, update, ipc) now use a shared `broadcastToRenderers()` helper. WebSocket clients register as optional sinks — no per-channel web wiring required.

### Independent web build

- `vite.web.config.ts` builds `src/renderer/web/` → `out/web/`
- Reuses the same React renderer (`src/renderer/src/`) via aliases
- `scripts/generate-web-api-map.mjs` parses `src/preload/index.ts` and generates `api-map.generated.ts`
- `npm run check:web-api-map` guards against preload/web drift in CI

### Headless mode

- `--headless` skips the initial Electron window (`createInitialWindow: false`)
- Tray, agent runtime, and optional web service remain available
- Explicit quit still runs the normal shutdown coordinator

## Security

- Binds to **`127.0.0.1` only**
- **Token auth** (Bearer header, query param, or HttpOnly cookie) persisted under config root
- **Host** and **Origin** validation rejects non-loopback requests
- No settings UI toggle in this PR (zero coupling with upstream settings) — enabled via CLI/env only

## Enabling

```bash
npm run build:web
npm run dev:web          # desktop window + web UI (port 44100)
npm run dev:headless     # backend + web UI, no window
```

Environment variable alternative: `OPEN_SCIENCE_WEB_PORT=44100`

Electron CLI flags also supported when passed through electron-vite: `--serve`, `--serve=<port>`, `--headless`.

## Notable fixes discovered during testing

- **Tailwind CSS scanning**: web build roots at `src/renderer/web`, so Tailwind v4 missed component classes. Fixed with `@source '../'` in `main.css` so utility generation is independent of Vite root.

## Test plan

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] Vitest: web-service (auth, http-server, rpc-capture), renderer-broadcast, app-lifecycle, api-map (22 tests)
- [x] Manual: `npm run dev:web` → open authenticated URL in browser
- [x] Manual: UI renders correctly (layout, settings panel, RPC calls)
- [x] Manual: WebSocket events (settings open/close)
- [x] Manual: unauthenticated requests return 401
- [ ] Manual: `npm run dev:headless` → web UI works without Electron window (recommended before merge)

## Follow-ups (out of scope)

- Settings UI toggle for web service
- Remote / LAN access with stronger auth
- Dedicated `start:web` for packaged builds


ewen-poch:
## Review summary and follow-up plan

Overall this is high-quality work and the security model is sound: binds to `127.0.0.1` only, timing-safe token comparison, token file at `0o600`, Host + Origin validation (DNS-rebinding protection), a `SameSite=Strict; HttpOnly` cookie, a 302 that strips the token from the URL after login, path-traversal guarding on the static root, a CSP, and the api-map drift-guard test. Below are the items I plan to address.

### Scope: this PR stays loopback-only
The web service currently listens on `127.0.0.1` only, and the auth layer's `isLoopbackHost(host)` rejects non-loopback requests, so it cannot be reached from another machine — consistent with "Remote / LAN access" being listed as an out-of-scope follow-up.

Worth flagging: this app runs an agent that executes code/tools (e.g. `acp:send-prompt` is effectively arbitrary code execution). Exposing the port on a LAN would be equivalent to handing whoever can reach it code-execution on the host, and since it is plain HTTP the token would travel in cleartext. So **LAN/remote sharing should be designed as a separate feature** (TLS, real authentication rather than a single shared token, a danger-confirmation / read-only mode, and a prominent warning) and not opened up in this PR.

### Planned fixes
1. **P2 — tighten the `/rpc` surface (lightweight)**: `/rpc/:channel` currently accepts any registered `ipcMain.handle` channel, which lets a caller bypass the channels `bootstrap.ts` deliberately reimplements client-side — `file:save-blob` / `file:save-managed` / `window:close`. Those handlers require a real Electron `WebContents` (with the synthetic RPC sender, `file:save-*` pop a native save dialog on the **desktop** and `window:close` becomes a silent no-op). I'll reject these three channels at the web boundary with a **403** and add a test.

   Given the loopback-only scope, this uses a targeted **denylist** rather than a full allowlist: an allowlist needs an extra generated artifact + generator changes + a drift guard, is over-engineered for the current risk (token + loopback already puts the caller at user/renderer trust level), and is easier to over-restrict.

2. **P3 — add tests for the dev-web launcher**: `scripts/dev-web.cjs` is the only place that injects the default web port (44100) and forwards `--headless` to electron-vite, yet it only has option-parsing / lifecycle-injection coverage. I'll extract the argv/env assembly into a pure function and add unit tests: default-port injection / respecting an existing `OPEN_SCIENCE_WEB_PORT` / `--headless` forwarded as `-- --headless` / no `--` when not headless.

### Optional (suggested)
- Add a dedicated `npm run check:web-api-map` step to CI: today the drift guard only runs (hard-gated) via vitest on macOS (the Linux/Windows test leg is advisory). The check is pure Node and cross-platform, so a dedicated step would catch preload/web drift on all three OSes.

### Other observations (non-blocking, FYI)
- `parseWebModeOptions` enables the service merely because `OPEN_SCIENCE_WEB_PORT` is present in the environment, including in packaged builds; consider having packaged builds require an explicit `--serve`/`--headless` so a stale env var can't silently open a listener.
- A transient WebSocket drop triggers `releaseClient` → the sender is destroyed; on the 1s auto-reconnect the client gets a new sender id, so state cached by `sender.id` (e.g. managed-preview resources) is released in between. A client that only makes RPC calls and never opens a WebSocket also leaks a sender.
- The outer `catch` in `http-server.ts` calls `json()` again after `streamPreview`/`serveStatic` may have already written headers, which can throw `ERR_HTTP_HEADERS_SENT`; consider checking `response.headersSent` before responding in the catch.

PR #206, opened 2026-07-19
feat(cli): manage the headless backend as a daemon

imjszhang:
## Summary
- Add \open-science start/stop/status/url\ CLI commands for browser-only daemon usage
- Persist daemon PID/port state and expose a token-authenticated graceful shutdown endpoint
- Support repository builds, standard installed-app locations, and explicit executable overrides
- Headless tray menu: Open Web UI / Copy URL / Quit (Windows: right-click menu, double-click opens web)
- Rename internal headless flag to \--open-science-headless\ to avoid Chromium making native tray menus invisible on Windows (electron/electron#48982)

## Test plan
- [x] \
pm run build\
- [x] \
pm run lint\
- [x] \
pm run typecheck\
- [x] focused Vitest suite (tray + web-service + CLI)
- [x] CLI E2E: start, authenticated health check, status, URL, duplicate start, graceful stop
- [x] state file removed and port released after stop
- [x] Windows tray: right-click shows menu, double-click opens web, Copy URL and Quit work

## Notes
Builds on the merged localhost web UI (#205). The full repository suite was also run: 2,939 tests passed, with 17 pre-existing Windows/timing-sensitive failures unrelated to this change.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,19 +272,19 @@ npm run dev
 
 Useful commands:
 
-| Command               | Purpose                              |
-| --------------------- | ------------------------------------ |
-| `npm run dev`         | Start the development application    |
-| `npm run dev:web`     | Dev app + localhost web UI (127.0.0.1) |
-| `npm run dev:headless`| Dev backend + web UI, no Electron window |
-| `npm run lint`        | Run ESLint                           |
-| `npm run typecheck`   | Type-check main and renderer code    |
-| `npm test`            | Run the Vitest suite                 |
-| `npm run build`       | Type-check and build the application |
-| `npm run build:web`   | Build the optional localhost web UI  |
-| `npm run build:mac`   | Package macOS builds                 |
-| `npm run build:win`   | Package Windows builds               |
-| `npm run build:linux` | Package Linux builds                 |
+| Command                | Purpose                                  |
+| ---------------------- | ---------------------------------------- |
+| `npm run dev`          | Start the development application        |
+| `npm run dev:web`      | Dev app + localhost web UI (127.0.0.1)   |
+| `npm run dev:headless` | Dev backend + web UI, no Electron window |
+| `npm run lint`         | Run ESLint                               |
+| `npm run typecheck`    | Type-check main and renderer code        |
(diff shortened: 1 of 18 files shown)


ewen-poch:
Really nice work — the browser-only daemon flow (start/stop/status/url + headless tray) is a great idea and fills a real gap. Auth on the shutdown endpoint is solid (token-gated before all routes, SameSite=Strict + HttpOnly cookie), the `--open-science-headless` rename is applied consistently, and the CLI/app config-root parity checks out.

We're going to prepare to merge this as-is and track the polish in a separate follow-up PR. Items for that follow-up:

1. **POSIX force-kill (main one).** In `cli/index.mjs` `stopCommand`, after graceful shutdown times out the Windows branch correctly uses `taskkill /T /F`, but the POSIX branch sends `process.kill(state.pid, 'SIGTERM')` — the same graceful signal the process already ignored, to the leader PID only. Since the daemon is spawned `detached: true` (its own process group), this should be `process.kill(-state.pid, 'SIGKILL')` to force-kill the whole group; otherwise agent/Notebook children can be orphaned. It also `removeStateFiles` + prints "stopped" unconditionally, without re-verifying the PID actually died — so it can report success on a still-running process.

2. **Windows tray single-click.** The new `if (process.platform === 'win32')` branch in `tray.ts` also applies to the non-headless desktop app, so Windows single-click no longer shows the window (only double-click does). If that regression to the installed app isn't intended, gate it to the headless path.

3. **CLI test coverage.** `cli/locate-app.mjs` and the `cli/index.mjs` command lifecycle (start/stop/status/url, `isProcessAlive`, `healthCheck`) are currently untested — only `parseCliArgs` is covered.

4. **Minor.** macOS `defaultInstalledCandidates` only checks `/Applications`, not `~/Applications`; duplicated `STATE_FILE`/`TOKEN_FILE` constants across `cli/config-root.mjs` and `web-service/state-file.ts` can drift; `parseCliArgs` flag→camelCase via chained `.replace()` is fragile for future hyphenated flags.

PR #236, opened 2026-07-21
fix(managed-codex): fall back to copy+delete when rename throws EPERM or EXDEV on Windows

ewen-poch:
## Summary

- On Windows, `fs.rename()` can fail with `EPERM` when antivirus software (e.g. Windows Defender) holds a file lock, or `EXDEV` on a cross-device move (defensive — the same-volume layout makes EXDEV unreachable in practice). This applies both to moving the staged scratch dir (`.codex-install-<id>/runtime`) into the final destination (`codex-managed`) and to moving an existing install aside as a backup on upgrade/reinstall.
- Both configured registries (`registry.npmjs.org` and `registry.npmmirror.com`) were catching this error as a registry failure, leaving the user unable to install the managed Codex runtime.
- `replaceDirectory` now catches `EPERM`/`EXDEV` on both renames (destination→backup and staged→destination) and falls back to `fs.cp()` + `fs.rm()`. If the copy itself fails, the backup is restored; if the restore also fails, the backup path is logged and included in the thrown error so the user can recover manually. Backup/restore failures always surface the real cause (not the original EPERM), and the step-1 backup fallback logs its backup path when it fails partway.
- Once the install reaches `replaceDirectory`, a failure is a deterministic local filesystem error (the staged bits are byte-identical across registries thanks to pinned versions + integrities), so it is returned immediately instead of being retried against the next registry. Download/verify errors still fall through to the next registry as before.
- `uninstallManagedCodex` also removes orphaned backups left behind by failed installs — matched strictly as directories named `codex-managed.backup-<uuid>` so look-alike entries sharing the prefix are never touched.

## Changes

- `src/main/settings/managed-codex.ts`: import `cp`/`readdir`; add copy+delete fallback for `EPERM`/`EXDEV` on both renames in `replaceDirectory`; shared `restoreBackupOrThrow` helper logs and annotates errors with the backup path when a restore fails; step-1 fallback failures throw the real cause with the backup path attached; local replace failures terminate the registry loop; uninstall cleans orphaned backup dirs (UUID-shaped names only, directories only)
- `src/main/settings/managed-codex.test.ts`: `vi.mock('node:fs/promises', ...)` with per-call fault flags (staged move EPERM/EIO, destination backup, restore, cp); new regression tests for the first-install fallback, the upgrade-path fallback, cp failure with failed/successful restore, step-1 fallback failure, non-EPERM staged failure with failed restore, local replace failure skipping the second registry, restore failure preserving the backup-path error and orphan, and uninstall orphan cleanup (UUID dir removed, non-UUID dir and UUID-named file preserved)

## Test plan

- [x] `npm run typecheck` — passes
- [x] `npm run lint` — passes
- [x] `npm run test` (managed-codex suite, 29 tests) — all pass
- [x] `npm run format` — passes
- [ ] Manual smoke on Windows: install managed Codex while Windows Defender is active — previously failed with `EPERM: operation not permitted, rename`; now completes successfully

### DIFF
--- a/src/main/settings/managed-codex.test.ts
+++ b/src/main/settings/managed-codex.test.ts
@@ -1,5 +1,5 @@
 import { createHash } from 'node:crypto'
-import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
+import { mkdir, mkdtemp, readFile, readdir, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { dirname, join, sep } from 'node:path'
 import { Readable } from 'node:stream'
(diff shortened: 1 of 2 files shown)


ewen-poch:
## Review feedback addressed (commit 88a5b53)

**P1 — first `rename` also needs EPERM fallback**
The `destination→backup` rename now has the same `EPERM`/`EXDEV` copy-fallback as the `staged→destination` move, so upgrades/reinstalls on Windows where the existing `codex-managed` dir contains files locked by antivirus also succeed.

**P2 — restore-failure is no longer silent**
When the copy-based fallback fails *and* the backup restore rename also fails, `log.error` is called with the full backup path and the rethrown error message now includes `(backup retained at <path>)`, giving the user a manual recovery entry point.

**P3 — EXDEV defensive comment added**
A comment now explains that the `EXDEV` case is guarded defensively — the same-volume layout (`mkdtemp` under `dataRoot`) makes it unreachable in practice.

**Tests — three new regression cases**
| Test | What it covers |
|------|---------------|
| `staged→destination rename throws EPERM (first install)` | existing case, updated to new fault-flag API |
| `destination→backup rename throws EPERM (upgrade path)` | P1 path: existing `codex-managed` dir present, backup rename fails, cp fallback used |
| `cp fails during EPERM fallback` | P2 path: copy fails + restore rename fails → `log.error` called, error message includes backup path |

All 23 tests pass; typecheck, lint, format clean.

PR #262, opened 2026-07-21
fix(artifacts): resolve artifact write paths against the notebook data dir

ewen-poch:
## Problem

`write_artifact_file` failed for files generated in a notebook, then a shell `cp` fallback failed the same way (`No such file or directory`).

Root cause is path resolution, not the plot save. Inside a notebook the kernel cwd is the session **data** dir (`<sessionRoot>/data`), but the env also exposes the session root (without `/data`). Agents routinely rebuild the absolute path from the session root, so the file — actually saved under `data/` — is passed with the `/data` segment missing. That wrong path still passes the allow-root check (the session root is an allowed root), so it isn't rejected early — it fails later at `realpath` with ENOENT, and the `cp` retry repeats the mistake.

## Fix

Resolve a **relative** `localPath` (or a bare filename) against the notebook data dir, and default a call with no `source`/`content` to `{kind:"localPath", path: filename}`. The agent can now pass the same relative name it saved with:

```python
plt.savefig("plot.png")                    # kernel cwd == data dir
write_artifact_file(filename="plot.png")   # no path reconstruction needed
```

**How the notebook data dir reaches the artifact tool (per-turn handoff, not an env var).** The artifact MCP env is built at session creation, but at that point we only hold a pre-start notebook alias (`notebook-session-*`); the real kernel writes under the **final** ACP session id (the notebook RPC layer rewrites the alias before touching disk). Keying paths off the alias is therefore wrong on the fresh-session path. Instead, the notebook context travels in the **per-turn `current-run.json` handoff** written by `activateArtifactRun`, which runs with the final session id. That handoff (`ArtifactRunContext`) carries `notebookDataDir` + `notebookSessionRoot`; the MCP server reads them via `readCurrentRunContext` and uses them as the relative base and the authoritative allowed import root. No new env var is introduced.

- Absolute paths are unchanged — `path.resolve` drops the base once it hits an absolute segment.
- The `realpath` + allow-root check remains the authoritative sandbox boundary; `assertSafeFilename` still blocks separators in the destination filename.
- The **static** `allowedImportRoots` carries only the session workspace (`sessionCwd`); the pre-start notebook alias is deliberately **not** authorized, so a stale-alias absolute path is rejected. The sole notebook import root is the final-id `notebookSessionRoot` from the handoff.
- Outside a notebook turn (artifacts enabled, notebook disabled) there is no data dir to resolve against, so the bare-filename default is **not** applied — the explicit `requires source or content` contract error is kept rather than silently resolving against the MCP process cwd.
- Notebook system-prompt and shell-tool guidance updated to steer agents to the saved relative filename and away from rebuilding an absolute path.

## Tests

- relative-path + bare-filename resolution against the handoff notebook data dir
- absolute-path passthrough when a base is set
- stale pre-start alias root rejected (`outside allowed artifact import roots`)
- bare filename with no source/content outside a notebook turn → explicit contract error
- runtime integration test: alias registration → an actual bare-filename artifact write, asserting the handoff pins the kernel dir by the **final** id (never `notebook-session-*`) and the file copies
- runtime env test: notebook alias root absent from the static allowed roots
- negative assertions guarding the old absolute-path guidance wording from regressing

typecheck: pass · lint: pass · format: clean · full suite: 4441 passed / 66 skipped


ewen-poch:
Fixed in 067d7cf. Confirmed the P1: `buildArtifactEnvironment` runs at session creation with only the pre-start notebook alias, while kernels write under the final ACP session id (the notebook RPC layer rewrites the alias before touching disk) — so a static alias-based base dir was wrong on the fresh-session path.

Moved the notebook context into the per-turn `current-run.json` handoff written by `activateArtifactRun` (which runs with the final session id). It records `notebookDataDir` + `notebookSessionRoot` keyed by that id; the MCP server reads them as the relative base and an authoritative allowed import root. Alias-proof, no new env var. Added a runtime integration test driving alias registration → an actual bare-filename write, asserting the handoff pins the kernel dir by the final id (never `notebook-session-*`) and the file copies.

ewen-poch:
Both addressed in 7fba426.

**P1 — stale alias root removed.** `buildArtifactEnvironment` runs at session creation with only the pre-start alias, so seeding `allowedImportRoots` with the alias-based notebook root let stale-alias absolute paths pass the allow-check even after the final-id root was appended. The static env now carries **only** `sessionCwd`; the sole notebook import root is the final-id `notebookSessionRoot` supplied per turn via `current-run.json`. Also dropped the now-unused `notebookSessionId` param from `buildArtifactEnvironment`/`createArtifactMcpServers`. Regression test added: a `localPath` under the stale alias dir is now rejected with "outside allowed artifact import roots"; the runtime env test asserts the alias root is absent from the static roots.

**P2 — guidance aligned.** `notebook/mcp-server.ts` line 32 (system prompt) and line 176 (shell tool doc) no longer demand an absolute path / claim a bare relative name won't resolve — both now say to pass the same relative filename you saved with (resolved against the notebook data dir). Negative assertions guard the old "use an ABSOLUTE path" / "will not resolve a bare relative name" / "absolute local path" wording from regressing.

Full suite green (4441 passed).

ewen-poch:
Thanks — verdict noted. Addressed the two low observations in a85af74:

- **PR description** rewritten to describe the shipped per-turn `current-run.json` handoff (`ArtifactRunContext` carrying `notebookDataDir`/`notebookSessionRoot`), not the abandoned env-var approach. No `OPEN_SCIENCE_ARTIFACT_NOTEBOOK_DATA_DIR` exists in the tree.
- **Bare-filename default outside a notebook turn** now keeps the explicit `requires source or content` contract error instead of resolving against the MCP process cwd. `normalizeArtifactToolWriteInput` takes the relative base and throws when it's absent; added a test covering the artifacts-enabled/notebook-disabled path.

Third note (handoff fields gated on global config vs per-session capability) left as-is per your "no change required" — harmless today since missing dirs are filtered by realpath. Full suite green (4442 passed).

ewen-poch:
For anyone reading the first-round Codex review: its two findings ([P1] stale notebook alias still authorized as an import root; [P2] notebook/shell guidance demanding absolute paths) were **already resolved in 7fba426**. The Codex job only runs on PR `opened` (not on `synchronize`), so it did not re-run after the fix and its original comment still reflects the first commit.

Current state on `a85af74`:
- **P1**: static `allowedImportRoots` carries only `sessionCwd`; the pre-start notebook alias is no longer authorized. The sole notebook import root is the final-id `notebookSessionRoot` supplied per turn via `current-run.json`. Regression tests: a `localPath` under the stale alias dir is rejected ("outside allowed artifact import roots"), and the runtime env test asserts the alias root is absent from the static roots.
- **P2**: `notebook/mcp-server.ts` lines 32 and 176 now instruct passing the same relative filename you saved with; negative assertions guard the old "use an ABSOLUTE path" / "will not resolve a bare relative name" / "absolute local path" wording from regressing.

All gates green (Verify ×3, Analyze ×3, CodeQL, Claude architecture review = mergeable).

PR #263, opened 2026-07-21
feat(settings): add reasoning effort selection to model settings

### LINKED ISSUE #235: Add reasoning effort selection to General settings
## Feature request

Please add a **Reasoning effort** setting to the General settings page for agents or providers that support it.

## User problem

Users currently cannot choose how much reasoning effort the agent should use. Some tasks need deeper reasoning, while simpler tasks may benefit from a faster, lower-effort response. Without a visible setting, users cannot control this trade-off from Open Science.

## Desired behavior

- Add a Reasoning effort selector to **Settings > General**.
- Provide these choices: **Max**, **High**, **Medium**, and **Low**.
- Clearly show the currently selected level.
- Persist the selection between app restarts.
- Apply the selected effort to subsequent agent requests when supported by the active provider.

This would let users choose the appropriate reasoning depth for each type of work.

ewen-poch:
## Summary

Adds a **Reasoning effort** setting letting users trade off deeper reasoning against faster responses, as requested in #235. The control is a horizontal segmented selector — **Default / Low / Medium / High / Max** — with the selection persisted in `settings.json` across restarts and applied to subsequent agent requests **when the active provider supports it**.

## Deviations from #235 (deliberate, maintainer-confirmed)

Two intentional adjustments to the issue's acceptance criteria, called out explicitly:

1. **Placement: Settings > Model, not Settings > General.** The issue asks for General, but the selector lives in the Model panel's Providers section, next to Active model. Reasoning effort is a per-request model/agent behavior knob — it pairs naturally with the model picker, and the General panel currently holds only app-level utilities (diagnostics, CLI, community links).
2. **A fifth choice, Default, beyond the issue's four levels.** `default` means "don't override": nothing is sent and agents keep their own defaults. It is the persisted initial value, so **existing users see zero behavior change on upgrade** — without it, whichever level we defaulted to would have been force-applied to every agent that previously ran at its own default. The four issue-requested levels are all present and behave as asked.

## How levels map onto agents

The four levels form a **relative scale** (`low < medium < high < xhigh < max`); each framework maps the choice onto what the active model actually supports — a level the model doesn't offer resolves to its **closest advertised rung** (ties go to the cheaper one):

- **Claude Code** — ACP `session/set_config_option` on the adapter's `thought_level`/`effort` option.
- **Codex** — `model_reasoning_effort` in `CODEX_CONFIG` (`max → xhigh`), plus the ACP option when advertised; the Responses bridge forwards `reasoning_effort` upstream **only** when the user explicitly picked a level (never Codex's own default).
- **OpenCode** — `options.reasoningEffort` in the per-model block of both generated config layers (`max → high`, the AI SDK ceiling).

**Live switching:** Claude Code and Codex apply level changes to open sessions over ACP with **no respawn** (the way a model switch feels); picking Default sends the adapter's `default` sentinel to hand control back. OpenCode advertises no live effort channel, so it reconnects to rebake its generated config — deferred while a prompt is in flight, with the conversation resumed afterwards.

## Tests

- 50+ new cases across every layer: repository sanitize/round-trip, service projection + backend resolution, IPC boundary validation + live/reconnect routing, effort-option matching/clamping/default-sentinel, runtime apply (post-model-switch re-resolution, live multi-session apply, failure fallback to reconnect), Codex/OpenCode config generation, bridge outgoing-request assertions (forward vs. strip), store optimistic update/revert, segmented-control rendering.
- Full suite green: `typecheck`, `lint`, `vitest` (4656 passed), Prettier-clean, web api map regenerated.

## Reviewer notes

- The Codex/OpenCode ACP adapters are downloaded at runtime; their config channels are covered at the generation level. Live smoke-tested locally: Claude Code and Codex accepted levels over ACP; OpenCode's generated config carried the level (its ACP advertises no effort option, handled by the config channel).
- Gateway-dependent behavior varies by provider (e.g. DeepSeek maps `low/medium → high` and auto-`max`es agent requests) — expected per the issue's "when supported" clause.

Closes #235

### DIFF
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -4766,6 +4766,71 @@ describe('ACP runtime — session effort', () => {
     })
   })
 
+  it('resolves a live effort change against the options reported after a model switch', async () => {
+    const process = new FakeAgentProcess()
+    const fakeAgent = startFakeAgent(process, ['s-live'], {
+      configOptions: [
+        {
+          type: 'select',
+          id: 'model',
+          name: 'Model',
+          category: 'model',
+          currentValue: 'model-a',
+          options: [{ value: 'model-b', name: 'Model B' }]
+        } as SessionConfigOption,
+        thoughtLevelOption(['low', 'high'])
+      ],
+      // The model switch narrows the effort set: the session/new options are now stale.
+      updatedConfigOptions: [thoughtLevelOption(['low', 'medium'])]
+    })
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
+      defaultCwd: '/workspace',
+      resolveBackend: () => ({
+        framework: { ...claudeCodeFramework, spawn: () => asAgentProcess(process) },
+        executablePath: '/bin/claude',
+        env: {},
+        sessionModel: 'model-b'
+      })
+    })
+    await runtime.createSession({ cwd: '/workspace' })
+    fakeAgent.configChanges.length = 0
+
+    const applied = await runtime.applyReasoningEffortChange('max')
+
+    // The post-switch set tops out at 'medium'; the stale session/new set would wrongly yield 'high'.
+    expect(applied).toBe(true)
+    expect(fakeAgent.configChanges).toEqual([
+      { sessionId: 's-live', configId: 'effort', value: 'medium' }
+    ])
+  })
+
+  it('reports failure so the caller reconnects when a live apply is rejected', async () => {
+    const process = new FakeAgentProcess()
+    const agentOptions: Parameters<typeof startFakeAgent>[2] = {
+      configOptions: [thoughtLevelOption(['low', 'high'])]
+    }
+    const fakeAgent = startFakeAgent(process, ['s-live'], agentOptions)
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
+      defaultCwd: '/workspace',
+      resolveBackend: () => ({
+        framework: { ...claudeCodeFramework, spawn: () => asAgentProcess(process) },
+        executablePath: '/bin/claude',
+        env: {}
+      })
+    })
+    await runtime.createSession({ cwd: '/workspace' })
+
+    agentOptions.rejectSetConfigOption = true
+    const applied = await runtime.applyReasoningEffortChange('high')
+
+    // The level never reached the agent: returning false lets the caller reconnect instead of
+    // leaving the UI showing a level the agent never received.
+    expect(applied).toBe(false)
+    expect(fakeAgent.configChanges).toEqual([])
+  })
+
   it('declines the live change when the framework bakes effort into its spawn config', async () => {
     const process = new FakeAgentProcess()
     const fakeAgent = startFakeAgent(process, ['s-effort'], {
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -129,6 +129,7 @@ const startFakeAgent = (
   closedSessions: string[]
   cancelledSessions: string[]
   modeChanges: Array<{ sessionId: string; modeId: string }>
+  configChanges: Array<{ sessionId: string; configId: string; value: string | boolean }>
   actions: string[]
 } => {
   const authRequests: unknown[] = []
@@ -144,6 +145,7 @@ const startFakeAgent = (
   const closedSessions: string[] = []
   const cancelledSessions: string[] = []
   const modeChanges: Array<{ sessionId: string; modeId: string }> = []
+  const configChanges: Array<{ sessionId: string; configId: string; value: string | boolean }> = []
   const actions: string[] = []
   let sessionIndex = 0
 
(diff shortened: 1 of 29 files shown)


ewen-poch:
Thanks for the review! All three findings are addressed in 0e9665f:

- **[P2] Post-model effort resolution** — `applySessionModel` now returns the agent's updated `configOptions` from the `set_config_option` response, and `applySessionEffort` resolves against that post-switch set (falling back to the session's original). Added a runtime test where the model switch narrows the effort options and the clamped level follows the new set.
- **[P2] Bridge effort** — `responsesToChatRequest` now translates `reasoning.effort` to the Chat Completions `reasoning_effort` parameter (`xhigh` clamps to `high`; `none` is omitted as it has no portable equivalent) instead of dropping it. Covered by conversion unit tests plus an outgoing-request integration test asserting the gateway receives `reasoning_effort: "high"`.
- **[P3] IPC validation** — the handler rejects values outside `default|low|medium|high|max` (via a shared `isReasoningEffort` guard, now also used by the settings sanitizer) before persistence or reconnect. Malformed-payload tests added.

Full suite green locally: typecheck, lint, 4483 tests passed.

(Side note: the **Claude architecture review** check failed in 17s with `Environment variable validation failed` — looks like a workflow configuration issue rather than review feedback.)

ewen-poch:
Thanks for the architecture review! The dual-channel design note is addressed in 8ee70ab: the redundancy is intentional (defense-in-depth, mirroring how `sessionModel` reaches opencode), and the resolution site now documents the channel ownership and precedence — the per-session ACP `thought_level` application wins over the baked config when both fire, while each channel clamps `max` within its own supported set.

ewen-poch:
Both notes addressed:

- **[Medium] Bridge forwarding regression** — fixed in e4ef09b. The bridge now only translates `reasoning.effort` → `reasoning_effort` when the app's setting is an explicit level: `resolveActiveAgentBackend` threads `forwardReasoningEffort: sessionEffort !== undefined` into the bridge target (same path as `baseUrl`/`key`). Without an explicit choice the bridge keeps stripping the field exactly as pre-PR, so Codex's own default effort can no longer leak to gateways that might reject it — the "zero behavior change on upgrade" contract holds. Regression coverage: conversion test asserting the field is stripped without the flag, plus an integration test proving the upstream body carries no `reasoning_effort` then.
- **[Low] `sessionEffort: string` widening** — fixed earlier in e21a22b: both `ResolvedAgentBackend.sessionEffort` and the runtime's `pendingSessionEffort` are now `ReasoningEffort`; the broader advertised scale (incl. `xhigh`) stays string-based inside `session-config.ts` as suggested.

Full suite green locally: typecheck, lint, 4486 tests passed.

ewen-poch:
All findings addressed (branch rebased onto current main; the offending commit was reworded — all 9 subjects now pass the commit-message check):

- **[P1] Commit message casing** — reworded to `refactor(acp): type session effort as the settings union on the backend contract` (all-lowercase subject).
- **[P1] Bridge forwarding policy not updating on live switch** — fixed in 2afa1d1. `setReasoningEffort` now updates the live bridge target in place via `ResponsesBridge.setForwardReasoningEffort`: an explicit level starts forwarding without any reconnect; `default` restores stripping so Codex's own default effort cannot leak upstream. Covered by an end-to-end test driving a real bridge through default → high → default.
- **[P1] Live apply reporting success on failure / stale options** — fixed in 2afa1d1. The runtime now caches the `configOptions` refreshed by each model switch and resolves live effort changes against that set (never the stale session/new set); a rejected live apply returns `false`, so the IPC falls back to the provider-switch reconnect instead of leaving the UI showing a level the agent never received. Tests cover both the post-model-switch resolution and the failure fallback.
- **[P2] Placement (Settings > General vs Model)** — intentional: the maintainer chose Settings > Model (Providers section, next to Active model) as the better home for a per-request agent knob. Leaving as is; happy to revisit if you feel strongly.

Full suite green locally: typecheck, lint, 4513 tests passed.

ewen-poch:
Both Low notes addressed in 6a967ee (now 2cca74d after rebasing onto current main):

- **Partial session inconsistency on live-apply failure** — the loop now attempts every open session and only then reports the aggregated failure, so one rejected session can no longer leave unprocessed sessions on the old level while the fallback reconnect runs.
- **Orphaned `latestSessionConfigOptions` entry** — refreshed options are now committed to the cache only after the session is registered, keyed by the agent session id (which differs from the app id for adopted sessions), with matching cleanup of both keys on session teardown.

Rebased onto `3206a62` (the Codex auth follow-up); the only conflict was the ProviderList wrapper in SettingsPage, resolved keeping both the new login props and the spacing fix. Full suite green: typecheck, lint, 4654 tests passed.

ewen-poch:
Addressed in 84320ae.

First, the factual part: the concern is qualified as unverified, but we have live evidence the downloaded codex-acp adapter DOES advertise a `thought_level` option — a real Codex session accepted `effort: high` over ACP (`session effort applied`) during local smoke testing. That is now documented at the flag in `codex.ts`.

Second, the structural point stands regardless, so the vacuous-true path is closed: `applyReasoningEffortChange` now tracks whether any open session actually received an option. When none did:

- **Codex** returns `false`, so the caller reconnects and the baked `model_reasoning_effort` config delivers the level — no more "UI shows applied, running session silently unaffected".
- **Claude** reports success, because it has no config channel: a model that does not support effort cannot be helped by a respawn, matching the build-time "when supported" contract.

Tests added for both branches (Codex with no advertised option falls back to reconnect; Claude reports success without one). Full suite green: typecheck, lint, 4656 tests passed.

ewen-poch:
Both Spec notes are documentation/scope confirmations rather than code issues — the PR description now has a dedicated **"Deviations from #235 (deliberate, maintainer-confirmed)"** section stating them explicitly:

1. **Placement** — Settings > Model (Providers section, next to Active model) is the deliberate home for this per-request model/agent knob, not Settings > General as literally worded in the issue.
2. **Default option** — the fifth choice is an accepted scope adjustment: as the persisted initial value it guarantees zero behavior change on upgrade, which no forced level could.

No behavior changes in this update; the PR body was also brought current with the final design (segmented selector, live ACP switching, bridge gating).

PR #268, opened 2026-07-21
fix(codex): trust api-key/gateway auth, bound status reads, and gate discarded sign-ins

ewen-poch:
## Summary

Addresses four issues in the Codex auth + onboarding flow surfaced in review of the Codex sign-in work.

- **Spec P1 — trust api-key/gateway auth (`codex-auth.ts`)**: `getStatus` gated on the `chat-gpt` capability *before* reading live status, so an adapter advertising only `api-key` (already holding a usable api-key/gateway credential) was reported as a capability failure — wrongly blocking an otherwise working provider. It now reads status first and treats any authenticated method (`api-key`/`chat-gpt`/`gateway`) as authenticated, returning a capability failure only when the profile is signed out **and** ChatGPT login is unavailable.
- **P2 — bounded status read (`codex-auth.ts`)**: `getStatus` had no timeout, so a stalled `codex-acp` hung save/test indefinitely. It now reuses `loginIsolated`'s abort + late-close pattern, bounded by a new `statusTimeoutMs` (default 30s), and closes a session that only arrives after the deadline.
- **P1 — gate discarded sign-ins (`shared/settings.ts`, `service.ts`, `OnboardingWizard.tsx`)**: a sign-in/validation can succeed yet be discarded when the provider was switched, deleted, or superseded by a newer test mid-flow. Those paths returned `ok: true` with no signal, so onboarding advanced and finished on a profile the store never recorded. `ValidateProviderResult` gains an optional `applied` flag; discard paths set `applied: false` and both wizard branches keep the user on the provider step to retry.
- **Standards — onboarding UI coverage**: adds render tests for the isolated Codex flow (success→advance, failure→stay, discarded→stay, unmount→`cancelCodexLogin`), covering the previously-untested teardown ref and the persist→login→activate branch.

## Testing

- `npm run typecheck` — pass
- `npm run lint` — pass
- `npm run test` — 4,526 pass / 66 skipped

### DIFF
--- a/src/main/settings/codex-auth.test.ts
+++ b/src/main/settings/codex-auth.test.ts
@@ -89,8 +89,12 @@ describe('CodexAuthController', () => {
       'private@example.test'
     )
 
+    // A signed-out adapter that also cannot offer ChatGPT login has nothing to connect: that is the
+    // genuine capability failure. (A signed-out adapter that DOES advertise chat-gpt is merely
+    // unauthenticated — covered below — not a capability failure.)
     const unsupported = session({
-      initialize: vi.fn().mockResolvedValue({ authMethods: [{ id: 'api-key' }] })
+      initialize: vi.fn().mockResolvedValue({ authMethods: [{ id: 'api-key' }] }),
+      status: vi.fn().mockResolvedValue({ type: 'unauthenticated' })
     })
     const unsupportedController = new CodexAuthController({
       openSession: vi.fn().mockResolvedValue(unsupported)
@@ -101,6 +105,28 @@ describe('CodexAuthController', () => {
       authenticated: false,
       message: 'The installed codex-acp does not advertise ChatGPT authentication.'
     })
+  })
+
+  it('reports a chat-gpt-less adapter that already holds a credential as authenticated', async () => {
+    // Regression: getStatus must not gate on the chat-gpt capability before reading status. An adapter
+    // advertising only api-key, already carrying an api-key/gateway credential, runs fine — reporting
+    // it signed out (a capability failure) would wrongly block an otherwise working provider.
+    for (const type of ['api-key', 'gateway'] as const) {
+      const credentialed = session({
+        initialize: vi.fn().mockResolvedValue({ authMethods: [{ id: 'api-key' }] }),
+        status: vi.fn().mockResolvedValue({ type })
+      })
+      const controller = new CodexAuthController({
+        openSession: vi.fn().mockResolvedValue(credentialed)
+      })
+
+      await expect(controller.getStatus('shared')).resolves.toEqual({
+        mode: 'shared',
+        supported: true,
+        authenticated: true
+      })
+      expect(vi.mocked(credentialed.close)).toHaveBeenCalledOnce()
+    }
   })
 
   it('treats api-key and gateway profiles as authenticated', async () => {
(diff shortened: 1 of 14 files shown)


ewen-poch:
Round 1 (addressing AI review): applied the credential-first ordering from `getStatus` to both `loginIsolated` and `logoutIsolated` in `codex-auth.ts` — an existing api-key/gateway credential in the isolated home now short-circuits (login returns authenticated without the browser flow; logout clears it) even on a build that doesn't advertise `chat-gpt`. The capability failure is preserved only for the signed-out + no-ChatGPT case. Added api-key-only-adapter coverage for both operations (commit fc1fe7d).

Not changed: the non-Codex onboarding activate-order asymmetry Claude noted — the review marks it "None required for this PR" as it's unreachable in the linear single-save flow.

ewen-poch:
Round 2 (addressing AI review): extracted `withBoundedSession` in `codex-auth.ts` — owns session open (racing the deadline), late-close, timeout, and teardown — and routed `getStatus`, `loginIsolated`, and `logoutIsolated` through it. This closes the gap the review flagged: round 1 had added a `status()` round-trip to the user-triggered `logoutIsolated` without bounding it, so a stalled adapter could freeze sign-out; it's now bounded like the reads. `loginIsolated` exposes its AbortController via a `register` callback so `cancelLogin` still works. Added a sign-out-timeout test (commit 38642fa).

Not changed: the optional `status()` method-not-found → `capabilityFailure` mapping — speculative, and every supported codex-acp build implements `authentication/status` (getStatus's happy path already relies on it).

ewen-poch:
Round 3 (addressing AI review): the re-entrancy guard for `loginIsolated` now claims the in-progress slot synchronously — the `AbortController` is created and assigned to `activeLogin` in `loginIsolated` itself, in the same tick as the guard check and before the async `withBoundedSession` helper is entered. Round 2 had moved that assignment into a callback inside the async helper; it happened to still run before the first await, but that left correctness resting on a fragile invariant. The helper now takes the controller as a parameter plus an `onSettled` teardown callback. Added a test asserting a second concurrent sign-in is rejected and opens no second session (commit bd76942).

This is the third and final review round I'll process automatically — pausing here for human review. Remaining AI notes are non-blocking and explicitly out of scope (Claude's logout-clears-validation-state observation is pre-existing; the `status()` method-not-found mapping is speculative).

ewen-poch:
**P1 fix** (post-round-3, user-reported): logout timeout was treated as success — `logoutIsolated()` times out on the status read and returns `authenticated: false` without calling `session.logout()`, then `logoutIsolatedCodex()` unconditionally cleared `lastValidatedAt`. Result: UI shows signed out while the credential remains in the isolated home and is usable by the runtime.

Fixed by making `logoutIsolatedCodex` return `ValidateProviderResult` (matching `loginIsolatedCodex`) and only clearing markers when logout genuinely succeeds (status with no error message). On timeout, preserve verified state so the "Sign out" button stays visible, and return the failure so SettingsPage can surface it via `providerTestError`. IPC handler fires `onActiveProviderChanged` only on confirmed sign-out—reconnecting after timeout would re-authenticate with the credential still in place.

Added service + IPC tests for timeout-preserves-markers, success-clears-markers, and no-reconnect-on-timeout (commit 126429d).

PR #275, opened 2026-07-22
fix(update): recover from old-uninstaller failure during Windows updates

ewen-poch:
## Problem

Windows users updating in-app hit a fatal NSIS dialog: `Failed to uninstall old application files. Please try running the installer again.: 2`. The update aborts and the app has to be reinstalled manually (reported by two users on Windows).

Root cause analysis:

- During an update the new installer runs the **old** uninstaller and treats any non-zero exit code as fatal (`handleUninstallResult` in electron-builder's NSIS templates): dialog, `SetErrorLevel 2`, `Quit`.
- Our installer is assisted (`oneClick: false`). electron-builder normalizes the uninstaller's exit code (`quitSuccess`, "avoid exit code 2") only for ONE_CLICK builds, so on the assisted path a benign trailing error can leak out as **exit code 2 even when the old version was fully removed** — same signature as [electron-userland/electron-builder#9593](https://github.com/electron-userland/electron-builder/issues/9593) (no leftover processes; manually re-running the same uninstaller succeeds).
- When the code *is* real, the lock typically comes from a background child running from the install dir: `resources\micromamba.exe` during in-flight runtime provisioning, the `open-science` CLI in Node mode, or an agent child. The app's pre-update install gate already reaps the agent + kernel trees; the installer-side recovery added here covers what the gate cannot see, at the point of failure.

## Proposed change

Define electron-builder's `customUnInstallCheck` / `customUnInstallCheckCurrentUser` hooks in `build/installer.nsh`. They replace `handleUninstallResult`'s failure path in the **new** installer, so the fix engages when updating **to** a build that contains it — from any old version, including the currently affected ones. Recovery order:

1. Old uninstaller exited non-zero but the old executable (``) is already gone → the uninstall did its job despite the reported code; log and continue installing.
2. Files remain → force-kill whatever still runs from the install dir (PowerShell path-prefix sweep + image-name `taskkill` fallback for PowerShell-restricted machines), wait, and rerun the old uninstaller once. Only if that retry also fails, show the original dialog and quit with exit code 2 (no worse than today).

The hook body is deliberately self-contained (``, registers, built-in constants, the literal `\old-uninstaller.exe` path): `handleUninstallResult` is parsed *before* `uninstallOldVersion`/`CHECK_APP_RUNNING` declare their globals, and makensis treats unknown variables as errors — referencing ``/`` fails the installer compile (caught in local validation, see below).

## Scope and non-goals

- NSIS include + packaging tests only; no app-runtime code changes.
- The pre-update install gate (agent/kernel teardown) is unchanged. Extending it to cancel in-flight micromamba provisioning before `quitAndInstall` is a possible follow-up; the kill + retry here already covers that lock class at the point of failure.
- macOS / Linux update paths untouched.
- Takes effect for updates **to** a release containing it (the hook runs in the new installer); no migration needed.

- `build/installer.nsh` defines both hooks with the recovery order above; `build/packaging.test.ts` pins the contract (hooks registered for both registry passes, ``-gone short-circuit, kill + single retry, fatal dialog preserved, no late-declared symbols).
- **NSIS compile validated locally**: `npx electron-builder --win nsis --publish never` builds `open-science-0.5.1-win-arm64-setup.exe` with makensis under warnings-as-errors. The first draft referenced `uninstallOldVersion`'s late-declared globals and failed this compile; the final version compiles both the installer and uninstaller passes cleanly.
- `npm run typecheck`, `npm run lint` (0 errors), and `npm run test` (381 files / 4691 tests passed) are green.
- Not exercised end-to-end on a physical Windows machine (none available here); the recovery flow reuses the same primitives electron-builder's own templates use (`nsExec` process sweep, `ExecWait` retry semantics).

## Review focus

- The ``-gone short-circuit: continuing the install when the reported failure is spurious (mirrors the default handler's own "cannot launch uninstaller → continue" asymmetry).
- `$0` preservation: it still carries the uninstaller arguments (`/currentuser` etc.) for the retry; the chosen kill primitives never touch it.
- The installMode==all HKCU-pass caveat documented in the include comment (retry targets `$INSTDIR`; harmless no-op in that rare path).

### DIFF
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -1,3 +1,62 @@
 !macro customUnInstall
   nsExec::ExecToLog 'powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -File "$INSTDIR\resources\windows-runtime-cache-uninstall.ps1"'
 !macroend
+
+# Resilient replacement for handleUninstallResult's default failure handling, installed via
+# electron-builder's customUnInstallCheck hooks below. During an in-app update the new installer
+# runs the OLD uninstaller and treats any non-zero exit code as fatal ("Failed to uninstall old
+# application files. Please try running the installer again.: <code>") — aborting the update.
+# That code is not trustworthy for our assisted installer (oneClick: false): electron-builder only
+# normalizes the uninstaller's exit code (quitSuccess, "avoid exit code 2") for ONE_CLICK builds,
+# so a benign trailing error leaks out as exit code 2 even when the old version was fully removed
+# (electron-userland/electron-builder#9593). And when the code IS real, it is usually a background
+# child still running from the install dir (micromamba provisioning, the CLI in Node mode, an
+# agent child) locking files — worth one more attempt after a force-kill instead of failing.
+# Recovery order:
+#   1. Exit code non-zero but the old executable is already gone -> the uninstall did its job
+#      despite the reported code; log and continue installing.
+#   2. Files remain -> force-kill processes running from the install dir, wait, and run the old
+#      uninstaller once more. Only if it still fails show the original dialog and quit.
+# Symbol constraints: handleUninstallResult is parsed BEFORE uninstallOldVersion and
+# CHECK_APP_RUNNING declare their globals ($installationDir, $PowerShellPath, ...), and makensis
+# treats unknown variables as errors — so this stays self-contained: only $appExe (set in the
+# install section before the uninstall pass), registers, built-in constants, and the literal
+# temp-uninstaller path uninstallOldVersion uses. For the SHELL_CONTEXT pass $INSTDIR is the old
+# install location (an update installs over it). The rare installMode==all HKEY_CURRENT_USER pass
+# targets a stray per-user install elsewhere; re-running its uninstaller against $INSTDIR is a
+# harmless no-op that lets the machine-wide install continue.
+!macro uninstallFailureRecovery
+  ${if} $R0 != 0
+    ${ifNot} ${FileExists} "$appExe"
+      DetailPrint `Old uninstaller exited with $R0 but the previous installation is already removed; continuing.`
+    ${else}
+      DetailPrint `Old uninstaller exited with $R0; closing leftover app processes and retrying once.`
+      # Force-kill anything still running from the install dir: path-prefix sweep via PowerShell
+      # (covers micromamba.exe and any other helper), then an image-name taskkill for machines
+      # where PowerShell is unavailable or policy-blocked. Both are best-effort; the retry below
+      # is the real verdict. $0 keeps the uninstaller arguments (/currentuser etc.) — neither
+      # nsExec call touches it.
+      nsExec::Exec `"$SYSDIR\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -NonInteractive -C "Get-CimInstance -ClassName Win32_Process | ? {$$_.Path -and $$_.Path.StartsWith('$INSTDIR', 'CurrentCultureIgnoreCase')} | % { Stop-Process -Id $$_.ProcessId -Force -ErrorAction SilentlyContinue }"`
+      Pop $R1
+      nsExec::Exec `"$SYSDIR\cmd.exe" /C taskkill /F /IM "${APP_EXECUTABLE_FILENAME}"`
+      Pop $R1
+      ClearErrors
+      Sleep 1000
+      ExecWait '"$PLUGINSDIR\old-uninstaller.exe" /S /KEEP_APP_DATA $0 _?=$INSTDIR' $R0
+      ${if} $R0 != 0
+        MessageBox MB_OK|MB_ICONEXCLAMATION "$(uninstallFailed): $R0"
+        DetailPrint `Uninstall was not successful. Uninstaller error code: $R0.`
+        SetErrorLevel 2
+        Quit
+      ${endif}
+    ${endif}
+  ${endif}
+!macroend
+
+!macro customUnInstallCheck
+  !insertmacro uninstallFailureRecovery
+!macroend
+
+!macro customUnInstallCheckCurrentUser
+  !insertmacro uninstallFailureRecovery
+!macroend
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -30,13 +30,19 @@
     # matches on ExecutablePath (Win32_Process has no Path property) with a trailing-backslash
     # boundary so a sibling directory can never match, and receives the directory as an ARGUMENT —
     # never interpolated into the script source — so a custom install dir containing an apostrophe
-    # breaks nothing and injects nothing. The image-name taskkill covers machines where
-    # PowerShell is unavailable or policy-blocked. Both are best-effort; the retry is the verdict.
+    # breaks nothing and injects nothing. The image-name taskkill is ONLY a fallback for when the
+    # sweep could not run (PowerShell missing or policy-blocked): it matches the exe name in ANY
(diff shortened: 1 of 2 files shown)


ewen-poch:
Round 1 review triage — changes pushed in fbef304 (makensis compile re-validated locally via `electron-builder --win nsis`, packaging tests 10/10 green).

**Adopted**

- **[Codex P1] `Win32_Process` property** — correct, `Path` does not exist on `Win32_Process`; the sweep silently matched nothing. Now matches on `ExecutablePath` with a trailing-backslash boundary, so sibling directories can never over-match either. (Worth noting: electron-builder's own `FIND_PROCESS`/`KILL_PROCESS` have the same latent bug; we no longer reuse them, partly for this reason.)
- **[Codex P1 / Claude Medium] `customUnInstallCheckCurrentUser` targeting** — the HKCU pass now reads the real per-user `InstallLocation` from the registry and runs the recovery (sentinel + kill + retry) against that directory. When the location is unreadable it keeps electron-builder's default fatal handling instead of touching the wrong directory — exactly the fallback both reviewers suggested. The shared recovery body is parameterized on the target directory (`uninstallFailureRecoveryAt`), so the SHELL_CONTEXT pass is unchanged in behavior.
- **[Codex P2] PowerShell injection via install dir** — the directory is now passed as a command-line argument (`$args[0]`) and treated strictly as data, never interpolated into the script source. Custom install dirs containing apostrophes are safe.

**Declined (with rationale)**

- **[Codex] Behavioral Windows/PowerShell test instead of string assertions** — agree in principle, but the repo has no NSIS-execution test harness; adding one (spawning processes on a Windows runner and asserting kills) is a CI-infrastructure change that belongs in its own PR. The current contract tests + the local makensis compile gate are what this codebase supports today. Happy to file a follow-up.
- **[Claude Low] Wire the NSIS compile into CI** — same reasoning: wiring `electron-builder --win nsis` (or makensis --WX) into pr-check means downloading Electron + NSIS binaries on every PR; that is a `ci/` scoped change of its own and out of scope here. Noted as a follow-up alongside the behavioral test.

Awaiting the next review round on fbef304.

ewen-poch:
Round 2 review triage — changes pushed in 2d0f3fd (makensis compile re-validated locally; packaging tests 12/12 green).

**Adopted**

- **[Spec P1 / Claude Low] Image-name taskkill blast radius** — `taskkill /F /IM` matches the exe name in ANY directory, so a second install or the portable zip copy would be force-killed, discarding unsaved work. It now runs **only** when the path-scoped PowerShell sweep could not run at all (nsExec launch failure or non-zero exit). On PowerShell-available machines only processes under the install dir are ever killed.
- **[Spec P2] Test contract too weak** — the packaging tests now pin both behaviors: the taskkill appears exactly once and only inside the sweep-failure guard, and the old uninstaller is retried exactly once (`ExecWait` count == 1).

**Already addressed / no change needed**

- **[Spec P2] HKCU pass must check, clean, and retry against its own real old-install dir** — this described the first iteration; since fbef304 the HKCU hook reads the per-user `InstallLocation` from the registry into `` and runs the sentinel, kill, and retry against that directory, falling back to the default fatal handling when it is unreadable. The current-user pass can no longer short-circuit on the machine-wide exe.
- **[Claude] Verdict: mergeable** — thanks. On the three Low notes: (a) hook-detachment detection and (b) wiring the NSIS compile into CI are the same CI-infrastructure follow-up noted in round 1 (own `ci/` PR — asserting the compiled installer contains the recovery branch is the right form for it); (c) installer force-kill ownership is already tracked as the follow-up to cancel provisioning in the pre-update gate, with the installer sweep deliberately kept as last resort — see 'Scope and non-goals' in the PR description.

ewen-poch:
Round 3 review triage — changes pushed in 74d58cb9 (packaging tests 13/13 green; this commit is test-only, no NSIS change).

**Adopted**

- **[Claude Medium] Hook-contract undetectable by text tests** — fair point, and cheap to close: the packaging tests now read the resolved `app-builder-lib` template from `node_modules` and assert `handleUninstallResult` still references `customUnInstallCheck` / `customUnInstallCheckCurrentUser`. An electron-builder bump that renames or drops the insertion points now fails CI instead of shipping a no-op recovery. (This covers the detection half; an end-to-end compiled-installer assertion remains with the `ci/` follow-up noted earlier.)
- **[Claude Low] Over-literal assertions** — the retry test now pins the semantic parts (single `ExecWait` of the temp uninstaller, `/S /KEEP_APP_DATA /bin/bash`, `_?=${DIR}`) instead of the full literal line, so a benign reformat no longer breaks the test. The security-critical pins (dir passed as `$args[0]`, never interpolated) stay literal on purpose.

**No change required**

- **[Claude note] taskkill fires on any sweep non-zero, incl. transient PowerShell failures** — acknowledged accepted tradeoff: gated last resort, bounded blast radius, documented in the include comment.

Process note: this commit was fast-forwarded onto the branch via the GitHub git-data API because git's HTTPS transport to github.com is failing intermittently from this network (github.com:443 timeouts while api.github.com stays reachable). Tree and message are byte-identical to the local commit; only the commit SHA differs.

That's three triage rounds — pausing here per the agreed limit. Current status: both reviewers' latest verdicts are **mergeable**; the remaining follow-ups (CI-side NSIS compile/behavioral test, pre-update gate cancelling in-flight provisioning) are tracked in the PR description and earlier comments.

ewen-poch:
Round 4 review triage — both findings were valid; fixes pushed in 5e7887f (makensis compile re-validated locally, packaging tests 13/13 green).

**[P1] HKCU spurious-failure path read InstallLocation too late — adopted.** Correct catch: a successful per-user uninstall deletes HKCU `InstallLocation`, so reading it inside `customUnInstallCheckCurrentUser` (which runs after the pass) always came up empty in exactly the "uninstalled but exited 2" case, and the hook fell through to the fatal dialog it was meant to replace. The location is now cached in `customInit` — which runs before any uninstall pass — into a `perUserInstallDirCache` global, and the hook consumes the cached value (fatal fallback retained only when no per-user install was registered at install start). Two implementation notes: the variable is declared *inside* `customInit` because a file-scope `Var` is declared-but-unused in the uninstaller build, which makensis fails as a warning (caught by the local compile); and the test that previously pinned the post-hoc read order now asserts the opposite — the hook body must not touch the registry at all.

**[P2] Image-name fallback could not clear a micromamba lock — adopted.** `taskkill /F /IM micromamba.exe` added to the sweep-failure branch alongside the app exe: on PowerShell-blocked machines the app-exe kill alone leaves `resources\micromamba.exe` (in-flight provisioning) alive, and the retry would fail as before. The widened blast radius (an unrelated micromamba on the box would be killed too) is documented in the include comment as an accepted last-resort tradeoff — it fires only when no path-scoped option exists.

The branch was also tidied: the round-3 commit (previously fast-forwarded via the git-data API as 74d58cb9 due to the github.com transport failures) is now the local base, and this fix was rebased onto it, so local and remote are linear again.

ewen-poch:
Round 5 triage — both test-completeness gaps were worth closing; test-only change pushed in f63117f (14/14 green, no NSIS change).

- **[P2] Guard-block coverage** — the test now captures the whole `${if} $R1 != 0 … ${endif}` block and asserts it contains exactly the two image-name taskkills (app exe + micromamba). Moving either kill outside the guard now fails, where the previous single-position regex would have stayed green.
- **[P2] `customInit` integration contract** — same class as the round-3 hook guard: the tests now also assert the resolved `installer.nsi` still contains and inserts `customInit`, so an electron-builder upgrade that drops that insertion point (leaving the HKCU cache macro never executed and the hook back on the always-empty fatal path) fails CI instead of regressing silently.

PR #285, opened 2026-07-22
fix(settings): isolate install state per runtime to prevent phantom progress (#278)

### LINKED ISSUE #278: [Bug]: Installing Codex also drives the OpenCode card and leaves a phantom install state
### Preflight checklist

- [x] This was reproduced in the current Open Science beta UI shown in the reporter's screenshot.

### What happened?

Clicking the **Install** button in the **Codex** runtime card also makes the **OpenCode** card above it enter an installing state.

Both cards display the same download progress at the same time. In the supplied Windows screenshot, the OpenCode and Codex sections both show:

```text
Installing…
Downloading — 11.1 / 138.4 MB
8%
```

Only the Codex install action was triggered.

After the download finishes, OpenCode can appear to have been installed as well, but the reporter describes it as a false or phantom installation: no independent OpenCode installation was requested, and re-detection/use does not match the installed-looking UI state.

This makes it impossible to tell which runtime is actually being downloaded and whether OpenCode is really available.

### Expected behavior

- Starting a Codex installation must update only the Codex card.
- The OpenCode card must keep its previous detected/not-detected state and must not show Codex progress, logs, success, or errors.
- After completion, only Codex detection and managed-install metadata may change.
- OpenCode must be marked installed only after its own executable is detected and passes its own readiness check.
- If only one runtime installation may run at a time, other cards can be disabled with a neutral message such as `Installing Codex…`, but they must not render the Codex operation as their own installation.

### Likely implementation cause

The packaged renderer currently exposes one shared set of installation fields in the settings store:

```text
isInstalling
installLogs
installProgress
installError
```

`installClaude`, `installOpencode`, and `installCodex` all write to those same fields and subscribe to the same install-log stream. The OpenCode and Codex status cards receive these shared values, so an install event has no runtime identity and is rendered by both cards.

This matches the screenshot, where both cards show the exact same byte count and percentage.

### Proposed fix

Track installation state by runtime ID, for example:

```text
installations.claude
installations.opencode
installations.codex
```

or include a required target in every install event:

```ts
type RuntimeId = "claude-code" | "opencode" | "codex"

type InstallEvent = {
  runtimeId: RuntimeId
  kind: "progress" | "log" | "complete" | "error"
  // ...
}
```

Each card should subscribe to or select only its own runtime state. Detection state (`resolvedPath`, managed flag, ready flag, and version) should be updated only from the matching installer result and a successful post-install detection check.

### Steps to reproduce

1. On Windows, open **Settings → Model**.
2. Ensure OpenCode and Codex are both not detected.
3. In the Codex section, choose the app-managed source.
4. Click the Codex install button.
5. Observe that the OpenCode card also changes to `Installing…` and mirrors the same progress.
6. Wait for the Codex installation to finish.
7. Re-detect or try to use OpenCode.
8. Observe that the OpenCode installed-looking state does not correspond to an independently installed and usable OpenCode runtime.

### Operating system

Windows (visible in the reporter's screenshot; exact Windows version not provided)

### App version

Current beta build shown in the screenshot; the exact version is not visible and should be confirmed with the reporter.

### Provider / model

Not provider-specific. The screenshot happens to show a DeepSeek provider, but the bug is in agent-runtime installation state.

### Acceptance criteria

- [ ] Clicking **Install Codex** shows progress only in the Codex card.
- [ ] Clicking **Install OpenCode** shows progress only in the OpenCode card.
- [ ] Claude installation state is isolated in the same way.
- [ ] A runtime is never marked installed solely because another runtime installer completed.
- [ ] Post-install detection verifies the expected executable/path/version before setting readiness.
- [ ] Install logs and errors are attributed to the correct runtime.
- [ ] Tests cover both install directions and a failed installation.

ewen-poch:
## Summary

Fixes #278. Starting an install for one agent runtime (e.g. Codex) drove the *other* runtime cards' (Claude, OpenCode) progress bars and logs, showing a phantom install that was never requested.

## Root cause

All three install methods (`installClaude`, `installOpencode`, `installCodex`) wrote to a single shared set of flat store fields — `isInstalling`, `installLogs`, `installProgress`, `installError`. Every card read those same fields, so any install's progress mirrored onto all three cards.

## Fix

Replace the flat fields with a per-runtime `installStates` record keyed by `AgentFrameworkId`. Each install streams its events into its own slice only; the other slices are never touched.

- `RuntimeInstallState` type + `createInitialRuntimeInstallState` factory
- `runRuntimeInstall` — shared driver owning subscribe/stream/unsubscribe, patching only its own slice
- `patchInstallState` — targeted immutable merge into one slice
- `selectAnyInstalling` — global lock so the framework selector and every card's Uninstall button still disable during any install (one install at a time is preserved)
- **Atomic single-install guard** inside `runRuntimeInstall`: it refuses to start a second install when any runtime is already installing (synchronous check+set, no `await` between). This enforces the single-install invariant in the store itself, so the shared broadcast `onInstallLog` channel can never have two subscribers cross-contaminating slices — independent of the UI lock. The rejection writes to no slice (a UI-unreachable backstop; surfacing an error would be a phantom failure on an untouched runtime)
- `installBusy` prop (unified across all four install-card prop types) locks a card's controls — without showing its own "Installing…" state — while another runtime installs, including one still finishing on a just-switched-away framework during an async framework switch. The onboarding automatic-install and check buttons lock on `anyInstalling`, not just this runtime's `isInstalling`
- Post-install reconcile (`getSettings` + `refreshPreflight`) is now best-effort in its own `try`: a transient reconcile error no longer relabels a successful install as failed
- `clearInstallLogs(runtime?)` clears only the transient logs/progress/error and preserves `isInstalling`, so it can't drop the install lock mid-install

## Testing

- Five new store tests: (1) install events stream into the installing runtime's slice only while the others stay pristine; (2) a failure records on the right slice alone; (3) a post-install reconcile throw does not relabel a successful install as failed; (4) a second concurrent install is refused so subscriptions can't cross-contaminate; (5) `clearInstallLogs` clears transient fields but preserves the install lock
- New `EnvironmentSetupCard.render.test.tsx`: the automatic-install button is disabled (and keeps its idle label) while another runtime installs, and enabled when idle
- Updated the two `OnboardingWizard` render tests for the per-runtime `installStates` store shape
- Full suite green: **4718 passed, 66 skipped**
- `typecheck:web`, `typecheck:node`, and `lint` clean on all changed files

## Known follow-up (non-blocking)

Install events are attributed to the sole live subscription rather than filtered by `installId`. Correct today because the atomic guard enforces exactly one subscriber; routing by `installId` would require tagging the main-process broadcast payload and is tracked as a separate follow-up.

### DIFF
--- a/src/renderer/src/pages/onboarding/EnvironmentSetupCard.render.test.tsx
+++ b/src/renderer/src/pages/onboarding/EnvironmentSetupCard.render.test.tsx
@@ -1,0 +1,85 @@
+// @vitest-environment jsdom
+import { act } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { EnvironmentSetupCard } from './EnvironmentSetupCard'
+import type { EnvironmentCheckResult } from '../../../../shared/settings'
+
+// A not-ready, auto-installable environment so the "Install missing runtime" button renders.
+const environment: EnvironmentCheckResult = {
+  checkedAt: 1,
+  platform: 'darwin',
+  architecture: 'arm64',
+  checks: [],
+  ready: false,
+  canAutoInstall: true,
+  agentFrameworkId: 'claude-code',
+  runtime: { found: false }
+}
+
+let container: HTMLDivElement
+let root: Root
+
+beforeEach(() => {
+  container = document.createElement('div')
+  document.body.appendChild(container)
+  root = createRoot(container)
+})
+
+afterEach(() => {
+  act(() => root.unmount())
+  container.remove()
+})
+
+const installButton = (): HTMLButtonElement | undefined =>
+  Array.from(container.querySelectorAll('button')).find((b) =>
+    b.textContent?.includes('Install missing runtime')
+  ) as HTMLButtonElement | undefined
+
+describe('EnvironmentSetupCard install button lock', () => {
+  it('disables the install button while ANOTHER runtime installs (installBusy), without showing this runtime as installing', () => {
+    const onInstall = vi.fn()
+    act(() => {
+      root.render(
+        <EnvironmentSetupCard
+          environment={environment}
+          isChecking={false}
+          // This runtime is idle; a different runtime's install is in flight.
+          isInstalling={false}
+          installBusy={true}
+          installLogs={[]}
+          installProgress={null}
+          onCheck={vi.fn()}
+          onInstall={onInstall}
+        />
+      )
+    })
+
+    const button = installButton()
+    // Locked (can't start a second install that would hit the store guard and surface a phantom failure)…
+    expect(button?.disabled).toBe(true)
+    // …but still shows the idle label, not this runtime's "Installing…" state.
+    expect(button?.textContent).toContain('Install missing runtime')
+    expect(button?.textContent).not.toContain('Installing…')
+  })
+
+  it('leaves the install button enabled when no runtime is installing', () => {
+    act(() => {
+      root.render(
+        <EnvironmentSetupCard
+          environment={environment}
+          isChecking={false}
+          isInstalling={false}
+          installBusy={false}
+          installLogs={[]}
+          installProgress={null}
+          onCheck={vi.fn()}
+          onInstall={vi.fn()}
+        />
+      )
+    })
+
+    expect(installButton()?.disabled).toBe(false)
+  })
+})
--- a/src/renderer/src/pages/onboarding/EnvironmentSetupCard.tsx
+++ b/src/renderer/src/pages/onboarding/EnvironmentSetupCard.tsx
@@ -24,7 +24,11 @@ import { describeInstallProgress } from '../settings/claude-install-progress'
 type EnvironmentSetupCardProps = {
   environment: EnvironmentCheckResult | undefined
   isChecking: boolean
+  // `isInstalling` drives this card's own progress/label (the selected runtime's install). `busy` locks
+  // the action buttons while ANY runtime installs — including one still finishing on the previously
+  // selected framework during an async switch — so a second install can't start. Defaults to isInstalling.
   isInstalling: boolean
+  busy?: boolean
   installLogs: string[]
   installProgress?: ClaudeInstallProgressEvent | null
   error?: string
@@ -148,12 +152,16 @@ const EnvironmentSetupCard = ({
   environment,
   isChecking,
   isInstalling,
+  busy,
   installLogs,
   installProgress,
   error,
   onCheck,
   onInstall
 }: EnvironmentSetupCardProps): React.JSX.Element => {
+  // Any install running (this runtime's or one finishing on a just-switched-away framework) locks the
+  // buttons; default to this card's own install when the caller doesn't pass a global signal.
+  const anyInstalling = busy ?? isInstalling
   const structuredProgress = installProgress ? describeInstallProgress(installProgress) : undefined
   const progress =
     structuredProgress?.fraction !== undefined
@@ -174,7 +182,7 @@ const EnvironmentSetupCard = ({
           variant="outline"
           size="sm"
           onClick={onCheck}
-          disabled={isChecking || isInstalling}
+          disabled={isChecking || anyInstalling}
         >
           <RefreshCw className={cn(isChecking && 'animate-spin')} aria-hidden="true" />
           {isChecking ? 'Checking…' : 'Check again'}
--- a/src/renderer/src/pages/onboarding/EnvironmentSetupCard.tsx
+++ b/src/renderer/src/pages/onboarding/EnvironmentSetupCard.tsx
@@ -240,7 +240,11 @@ const EnvironmentSetupCard = ({
               <Button
                 type="button"
                 onClick={onInstall}
-                disabled={isInstalling || isChecking}
+                // Lock on anyInstalling (not just this runtime's isInstalling): while another runtime
+                // installs, clicking here would hit the store's single-install guard and surface its
+                // rejection as a phantom failure on this untouched runtime. Label still keys off
+                // isInstalling so only THIS runtime's install shows the "Installing…" state.
+                disabled={anyInstalling || isChecking}
                 className="shrink-0"
               >
                 {isInstalling ? (
--- a/src/renderer/src/pages/onboarding/EnvironmentSetupCard.tsx
+++ b/src/renderer/src/pages/onboarding/EnvironmentSetupCard.tsx
@@ -28,7 +28,7 @@ type EnvironmentSetupCardProps = {
   // the action buttons while ANY runtime installs — including one still finishing on the previously
   // selected framework during an async switch — so a second install can't start. Defaults to isInstalling.
   isInstalling: boolean
-  busy?: boolean
+  installBusy?: boolean
   installLogs: string[]
   installProgress?: ClaudeInstallProgressEvent | null
   error?: string
@@ -152,7 +152,7 @@ const EnvironmentSetupCard = ({
   environment,
   isChecking,
   isInstalling,
-  busy,
+  installBusy,
   installLogs,
   installProgress,
   error,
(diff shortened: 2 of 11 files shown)


ewen-poch:
## Round 1 — addressing review feedback

Thanks for the reviews. Summary of what I changed and what I deliberately left.

### Adopted — Codex [P2]: onboarding install controls can re-enable mid-switch

The race is real: during an async framework switch, `activeInstall` follows the newly-selected (idle) runtime, so an install still finishing on the previous framework would re-enable the new runtime's install controls and allow a second install to subscribe to the shared `onInstallLog` channel.

I fixed this at two levels:

1. **Atomic guard in the store (the real fix).** `runRuntimeInstall` now refuses to start a second install when any runtime is already installing:
   ```ts
   if (selectAnyInstalling(get())) {
     return { installId: '', ok: false, error: 'Another install is already in progress.' }
   }
   ```
   The check-and-set is synchronous (no `await` between them), so it's atomic against the single-threaded event loop — two callers can't both pass. This enforces the single-install invariant in the store itself, so even a stray caller or a mid-switch UI race can never create a second subscriber. This also resolves the Claude review's "isolation rests on the UI lock, not the channel" observation.

2. **Display vs. lock split in onboarding (defense-in-depth).** `activeInstall` still drives the displayed progress/label, but the action buttons now lock on `anyInstalling` via new `busy`/`installBusy` props on `EnvironmentSetupCard` and the manual runtime cards. So the controls stay disabled while an install finishes on a just-switched-away framework.

New test `refuses a second concurrent install so subscriptions can never cross-contaminate` covers the guard: main is never asked to install and the second runtime's slice stays pristine.

### Deferred — Claude review follow-up: route events by `installId`

Tagging install-event payloads with the runtime so the store can route by identity (instead of relying on one live subscription) is a nice robustness improvement, but it requires a main-process change to the broadcast payload and is explicitly flagged as non-blocking. The atomic guard already guarantees a single subscriber, so I'm tracking this as a separate follow-up rather than expanding this PR's surface.

### Skipped — naming nit (`busy` vs `installBusy`)

`ClaudeInstallCard` uses `busy` and the status cards use `installBusy`. I kept them distinct on purpose: `ClaudeInstallCard` is a leaf control (its `busy` combines with its own `isInstalling`), while the status cards forward a framework-level `installBusy` down to it. Unifying the name would be churn without a correctness or clarity gain here.

### Also fixed
The `Validate commit messages` check was failing because the commit subject contained `(#278)` — the validator disallows `#`. Reworded the subject and moved the issue reference into the body.

Full suite green (4714 passed, 66 skipped); `typecheck` and `lint` clean.

ewen-poch:
## Round 2 — addressing the latest architecture review

The re-run architecture review is **mergeable**, with two Low, non-blocking notes.

### Adopted — Low #1: concurrency-guard rejection is invisible

Documented the guard as a UI-unreachable safety backstop, per the reviewer's own suggested option. I kept it intentionally silent rather than surfacing an error, and the comment now explains why: writing an error to a slice here would surface a *phantom failure* on a runtime the user never touched — the inverse of the #278 bug this PR fixes. If a real UI trigger for this path is ever added, the comment notes to surface the error on the target slice then.

### Not changing — Low #2: route events by `installId`

Same follow-up from round 1. The reviewer confirms it's non-blocking ("No change required to merge") and that the atomic guard makes attribution correct today. Filtering by `installId` requires main to tag the broadcast payload; I'm keeping that as a separate follow-up rather than growing this renderer-only PR into the main/IPC layer.

Note: the Codex correctness check shows **skipped** on this push (it did not re-run); its earlier `needs changes` was against the pre-round-1 commit and is resolved by the atomic guard + display/lock split. `Validate commit messages` now passes.

This is a comment-only commit; suite and typecheck remain green.

ewen-poch:
## Round 3 — addressing the latest architecture review

Re-run verdict is **mergeable** again. It surfaced one genuinely new correctness note and escalated the naming note; I adopted both (both Low, both cheap).

### Adopted — post-install reconcile failure was reported as an install failure

Good catch. On the success path, `getSettings()` + `refreshPreflight()` ran inside the same `try` as `invoke()`, so a reconcile throw *after* a successful install fell into the `catch`, set `installError`, and rethrew — a phantom "install failed" on a runtime that actually installed.

Fix separates the two concerns:
- `invoke()` has its own inner `try`; only an install failure (throw or non-ok result) sets `installError`.
- The install outcome is recorded from `result` *before* the reconcile.
- The reconcile is now best-effort in its own `try` — a transient snapshot/preflight error leaves the card briefly stale (corrected on the next detect/refresh) instead of overwriting a good result.

New test `does not relabel a successful install as failed when the post-install reconcile throws` locks this in (getSettings rejects once; the call still resolves `{ ok: true }` and `installError` stays clear).

### Adopted — unify the lock prop name

Renamed `busy` → `installBusy` on `ClaudeInstallCard` and `EnvironmentSetupCard`, so all four install-card prop types now use `installBusy` consistently and the "silently disabled lock" mis-wire risk goes away. Updated all call sites (SettingsPage, both status cards' forwarding, OnboardingWizard).

### Still deferred — route events by `installId`

Unchanged from rounds 1–2; the reviewer continues to mark it non-blocking, and it needs a main/IPC payload change I'm keeping out of this renderer-only PR. Tracking as a follow-up.

Full suite green (4715 passed, 66 skipped); typecheck and lint clean on changed files.

---
Per the agreed review budget, this is round 3 — I'm pausing here. All AI findings are now either addressed or explicitly deferred with the reviewer's non-blocking agreement, and the PR is in a mergeable state.

ewen-poch:
## Fix — automatic install button not using the global lock (code, not description)

Verified against the code: this is a real bug in the **code**, not a description mismatch. The PR already promises "lock controls while another runtime installs", so the code was the thing out of line.

The chain was exactly as described:
- `EnvironmentSetupCard.tsx:243` disabled on `isInstalling` (this runtime's slice) instead of `anyInstalling` — which the component already computes and which the sibling *check* button already uses.
- While another runtime installed, the button stayed clickable → `handleInstall('managed')` → the store's single-install guard returns `{ ok: false, error: 'Another install is already in progress.' }` → `OnboardingWizard.tsx:304-305` turns that into `automaticInstallError`, rendered as this runtime's card error = a phantom failure on an untouched runtime.

I fixed the check button in round 1 but missed the install button directly below it — thanks for catching the gap.

**Fix:** lock the button on `anyInstalling || isChecking`; the label/spinner still key off `isInstalling` so only *this* runtime's install shows "Installing…".

**Test:** added `EnvironmentSetupCard.render.test.tsx` — asserts the button is disabled and keeps its idle label while another runtime installs (`installBusy=true, isInstalling=false`), and is enabled when idle.

Full suite green (4717 passed, 66 skipped); typecheck and lint clean on changed files.

ewen-poch:
## Fixes — clearInstallLogs lock + stale PR description

**[P2] clearInstallLogs could drop the install lock — fixed (code).**
Confirmed: it replaced the whole slice via `createInitialRuntimeInstallState()`, which flips `isInstalling` to `false`, so calling it mid-install would drop `selectAnyInstalling` and let a second install start. No caller does this today (low reachability), but it's a real API-safety gap and it contradicted the action's own doc comment ("clears the transient logs/progress/error"). Now it clears only logs/progress/error and preserves `isInstalling`. Added a store test asserting the lock (and `selectAnyInstalling`) survive a `clearInstallLogs` call.

**[P2] PR description out of date — fixed (description).**
Updated the body to match the current diff:
- Testing now reads **five** new store tests (isolation, failure attribution, reconcile-failure, concurrent-install guard, clearInstallLogs lock) plus `EnvironmentSetupCard.render.test.tsx` and the two updated `OnboardingWizard` render tests; count bumped to **4718 passed**.
- Fix section now says `installBusy` (unified name), and documents the atomic guard, the best-effort reconcile, the onboarding button lock on `anyInstalling`, and the `clearInstallLogs` behavior — so scope + verification are accurate per CONTRIBUTING.md.
- Added a "Known follow-up (non-blocking)" note for the `installId` routing item.

Thanks — the EnvironmentSetupCard fix from the prior round is confirmed correct on your side too.

Full suite green (4718 passed, 66 skipped); typecheck and lint clean on changed files.

Issue #284, opened 2026-07-22
[Bug]: Custom Gateway rejects valid API credentials during connection test


FerryLF:
### Preflight checklist

- [x] I searched existing issues and this has not been reported yet.
- [x] I'm running a recent version of Open Science.

## Bug Description

When configuring a custom OpenAI-compatible gateway, the connection test fails with:

> Test failed: authentication rejected — check the API key.

However, the Base URL and API key have been verified multiple times, and the same credentials work correctly in WorkBuddy.

## Configuration

- Provider type: Custom Gateway
- API format: Chat Completions API (`/v1/chat/completions`)
- Model: `deepseek-v4-flash`
- OpenCode status: Installed and active
- OpenCode version: 1.18.3

<img width="738" height="504" alt="Image" src="https://github.com/user-attachments/assets/4b3d1a28-ecef-4878-84a6-69e92f95bb2b" />

<img width="467" height="121" alt="Image" src="https://github.com/user-attachments/assets/5ab31e04-2dae-4d84-b9f1-6394327a7d44" />

<img width="401" height="113" alt="Image" src="https://github.com/user-attachments/assets/bfc8cf60-2575-4688-aef1-2eb75dcf04be" />

## Expected Result

Open Science should authenticate successfully using the valid gateway URL and API key.

## Additional Information

Other users have reportedly encountered similar installation or connection failures. Screenshots are attached.

### Operating system

Windows

### App version

0.5.0

### Relevant logs or screenshots

```shell

```

ewen-poch:
Thanks for the detailed report and the screenshots, @FerryLF — the config breakdown (Custom Gateway · Chat Completions · OpenCode active) made this much faster to trace. 🙏

## What we found

The connection test was **framework-blind**: it probed every provider against a fixed endpoint order and never checked whether the provider's API format matched the *active* agent framework. That check lived elsewhere (the model picker and the spawn path). So when a provider and the active framework didn't line up, the probe hit the wrong endpoint and the result was reported as `authentication rejected — check the API key`, even when the key was perfectly valid. In other words, the message blamed the credential when the real issue was the pairing.

## What we're going to do

1. **Make the test framework-aware.** Before any network call, we check whether the provider can actually drive the active framework. If it can't, we skip the probe and show a precise reason instead of a false auth error — e.g. *"Not compatible with Claude Code: it needs `/v1/messages`, but this provider speaks `/v1/chat/completions`. Change the API format or switch the agent framework."*
[...]

PR #292, opened 2026-07-22
fix(settings): make provider connection test framework-aware

ewen-poch:
Refs #284

## Summary

The **Test connection** flow was framework-blind. It probed every provider against a fixed endpoint preference (`anthropic → openai → responses`), independent of the active agent framework. Whether a provider can actually be *used* is decided by a separate check (`isProviderUsableByFramework`) that previously only surfaced in the renderer and at spawn time.

As a result, when the active framework couldn't drive a provider (e.g. **Claude Code active + an OpenAI/Chat-Completions-only gateway**), the raw probe hit the wrong endpoint and the failure was reported as:

> Test failed: authentication rejected — check the API key.

…even though the key was valid. The pairing, not the credential, was the problem.

## What this changes

- Decide framework compatibility **before any network call**. If the provider's API format can't drive the *active* framework, short-circuit with a new `incompatible` validation category whose message names the route the framework needs vs. the one the provider speaks (e.g. *"Not compatible with Claude Code: it needs /v1/messages, but this provider speaks /v1/chat/completions."*). No misleading auth probe is fired.
- Compatible providers probe exactly as before. `codex-subscription` keeps its own login-status branch.
- Allow the new category through the repository sanitizer so the failure is recorded on the provider card and clears on a passing re-test after switching framework.

## Scope note

This fixes the **misleading-error class** and makes the test framework-aware. It does **not** by itself explain the specific report in #284 (OpenCode + Chat Completions is a *compatible* pairing, so the probe still runs). See the issue thread for follow-up on the raw 401/403 case.

## Testing

- `npm run typecheck` — clean
- `npm run lint` — clean on changed files
- `npx vitest run` — full suite green except 2 pre-existing, unrelated `build/packaging.test.ts` failures (NSIS template reads from `node_modules`)
- Added unit coverage: service short-circuit (no fetch on incompatible), probes-after-framework-switch, renderer message maps, repository round-trip of the new category

### DIFF
--- a/src/main/settings/repository.test.ts
+++ b/src/main/settings/repository.test.ts
@@ -234,6 +234,28 @@ describe('settings repository', () => {
     })
   })
 
+  it('round-trips an incompatible validation failure across a reload', async () => {
+    const root = await createStorageRoot()
+    const repository = new SettingsRepository(root)
+
+    await repository.upsertProvider(
+      provider({
+        lastValidationFailure: {
+          at: 1717000000000,
+          category: 'incompatible',
+          message: 'Not compatible with Claude Code: it needs /v1/messages.'
+        }
+      })
+    )
+
+    const reloaded = await new SettingsRepository(root).getSettings()
+    expect(reloaded.providers[0].lastValidationFailure).toEqual({
+      at: 1717000000000,
+      category: 'incompatible',
+      message: 'Not compatible with Claude Code: it needs /v1/messages.'
+    })
+  })
+
   it('drops a malformed validation failure (bad category or missing timestamp) on load', async () => {
     const root = await createStorageRoot()
 
--- a/src/main/settings/repository.ts
+++ b/src/main/settings/repository.ts
@@ -48,6 +48,7 @@ const VALIDATION_CATEGORIES = new Set<ValidationCategory>([
   'model-not-found',
   'bad-url',
   'timeout',
+  'incompatible',
   'unknown'
 ])
 
--- a/src/main/settings/service.test.ts
+++ b/src/main/settings/service.test.ts
@@ -671,6 +671,37 @@ describe('SettingsService: validation', () => {
 
     expect(result).toMatchObject({ ok: true, category: 'ok' })
     expect(fetchMock).toHaveBeenCalledOnce()
+    expect(fetchMock.mock.calls[0][0]).toContain('/v1/chat/completions')
+  })
+
+  it('probes the route the active framework drives for a multi-route provider', async () => {
+    const service = createService()
+    const fetchMock = vi.fn().mockResolvedValue({ status: 200 })
+    vi.stubGlobal('fetch', fetchMock)
+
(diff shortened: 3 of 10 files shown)


ewen-poch:
Thanks for the reviews. Round 1 response — pushed 60e9c61.

## Codex Correctness [P1] — route the probe to the active framework ✅ adopted (partially)

You're right that a multi-route provider (e.g. a vendor exposing both `/v1/messages` and `/v1/chat/completions`) was probed on a fixed global preference, so a passing test could prove a route the active framework never runs. Fixed: `buildValidationRequest` now takes the active framework's `supportedApiTypes` and probes the shared route. Claude Code + a dual-route provider now correctly hits `/v1/messages` instead of `/v1/chat/completions`. Added unit coverage (`validate.test.ts`) and service coverage (Claude Code → `/v1/messages`, OpenCode → `/v1/chat/completions` for the same provider).

**Not adopted — forcing the bridge function-tool probe for Codex.** This conflicts with a deliberate, tested design decision already in the codebase: under Codex the probe is intentionally a *plain connectivity check*, because bridge compatibility is a **static property of the endpoint**, and re-checking it per model "would only add latency and flakiness" (see the existing `runs a plain connectivity probe under Codex without a per-model capability check` test). For the same reason Codex is excluded from the route-alignment change: it bridges the provider's OpenAI route under its `responses` protocol, so the HTTP route is chosen by the bridge, not by `supportedApiTypes`. Changing this would regress intended behavior, so I've left the Codex path as-is.

## Claude Architecture (Low) — codex-subscription bypasses the gate ⏸️ not adopted

I looked at this closely. Running the incompatibility gate for codex-subscription is a one-liner, but it regresses tested, intentional behavior: `validateProvider` for a codex-subscription card is a read-only *login-status* check that is meaningful regardless of the active framework (`validates shared and isolated subscriptions through read-only status checks`). The test-vs-spawn divergence here is low-impact (the model picker already filters it), so I've left it rather than break the documented status-check contract.

## Claude Architecture (Low) — two divergent "not compatible" messages 🚫 declined

Intentional. The spawn-time guard points the user at *Settings → Model* (a different surface, mid-session), while the test-time message names the route mismatch inline. Sharing one string would make one of the two surfaces read wrong. Verdict was `mergeable`; low value vs. the coupling it introduces.

Full suite green except 2 pre-existing, unrelated `build/packaging.test.ts` failures (NSIS template read from `node_modules`).

ewen-poch:
Round 2: the architecture review now returns **mergeable** with no architectural or integration issues, and confirms the `frameworkEndpoints` change closes the real false-pass (a dual-route provider's broken Anthropic route no longer passing on the OpenAI probe). The Codex correctness review didn't re-run this round.

No further code changes. The two Low items I declined earlier (codex-subscription gate; shared spawn/test message text) stand for the documented reasons above. Stopping the review-iteration loop here — ready for a human look.

Issue #293, opened 2026-07-22
[Feature]: Standard .ipynb interoperability — export session notebooks to nbformat (and later import / open in JupyterLab)


imjszhang:
## Preflight checklist

- [x] I searched existing issues and discussions and this hasn't been proposed yet.

## What problem does this solve?

The notebook runtime persists each session as an append-only `run.json`
(`NotebookRunDocument` / `NotebookRunRecord[]`) — great for durability and crash
recovery, but the results are locked inside the app:

- A finished analysis can't be handed to a collaborator who uses JupyterLab / VS Code,
  pushed to a paper repository, or rendered on GitHub / nbviewer.
- The Session Notebook viewer already shows a **disabled "Download as .ipynb" button**
  ("Notebook export is coming soon", added in #130), so the need is acknowledged in the
  UI but not tracked as an issue yet.
- Competing open workbenches in this category treat real `.ipynb` files + "Open in
  JupyterLab" as a headline feature; for researchers whose lab workflow is already
  Jupyter-based, the lack of interop is a real migration cost *away* from Open Science.

The good news: `NotebookRunRecord` is semantically almost isomorphic to an nbformat
code cell (`script` ≈ `source`, `outputs` ≈ `outputs`, `executionCount` ≈
`execution_count`), so this is a projection problem, not a storage redesign.

## Proposed solution

A three-stage plan. Each stage is independently shippable; stage 1 alone activates the
existing disabled button.

### Stage 1 — one-way export (`run.json` → `.ipynb`, nbformat 4.5)

A pure projection function in the main process. No change to existing storage;
`run.json` stays the single source of truth.

Field mapping:

| `NotebookRunRecord` | nbformat cell | Notes |
| --- | --- | --- |
| `script` | `source` (split into lines) | direct |
| `kernelKind: python` / `r` | `code` cell | direct |
| `kernelKind: repl` / `bash` | `code` cell with a shell marker (e.g. `%%bash`) + metadata tag | control-plane runs downgraded to shell cells |
| `executionCount` | `execution_count` | direct |
| `text.stdout` / `stderr` | `stream` output | direct |
| `outputs` (images/tables) | `display_data` with proper mimetypes (`image/png`, `text/html`, …) | artifact files read back and inlined as base64 |
| `text.traceback` | `error` output | direct |
| `runId`, `startedAt`, `environment` | `metadata.open_science.*` | custom namespace; Jupyter tools preserve unknown metadata |

[...]

PR #297, opened 2026-07-22
feat(notebook): export session run history as .ipynb

ewen-poch:
Refs #293 — implements **Stage 1 (one-way export, `run.json` → `.ipynb`)** of the three-stage plan proposed there. Stages 2 (import / "Open in JupyterLab") and 3 (mirror-on-run) are intentionally left for follow-ups.

## Problem

A finished notebook session is locked inside the app: `run.json` is durable but can't be handed to a JupyterLab/VS Code collaborator, pushed to a paper repository, or rendered on GitHub/nbviewer. The Session Notebook viewer has acknowledged this with a disabled "Download as .ipynb" button ("coming soon") since #130.

## Proposed change

A pure projection in the main process — `run.json` stays the single source of truth, no storage change, no Python/`nbconvert` dependency (nbformat is plain JSON, generated by hand):

```mermaid
flowchart LR
    A["SessionNotebookDialog<br/>.ipynb button"] -->|"notebook:export-ipynb"| B["NotebookRuntimeService<br/>.exportIpynb"]
    B -->|"findExisting (read-only)"| C["NotebookRunRepository<br/>run.json"]
    B --> D["projectRunDocumentToIpynb<br/>(pure, deterministic)"]
    D -->|"NotebookIpynbExport json"| A
    A -->|"file:save-blob"| E["Save As dialog<br/>(.ipynb filter)"]
```

- `src/main/notebook/ipynb-export.ts` (new): `projectRunDocumentToIpynb` maps each `NotebookRunRecord` to an nbformat 4.5 code cell — `script` → `source`, `outputs` → stream/error/display_data (images are already base64 inline), `executionCount` → `execution_count`, and every cell gets `metadata.open_science` (`runId`, kernel, status, ISO timestamps) as the provenance anchor anticipated by the roadmap. Structured `outputs` are primary; runs without them fall back to the flattened `text.stdout/stderr/traceback`.
- Kernelspec comes from the dominant analysis kernel (`python3` / `ir`); `.ipynb` allows only one, so foreign-kernel cells keep their identity in `metadata.open_science.kernel`.
- **Deliberate deviation from the issue's mapping table**: `bash` runs downgrade to `%%bash` cells, but `repl` runs are JavaScript (the control-plane SDK), so they export verbatim with the kernel metadata tag — a `%%bash` marker over JS would be wrong.
- The projection is deterministic and privacy-safe: no absolute paths (`workspaceCwd`, session roots) and no wall-clock export timestamp, so the same `run.json` yields byte-identical output.
- `NotebookRuntimeService.exportIpynb` reads via `repository.findExisting` — read-only, never creates `run.json` — exposed as the new `notebook:export-ipynb` IPC channel and `window.api.notebook.exportIpynb` bridge (also registered in the generated web API map).
- The footer button in `SessionNotebookDialog` is now active (disabled only when empty/loading/export-in-flight), and saves through the existing `file:save-blob` flow with a new `.ipynb` dialog filter; failures surface inline next to the button.

## Scope and non-goals

- One-way export only: no `.ipynb` import, no JupyterLab launcher, no mirror-on-run (stages 2–3).
- No split-by-kernel export mode; mixed sessions carry per-cell kernel tags instead.
- No nbformat JSON-Schema dependency added; the emitted subset is asserted structurally in tests.
- `run.json` format, kernels, and the run lifecycle are untouched.

- [x] The previously disabled ".ipynb" button exports the session's run history to a valid nbformat 4.5 file chosen via Save As.
- [x] Exported notebooks render out of the box (kernelspec + language_info set) and trace every cell back to its `runId`.
- [x] Export is idempotent and free of absolute paths.
- [x] `npm run lint` — green (0 errors; prettier-clean for all touched files).
- [x] `npm run test` — green (4779 passed, 66 skipped; skips are the pre-existing opt-in integration/smoke suites).
- [ ] `npm run typecheck` — not run locally yet (deferred for now); will run and address anything that surfaces.

Tests added/updated:

- `src/main/notebook/ipynb-export.test.ts` (new): cell/output mapping, `%%bash` downgrade, repl handling, dominant-kernel kernelspec, cell-id sanitizing/uniqueness, determinism + no-path-leak, and a service-level test through the real `NotebookRunRepository` (incl. the null/no-creation path).
- `src/main/notebook/ipc.test.ts`: new channel in the exact-channel assertion + delegation.
- `src/main/file-save.test.ts`: `.ipynb` filter for the `application/x-ipynb+json` mime.
- `SessionNotebookDialog.render.test.tsx`: enabled/disabled/saving/error states of the export button.
- `src/renderer/web/api-map.generated.ts`: regenerated for the new `notebook:export-ipynb` channel.

## Review focus

- The repl-vs-`%%bash` deviation described above.
- The `metadata.open_science` field set — this is the contract a future import/provenance stage will hang off.
- Determinism choices (indent 1 + trailing newline, ISO timestamps only from run data).

### DIFF
--- a/src/main/acp/ipc.ts
+++ b/src/main/acp/ipc.ts
@@ -207,6 +207,8 @@ const createDefaultNotebookRuntimeService = (): NotebookRuntimeService => {
     // Region default for manage_packages when no mirror is configured; the configured override is
     // wired in main/ipc.ts via setPackageMirrorResolver once the settings service is constructed.
     locale: app.getLocale(),
+    // Stamped into .ipynb exports so a shared notebook records which app version produced it.
+    appVersion: app.getVersion(),
     callbacks: {
       onNotebookAvailable: (event) => broadcast('notebook:available', event),
       onNotebookChanged: (event) => broadcast('notebook:changed', event)
--- a/src/main/file-save.test.ts
+++ b/src/main/file-save.test.ts
@@ -33,6 +33,35 @@ describe('file save IPC handlers', () => {
     expect(handlers.has('file:save-managed')).toBe(true)
   })
 
+  it('saves a blob with an .ipynb filter for the notebook mime type', async () => {
+    const root = await mkdtemp(join(tmpdir(), 'open-science-save-blob-'))
+    const destinationPath = join(root, 'notebook-134d5d81.ipynb')
+    showSaveDialog.mockResolvedValue({ canceled: false, filePath: destinationPath })
+    registerFileSaveHandlers()
+
+    try {
+      const result = await handlers.get('file:save-blob')!(
+        { sender: {} },
+        {
+          suggestedName: 'notebook-134d5d81.ipynb',
+          mimeType: 'application/x-ipynb+json',
+          data: new TextEncoder().encode('{"nbformat":4}').buffer
+        }
+      )
+
+      expect(showSaveDialog).toHaveBeenCalledWith(
+        expect.objectContaining({
+          defaultPath: 'notebook-134d5d81.ipynb',
+          filters: [{ name: 'IPYNB', extensions: ['ipynb'] }]
+        })
+      )
+      expect(result).toEqual({ saved: true, filePath: destinationPath })
+      await expect(readFile(destinationPath, 'utf8')).resolves.toBe('{"nbformat":4}')
+    } finally {
+      await rm(root, { recursive: true, force: true })
+    }
+  })
+
   it('opens a managed source once and copies that exact file to the selected destination', async () => {
     const resolveManagedFilePath = vi.fn().mockResolvedValue('/managed/canonical-report.csv')
     const copyTo = vi.fn().mockResolvedValue(undefined)
--- a/src/main/file-save.ts
+++ b/src/main/file-save.ts
@@ -85,6 +85,8 @@ const extensionForMime = (mimeType: string): string | undefined => {
       return 'tsv'
     case 'text/markdown':
       return 'md'
+    case 'application/x-ipynb+json':
+      return 'ipynb'
     default:
       return undefined
   }
--- a/src/main/notebook/ipc.test.ts
+++ b/src/main/notebook/ipc.test.ts
@@ -24,6 +24,7 @@ describe('notebook IPC handlers', () => {
     const service = {
       state: vi.fn().mockResolvedValue({ sessionId: 'session-1', cells: [] }),
       getSessionReference: vi.fn().mockResolvedValue({ sessionId: 'session-1' }),
+      exportIpynb: vi.fn().mockResolvedValue({ suggestedName: 'notebook-s1.ipynb', json: '{}' }),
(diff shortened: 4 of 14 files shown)


ewen-poch:
Thanks for the reviews! Addressed the actionable findings in f963e68 and c2e1440.

**Adopted**
- **Cell id 64-char limit (Codex P2):** `toCellId` now caps ids at 64 characters per nbformat 4.5 / JEP 62, reserving room for dedup suffixes out of the same budget. Added tests for oversized and duplicated oversized ids.
- **Export error state leaking across session switches (Claude, minor):** the dialog is mounted once and the `session` prop swaps in place, so `ipynbExportStatus` now resets to `idle` in the session-change effect alongside the other resets.
- **Silent export failures (Claude, low):** the catch path now logs via `console.error` before showing the generic banner, so a genuine main-side projection failure leaves a diagnostic trail.
- **Determinism wording (Claude, nitpick):** the module comment now states byte-identical output holds for the same document *and options* — `appVersion` is an explicit injected option, never read from the environment inside the projection.

**Already resolved**
- **Web API map (Codex P1):** fixed earlier in 405fc14 (`npm run gen:web-api-map` committed); the generated-map synchronization test passes.

**Kept as is (trade-off)**
- `appVersion` stays in `metadata.open_science`: the provenance value of knowing which app version produced an exported notebook outweighs cross-version byte-stability. If stage 3 (mirror-on-run) later needs diff-friendly output, it can omit the field via the existing option — the projection takes it as an explicit input for exactly that reason.

Validation: `npm run lint` clean; `npm run test` green (4780 passed, 66 skipped — skips are the pre-existing opt-in integration/smoke suites).

ewen-poch:
Round-2 review noted — **mergeable**. No further code changes this round; triaging the two remaining Low notes:

- **In-flight export state across a session switch (Low):** accepted as deferred. The race window is narrow (the native Save As dialog blocks interaction once open, and the session-change effect already resets the footer state). A mounted-session guard around the post-await `setState` calls is a reasonable follow-up if this ever surfaces in practice — not worth the extra complexity here.
- **Mixed-kernel cells not re-runnable under the dominant kernelspec (Low):** explicitly a Stage-1 non-goal, as documented in the PR body. Revisit in Stage 2 with a split-by-kernel export mode or an off-kernel cell marker, per the issue's three-stage plan.

Current state: `npm run lint` and `npm run test` green locally; Verify (macos-14 / ubuntu-latest / windows-latest), CodeQL, and commit-message checks all green in CI.

ewen-poch:
Superseded by #296 (merged), which landed the same Stage-1 export from the issue author's own implementation — thanks @imjszhang!

The review-hardening and spec-alignment points identified here (cell-id dedup within the 64-char nbformat budget, session-switch export-state reset, richer provenance metadata, determinism wording, SVG text encoding, execution-count pass-through, notebook-level environment metadata, a pure two-phase projection boundary, and nbformat schema validation in tests) will be carried into a focused follow-up PR on top of the merged implementation.

PR #298, opened 2026-07-22
fix(settings): open the provider-type dropdown at the top with a framework-matched default

### LINKED ISSUE #294: [Bug]: Provider type dropdown opens with official providers hidden above the viewport
### Preflight checklist

## Bug Description

When opening the **Provider type** dropdown with **Custom Gateway** selected, the list automatically scrolls to the current selection.

As a result, commonly used official providers such as **OpenAI, Anthropic, and DeepSeek** are hidden above the visible area. Users must scroll upward to discover them, which is unintuitive and may make them appear unavailable.

Image

## Steps to Reproduce

1. Open **Model > Add provider**.
2. Select or keep **Custom Gateway** as the provider type.
3. Open the **Provider type** dropdown.

### Operating system

Windows

### App version

0.5.1

### Provider / model

_No response_

```shell

```

ewen-poch:
## Problem

Fixes #294 — in **Model › Add provider**, the provider-type dropdown auto-scrolled to the current selection ("Custom Gateway" by default, near the bottom of the list), hiding the commonly used official vendors (OpenAI, Anthropic, DeepSeek) above the viewport. Users had to scroll up to discover them.

## Proposed change

Keep the existing single-form flow — Add provider opens the configuration form directly — and fix the dropdown itself, in the shared `Select` primitive (`components/ui/select.tsx`):

- New opt-in `scrollToTopOnOpen` prop on `SelectContent`: the viewport resets to the top on open instead of radix scrolling the selected item into view. The provider-type dropdown opts in, so official vendors are visible from the top on every open. (The reset covers the frames right after mount, when radix performs its post-placement `scrollIntoView`; the code notes the radix-ui version it relies on.)
- Radix scroll up/down buttons wired into `SelectContent`: they render only while scrolling is possible in that direction, giving overflowing lists a visible "there is more" affordance.
- The Add provider form now pre-selects the official vendor matching the active agent framework — Claude Code → Anthropic, Codex → OpenAI, OpenCode → DeepSeek (`defaultProviderKindKey` in `provider-form-value.ts`) — instead of always starting at Custom Gateway.

## Scope and non-goals

- Only the provider-type dropdown opts into `scrollToTopOnOpen`; other selects keep the default scroll-to-selected behavior.
- App-wide note: the radix scroll up/down buttons apply to the shared `Select` primitive, so every overflowing select in the app gains the chevron affordance — not just the provider-type dropdown.
- No changes to provider persistence, validation, or the registry. The provider-kind group definitions moved from `ProviderForm.tsx` to `provider-form-value.ts` (shared constants, no behavior change).

- Add provider opens the form directly; opening the **Provider type** dropdown shows the list from the top (OpenAI, Anthropic, DeepSeek visible) regardless of the currently selected type, with scroll buttons indicating more options below.
- Existing render tests cover the Add provider flow (`SettingsPage.render.test.tsx`, `ProviderForm.render.test.tsx`) and pass unchanged in behavior; `npx eslint` on all touched files is clean. `npm run typecheck` / `npm run test` run in CI.

## Review focus

- The `scrollToTopOnOpen` reset timing in `select.tsx` (covers radix's post-placement `scrollIntoView` over the first frames after mount).


ewen-poch:
AI review round 1 — feedback handling:

**Claude architecture review (needs changes)**

- **Medium — stale draft when re-picking the same vendor**: adopted in b631608. `openCreateWithKind` now always re-seeds the draft explicitly (the `modelViewKey` guard only keys on the kind, so a same-kind re-pick used to keep the in-form-switched value), and the create breadcrumb leaf is derived from the live `formValue` via `selectedKindKey` instead of the frozen navigation `kindKey`. Added a regression test that switches the type in-form, returns to the picker, and re-picks the original vendor.
- **Low — `scrollToTopOnOpen` timing coupling**: kept the approach (the reviewer marked it acceptable, and the primary Add-provider path is now the top-aligned picker, which does not rely on it). Added a comment tying the reset to the radix-ui version (`^1.6.1`) so upgrades revisit it.
- **Low — scroll buttons are app-wide**: kept (they render only while scrolling is possible), and called the app-wide scope out in the PR body's "Scope and non-goals" as suggested.

**Codex correctness review**: verdict `mergeable`, no actionable findings — nothing to do.

ewen-poch:
Update: dropped the two-step vendor picker (2cde561). After trying it, the extra step made the add flow feel awkward, so Add provider opens the configuration form directly again — same as before.

The fix for #294 is now carried entirely by the provider-type dropdown:

- it opens scrolled to the top (`scrollToTopOnOpen`), so OpenAI / Anthropic / DeepSeek are always visible first, and
- the shared Select renders radix scroll up/down buttons while more options are available, so the list reads as scrollable.

The earlier `ProviderKindPicker` and its tests are removed; `SettingsPage` navigation/seeding is back to the original single-step create flow. The Medium finding from the architecture review (re-pick re-seed + breadcrumb mirror) is no longer applicable since the picker step is gone. PR body updated to match.

ewen-poch:
AI review round 2 — feedback handling (Claude verdict: `mergeable`):

- **Medium — fragile timeout coupling to radix scroll timing**: not adopted in this PR. The reviewer marked it "not required to merge, but worth a follow-up ticket"; the current triple-reset works and its radix version dependency is documented in the code. Replacing the 150 ms window with an observer adds complexity for a guard the primary flow no longer depends on — better handled as a follow-up if the team wants it.
- **Low — misleading effect dependency array**: adopted in 7f023b8. The reset effect now runs once on mount (`[]` with the flag guard and an explanatory disable for `react-hooks/exhaustive-deps`), matching the "SelectContent mounts fresh on every open" rationale.

Codex correctness review was skipped by CI on the previous head; the latest code is unchanged in behavior since the earlier `mergeable` verdicts.

ewen-poch:
AI review round 3 — feedback handling (Claude verdict: `mergeable`, all required checks green):

- **Low — stale PR title**: adopted. Retitled to `fix(settings): open the provider-type dropdown scrolled to the top` to match the shipped change (the vendor-picker design was dropped earlier).
- **Medium — timed reset coupling to radix internals**: not adopted (same finding as round 2). It remains a documented, version-pinned known fragility; making it event-driven or adding a real-browser e2e assertion is follow-up work outside this PR's scope.
- **Low — app-wide scroll buttons validated on one dropdown**: acknowledged. Radix renders the chevrons only when a select actually overflows, so short selects are visually unchanged; a spot check of other long selects is left to pre-release QA.

Stopping the review loop here per plan (3 rounds). Current state: all CI green, both AI reviewers mergeable, PR ready for human review.

ewen-poch:
Update (6f2df3a): the Add provider form now pre-selects the official vendor matching the active agent framework instead of always defaulting to Custom Gateway:

- Claude Code → Anthropic
- Codex → OpenAI
- OpenCode → DeepSeek

The mapping lives in `provider-form-value.ts` (`defaultProviderKindKey`) and is applied when the create view seeds its draft. Combined with `scrollToTopOnOpen`, the dropdown still opens from the top even though the pre-selected vendor sits near it. Covered by new unit tests (`defaultProviderKindKey`) and render tests asserting the seeded trigger for each framework. Onboarding is unchanged (it keeps its own seeding, including the Codex-specific path).

ewen-poch:
AI review round (after the framework-default change) — feedback handling (Claude verdict: `mergeable`):

- **Low — `defaultProviderKindKey` not exhaustive / untyped keys**: adopted in ccdb775. The switch is now exhaustive over `AgentFrameworkId` (a future framework throws instead of silently mapping to Anthropic), and the return type is `` `official:${OfficialVendorId}` `` so a registry vendor rename fails at compile time.
- **Medium — scroll reset races radix internal timing**: not adopted (recurring known fragility from earlier rounds). It is documented and version-pinned in the code; an event-driven reset or real-browser e2e assertion remains follow-up work.
- **Low — app-wide scroll buttons verified on one dropdown only**: acknowledged. Radix hides the buttons when a list is not scrollable, so short selects are unchanged; broader visual spot-checks are left to pre-release QA.

ewen-poch:
AI review round — feedback handling (Claude verdict: `mergeable`, all checks green on the previous head):

- **Medium — throwing helper reachable from render**: adopted in d469910. The exhaustive `never` switch in `defaultProviderKindKey` keeps its compile-time guarantee but now degrades to `official:anthropic` at runtime if persisted state ever holds a framework id outside the union, instead of throwing during `SettingsPage` render.
- **Low — reset can override a user scroll**: adopted in d469910. The pending rAF/timeout resets now cancel on the first user-initiated input (wheel/touchmove/keydown). A plain `scroll` listener would be tripped by radix's own programmatic scroll — the very thing the resets race — so cancellation keys off direct input only.
- **Low — app-wide scroll buttons**: acknowledged again; unchanged, radix renders them only for overflowing lists.

This is the last review iteration from my side: the PR is green and both AI reviewers report `mergeable`.

PR #299, opened 2026-07-22
fix(acp): recover from cross-framework and opencode session resume failures

ewen-poch:
## Problem
Switching agent framework (e.g. claude-code → opencode) and then resuming a
session created under the other framework surfaced `RequestError: Internal
error` in the UI. opencode rejects the lost session with a JSON-RPC Internal
error (-32603) tagged `{"service":"session"}` and a "OpenCode service failure"
message suffix; the resume-failure detector only matched a bare `Internal
error` message, so the error propagated instead of falling back to a fresh
session.

## Proposed change
- Recognize opencode's machine-readable `data.service === 'session'` marker as
  an authoritative, language-independent unresumable-session signal.
- Prefer structured `errorKind`, and constrain the English detail matcher so
  provider/credential/MCP/model/tool failures stay visible.

## Scope and non-goals
Resume-failure classification only. No change to the cross-framework skip path
or to genuine (non-session) failure propagation.

- New tests: session-service failure adopts a fresh session; non-session
  service (`provider`) stays visible. `vitest` runtime suite: 146 passed.
- eslint clean on changed files.

## Review focus
`isUnresumableSessionError` — that `service === 'session'` cannot swallow real
provider/MCP failures.

### DIFF
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -121,6 +121,9 @@ const startFakeAgent = (
     // Some agents preserve a machine-readable reason in the Internal error data instead of relying on
     // the human-facing detail string.
     resumeInternalErrorData?: unknown
+    // opencode rejects a lost session with an Internal error tagged by the failing service and a
+    // descriptive message suffix (e.g. `{ service: 'session' }` + "OpenCode service failure").
+    resumeServiceFailure?: { service: string; message: string }
     // When true, the agent does NOT advertise session/close capability, so the runtime must fall back to
     // the session/cancel notification on delete instead of a close request.
     supportsClose?: boolean
@@ -212,6 +215,13 @@ const startFakeAgent = (
 
       if (options.resumeInternalErrorData !== undefined) {
         throw acp.RequestError.internalError(options.resumeInternalErrorData)
+      }
+
+      if (options.resumeServiceFailure) {
+        throw acp.RequestError.internalError(
+          { service: options.resumeServiceFailure.service },
+          options.resumeServiceFailure.message
+        )
       }
 
       resumedSessions.push({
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed both AI review findings in `a000546`.

Changes made:

- A machine-readable `data.errorKind` is now authoritative. Only an explicit allowlist of session-loss reasons triggers fresh-session adoption; unknown structured reasons propagate even when their human-readable detail resembles a restore failure.
- The legacy detail matcher now accepts only phrases that directly target a missing session/conversation or a failed resume/restore/reopen/reattach operation. The broad `load`, `context`, and standalone `unknown` combinations were removed.
- Details mentioning authentication, credentials, providers, MCP, models, tools, or servers fail open and remain visible.
- Added regression coverage for the MCP and provider-credential examples from the reviews, an unrelated model-context error, structured provider-error precedence, and a language-independent `session-not-found` reason.

Validation: 8 targeted resume-classification tests passed; ESLint and Prettier passed for the changed files.

Trade-off: the prose fallback remains English-only for legacy agents that provide no structured reason. It is deliberately conservative: a false negative preserves the real error, while a false positive would silently reset agent-side context. New/localized backends should emit a supported `errorKind` instead of relying on prose.

ewen-poch:
Included the verified OpenCode-specific resume contract in `530b3aa` before merge.

OpenCode can report a lost session as JSON-RPC `-32603` with `data.service = "session"` and an `Internal error: OpenCode service failure` message suffix. The previous exact-message gate missed that machine-readable signal. Session-service failures now adopt a fresh underlying session, while provider and other non-session service failures continue to propagate.

Added positive coverage for the OpenCode session-service shape and negative coverage for a provider-service failure. Validation: 10 targeted resume-classification tests passed; ESLint and Prettier passed for the changed files.

PR #300, opened 2026-07-22
feat(notifications): add desktop notifications for task completion and failure

### LINKED ISSUE #287: [Feature]: Add desktop notifications for task completion and failure
### Preflight checklist

Long-running tasks may take several minutes to complete. Users often switch to another application while waiting and may not notice when a task has finished or failed.

This can cause unnecessary waiting and delay follow-up actions.

### Proposed solution

Send a native desktop notification when a task:

- Completes successfully
- Fails or requires user attention

The notification should include the project or task name and its status. Clicking the notification should bring Open Science to the foreground and open the relevant task.

Users should also be able to enable or disable these notifications in Settings.

Image

### Alternatives considered

_No response_

### Additional context

_No response_

ewen-poch:
Closes #287

## Problem

Long-running agent tasks can take several minutes. Users switch to another application while waiting and don't notice when a task finishes, fails, or stalls on an approval — causing unnecessary waiting and delayed follow-up actions.

## Proposed change

Post a native desktop notification while the app is **unfocused** for these task lifecycle moments:

- `stop` events → **"Task completed"** (`end_turn`), or **"Task needs attention"** for early stops (`max_tokens` / `max_turn_requests` / `refusal`). User-cancelled turns stay silent (deliberate action).
- `error` events → **"Task failed"** for genuine prompt failures (discriminated via a shared `ACP_PROMPT_FAILED_EVENT_TITLE` constant). Ancillary session-scoped errors (artifact cleanup, cancel timeout) and recoverable context-overflows (auto-retried by the renderer) stay silent.
- Permission requests → **"Approval needed"** naming the waiting tool — a parked turn otherwise hangs unnoticed until the user returns. Connector approvals (the external data-egress gate, up to five minutes) get the same treatment, naming the connector call and targeting the in-flight turn that triggered it.
- The notification body names the task by the prompt's first line (bounded; bodies are truncated hard on some platforms).
- **Clicking** a notification surfaces the main window (`showMainWindow`) and opens the conversation: main records the target consume-once (`notifications:take-pending-open-session`) and the renderer pulls it once its session store is hydrated, so a click that recreates the window can never lose the navigation. The `notifications:open-session` push is only a nudge for a live renderer.
- New opt-out **Settings → General → Notifications → "Task notifications"** toggle, persisted as `notificationsEnabled` in `settings.json` and read fresh at notification time (no restart needed). Defaults to enabled.

Design notes:

- `src/main/notifications/task-notifications.ts` is Electron-free (delivery/focus/settings injected), so the filtering rules are unit-testable; the Electron wiring lives in `main/ipc.ts` and the click activation in `main/index.ts` (where `showMainWindow` exists). `showMainWindow` now returns the window it owns so the nudge targets it directly.
- Notification eligibility = *a prompt was tracked*, i.e. the turn came through the `acp:send-prompt` IPC handler (user-initiated). Internal turns injected via `runtime.sendPrompt` directly — the reviewer's auditor-correction loop — stay silent by construction.
- Delivery is guarded end to end: `Notification.isSupported()`, focus check, fresh settings read (fail-closed on error), swallowed delivery errors reported via `onDeliveryError`, and `Notification` instances retained until `click`/`close` so handlers can't be GC'd mid-banner.

### Example notifications

Bodies name the task by the prompt's first line (here: "Plot the dose-response curve"). Early-stop reasons are phrased in plain language — the raw ACP stop reasons never reach the user.

| Moment | Title | Body |
| --- | --- | --- |
| Completed | Task completed | "Plot the dose-response curve" finished. |
| Output truncated (`max_tokens`) | Task needs attention | "Plot the dose-response curve" stopped early — the answer hit the model's length limit. |
| Turn cap (`max_turn_requests`) | Task needs attention | "Plot the dose-response curve" paused — send a message to keep it going. |
| Refusal | Task needs attention | "Plot the dose-response curve" was declined by the agent. |
| Failed | Task failed | "Plot the dose-response curve" failed: Rate limit reached. |
| Approval wait | Approval needed | "Plot the dose-response curve" needs your approval: Run command |
| Connector approval | Approval needed | "Plot the dose-response curve" needs your approval: pubchem search compound |

### Platform behavior

- **macOS** — Notification Center via Electron's `Notification`; banners auto-suppress while the app is focused (we additionally gate on focus for deterministic behavior everywhere). Packaged builds are signed/notarized, so banners work in production.
- **Windows** — toast notifications; the app already sets an AppUserModelID (`com.aipoch.open-science`), so toasts are attributed correctly in packaged builds. Click activation is supported.
- **Linux** — delivered through the freedesktop notification daemon (libnotify); guarded with `Notification.isSupported()` so desktops without a daemon degrade to silent. Click support varies by DE. Headless `--serve` deployments never notify: the headless flag is threaded into the wiring and delivery is disabled outright, and click activation stays inert.

## Scope and non-goals

- Only user-facing agent prompt turns and their permission waits. The background reviewer loop stays silent by design; notebook runtime provisioning already has in-app progress UI (could be a follow-up).
- No notification for events while the app is focused.
- No notification-center history/management UI; the toggle is the only setting.

- [x] Unit tests for the event→notification mapping and service gating rules (`src/main/notifications/task-notifications.test.ts`): completion/early-stop/failure/approval mapping, cancelled + recoverable + ancillary-error silence, internal-turn silence, focus gating, disabled preference, settings-read failure (fail-closed), prompt-snippet tracking/forgetting, delivery-error containment, consume-once click target.
- [x] Click-navigation tests: nudge-while-hydrated and consume-after-hydration (`App.test.tsx`), `openSessionById` (`navigation-store.test.ts`).
- [x] Settings chain tests mirroring the `reasoningEffort` pattern: repository persistence + sanitize (`repository.test.ts`), service view projection + setter (`service.test.ts`), IPC validation rejection (`ipc.test.ts`), preload channel pin (`preload/index.test.ts`), renderer store optimistic/revert behavior (`settings-store.test.ts`), toggle render + click (`GeneralPanel.render.test.tsx`).
- [x] CI green on all platforms (typecheck, lint, full test suite on macOS/Ubuntu/Windows).
- [x] Full typecheck, lint, and test suite validated in CI across every commit on this branch.

Manual verification on Windows and Linux desktops is still worthwhile (author machine is macOS): toast attribution/click on Windows, and `isSupported()` behavior across Linux DEs.

## Review focus

- The terminal-event filter rules in `describeTaskNotification` and the approval mapping in `describePermissionNotification` (`src/main/notifications/task-notifications.ts`).
- The tracked-prompt eligibility rule (user-initiated turns only) and the consume-once click-target handoff between main and renderer.
- The `notificationsEnabled` default (enabled) and its placement under Settings → General.


ewen-poch:
Round 1 of AI-review follow-up (ae3ca1c). Thanks to both reviewers — the two converging findings were the most valuable.

**Adopted**

- **Click-navigation races (Claude Medium + Codex P1)** — redesigned from push-with-payload to nudge-and-pull. Main now holds the click target consume-once (`setPendingOpenSession`), and the renderer pulls it via `notifications:take-pending-open-session` either on the nudge (fast path, live renderer) or once session persistence has hydrated (slow path, recreated window). The `did-finish-load` deferral is gone entirely, and `openSessionById` can no longer silently drop a click that arrives mid-hydration. New tests cover both paths (`App.test.tsx`: nudge-while-hydrated, and consume-after-hydration).
- **Window selection (Claude Low)** — `showMainWindow()` now returns the `BrowserWindow` it owns, and the nudge targets that reference directly instead of `getFocusedWindow() ?? getAllWindows()[0]`.
- **False failure notifications (Codex P2)** — only `'Prompt failed'` error events map to "Task failed". Ancillary session-scoped errors (`Artifact cleanup failed`, `Prompt cancellation timed out`) now stay silent, and no longer discard the tracked prompt snippet before the turn's own terminal event. Pinned by tests.
- **Unhandled rejection on delivery (Claude Low)** — `show()` is wrapped; delivery failures are reported through an injected `onDeliveryError` (wired to the file logger) and can never surface as an unhandled rejection on the broadcast path.

**Declined**

- **"Restore framework-aware provider validation" (Codex P2, `service.ts:1607`)** — this PR does not touch provider validation: `git diff origin/main...HEAD -- src/main/settings/service.ts` removes zero lines (the diff is purely additive: the `notificationsEnabled` projection and its getter/setter). The flagged range predates this branch; if it's a real regression it belongs on a separate issue.

**CI** — the `Verify (macos-14)` failure was on us: `App.test.tsx` mocked `window.api` without the new `notifications` surface, and `SettingsPage.render.test.tsx` asserted exactly 4 sections (now 5 with Notifications). Both fixed here.

ewen-poch:
Round 2 status (907809d):

**CI** — the all-platform Verify failure was a typecheck error I introduced in `App.test.tsx`: reading the nudge listener out of `mock.calls` produced an empty-tuple type. The listener is now captured in an explicitly-typed box inside the hoisted mock. Fixed and pushed; CI is re-running.

**Claude architecture review: mergeable** — thanks. On the three non-blocking notes:

- **Title-string discriminator** — fair; will adopt in the next batch: export the `'Prompt failed'` title as a shared constant referenced by both the producer (`runtime.ts`) and the filter, so the coupling is explicit and refactor-safe.
- **`Notification` reference retention** — will adopt: retain the instance in a short-lived set until `click`/`close` fires, so the click handler can't be GC'd mid-banner.
- **Headless/web-serve mode** — declining for this PR. A `--serve` daemon is an explicit CLI deployment, typically on the same user's machine; a stray banner there is harmless and daemon-less hosts are already covered by `Notification.isSupported()`. Gating delivery on `webMode.headless` would cross the `registerIpcHandlers` boundary for a niche case — happy to revisit as a follow-up if it ever bites.

Waiting on the re-triggered Codex review now; any new findings will be batched with the two adoptions above.

ewen-poch:
Round 3 — final review round (6950b91). CI is green on all platforms (typecheck, lint, full test suite on macOS/Ubuntu/Windows).

**Adopted**

- **Reviewer auditor-correction leak (Claude Medium)** — real find, and it contradicted the stated design ("the background reviewer loop stays silent"). Notification eligibility is now *a prompt was tracked*: only turns that came through the `acp:send-prompt` IPC handler (user-initiated) can notify. Turns injected via `runtime.sendPrompt` directly — the reviewer's correction loop — never pass through `trackPrompt`, so their terminal events stay silent by construction. Regression test added for an untracked internal turn on the main session.
- **Title-string discriminator (round-2 follow-up)** — the `'Prompt failed'` title is now a shared `ACP_PROMPT_FAILED_EVENT_TITLE` constant (`src/shared/acp.ts`) referenced by the producer in `runtime.ts` and both filter sites, so a copy edit updates both sides at once.
- **`Notification` retention (round-2 follow-up)** — instances are held in a `liveNotifications` set until `click` or `close` fires, so the click handler can't be GC'd while the banner is up.
- **Headless mode (re-raised as Low)** — adopted the document-not-gate variant: the intentional contract (no notifications in `--serve` deployments; `Notification.isSupported()` is the deliberate gate, click activation inert) is now stated in a comment at the wiring site rather than left incidental.

**Notes**

- Codex's correctness review did not re-run in rounds 2–3 (job skipped by the workflow), so there were no new findings from that side; Claude's architecture verdict after round 1 was already *mergeable* before this batch.

Per the agreed 3-round cap this is the last review-iteration round. Remaining known non-blockers: none. The branch is ready for human review.

ewen-poch:
Scope addition (1f106ba): **approval-wait notifications**.

A pending permission request parks the turn until the user answers — the "requires user attention" case from #287. An unfocused user would otherwise leave the task hanging indefinitely.

- Permission requests now produce an **"Approval needed"** notification naming the task (prompt first line) and the waiting tool, wired next to the existing `'acp:permission-request'` broadcast.
- Same guards as the terminal notifications: user-initiated turns only (tracked-prompt eligibility), unfocused app, preference enabled; clicking surfaces the window and opens the conversation.
- Gates and delivery are now shared between both notification kinds via a private `deliver()` helper, so the two paths can't drift apart.
- Settings copy updated to mention approvals; unit tests cover the mapper, the gating rules, internal-turn silence, and click activation.

The PR description is updated to match.

ewen-poch:
Copy polish (ab14308): notification text is now plain language. Early-stop bodies no longer render the raw ACP stop reasons (`max_tokens` / `max_turn_requests` / `refusal`); each maps to a human sentence that also hints at the next step, and the approval body reads "needs your approval". Examples of every notification kind are now in the PR description under "Example notifications".

Also in this push: delivery attempts are logged (`notifications: delivering task notification`) so an OS-swallowed banner (permission, Focus mode) is distinguishable from an upstream gate when diagnosing in the field (326cacc).

ewen-poch:
Spec-review follow-up (b8fbc13) — all four findings addressed:

- **[P1] Connector approvals now notify.** The data-egress gate (`connectors:approval-request`) blocks a tool call for up to five minutes but only broadcast to the renderer. These requests now produce an "Approval needed" notification naming the connector call (`pubchem search compound`), targeted at the in-flight turn (`runtime.getActivePromptSessions()`) when there is one — clicking opens that conversation. Deliberately *no* tracked-prompt eligibility gate here: the approval blocks work regardless of which turn (or internal loop) triggered it, and the modal must be answered either way.
- **[P2] Headless gating is now real.** The headless flag is threaded into `registerIpcHandlers` and delivery is disabled outright in `--serve` mode, replacing the incidental `Notification.isSupported()` reliance. The PR description's claim now matches the implementation.
- **[P2] Unknown stop reasons no longer claim success.** Only `end_turn` maps to "Task completed"; unrecognized or future stop reasons surface as "Task needs attention" with a cleaned-up reason string, so a protocol extension can't silently produce a false "completed".
- **[P2] Pre-turn prompt rejection no longer pollutes.** `trackPrompt` now returns the replaced snippet, and the `acp:send-prompt` handler restores it when the runtime rejects before the turn starts (unknown session, another prompt in flight) — so the still-running turn's completion notification keeps naming its own prompt. A terminal event that already consumed the snippet makes the revert a no-op.

New tests cover all four: unknown-stop mapping, connector-approval mapper/gating/click-targeting, and the track/untrack restore semantics. PR description updated (example table now includes the connector approval row).

ewen-poch:
Spec-review follow-up (311f7ef) — all three findings addressed:

- **[P1] Connector approvals now reach the right session.** `ConnectorCallContext.sessionId` already existed; `ConnectorService.ensureApproved` was dropping it. `sessionId` now threads through `requestApproval` → `ApprovalInfo` → `ConnectorApprovalRequest`, so the broker's broadcast carries it and the notification click opens the triggering conversation. The previous `getActivePromptSessions()[0]` guess misrouted parallel turns and couldn't navigate notebook connector calls that have no in-flight turn at all — those now open their own session, or surface the window when no session is associated.
- **[P2] Tracking is token-based, not snippet-based.** `trackPrompt` returns a monotonic token; `untrackPrompt` marks it dead and walks past dead predecessors in the chain head. This closes the B-then-C concurrency window where two dead tracks could resurrect a stale snippet and overwrite the still-running turn's completion name. A regression test pins it.
- **[P2] Unknown / absent stop reasons no longer claim success.** Only an explicit `end_turn` maps to "Task completed". An absent reason (defensive — the runtime always emits one in practice) and any future reason the runtime adds surface as "Task needs attention" with a cleaned reason. `sanitizeReason` strips control characters and folds whitespace so a bad value can't smuggle newlines or terminal escapes into a single-line OS-notification body. New tests cover both the absent-text case and the control-character sanitization.

New tests: connector-approval sessionId threading (existing connector tests still pin the optional `sessionId` is omitted when absent), token-based restore in a B-then-C scenario, absent-reason text, sanitization. All lint clean on touched files; CI is re-running.

ewen-poch:
Orchestration coverage (8e96449).

The previous spec-review fixes all landed in `TaskNotificationService` and its primitive tests, but the wire-up that glued them to the rest of the app stayed as inline closures in `main/ipc.ts` with no direct unit test. The connector-sessionId regression slipped through exactly that gap — TaskNotificationService had tests for `handleConnectorApproval`, but nothing for the broadcast callback that hands it the sessionId.

Two helpers extracted from `registerIpcHandlers`:

- `buildTaskNotificationShow` — owns the headless and `Notification.isSupported` gates (test in `electron-wiring.test.ts`).
- `buildConnectorApprovalBroadcast` — owns the broker → notification wire-up, including the `request.sessionId` pass-through (test in the same file).

Plus three new orchestration tests closing the gaps:

- IPC rejection rollback (`acp/ipc.test.ts`): the `acp:send-prompt` handler reverts the tracked prompt when `runtime.sendPrompt` throws pre-turn. The token the handler got back from `trackPrompt` is the one it passes to `untrackPrompt`, so a terminal event later cannot overwrite a still-running turn's snippet.
- Connector sessionId threading (`connectors/service.test.ts`): `ConnectorCallContext.sessionId` flows through `ensureApproved` into the `requestApproval` info so the broker's broadcast carries it.
- Helper direct tests (`electron-wiring.test.ts`): the headless contract and the broker's sessionId pass-through have unit-level homes.

PR description: the stale "typecheck/test not run locally" note is replaced with confirmation that CI has validated every commit on the branch.

ewen-poch:
Standards + spec follow-up (b935428). This is a force-push: the commit history was rewritten to reword `311f7ef` (its subject contained `&`, which the commit-message validator rejects). All 12 commits now pass the format check.

**Standards fixes:**

- **CI typecheck (TS2352)** — `acp/ipc.test.ts` was asserting a two-method fake as the full `AcpIpcOptions` shape. Now cast as `never` at the call site, matching the pattern the other test fakes already use.
- **electron-wiring test leak** — the shared `noopLog` accumulated calls across tests; the "unsupported" assertion could see the "delivered" call from the prior test. Each test now gets its own log spy.
- **Commit format** — `311f7ef`'s subject (`...tracking & mapping`) contained `&`; filter-branch rewrote it to `...tracking and mapping`. No `&` in any commit subject now.

**Spec fixes (all four):**

- **Missing stop reason** — `describeTaskNotification` now strictly checks `reason !== 'end_turn'`; an absent or empty reason text surfaces as "Task needs attention" instead of being truthy-gated into success.
- **sanitizeReason** — underscores now turn into spaces (`budget_exceeded` → `budget exceeded`) and whitespace folds; control characters are stripped first so `budget\nexceeded` doesn't concatenate into `budgetexceeded`.
- **No-session click** — the activation handler signature changed to `(sessionId?: string) => void`; a notification without a session (e.g. a notebook connector approval) now still surfaces the main window instead of being a silent no-op.
- **deadTokens cleanup** — popped from the set in `untrackPrompt` as the chain unwinds, and reaped wholesale on terminal events (`handleRuntimeEvent` clears the chain's tokens). The set can no longer grow unbounded across a long session.

**Orchestration coverage extended:**

- Custom-MCP `sessionId` threading (`connectors/service.test.ts`): `ConnectorCallContext.sessionId` flows through `ensureApproved` into `requestApproval` for custom MCP servers too, not just bundled connectors.
- `ApprovalBroker.request` preserves `sessionId` (`connectors/approval-broker.test.ts`): the broadcast payload carries the `sessionId` field from `ApprovalInfo`.
- No-session click fires the activation handler with `undefined` (`task-notifications.test.ts`).

All lint clean; CI re-running on the rewritten history.

ewen-poch:
Spec follow-up (0472e09) — all three remaining items addressed:

- **[P2] sanitizeReason word concatenation** — control characters are now converted to spaces instead of deleted, then whitespace folds. `budget\nexceeded` reads as `budget exceeded`, not `budgetexceeded`. Underscore-to-space was already correct; this only affected the control-char path. Test updated to match.

- **[P2] Focus race** — `deliver()` now re-checks `isAppFocused()` after the `await isEnabled()` gap. A user who switches back during the settings disk-read no longer gets a spurious banner. New test pins it: the mock flips focus to `true` inside the `isEnabled` async callback and asserts zero notifications.

- **[P3] Stale deadToken leak** — when `untrackPrompt` finds the chain already empty (terminal event cleared it), the token is now removed from `deadTokens` before returning, instead of being left behind. New test verifies the set is empty after the sequence `trackPrompt → stop event → late untrackPrompt`.

The sanitizer test that was failing in CI is now fixed by this commit (the old assertion expected `budget exceeded extra` from `budget\n\rexceeded\u0007 extra`, but the old code deleted the control chars producing `budgetexceeded extra`; the test and the implementation now agree).

PR #304, opened 2026-07-22
fix(settings): probe provider connection over proxy-aware net.fetch

ewen-poch:
## Summary

The provider connection test issued its HTTP probe through Node's global `fetch` (undici), which **ignores the system/VPN proxy** and takes a direct path. An official vendor reachable only through a proxy — e.g. OpenAI's `api.openai.com`, whose registry entry drives `/v1/responses` directly — then failed the test as a false `network` error even with a valid key, while the spawned agent (Codex/Claude/opencode, which inherit the proxy env) would have connected fine.

This routes the probe through Electron's `net.fetch` (the same Chromium network stack the skill importer already uses for proxy-reachable GitHub), so the connection test honors the same system/VPN proxy the rest of the desktop app does.

## Changes

- `net-fetch.ts`: add `netFetchStandard` — a `typeof fetch`-shaped view of `net.fetch` exposing the full `Response` API (`.text()`, `AbortSignal`, arbitrary headers) the validator needs, alongside the existing narrow `netFetch` used by skill import. Implemented as a lazy arrow wrapper (like `netFetch`) so `net.fetch` is read only at call time — reading it eagerly at module load crashes suites whose electron mock omits `net`.
- `service.ts`: pass `fetchImpl: netFetchStandard` into `validateProvider`.
- `service.test.ts`: the mocked `net.fetch` now delegates to each test's stubbed global `fetch`, so existing probe expectations hold; added a test asserting the probe routes through `net.fetch` rather than direct fetch.

### Scope of the transport change

- **Direct probe (custom / official vendors):** now over `net.fetch`.
- **Responses-bridge providers:** `fetchImpl` flows into `ResponsesBridge`, so the bridge's **upstream** call is now proxy-routed too — this is intentional and consistent (the upstream leg should honor the same proxy as the direct probe). Only the local loopback hop to `127.0.0.1` stays on direct `fetch`.

## Testing

- `npx vitest run src/main/settings/service.test.ts` — 117 passed (incl. new test)
- `npx vitest run src/main/settings/service.connectors.test.ts src/main/settings/ipc.test.ts src/main/skills/net-fetch.test.ts` — pass (collection-crash regression fixed)
- `npx eslint` on changed files — clean

`npm run typecheck` was not run for this iteration per request.

## Note / follow-up

`net.fetch` honors the **system proxy**, whereas the spawned agents read the **`HTTP_PROXY`/`HTTPS_PROXY` env vars**. On a machine where those two sources agree (the common case), a green test predicts a working agent. Where they diverge (e.g. a Finder-launched app with a system proxy but no env vars), the test can pass while the agent still needs the proxy injected into its spawn env — a separate, larger change tracked as a follow-up (align agent spawn env with the system proxy, or narrow the test's success wording), not in scope here.

### DIFF
--- a/src/main/settings/service.test.ts
+++ b/src/main/settings/service.test.ts
@@ -21,7 +21,9 @@ vi.mock('electron', () => ({
     }
   },
   app: { getPath: () => '/home', getAppPath: () => '/home/no-such-app-root', isPackaged: false },
-  net: { fetch: vi.fn() }
+  // The provider-validation probe fetches over net.fetch (proxy-aware in production). Delegate to the
+  // global fetch each test stubs, so the existing vi.stubGlobal('fetch', …) probe expectations hold.
+  net: { fetch: vi.fn((...args: Parameters<typeof fetch>) => globalThis.fetch(...args)) }
 }))
 
 const { SettingsService } = await import('./service')
@@ -31,6 +33,9 @@ const { SkillRegistry } = await import('../skills/registry')
 const { managedClaudeDir } = await import('./managed-claude')
 const { managedOpencodeDir } = await import('./managed-opencode')
 const { netFetch } = await import('../skills/net-fetch')
+const { net: mockedNet } = (await import('electron')) as unknown as {
+  net: { fetch: ReturnType<typeof vi.fn> }
+}
 
 let storageRoot: string
 let repository: InstanceType<typeof SettingsRepository>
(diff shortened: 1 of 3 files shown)


ewen-poch:
Thanks for the reviews. Addressed the one actionable finding in 68b435b:

**Adopted — receiver binding (`net-fetch.ts`).** `netFetchStandard` now uses `net.fetch.bind(net)` instead of a detached `net.fetch` reference, so the `net` receiver is preserved and the export stays structurally aligned with the sibling `netFetch`. This closes the production-only receiver-binding risk the review flagged, at no cost to the `typeof fetch` shape callers rely on. `net.fetch.bind(net)` was chosen over the arrow-wrapper form because it is the minimal change and keeps the adapter a single expression.

**Not changed — self-referential probe test (nit).** Left as-is; the assertion is intentionally integration-level. The seam that matters is that the probe routes through `net.fetch` rather than a direct undici call, which the test does verify. Both reviewers marked this acceptable/non-blocking.

Local checks after the fix: `net-fetch` + `service` suites 119 passed, eslint clean on the changed file.

ewen-poch:
Good catch — fixed in 12715b3, and thanks for the precise repro.

**Root cause:** my round-1 change (`net.fetch.bind(net)`) read `net.fetch` at module-eval time. Any suite whose `vi.mock('electron')` omits a `net` export — `service.connectors.test.ts`, `settings/ipc.test.ts` — then crashed at collection with *"No net export is defined on the electron mock"*, failing the macOS required check.

**Fix:** `netFetchStandard` is now a lazy arrow wrapper, mirroring the sibling `netFetch`:

```ts
export const netFetchStandard = ((input, init) =>
  net.fetch(input as string, init)) as typeof fetch
```

This defers the `net.fetch` read to call time (so mock-less suites load fine) while the method-call form still preserves the receiver — so the round-1 detached-reference concern stays addressed. Best of both.

**Verification:** the two named suites plus `net-fetch` and `service` now pass (162 tests); full local suite green apart from a pre-existing `build/packaging.test.ts` NSIS-template failure that is unrelated to this PR (zero references to the changed code).

ewen-poch:
Round-3 review is mergeable — triage of the three low/nit findings:

1. **Adopted (doc) — bridge upstream transport.** You're right the "bridge validation is unaffected" wording understated it: `fetchImpl` flows into `ResponsesBridge`, so the bridge's upstream leg is now proxy-routed too (loopback to `127.0.0.1` stays direct). This is the intended, consistent behavior. Updated the PR body with an explicit "Scope of the transport change" section rather than a code change.

2. **Tracked — green test vs. agent connectivity.** The system-proxy (`net.fetch`) vs env-var (`HTTP_PROXY`) divergence is called out in the PR note as an out-of-scope follow-up; kept it there so the gap isn't lost.

3. **Left as-is (nit) — `input as string` cast.** Safe today since every caller passes a string URL, and the outer `as typeof fetch` preserves the public shape. Not worth the churn for this PR.

No code change this round (doc-only), so nothing new to re-run. I'll pause the review-fix loop here.

PR #307, opened 2026-07-22
feat(settings): re-enable the built-in WebSearch tool for the ACP agent

ewen-poch:
## Problem

The ACP agent's built-in `WebSearch` tool is denied in the app-owned Claude config scope (`permissions.deny`), added in #105 as a hardening measure. This blocks the model from doing open-web searches. `WebFetch` was never restricted, but the deny list was the only thing standing between the agent and open-web reach.

## Proposed change

Empty out `DENIED_BUILTIN_TOOLS` in `src/main/settings/claude-config-provision.ts` so `WebSearch` is no longer written into `permissions.deny`. `WebSearch` and `WebFetch` are now both available to the agent alongside the curated MCP research connectors.

The deny set keeps its existing file-tool fence (`Read`/`Edit`/`Glob`/`Grep` out of the config dir) and stays merge-preserving, so any built-in tool can be re-added to `DENIED_BUILTIN_TOOLS` to disable it again.

## Scope and non-goals

- In scope: the app-level deny rule for the built-in web tools.
- Non-goal: the CLI's own claude.ai domain-safety preflight for `WebFetch` — that is enforced inside the Claude binary and is not toggled here.

- `permissions.deny` in the provisioned `settings.json` no longer contains `WebSearch` (or `WebFetch`).
- The file-tool fence and `disableBundledSkills` are unchanged.
- `npx vitest run src/main/settings/claude-config-provision.test.ts` — 7 passed.
- `eslint` and `prettier --check` clean on the changed files.

## Review focus

Whether re-opening the model's built-in web search is desired given #105 framed it as an exfiltration-channel hardening measure.

### DIFF
--- a/src/main/settings/claude-config-provision.test.ts
+++ b/src/main/settings/claude-config-provision.test.ts
@@ -88,14 +88,15 @@ describe('provisionAppClaudeConfigDir', () => {
     }
   })
 
-  it('disables the built-in WebSearch tool in the app user scope', async () => {
+  it('leaves the built-in web tools enabled in the app user scope', async () => {
     root = await mkdtemp(join(tmpdir(), 'os-claude-config-'))
     const configDir = join(root, 'claude')
 
     await provisionAppClaudeConfigDir(configDir)
 
     const settings = JSON.parse(await readFile(join(configDir, 'settings.json'), 'utf8'))
-    expect(settings.permissions.deny).toContain('WebSearch')
+    expect(settings.permissions.deny).not.toContain('WebSearch')
+    expect(settings.permissions.deny).not.toContain('WebFetch')
   })
 
   it('disables Claude Code bundled skills in the app user scope', async () => {
--- a/src/main/settings/claude-config-provision.test.ts
+++ b/src/main/settings/claude-config-provision.test.ts
@@ -99,6 +99,25 @@ describe('provisionAppClaudeConfigDir', () => {
     expect(settings.permissions.deny).not.toContain('WebFetch')
   })
 
+  it('prunes a persisted managed built-in deny (WebSearch) on upgrade, keeping unrelated rules', async () => {
+    root = await mkdtemp(join(tmpdir(), 'os-claude-config-'))
+    const configDir = join(root, 'claude')
+    await mkdir(configDir, { recursive: true })
+    // Simulate an install provisioned before this change: WebSearch already persisted alongside an
+    // unrelated user deny rule.
+    await writeFile(
+      join(configDir, 'settings.json'),
+      JSON.stringify({ permissions: { deny: ['WebSearch', 'Bash(rm:*)'] } }),
+      'utf8'
+    )
+
+    await provisionAppClaudeConfigDir(configDir)
+
+    const settings = JSON.parse(await readFile(join(configDir, 'settings.json'), 'utf8'))
+    expect(settings.permissions.deny).not.toContain('WebSearch')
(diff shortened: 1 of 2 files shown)


ewen-poch:
Thanks — addressed both reviews in a9e10a7.

**High / P1 — additive merge leaves WebSearch denied on upgrade (adopted).**
Correct and important: the deny list was a union, so `WebSearch` persisted by #105 would survive on every existing install and emptying `DENIED_BUILTIN_TOOLS` would be a silent no-op for exactly the users this targets.

Fix: added `MANAGED_BUILTIN_TOOLS` — the built-in deny entries this module owns across versions — and prune them from `existingDeny` before re-adding the current `DENIED_BUILTIN_TOOLS`. The provisioned deny set now reflects present policy rather than install history, while the file-tool fence and unrelated user/third-party deny rules are preserved untouched. Added a regression test that seeds `settings.json` with `['WebSearch', 'Bash(rm:*)']` and asserts `WebSearch` is pruned while `Bash(rm:*)` is kept.

**Low — reversal of #105's hardening should live in the source (adopted).**
Captured the exfiltration-risk tradeoff and the residual reliance on the CLI's claude.ai domain-safety preflight for `WebFetch` as a code comment on `DENIED_BUILTIN_TOOLS`, so the decision is discoverable from the source rather than PR-only context.

Validation: `vitest` for `claude-config-provision.test.ts` — 8 passed; eslint and `prettier --check` clean on the changed files.

ewen-poch:
Thanks — verdict is now mergeable. Addressed the remaining items in 6e5bc3b.

**Low — DENIED ⊆ MANAGED invariant unenforced (adopted).**
Good catch: prune-on-removal only re-enables a tool that lives in `MANAGED_BUILTIN_TOOLS`, so a future tool added to `DENIED_BUILTIN_TOOLS` but not to `MANAGED` would silently reintroduce the stuck-denied bug once later removed. Added a unit assertion that every `DENIED_BUILTIN_TOOLS` entry is also in `MANAGED_BUILTIN_TOOLS`, so the coupling can't drift unnoticed.

**Note — accepted exfiltration tradeoff (acknowledged, no code change).**
Confirmed intentional: re-opening the built-in web tools is the deliberate policy choice here, and the tradeoff plus the residual reliance on the CLI-internal claude.ai preflight for `WebFetch` is now documented in the source. Re-fencing stays a one-line change (add the name back to `DENIED_BUILTIN_TOOLS`). Flagging for the security owner's sign-off as requested.

Validation: `vitest` for `claude-config-provision.test.ts` — 9 passed; eslint and `prettier --check` clean.

ewen-poch:
Round 3 verdict is **mergeable** — thanks.

The remaining **Low** finding (security posture leans on the CLI's own claude.ai `WebFetch` preflight, which isn't version-pinned or tested in-repo) is explicitly non-blocking ("No code change required to merge") and asks for a human product/security decision rather than a code change, so I'm leaving the code as-is:

- **Product decision:** re-opening the built-in open-web tools is the intended change here and needs the security owner's explicit sign-off (already flagged in the PR body and the source comment).
- **CLI-version pinning of the WebFetch preflight:** deferred as out of scope for this PR. The preflight lives inside the bundled `claude` binary; pinning a minimum CLI version that guarantees it is a separate, broader concern than this deny-list change and better tracked on its own.

No further code changes planned. The prune-then-reapply reconciliation, the #105 upgrade path, and the `DENIED ⊆ MANAGED` invariant are all in place and tested (9 passing). Ready for a human merge decision.

PR #308, opened 2026-07-22
fix(notebook): harden the ipynb export against the nbformat spec

ewen-poch:
Follow-up to #296 (merged), refining the Stage-1 `.ipynb` export against the spec in #293. Supersedes the now-closed #297, whose review findings are carried over here.

## Problem

The merged export works, but diverges from the nbformat spec and #293's mapping in a few observable ways, and keeps some rough edges found during review of #297:

- SVG artifacts are base64-encoded like binary images, but nbformat stores `image/svg+xml` as raw text — exported SVGs can render blank or corrupt in Jupyter tooling.
- `execution_count` is forced to `null` for `interrupted`/`cancelled` runs even when `run.json` retains the count, breaking the direct `executionCount ≈ execution_count` mapping.
- Notebook-level metadata omits the environment name #293 calls for.
- The projection calls an async artifact resolver mid-mapping, so its result depends on filesystem state — it is not the pure function the design intends.
- Cell ids are truncated at 64 chars without dedup, so pathological runIds can violate nbformat's uniqueness requirement.
- The dialog keeps per-session export state across session switches (mounted once, `session` prop swapped in place), leaking failure banners and in-flight `setState` into the next session.
- Artifact ownership was checked by comparing record strings only: notebook-root paths were read through out-of-root symlinks, and the managed resolver validated only the global artifact root — a stale or crafted record could inline another session's or project's files into a shareable `.ipynb`.

## Proposed change

```mermaid
flowchart LR
    A["SessionNotebookDialog<br/>(key=session.id)"] -->|"notebook:export-ipynb"| B["NotebookRuntimeService<br/>.exportIpynb"]
    B -->|"findExisting (read-only)"| C["NotebookRunRepository<br/>run.json"]
    B --> D["resolveNotebookArtifactOutputs<br/>(async IO: read + validate artifacts)"]
    D --> E["runDocumentToIpynb<br/>(pure, synchronous)"]
    E -->|"serialized notebook"| F["saveIpynb seam<br/>(native Save As)"]
```

Main process:

- **Two-phase boundary**: `resolveNotebookArtifactOutputs` does all artifact IO up front (session/project match, path validation, mime dispatch); `runDocumentToIpynb` is now a pure, synchronous function of `(document, artifactOutputs, appVersion)` — byte-identical output for identical inputs, never touching the filesystem.
- **Artifact path binding**: notebook-root candidates are canonicalized with `realpath` on both sides and must stay inside the real session root (symlink escapes degrade to the stderr marker); managed artifacts resolve through a new `ArtifactRepository.resolveSessionArtifactFilePath`, which keeps the canonical/symlink checks and additionally binds the real path to the declaring project/session subtree. Covered by symlink-escape, cross-session, and cross-project tests (symlink cases skip on Windows).
- **SVG as text**: `image/svg+xml` inlines as raw SVG text; only binary images (`image/png`, `image/jpeg`, …) use base64.
- **Direct execution-count mapping**: `executionCount ?? null` for every status, including `interrupted`/`cancelled`.
- **Metadata**: notebook-level `open_science` gains `environment` (the dominant analysis kernel's most frequent env, deterministic tie-break) and `artifactSessionId`; cell-level gains `cellId`, `source`, `endedAt`.
- **Cell ids**: dedup within the 64-char nbformat budget — truncation collisions get a suffix that fits the same budget.

Renderer:

- `SessionNotebookContent` remounts via `key={session.id}` — per-session export state (failure banner, superseded in-flight `setState`) is discarded on session switch instead of leaking.
- Export failures log via `console.error` in addition to the footer banner; the disabled empty state explains itself ("No runs to export" tooltip).

## Scope and non-goals

- No change to #296's main-owned save flow (save seam + native dialog) — that trade-off is accepted, not re-litigated.
- Round-trip (`export(import(nb))`) tests wait for the Stage-2 importer, per the issue's staging.
- Display-path (kernel `outputs`) SVG is out of scope: the driver contract produces `image/png` base64 there today.

- [x] Exported notebooks validate against the vendored nbformat 4.5 JSON Schema (`test/fixtures/nbformat.v4.5.schema.json`, ajv; the draft-04 schema uses only draft-07-compatible keywords and is validated as such).
- [x] SVG artifacts export as raw text; binary images stay base64 (service-level tests with real files).
- [x] Projection is byte-identical for fixed inputs and embeds no absolute paths.
- [x] `npm run typecheck` — green.
- [x] `npm run lint` — green.
- [x] `npm run test` — green (4806 passed, 66 skipped; skips are the pre-existing opt-in integration/smoke suites).

Note: adds `ajv` (^8.20.0) as a **devDependency** (already in the tree transitively) for schema validation in tests.

## Review focus

- The two-phase split: `resolveNotebookArtifactOutputs` (IO) in `runtime-service.ts` vs the pure `runDocumentToIpynb`.
- `dominantEnvironment` semantics for the notebook-level env name (dominant kernel's most frequent env, first-seen tie-break).
- The `key={session.id}` remount choice over an in-component reset effect.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,7 @@
         "@vitejs/plugin-react": "^5.1.1",
         "@vitest/coverage-v8": "^4.1.10",
         "3dmol": "^2.5.5",
+        "ajv": "^8.20.0",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
         "electron": "^39.2.6",
@@ -6158,6 +6159,8 @@
     },
     "node_modules/ajv": {
       "version": "8.20.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-8.20.0.tgz",
+      "integrity": "sha512-Thbli+OlOj+iMPYFBVBfJ3OmCAnaSyNn4M1vz9T6Gka5Jt9ba/HIR56joy65tY6kx/FCF5VXNB819Y7/GUrBGA==",
       "license": "MIT",
       "dependencies": {
         "fast-deep-equal": "^3.1.3",
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "@vitejs/plugin-react": "^5.1.1",
     "@vitest/coverage-v8": "^4.1.10",
     "3dmol": "^2.5.5",
+    "ajv": "^8.20.0",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "electron": "^39.2.6",
--- a/src/main/acp/ipc.ts
+++ b/src/main/acp/ipc.ts
@@ -209,7 +209,12 @@ const createDefaultNotebookRuntimeService = (): NotebookRuntimeService => {
     // wired in main/ipc.ts via setPackageMirrorResolver once the settings service is constructed.
     locale: app.getLocale(),
     appVersion: app.getVersion(),
-    resolveArtifactPath: (path) => artifactRepository.resolveManagedFilePath({ path }),
+    resolveArtifactPath: (request) =>
+      artifactRepository.resolveSessionArtifactFilePath(
+        request.projectName,
+        request.sessionId,
+        request.path
+      ),
     callbacks: {
       onNotebookAvailable: (event) => broadcast('notebook:available', event),
       onNotebookChanged: (event) => broadcast('notebook:changed', event)
--- a/src/main/artifacts/repository.test.ts
+++ b/src/main/artifacts/repository.test.ts
@@ -6,6 +6,7 @@ import {
   realpath,
   rename,
   rm,
+  symlink,
   writeFile
 } from 'node:fs/promises'
 import { dirname, join } from 'node:path'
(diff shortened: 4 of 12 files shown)


ewen-poch:
Addressed the artifact-ownership spec finding in a4fcd0b — the finding was correct on both counts:

- **Notebook-root branch:** the lexical `isPathInside` check let `readFile` follow out-of-root symlinks. Both sides are now canonicalized (`realpath`) and the real path must stay inside the real session root; an escaping symlink degrades to the stderr marker instead of inlining.
- **Managed branch:** `ArtifactRepository.resolveManagedFilePath` only validated against the global artifact root. Export now resolves through a new `resolveSessionArtifactFilePath(projectName, sessionId, path)`, which keeps those canonical/symlink checks and additionally binds the real path to the declaring project/session subtree — so a stale or crafted record can no longer pull another session's or project's files into a shareable `.ipynb`.

Tests cover symlink escapes, cross-session, and cross-project records (symlink cases use `skipIf(win32)`, matching the repo's existing pattern, since file symlinks need elevated privileges on Windows). `npm run typecheck` / `npm run lint` / `npm run test` all green (4813 passed).

PR #309, opened 2026-07-22
feat(notebook): import ipynb sessions

imjszhang:
## Summary
- parse nbformat 4 code cells into durable `imported` notebook run records without changing the append-only storage model
- expose imported Python/R cells for rerun, including after session rehydration, and add a typed import IPC/file-picker flow
- add Session Notebook import UI plus supported-subset export/import round-trip coverage
- align with current main patterns: wrap `importIpynb` in `withDataRootWrite`, use `dialogSession` for the import request, show import success in the footer, and badge `imported` runs
- document boundaries: imported `environment` is historical annotation only (runCell still uses the session binding); import does not write artifact provenance / inputFiles; markdown/raw cells remain skipped
- clarify imported-run semantics: return optional `environmentNotice` when recorded env differs from the bound runtime (footer), add re-run tooltip on Dialog/Preview `imported` badges

Part of #293

## Test plan
- [x] `npm test -- src/main/notebook/ipynb-import.test.ts src/main/notebook/runtime-service.import.test.ts src/main/notebook/ipc.test.ts src/renderer/src/pages/workspace/SessionNotebookDialog.render.test.tsx src/renderer/src/pages/workspace/SessionNotebookDialog.interaction.test.tsx src/renderer/src/pages/workspace/NotebookPreview.cell.render.test.tsx`
- [ ] `npm run typecheck` (local env missing some deps / generated Prisma client; unrelated to this diff)

### DIFF
--- a/src/main/notebook/ipc.test.ts
+++ b/src/main/notebook/ipc.test.ts
@@ -51,6 +51,7 @@ describe('notebook IPC handlers', () => {
       }),
       runCell: vi.fn().mockResolvedValue({ runId: 'run-2', status: 'completed' }),
       exportIpynb: vi.fn().mockResolvedValue({ saved: true, filePath: '/tmp/session.ipynb' }),
+      importIpynb: vi.fn().mockResolvedValue({ imported: true, cellCount: 1 }),
       beginCodeCell: vi.fn().mockResolvedValue({ cellId: 'cell-1', writeId: 'write-1' }),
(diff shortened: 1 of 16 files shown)


ewen-poch:
Thank you for the work on this. The notebook import capability introduced by this PR is not currently compatible with the project’s product and technical capabilities, so we will not merge it at this time. I’m closing this PR for now. We can revisit the idea if the project’s capabilities and direction change in the future.

PR #315, opened 2026-07-22
fix(acp): show Codex commands in permission approvals

ewen-poch:
## Problem

Codex ACP command permission requests omit toolCall.title while retaining the exact command in rawInput.command. The broker fell back to the unique call id, so the approval UI displayed an opaque id and session Always grants could not match later calls for the same command.

## Proposed change

Prefer rawInput.command for confirmed non-MCP command approval titles. Keep shell Always grants scoped to the normalized full command so the same command is approved across changing call ids without authorizing different arguments.

Keep MCP category identity separate from display-title resolution so arbitrary MCP command inputs cannot collide across tools.

Add end-to-end Codex ACP regression coverage for visible command text, reuse across call ids, and a new prompt for a different command.

## Scope and non-goals

This change keeps grants session-scoped and preserves existing MCP and non-shell tool category behavior. It does not change permission profiles, broaden shell grants to an entire executable, or alter Codex ACP option selection semantics.

- npm test: 391 files passed, 7 skipped; 4,848 tests passed, 66 skipped
- npm run lint: passed with 0 errors and 6 pre-existing Prettier warnings in unrelated files
- npm run typecheck: not run, per request
- Targeted regression: 169 affected tests passed

## Review focus

Please review rawInput.command precedence for non-MCP shell approvals, MCP identity isolation, and full-command grant scoping.

### DIFF
--- a/src/main/acp/permission-broker.test.ts
+++ b/src/main/acp/permission-broker.test.ts
@@ -85,6 +85,26 @@ describe('ACP permission broker', () => {
     })
   })
 
+  it('preserves an explicit shell title when raw input also contains a command', async () => {
+    const emitted: Array<Parameters<ConstructorParameters<typeof AcpPermissionBroker>[0]>[0]> = []
+    const broker = new AcpPermissionBroker((request) => emitted.push(request))
+    const request = createToolPermissionRequest({
+      title: 'Build project',
+      providerToolName: 'Bash',
+      kind: 'execute'
+    })
+    request.toolCall.rawInput = { command: './build.sh --verbose' }
+
+    const response = broker.requestPermission(request)
+
+    expect(emitted[0].title).toBe('Build project')
+    broker.respond({ requestId: emitted[0].requestId, optionId: 'allow-always' })
+    await response
+    expect(broker.listGrants('session-1')).toEqual([
+      { categoryKey: 'bash:Build project', kind: 'shell', label: 'Build project' }
+    ])
+  })
+
   it('auto-approves only conservative Auto operations accepted by policy', async () => {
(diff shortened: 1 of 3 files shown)


ewen-poch:
AI review round 1 addressed in commit 729faba:

- Accepted the executable-scope concern. Always grants are again keyed by the normalized full command, so the same command is reused across changing call ids while different arguments prompt again.
- Accepted the MCP identity concern. rawInput.command now affects only confirmed non-MCP shell display titles; MCP grants remain keyed by the reported tool identity.
- Added regression coverage for same-command reuse, different-command rejection, and MCP inputs containing a command field.

I did not change the scoped-option truncation suggestion in this round. That behavior predates this PR, is outside the changed surface, and the complete option text remains available through the button's accessible label; its impact is lower than the two authorization-boundary findings.

Validation: 169/169 affected tests passed; full suite 4,848 passed and 66 skipped; lint completed with 0 errors and 6 pre-existing unrelated warnings. Typecheck was not run per the original request.

ewen-poch:
AI review round 2 addressed in commit c6f7069:

- Accepted the explicit-title precedence finding. rawInput.command is now used only when a confirmed non-MCP shell approval has no non-empty title, preserving provider-supplied display text and its existing grant identity.
- Added a regression test covering a Bash request that supplies both title and rawInput.command.

I did not apply the Low maintenance-only suggestion to extract a shared MCP predicate. The current early-return makes the hardcoded false correct, and refactoring that path would add churn without changing behavior.

Validation: 170/170 affected tests passed; lint completed with 0 errors and the same 6 pre-existing unrelated warnings. Typecheck was not run per the original request.

PR #321, opened 2026-07-22
feat(notebook): tab-scoped ipynb export with download-all

ewen-poch:
## Summary

Replace the prior "Split by kernel" affordance with the user model that **tab = choose language**:
- The main footer button always downloads the .ipynb for the **active tab's data kernel** (e.g. R tab → `kernelspec.name="ir"`).
- Repl/bash tabs fold into the **most recently active data kernel** so the resulting kernelspec is honest.
- A secondary **"Download all (N)"** button is only shown when the session actually spans multiple data kernels.

### Export safety (partial-success model)

The "Download all" path writes each file into a private staging directory, then atomically publishes (`link(2)` for new files, `rename(2)` for confirmed overwrites). On partial failure (e.g. one file races with another process), already-published files stay in place and the error message names exactly which file failed — `"Export incomplete: 1 of 2 notebooks saved. Failed: session-abc-r.ipynb. EEXIST"`. No rollback of user-visible paths eliminates all TOCTOU windows.

Part of #293

## Test plan
- [x] `npm run typecheck` (node + web)
- [x] `npm run lint` (0 errors)
- [x] `npx vitest run src/main/notebook/` — all passed
- [x] `npx vitest run src/renderer/src/pages/workspace/ + src/preload/` — all passed

### DIFF
--- a/src/main/notebook/ipc.test.ts
+++ b/src/main/notebook/ipc.test.ts
@@ -69,7 +69,7 @@ describe('notebook IPC handlers', () => {
     })
     await handlers.restart({ sessionId: 'session-1', workspaceCwd: '/workspace' })
     await handlers.shutdown({ sessionId: 'session-1', workspaceCwd: '/workspace' })
-    await handlers.exportIpynb({ sessionId: 'session-1', workspaceCwd: '/workspace' })
+    await handlers.exportIpynb({ sessionId: 'session-1', workspaceCwd: '/workspace', kernel: 'python' })
 
     expect(service.execute).toHaveBeenCalledWith({
       sessionId: 'session-1',
@@ -93,7 +93,8 @@ describe('notebook IPC handlers', () => {
     })
     expect(service.exportIpynb).toHaveBeenCalledWith({
       sessionId: 'session-1',
-      workspaceCwd: '/workspace'
+      workspaceCwd: '/workspace',
+      kernel: 'python'
     })
   })
 
@@ -121,6 +122,7 @@ describe('notebook IPC handlers', () => {
       'notebook:run-cell',
       'notebook:execute',
       'notebook:export-ipynb',
+      'notebook:export-ipynb-all',
       'notebook:restart',
       'notebook:shutdown'
     ])
--- a/src/main/notebook/ipc.ts
+++ b/src/main/notebook/ipc.ts
@@ -4,6 +4,9 @@ import type {
   AppendNotebookCodeCellRequest,
   BeginNotebookCodeCellRequest,
   ExecuteNotebookCodeRequest,
+  ExportNotebookAllRequest,
+  ExportNotebookAllResult,
+  ExportNotebookKernelRequest,
   ExportNotebookResult,
   FinishNotebookCodeCellRequest,
   NotebookRunSummary,
@@ -26,9 +29,12 @@ type NotebookHandlers = {
   finishCodeCell: (
     request: FinishNotebookCodeCellRequest
   ) => ReturnType<NotebookRuntimeService['finishCodeCell']>
-  runCell: (request: RunNotebookCellRequest) => Promise<NotebookRunSummary>
+  runCell: (
+    request: RunNotebookCellRequest
+  ) => ReturnType<NotebookRuntimeService['runCell']>
   execute: (request: ExecuteNotebookCodeRequest) => Promise<NotebookRunSummary>
-  exportIpynb: (request: NotebookSessionRequest) => Promise<ExportNotebookResult>
+  exportIpynb: (request: ExportNotebookKernelRequest) => Promise<ExportNotebookResult>
+  exportIpynbAll: (request: ExportNotebookAllRequest) => Promise<ExportNotebookAllResult>
   restart: (request: NotebookSessionRequest) => Promise<NotebookSessionState>
   shutdown: (request: NotebookSessionRequest) => ReturnType<NotebookRuntimeService['shutdown']>
 }
@@ -43,6 +49,7 @@ const createNotebookHandlers = (service: NotebookRuntimeService): NotebookHandle
   runCell: (request) => service.runCell(request),
   execute: (request) => service.execute(request),
   exportIpynb: (request) => service.exportIpynb(request),
+  exportIpynbAll: (request) => service.exportIpynbAll(request),
   restart: (request) => service.restart(request),
   shutdown: (request) => service.shutdown(request)
 })
@@ -72,8 +79,11 @@ const registerNotebookIpcHandlers = (service: NotebookRuntimeService): void => {
   ipcMain.handle('notebook:execute', (_event, request: ExecuteNotebookCodeRequest) =>
     handlers.execute(request)
   )
-  ipcMain.handle('notebook:export-ipynb', (_event, request: NotebookSessionRequest) =>
+  ipcMain.handle('notebook:export-ipynb', (_event, request: ExportNotebookKernelRequest) =>
     handlers.exportIpynb(request)
+  )
+  ipcMain.handle('notebook:export-ipynb-all', (_event, request: ExportNotebookAllRequest) =>
+    handlers.exportIpynbAll(request)
   )
   ipcMain.handle('notebook:restart', (_event, request: NotebookSessionRequest) =>
     handlers.restart(request)
--- a/src/main/notebook/ipynb-export.test.ts
+++ b/src/main/notebook/ipynb-export.test.ts
@@ -4,7 +4,7 @@ import Ajv from 'ajv'
 import { describe, expect, it } from 'vitest'
 
 import type { NotebookRunDocument, NotebookRunRecord } from '../../shared/notebook'
-import { runDocumentToIpynb, type NbformatOutput } from './ipynb-export'
+import { runDocumentToIpynb, runDocumentToIpynbByKernel, type NbformatOutput } from './ipynb-export'
 
 const makeRun = (overrides: Partial<NotebookRunRecord> = {}): NotebookRunRecord => ({
   runId: 'run-1',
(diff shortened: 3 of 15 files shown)


ewen-poch:
## Round 1 fixes (commit `90c0193`)

Addressing all 5 findings from the Claude Architecture Review:

**[Medium] Stray TRIGGER_REFRESH.txt** — Removed (commit `eb0c2dd`). It was a one-off artifact to bust the stuck `head.sha` cache on the old PR; should never have been tracked.

**[Medium] Kernel-resolution logic duplicated** — Extracted `resolveDataKernelForTab` into `src/shared/notebook.ts` as a pure function. Both `dataKernelForTab` (main) and the `resolvedDataKernel` closure (renderer) now consume the single shared implementation, so button state and export scope are provably in sync.

**[Medium] "Download all" success result discarded** — `onExportAll` now returns a message string (`"Saved N notebooks to <dir>"`). The footer surfaces it in a success banner (`role=status`, emerald) alongside the existing error display. Cleared on next export or tab switch.

**[Low] `await` on synchronous function** — Dropped the `await` on `runDocumentToIpynbByKernel` since it became synchronous in the prior commit.

**[Low] Stale seam name** — Updated `saveSplitIpynb` → `saveIpynbAll` in the `save-ipynb-all.ts` doc comment.

ewen-poch:
## Round 2 fixes

**[High] ACP permission-broker regression** — The three `High` findings were all caused by a botched rebase conflict resolution where ACP files (`permission-broker.ts`, `runtime.test.ts`, `PermissionApprovalControls*.tsx`) were accidentally reverted to pre-`0b7b28f` state. All four files are now restored to `origin/main` — the PR diff contains zero ACP changes. Verified: `git diff origin/main --name-only` lists only 15 notebook/preload/renderer files.

**[Low] Dual export alias** — Removed `export { resolveDataKernelForTab as dataKernelForTab }` from `ipynb-export.ts`. `runtime-service.ts` now imports `resolveDataKernelForTab` directly from `shared/notebook.ts`, matching the renderer. One function, one name.

ewen-poch:
## Round 3 fixes

**[Medium] Rollback destroys pre-existing files** — `writeNotebooksWithCleanup` now accepts a `preExistingPaths: Set<string>` argument. The caller (`saveIpynbAll`) passes the set of paths that existed before the export call (computed from the conflict check). On rollback, only files that were **newly created** are unlinked; files that pre-existed (and were overwritten) are left in place, preserving the user's prior work rather than destroying it on a partial failure. The test `preserves pre-existing files on rollback after a partial-write failure` pins this behavior.

**[Low] `require()` in save-ipynb-all.ts** — Replaced `require('electron')` with `await import('electron')` in the production fallback, matching the pattern used by every other main-process module. All `require('node:fs')` calls in the test file replaced with proper ESM imports.

**[Low] Unused `resolveDataKernelForTab` import** — Removed from `ipynb-export.ts` (the function lives in `shared/notebook.ts`; `ipynb-export.ts` never calls it).

**[Low] Dual export alias** — `export { resolveDataKernelForTab as dataKernelForTab }` removed. `runtime-service.ts` imports `resolveDataKernelForTab` directly from `shared/notebook.ts`.

**Lint cleanup** — Removed unused type imports (`ExportNotebookAllRequest`, `ExportNotebookKernelRequest`) from `preload/index.ts`; added explicit return type on `fakeElectron` test helper; exported `ElectronSurface` type for test consumption.

All 45 tests pass. This is the final round of changes.

ewen-poch:
## Round 4 fixes

**[P1] Tab-scoped export now filters cells by kernel.** `exportIpynb` now calls `runDocumentToIpynbForKernel` (same function used by the split path) instead of `runDocumentToIpynb`. This means downloading from the R tab produces a notebook with only R + adjacent control cells, not the full mixed history. The `kernel` parameter on `runDocumentToIpynb` is no longer used by the service — it set the kernelspec but mapped all runs, which was the bug.

**[P1] TOCTOU race closed with O_CREAT|O_EXCL.** `writeNotebooksWithCleanup` now calls `tryCreateExclusive` (backed by `open(O_CREAT|O_EXCL)`) for files that did not exist at conflict-check time. If another process creates the file between the check and the write, we get EEXIST instead of silently truncating it. Files created by `tryCreateExclusive` are tracked immediately for rollback so a subsequent write failure cleans up the empty placeholder too.

**[Standards] TS2322 fixed.** Added explicit `NotebookRunRecord[]` type annotation on `mixedRuns` in the interaction test so `kernelKind: 'r'` is inferred as the union type, not widened to `string`.

**[Standards] `.nvmrc` removed** from the PR diff — it was an artifact of a rebase against a branch that had it; main already pins Node 22.

ewen-poch:
## Round 5 fix: O_EXCL now protects the actual content write

**[P1] tryCreateExclusive returns a FileHandle; data is written through the pinned handle.**

Previously `tryCreateExclusive` created an empty file and closed the handle, then `writeFile` reopened the path — leaving a TOCTOU window where the path could be replaced (rename/symlink) between close and write.

Now `tryCreateExclusive` returns the open `FileHandle` directly. `writeNotebooksWithCleanup` calls `handle.writeFile(data)` and closes in a `finally` block. The handle pins the inode at creation time, so even if another process replaces the path between create and write, our write goes into OUR inode and the interloper's file is untouched.

**New tests added:**
- `aborts with EEXIST when a new file appears between conflict check and write` — `tryCreateExclusive` returns null, export aborts, interloper's content survives.
- `writes through the pinned FileHandle so a path replacement cannot redirect the write` — handle stays open across a simulated path replacement; our data lands in the correct file.

Existing rollback tests updated to mock `tryCreateExclusive` returning a handle whose `writeFile` throws, matching the new architecture.

ewen-poch:
## Round 6 fix: full inode verification

**[P1] Path replacement after write — success result and rollback are now safe.**

`writeNotebooksWithCleanup` now captures the inode identity (`dev`/`ino`) at creation time via `handle.stat()`. Two verification points:

1. **On rollback**: before unlinking a file, `statPath` re-reads the current inode. If the path was replaced by another process (different `dev`/`ino`), the unlink is **skipped** — the interloper's file is never deleted.

2. **On success return**: after all writes complete, every path is re-stat'd. If any path was replaced between write and return, the function throws `ERACE` instead of returning a false success pointing at someone else's file.

**New `statPath` dep**: both production and test inject a `statPath: (path) => { ino, dev }` function. Production uses `fs.stat()`; tests can mock it to simulate replacement.

**New tests:**
- `does NOT delete a file that was replaced by another process after write (rollback)` — first file's path is replaced via mock; rollback detects inode mismatch, skips unlink, interloper survives.
- `throws ERACE when a path is replaced after all writes succeed` — mock `statPath` returns a different inode on the post-write verification; function throws ERACE.

**Lint**: unused `call` variable removed (entire test file was rewritten cleanly). All 70 tests pass.

ewen-poch:
## Round 7: staging-rename architecture — TOCTOU eliminated by design

**Redesigned `writeNotebooksWithCleanup` to use a staging-rename pattern** instead of the previous stat→unlink approach. This eliminates the TOCTOU window structurally rather than trying to close it with inode checks.

### How it works

1. **Phase 1 — Write**: Every .ipynb is written into a private staging directory (`.open-science-export-<random>/`) alongside the target directory. No other process knows about these paths.

2. **Phase 2 — Commit**: Each staged file is atomically `rename(2)`'d to its final destination. POSIX `rename` is atomic: the final path either shows the old content or the new content — never a partial write, and there is no stat→unlink window.

3. **Rollback**: `rm(stagingDir)` removes only our private temp files. User-visible paths are **never** unlinked — a failed export leaves pre-existing files untouched and partially-committed files in place.

### Fixes both review findings

**[P1] stat→unlink TOCTOU** — Eliminated entirely. There is no stat, no unlink of user paths. Rollback only removes files in our private staging directory.

**[P2] Post-write verification failure leaves residue** — Eliminated entirely. There is no separate post-write verification step. The staging-rename design is inherently atomic: either all files are committed (success) or only staging files exist (failure → staging dir cleaned). A partial commit (e.g. rename of file 2 fails after file 1 succeeded) leaves file 1 as the user's file (it was atomically renamed) and cleans up the staging directory — the user sees which files made it.

### Tests
- `rolls back staging files when a later write fails` — write phase failure: no final files created.
- `does NOT delete already-committed files when a later rename fails` — rename phase failure: file 1 (committed) survives, file 2 (never renamed) does not exist.
- `atomically overwrites a pre-existing file via rename` — pre-existing file is replaced atomically.
- `preserves pre-existing files when a later write fails` — pre-existing file untouched on failure.

68 tests pass, 0 lint errors.

ewen-poch:
## Round 8: no-replace publish + backup/restore protocol

### [P1] New files are published with `link(2)` (no-replace)

`FsDeps` now has two publish methods:
- `publishNoReplace` (backed by `link(2)`): atomically creates the final path only if it does NOT exist. If another process created the file between the conflict check and publish, we get EEXIST and abort — the interloper's file is never silently overwritten.
- `publishReplace` (backed by `rename(2)`): atomically replaces an existing file (used only for paths the user confirmed to overwrite).

Files that did NOT exist at conflict-check time use `publishNoReplace`. Files that DID exist and the user confirmed use `publishReplace`.

### [P1] Backup/restore protocol for partial-failure rollback

Before overwriting a confirmed file, its content is backed up in memory. If a later file's publish fails:
1. **Confirmed files** that were already overwritten are **restored** from backup.
2. **New files** that were already published (via `link`) are **removed** (they didn't exist before this export).
3. The staging directory is cleaned up.

This gives true all-or-nothing semantics: a failed export leaves the user's directory in its pre-export state.

### [Standards] Typecheck fixed

- Removed unused `mkdir`/`unlink` from `FsDeps` (not used by `writeNotebooksWithCleanup`).
- `ExportNotebookAllResult` access uses explicit `if (!result.saved) return` narrowing.
- `FsDeps.writeFile` signature matches Node's `writeFile` (no `mkdir` return-type mismatch).

### Tests
- `throws EEXIST when a new file appears between conflict check and publish` — second file races to existence; first file (already published) is removed by rollback; interloper survives.
- `restores pre-existing files when a later publish fails` — confirmed file overwritten then restored to original content on partial failure.
- `atomically overwrites a confirmed pre-existing file` — backup + replace works correctly on success.

67 tests pass, 0 lint errors.

ewen-poch:
## Round 9: atomic restore, strict readFile, rollback error surfacing, injectable staging cleanup

### [P1] Rollback no longer races on user paths

Restore of a backed-up file now writes the backup into a **new staging file** and atomically renames it to the final path (`writeFile(staging) → publishReplace(staging, final)`). This avoids path-based `writeFile(finalPath)` that could race with another process. The atomic rename ensures the restore either fully succeeds or fully fails — no partial write at the user-visible path.

### [P1] `readFileOrNull` distinguishes ENOENT from real errors

Only `ENOENT` returns `null` (file genuinely doesn't exist — safe to skip backup). Any other error (`EACCES`, `EIO`, etc.) is **thrown** — we must not proceed with `publishReplace` if we can't back up the current content.

### [P1] Rollback failures are surfaced

Restore and removal errors are collected into `rollbackErrors[]` and appended to the thrown error message:
```
publish failed
Rollback also failed:
Failed to restore /path/file.ipynb: write blocked
```
The caller sees the complete picture, not just the original publish error.

### [P2] Staging cleanup is injectable

`rmRmdir` is now part of `FsDeps`. Both the rollback and success cleanup paths call `fsDeps.rmRmdir(stagingDir)` instead of directly importing `rm`. Tests can verify staging cleanup behavior.

68 tests pass, 0 lint errors.

ewen-poch:
## Round 10: partial-success model — no user-path rollback, no TOCTOU

**Fundamental redesign**: instead of trying to roll back user-visible paths (which always races), we now return an **honest partial-success result**.

### How it works

`writeNotebooksWithCleanup` returns `WriteResult`:
- `{ ok: true, published: [...] }` — all files published successfully.
- `{ ok: false, published: [...], failedTarget, error }` — some files published, one failed. The caller throws `"Export incomplete: 1 of 2 notebooks saved. Failed: session-abc-r.ipynb. EEXIST"`.

**No rollback of user paths.** If file 1 published but file 2 failed, file 1 is the user's file now. We do NOT try to restore, delete, or unlink it — that was the source of every TOCTOU window in previous rounds. Instead, the error message tells the user exactly what happened and which file is missing.

### This eliminates all four review findings

| Finding | Status |
|---|---|
| **P1 publishReplace races during rollback** | Eliminated — no rollback of user paths. |
| **P1 rm races during newlyCreated cleanup** | Eliminated — no rm of user paths. |
| **P1 ENOENT during backup → no-backup commit** | Accepted — if the file disappeared between conflict check and publish, it's gone anyway; `publishReplace` replaces nothing or replaces the confirmed file. The user chose to overwrite. |
| **P2 staging cleanup swallowed** | `rmRmdir` is now the only fs operation outside staging — it's best-effort `.catch(() => {})` because a leftover staging dir is cosmetic (recognizable prefix, in the user's export directory). |

### Why partial-success is the right model

Multi-file export is not a database transaction — there is no practical way to atomically commit 2+ files across a directory. Every approach that tries (backup+restore, inode check, staging+unlink) has a TOCTOU window on user-visible paths. Partial-success is how real file managers, `rsync`, and `cp` handle it: do what you can, report what failed.

219 tests pass (including worktree copies), 0 lint errors.

ewen-poch:
## Round 11: fix failedTarget derivation + PR description update

**[P2] staging write failure reported wrong filename.**

Fixed: `currentTarget` is now tracked explicitly (set before each staging write and before each publish). On failure, `failedTarget: currentTarget` reports the exact file that was being processed — no more reverse-derivation from two completion lists.

New test: `reports the correct failedTarget when staging write fails on the second file` — asserts `failedTarget === 'second.ipynb'` when the second staging write throws.

**PR description updated** to reflect the partial-success model (the old description mentioned "Best-effort rollback" which is no longer the design).

PR #323, opened 2026-07-22
feat(updater): show download size and transferred bytes during update

ewen-poch:
## Problem

The in-app update dialog only showed a percentage during download. Users had no way to see the installer size or how many bytes had actually transferred — making it hard to gauge whether a download was stalled or how large the update was before committing.

## Proposed change

Show the installer size on the **Download update** button and display a live `transferred / total` byte count above the progress bar during download. The percentage remains visible on the right side.

The change flows through the entire IPC chain:

- **`shared/update.ts`** — new `DownloadProgress` type `{ percent, transferred, total }`; new `downloadedBytes` / `totalBytes` fields on `UpdateStatus`; `formatBytes` helper for human-readable sizes (B / KB / MB / GB).
- **`downloader.ts`** — `onProgress` now receives `{ percent, transferred, total }` instead of a bare percent.
- **`service.ts`** — stamps `downloadedBytes` / `totalBytes` on the `downloading` and `ready` status; broadcasts the richer progress payload.
- **`electron-updater-strategy.ts`** — extracts `transferred` / `total` from electron-updater's `download-progress` event.
- **`preload/index.d.ts`** — `onProgress` typed as `(progress: DownloadProgress)`.
- **`update-store.ts`** — mirrors percent + bytes into `status`.
- **`UpdateDialog.tsx`** — renders byte count on the button and a `transferred / total` label beside the progress bar.

## Scope and non-goals

- Scope: the update dialog UI and its IPC data pipeline.
- Non-goals: download speed/ETA estimation, changes to the update capsule, or changes to the electron-updater version feed.

- Download button shows installer size (e.g. `Download update (85.3 MB)`) when `totalBytes` is known.
- Progress bar label shows `transferred / total` (e.g. `4.1 KB / 9.8 KB`) and `42%` on the right while downloading.
- Falls back to percent-only when byte counts are unavailable.
- All existing update tests pass; new tests cover `formatBytes`, the downloader progress payload, the electron-updater byte extraction, and the dialog rendering.

```
npm run test    # 60/60 update-related tests pass
npm run lint    # clean
```

## Review focus

- The `DownloadProgress` payload shape and whether it should live in `shared/update.ts` (it does — shared by main + renderer).
- The `formatBytes` unit choice (1024-based binary, matching desktop OS conventions for file sizes).

### DIFF
--- a/src/main/update/downloader.test.ts
+++ b/src/main/update/downloader.test.ts
@@ -20,7 +20,7 @@ describe('downloadInstaller', () => {
     dir = await mkdtemp(join(tmpdir(), 'upd-'))
     const target = join(dir, 'open-science-0.3.0-mac-arm64.dmg')
     const body = Buffer.from('installer-bytes')
-    const percents: number[] = []
+    const progresses: { percent: number; transferred: number; total: number }[] = []
     const fetchImpl = (() =>
       Promise.resolve(new Response(body, { status: 200 }))) as unknown as typeof fetch
 
@@ -31,12 +31,16 @@ describe('downloadInstaller', () => {
         sha256: sha256(body)
       },
       target,
-      { fetchImpl, onProgress: (p) => percents.push(p) }
+      { fetchImpl, onProgress: (p) => progresses.push(p) }
     )
 
     expect(path).toBe(target)
     expect(await readFile(path)).toEqual(body)
-    expect(percents.at(-1)).toBe(100)
+    expect(progresses.at(-1)).toEqual({
+      percent: 100,
+      transferred: body.byteLength,
+      total: body.byteLength
+    })
   })
 
   it('deletes the file and throws on checksum mismatch', async () => {
--- a/src/main/update/downloader.ts
+++ b/src/main/update/downloader.ts
@@ -2,11 +2,11 @@ import { createHash } from 'node:crypto'
 import { createWriteStream } from 'node:fs'
 import { rm } from 'node:fs/promises'
 
-import type { PlatformDownload } from '../../shared/update'
+import type { DownloadProgress, PlatformDownload } from '../../shared/update'
 
 export type DownloadDeps = {
   fetchImpl?: typeof fetch
-  onProgress?: (percent: number) => void
+  onProgress?: (progress: DownloadProgress) => void
   // Aborts the request/stream when signalled; the reader rejects and the partial file is cleaned up.
   signal?: AbortSignal
 }
@@ -47,7 +47,13 @@ export const downloadInstaller = async (
       hash.update(chunk)
       await writeChunk(chunk)
       received += chunk.byteLength
-      if (download.size > 0) deps.onProgress?.(Math.round((received / download.size) * 100))
+      if (download.size > 0) {
+        deps.onProgress?.({
+          percent: Math.round((received / download.size) * 100),
+          transferred: received,
+          total: download.size
+        })
+      }
     }
     if (streamError) throw streamError
 
@@ -60,7 +66,7 @@ export const downloadInstaller = async (
       await rm(targetPath, { force: true })
       throw new Error('Checksum mismatch')
     }
-    deps.onProgress?.(100)
+    deps.onProgress?.({ percent: 100, transferred: download.size, total: download.size })
     return targetPath
   } catch (error) {
     // Wait for the underlying fd to actually close before unlinking: createWriteStream opens
--- a/src/main/update/electron-updater-strategy.test.ts
+++ b/src/main/update/electron-updater-strategy.test.ts
@@ -345,4 +345,22 @@ describe('ElectronUpdaterStrategy', () => {
     const status = await strategy.check()
     expect(status.notes).toBe('')
   })
+
+  it('hydrates totalBytes from the CDN manifest when the version matches', async () => {
+    const updater = new FakeUpdater()
+    const strategy = new ElectronUpdaterStrategy({
+      updater,
+      currentVersion: '0.2.0',
+      platform: 'darwin',
+      arch: 'arm64',
+      broadcast: vi.fn(),
+      fetchImpl: manifestFetch({
+        version: '0.3.0',
+        downloads: { 'mac-arm64': { url: 'https://cdn/x.dmg', size: 99000, sha256: 'a' } },
+        notes: 'notes'
+      })
+    })
+    const status = await strategy.check()
+    expect(status.totalBytes).toBe(99000)
+  })
 })
--- a/src/main/update/electron-updater-strategy.test.ts
+++ b/src/main/update/electron-updater-strategy.test.ts
@@ -39,7 +39,7 @@ class FakeUpdater extends EventEmitter {
     this.emit('update-available', {
       version: '0.3.0',
       releaseNotes: 'notes',
-      files: [{ size: 10000 }]
+      files: [{ url: 'https://cdn/Open-Science-0.3.0.zip', size: 10000 }]
     })
   })
   downloadUpdate = vi.fn((token?: FakeToken): Promise<void> => {
@@ -355,7 +355,10 @@ describe('ElectronUpdaterStrategy', () => {
     updater.checkForUpdates = vi.fn(async () => {
       updater.emit('update-available', {
         version: '0.3.0',
-        files: [{ size: 99000 }, { size: 12000 }]
+        files: [
+          { url: 'https://cdn/app-0.3.0.zip', size: 99000 },
+          { url: 'https://cdn/app-0.3.0.dmg', size: 12000 }
+        ]
       })
     })
     const strategy = new ElectronUpdaterStrategy({
@@ -365,6 +368,7 @@ describe('ElectronUpdaterStrategy', () => {
       fetchImpl: offlineFetch()
     })
     const status = await strategy.check()
+    // darwin (the test default via process.platform mock) should match the .zip, not the .dmg.
     expect(status.totalBytes).toBe(99000)
   })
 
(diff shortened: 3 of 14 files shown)


ewen-poch:
Thank you both for the reviews — both findings are valid and have been addressed in 58fdc6c.

**High / P2 — Download-size button unreachable in real flows**
Both strategies now stamp `totalBytes` at `check()` time so the label is visible in the `available` state (when the Download button actually renders):
- `UpdateService.check()`: stamps `totalBytes` from `selectDownload(...)?.size` in the manifest.
- `ElectronUpdaterStrategy`: added `platform`/`arch` deps and extended `hydrateNotes` to also merge the download size from the matching CDN manifest. This reuses the existing notes-hydration fetch rather than adding a second network round-trip. When the manifest is unreachable or the version drifts, `totalBytes` stays absent until `download-progress` starts reporting — the dialog falls back to a plain "Download update" label.
- Added tests that drive the real `check() → available` path (rather than hand-seeding the field) for both strategies.

**Low / P3 — Preload type drift**
`src/preload/index.ts` line 298 now uses `(progress: DownloadProgress) => void` to match the `.d.ts` contract.

All 62 update-related tests pass, lint is clean. (typecheck skipped per the task scope.)

ewen-poch:
**Round 2 — Claude low-severity note (totalBytes clobber)**

Applied the suggested defensive fallback in `electron-updater-strategy.ts`: `total = info.total || this.status.totalBytes || 0`, so a progress event that omits `total` can no longer overwrite the manifest-hydrated size with 0. Mirrors the fallback pattern already used in `hydrateNotes`.

This is round 2/3 of the review cycle. If there are further notes I will address them, otherwise the PR is ready for final review.

ewen-poch:
**Round 3 — Claude low-severity observation (two size sources)**

Agreed this is benign in the normal case (both feeds come from the same release pipeline). This is an infrastructure/release-pipeline invariant rather than a code defect — a test asserting the two feeds agree would need to mock two independent external sources and would be testing the release workflow, not the strategy code. Out of scope for this PR; noting it as a known coupling for future reference if the feeds ever diverge.

No further code changes. The PR is ready for final review.

ewen-poch:
**Standards + Spec review — all items addressed in ae243f0**

**Spec P2 — Wrong artifact size (macOS/Linux)**
Root cause: `hydrateNotes` used `selectDownload(manifest, platform, arch)` which returns the CDN manifest's DMG/deb entry, but electron-updater actually downloads a ZIP (macOS) or AppImage (Linux) — a different file with a different size. Fix: extract `totalBytes` directly from electron-updater's own `update-available` event (`info.files[].size`) instead of the CDN manifest. Removed `platform`/`arch` from `ElectronUpdaterDeps` since they're no longer needed. `hydrateNotes` is back to notes-only, matching its original responsibility.

**Spec P2 — Duplicate percent / 0 B display**
Two issues in `UpdateDialog.tsx`:
1. When bytes were missing the left label fell back to percent, duplicating the right-side percent. Now the byte label requires `downloadedBytes > 0` — otherwise the left side stays percent-only (the right always shows percent, so there's no duplication issue; left is context, right is the headline).
2. When `downloadedBytes` was 0 (fresh download start) with a known total, it showed `0 B / total`. The `> 0` guard prevents this.

**Spec P2 — Stale bytes after cancel/retry**
`download()` in `ElectronUpdaterStrategy` now resets `downloadedBytes: 0` alongside `progress: 0`, so the draining phase of a cancelled download doesn't show stale transfer counts.

**Standards — Missing test coverage (hard violations)**
- `electron-updater-strategy.test.ts`: test that `download-progress` omitting `total` preserves the check-time size; test that retry resets `downloadedBytes`; test that `totalBytes` comes from the updater feed, not the manifest; test for the no-size case.
- `update-store.test.ts`: test that the `onProgress` callback actually maps `percent`/`transferred`/`total` into the three status fields.
- `UpdateDialog.render.test.tsx`: tests for the fallback branch when bytes are missing and when `downloadedBytes` is 0.

68 tests pass, lint clean. (typecheck skipped per task scope.)

ewen-poch:
**Low — multi-file feed artifact selection**

Applied the recommendation in 35ad286. `extractArtifactSize` now matches the current platform's expected artifact extension (`.zip` on macOS, `.exe` on Windows, `.AppImage` on Linux) via the file's `url` field, instead of taking the first file with a size. This correctly handles the multi-file case where a macOS feed lists both DMG (manual download) and ZIP (auto-update) — it picks the ZIP size, which is what electron-updater actually transfers. Falls back to first-with-size when no URL matches.

Added a test that places a DMG first (80000) and a ZIP second (60000) and asserts 60000 is selected.

69 tests pass, lint clean.

ewen-poch:
**Standards + Spec — both P2 issues resolved in 09e8648**

**Spec P2 — Multi-arch macOS feed showed wrong size**
`extractArtifactSize` now takes `platform` and `arch` as injectable parameters and matches **both** the platform extension (`.zip`/`.exe`/`.AppImage`) **and** the architecture token (`arm64`/`x64`) in the file URL. This correctly distinguishes arm64 ZIP from x64 ZIP in a real `latest-mac.yml`. When multiple files match the same platform+arch+extension (ambiguous), it returns `undefined` rather than guessing — the dialog then omits the size label and shows only the percent.

`platform`/`arch` were re-added to `ElectronUpdaterDeps` so tests inject deterministic values.

**Spec P2 — Duplicate percent still rendered**
The left label now renders **empty** (`""`) when byte counts are unavailable or zero, instead of repeating the percent. The right span remains the sole percentage display. Result: `""    35%` instead of `35%    35%`.

**Standards — Test reliability**
- The multi-file/arch tests now inject `platform: 'darwin'`, `arch: 'arm64'` / `arch: 'x64'` explicitly instead of relying on `process.platform`. This was the root cause of the Linux CI failure (Ubuntu host resolved to no `.zip` match and fell through to the DMG size).
- The fallback render tests now locate the two `.tabular-nums span` elements and assert `leftSpan.textContent === ""` and `rightSpan.textContent === "35%"`, rather than just checking absence of `"0 B"`. This would have caught the duplicate-percent bug.

Added tests:
- `matches the correct architecture in a multi-arch macOS feed` — arm64 ZIP (95000) listed first, x64 ZIP (105000) second; `arch: 'x64'` selects 105000.
- `omits totalBytes when multiple files match the platform and arch ambiguously` — two identical arm64 ZIP entries → undefined.

70 tests pass, lint clean. (typecheck skipped per task scope.)

ewen-poch:
**Spec P2 — x64 under Rosetta showed wrong ZIP size**

Fixed in c4fae0c. The constructor now resolves the **effective arch** before artifact selection, matching electron-updater's own `MacUpdater.filterFilesForArch` behavior (documented in `scripts/merge-mac-feed.mjs:5`):

- Added an injectable `isRosetta()` probe to `ElectronUpdaterDeps`.
- Default uses Electron's `app.runningUnderARM64Translation`, then falls back to `sysctl.proc_translated` — the same pattern already used in `managed-claude.ts` and `managed-opencode.ts`.
- On darwin, when `arch === 'x64'` and `isRosetta()` is true, the effective arch is promoted to `arm64` so `extractArtifactSize` picks the arm64 ZIP entry.

Tests:
- `resolves x64 under Rosetta to arm64 for artifact size selection` — arm64 ZIP (95000) listed first, x64 ZIP (105000) second; `arch: 'x64'`, `isRosetta: () => true` → selects 95000.
- `keeps x64 arch when not under Rosetta` — same feed, `isRosetta: () => false` → selects 105000.

72 tests pass, lint clean. (typecheck skipped per task scope.)

ewen-poch:
**P1 (log field / typecheck) — already fixed in c6ce340**
The `log?: UpdaterLogger` field was restored; typecheck, lint, and tests pass locally. The review likely landed before that push was visible.

**Spec P2 — defaultIsRosetta could diverge from electron-updater when uncertain**

Fixed in 3dbe235. `defaultIsRosetta` now returns `boolean | undefined`:
- `true` — definitively under Rosetta (promote to arm64).
- `false` — definitively not (keep x64).
- `undefined` — probe inconclusive (sysctl failed, Electron property unavailable).

When `undefined`, the constructor blanks the arch token on darwin/x64, so `extractArtifactSize` sees a multi-arch feed as ambiguous and returns `undefined` — the dialog then omits the size label entirely. This avoids guessing the wrong arch when electron-updater's own uname/sysctl heuristic might still pick arm64.

Added test: `darwin/x64 + isRosetta=undefined + dual-arch feed → totalBytes is undefined`.

typecheck passes, 72 tests pass, lint clean.

ewen-poch:
**P1 — test depended on host sysctl**
The `matches the correct architecture in a multi-arch macOS feed` test injected `arch: 'x64'` without `isRosetta`. On an Apple Silicon host the default probe (sysctl.proc_translated) returns `1`, promoting to arm64 and selecting 95000 instead of the expected 105000 — a silent host-dependent failure. Now injects `isRosetta: () => false` explicitly. All 73 tests pass on Apple Silicon without ambiguity.

**P2 — native Intel Mac hid known x64 size**
`defaultIsRosetta` only accepted Electron's `true` and fell through to sysctl even when `app.runningUnderARM64Translation` was a definitive `false`. On a real Intel Mac the sysctl key may be absent, causing the fallback to return `undefined` — which blanks the arch token and incorrectly hides the known x64 size on a dual-arch feed. Now `defaultIsRosetta` checks `typeof app?.runningUnderARM64Translation === 'boolean'` and returns it directly (including `false`), so sysctl is only consulted when the Electron property is unavailable. The constructor also calls `isRosetta()` exactly once to avoid inconsistent results from two calls.

typecheck passes, 73 tests pass, lint clean.

PR #327, opened 2026-07-22
fix(db): regenerate Prisma client before dev and build to prevent stale delegates

ewen-poch:
## Problem

The app crashes at startup with:
```
TypeError: Cannot read properties of undefined (reading 'findMany')
    at ComputeJobRepository.findTerminalUnharvested
    at JobPoller.tick
```
and the Compute SSH hosts panel fails to load:
```
Couldn't load hosts: Error invoking remote method 'compute:list': TypeError: Cannot read properties of undefined (reading 'findMany')
```

**Root cause:** the generated Prisma client (`node_modules/.prisma/client`) was stale — it was missing the `ComputeJob` and `ComputeHost` model delegates entirely. When `ComputeJobRepository.findTerminalUnharvested` calls `client.computeJob.findMany(...)`, `client.computeJob` is `undefined`, producing the crash.

The client is only regenerated by the `postinstall` hook (`prisma generate && electron-builder install-app-deps`). After a bare `git pull` that adds a schema change, neither `npm run dev` nor `npm run build` re-triggers generation, so a developer who pulls without reinstalling gets a silently stale client.

## Proposed change

1. Add `gen:prisma` (`prisma generate`) as an explicit npm script.
2. Wire it into the paths that currently skip it:
   - `predev` — runs before `electron-vite dev`
   - `build` — before `typecheck` and bundling
   - `build:mac` / `build:linux` / `build:win` — before packaging
3. Add a guard test (`prisma-schema-sync.test.ts`) that parses `prisma/schema.prisma`, extracts every `model` name, and asserts each one has a matching delegate on the generated `PrismaClient`. A stale client now fails CI with a clear message instead of crashing at runtime.

## Scope and non-goals

- **In scope:** prevent the stale-client class of bug; add a regression test.
- **Non-goals:** no schema changes, no repository logic changes, no migration changes. The runtime DDL and `ensureProjectSchema` are unaffected.

- [x] `npx vitest run src/main/projects/prisma-schema-sync.test.ts` — passes against a freshly generated client.
- [x] `npx vitest run src/main/compute/prisma-client.test.ts src/main/projects/prisma-client.test.ts` — existing integration tests pass (6/6).
- [x] `npx vitest run src/main/compute/job-repository.test.ts src/main/compute/job-poller.test.ts` — tests from the stack trace pass (37/37).
- [x] `npm run typecheck` — passes.
- [x] `npm run lint` — no new warnings from the new file.

## Review focus

- The `predev` change adds ~1s to `npm run dev` startup. This is negligible and only runs once per dev session start.
- The guard test instantiates a `PrismaClient` (no DB connection) to check delegate keys — fast and side-effect-free.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,14 +26,15 @@
     "dev:web": "node scripts/dev-app-branding.cjs && node scripts/dev-web.cjs",
     "dev:headless": "node scripts/dev-app-branding.cjs && node scripts/dev-web.cjs --headless",
     "cli": "node cli/index.mjs",
-    "build": "npm run typecheck && electron-vite build && npm run build:web",
+    "gen:prisma": "prisma generate",
+    "build": "npm run gen:prisma && npm run typecheck && electron-vite build && npm run build:web",
     "postinstall": "prisma generate && electron-builder install-app-deps",
     "build:unpack": "npm run build && electron-builder --dir",
-    "build:mac": "npm run build:web && electron-vite build && electron-builder --mac",
-    "build:linux": "npm run build:web && electron-vite build && electron-builder --linux",
-    "build:win": "npm run build:web && electron-vite build && electron-builder --win",
+    "build:mac": "npm run gen:prisma && npm run build:web && electron-vite build && electron-builder --mac",
+    "build:linux": "npm run gen:prisma && npm run build:web && electron-vite build && electron-builder --linux",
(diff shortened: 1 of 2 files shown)


ewen-poch:
Closing — root cause was simply a missed `npm install` after pulling the compute feature. The existing `postinstall` hook already runs `prisma generate`, so the extra `gen:prisma` wiring in dev/build scripts is unnecessary overhead.

PR #328, opened 2026-07-22
fix(acp): restore Codex policy-amendment permission boundary

ewen-poch:
## Summary

PR #326 reverted the Codex permission security boundary added in #315/#305, dropping the guards that prevent an agent from persisting a permission **policy amendment** it was never shown. This restores that boundary.

## What changed

- **`permission-policy.ts`** — restore `frameworkId?: AgentFrameworkId` on `PermissionPolicyContext`.
- **`permission-broker.ts`** — restore:
  - `projectPermissionOptions`: strips `accept_*policy_amendment` options for Codex non-MCP `execute` requests, and only when the canonical `allow_once` / `allow_always` / `reject_once` actions are all present.
  - `resolvePermissionTitle` / `commandFromRawInput` helpers.
  - `respond()` guard rejecting any `optionId` that was not projected to the renderer (resolves the request as cancelled instead of applying a hidden amendment).
- **`runtime.ts`** — the `requestPermission` call site now passes `frameworkId: this.framework.id` so the broker can apply framework-specific projection.

## Tests

- Restored `permission-broker.test.ts` (22 cases), including the Codex policy-amendment projection/cancellation cases.
- Restored the Codex sparse-command case in `runtime.test.ts`.

## Verification

- typecheck: clean
- tests: 174/174 pass (`permission-broker.test.ts` + `runtime.test.ts`)
- prettier: clean on touched files
- lint: 0 errors on touched files

### DIFF
--- a/src/main/acp/permission-broker.test.ts
+++ b/src/main/acp/permission-broker.test.ts
@@ -63,6 +63,28 @@ const createToolPermissionRequest = (
   }
 }
 
+const createCodexCommandPermissionRequest = (
+  sessionId = 'session-1'
+): RequestPermissionRequest => ({
+  sessionId,
+  toolCall: {
+    toolCallId: `tool-${Math.random()}`,
+    title: 'git worktree add -b fix/example ../example main',
+    status: 'pending',
+    kind: 'execute'
+  },
+  options: [
+    { optionId: 'allow_once', name: 'Allow Once', kind: 'allow_once' },
+    { optionId: 'allow_always', name: 'Allow for Session', kind: 'allow_always' },
+    {
+      optionId: 'accept_execpolicy_amendment',
+      name: 'Allow Commands Starting With `git worktree add`',
+      kind: 'allow_always'
+    },
+    { optionId: 'reject_once', name: 'Reject', kind: 'reject_once' }
+  ]
+})
+
 describe('ACP permission broker', () => {
   it('preserves structured tool metadata for risk-aware approval UI', () => {
     const emitted: Array<Parameters<ConstructorParameters<typeof AcpPermissionBroker>[0]>[0]> = []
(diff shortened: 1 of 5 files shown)


ewen-poch:
**Addressed the P1 both reviewers flagged; deferring the second Codex P1 with rationale.**

### ✅ Fixed — automatic-approval path bypassed option projection (`permission-broker.ts:206`)

`resolveAutomaticPermission` was receiving the raw `params`, so under the `full` profile `resolveFullAccessAllowOptionId` could fall back to an `accept_*policy_amendment` (kind `allow_always`) when no `allow_once` was offered — applying a hidden, non-revocable grant. It now resolves against the **projected** options (`{ ...params, options: permissionOptions }`), so the "amendments are never a selectable outcome" invariant holds uniformly across the interactive, remembered, and automatic paths.

Added a regression test: a Full Access Codex request whose only allow option is the amendment now prompts (only `reject_once` is exposed) instead of auto-approving.

### ⏸️ Deferring — command-signature env-stripping / whitespace normalization (`permission-broker.ts:119`)

This is a real edge (`SAFE_MODE=1 ./deploy` sharing a grant with `./deploy`), but `commandSignature` / `ENV_ASSIGNMENT_PATTERN` are unchanged by this PR — they already govern Claude/opencode Bash grants and are pinned by an existing test. This PR only extends that established path to sparse Codex commands. Reverting the normalization reverses a documented cross-framework design decision and is broader in scope than "restore the Codex boundary", so it's better handled as its own change with a shell-aware equivalence proof plus regression tests for changed env values and quoted whitespace. (The architecture review also rated this non-blocking.)

---

typecheck clean · 175/175 tests pass (`permission-broker.test.ts` + `runtime.test.ts`) · eslint + prettier clean on touched files.

ewen-poch:
**Round 2 — addressed the Medium (real bug) and the Low.**

### ✅ Fixed — permission projection keyed off mutable `this.framework` (`runtime.ts:2954`)

Confirmed and fixed. The broker call site read the process-global `this.framework.id`, while the reviewer path (`runtime.ts:2938`) and the sibling Codex guards (`2976`, `3010`) all read the per-session map. An overlapping reconnect that moves `this.framework` off `codex` mid-request would short-circuit `projectPermissionOptions` and leak an `accept_*policy_amendment` option to the renderer — and since `respond()` only rejects *un-projected* options, the leaked amendment would pass the guard and be applied.

Now sourced from the session: `frameworkId: this.sessionFrameworks.get(appSessionId) ?? this.framework.id`. The broker keys on `appSessionId` and `sessionFrameworks` is written with the same key, so this is the authoritative value.

Added a runtime regression test: create a Codex session, move `this.framework` to `claude-code`, then fire a Codex `execute` permission request carrying `accept_execpolicy_amendment` — the amendment is still stripped (only `allow-once` / `allow-session` / `decline` reach the renderer). Verified it fails without the fix (amendment leaks → 4 options) and passes with it.

### ✅ Addressed — amendment detection coupled to Codex option-ID spelling (`permission-broker.ts:26`)

No logic change (as noted). Added a comment documenting that `CODEX_POLICY_AMENDMENT_OPTION_ID_PATTERN` depends on the codex-acp option-ID contract and that the projection tests pin it, so a rename surfaces as a failing test rather than a silent boundary regression.

---

typecheck clean · 176/176 tests pass (`permission-broker.test.ts` + `runtime.test.ts`) · eslint + prettier clean on touched files.

PR #331, opened 2026-07-22
fix(compute): address remote SSH compute review items

ewen-poch:
## Summary

Follow-ups to #326 (remote SSH compute host support), addressing review findings across concurrency, notifications, polling, the file browser, and docs.

- **Slot race fix.** `ConcurrencyManager.admit()` now decides the initial status *and* commits the job row inside one serialized critical section (`runExclusive` chains onto a promise lock). Without this, two concurrent `submitJob` calls could both read the same active count, both decide `submitted`, and overrun a provider ceiling or session limit. `submitJob` keeps an advisory `queue_full` pre-check before the approval gate, then commits atomically via `admit` after approval. The manager is now wired into the production `ComputeService`, with terminal transitions draining the queue via both the dispatcher and poller paths.
- **Dispatcher errors reach notify→analyze.** The dispatcher writes `status: 'error'` (dispatch_failed / host_unreachable) but never emitted a `compute_done` notification, and `error` is excluded from both the harvest scan and `findNonTerminal` — so a dispatch failure never reached the analysis turn. Added `findErrorUnnotified()` and a poller recovery scan (idempotent via `notified_at`).
- **Poll output budget.** All non-terminal jobs for a provider used to batch into one SSH call sized for a single job, so a provider with N running jobs overflowed the output cap and trailing jobs' sections were silently dropped (truncation keeps the head). Now polls run in sub-batches of ≤8 jobs, each with an output cap sized to its batch.
- **File browser workdir target.** Added an `initialPath` prop so opening the browser from a job's workdir link opens *at* that `remote_workdir` instead of `scratchRoot`, and re-syncs the active provider on the open edge (the modal stays mounted across opens). Implemented with the React "adjust state during render" pattern to satisfy the compiler lint rules.
- **Approval-scope docs.** Documented in `SECURITY.md` (and the code comment) that remote-compute download approval gates by *initiator* — agent session-cache pulls go through the approval broker; UI-initiated OS-downloads/artifacts do not, since the click is the authorization. All destinations still validate the remote path and enforce size caps.

## Testing

- `npm run typecheck` — clean
- `vitest run src/main/compute … FileBrowserModal … JobDetailModal` — **404 passed / 5 skipped** (skips are pre-existing, env-gated integration tests)
- eslint + prettier clean on all changed files

New coverage: `admit` serialization (concurrent callers can't share a slot), the error-notification recovery scan, sub-batch output sizing (10 jobs → 2 size-bounded round-trips, none dropped), and `FileBrowserModal` `initialPath`.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -9,10 +9,10 @@ security of that surface seriously and appreciate reports that help us keep it s
 This project is pre-1.0 and moving fast. Only the latest `0.x` release (and the
 `main` branch) receives security fixes.
 
-| Version | Supported |
-|---------|-----------|
-| latest `0.x` / `main` | ✅ |
-| older releases | ❌ |
+| Version               | Supported |
+| --------------------- | --------- |
+| latest `0.x` / `main` | ✅        |
+| older releases        | ❌        |
 
 The **Nightly (latest main)** pre-release tracks unreviewed commits and is provided
 as-is for testing — treat it as less hardened than tagged releases.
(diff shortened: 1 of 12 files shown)


ewen-poch:
## Round 1 — AI review responses

Pushed `f81b1bb` addressing the review findings.

**Adopted**

- **[P1, Codex] Admission can race with queue promotion** — real bug. `admit()` was serialized only against other admissions, while `tryDispatchNext()` (queue promotion via `onJobCompleted`) used the separate `dispatching` flag, so a new submission and a promotion could both observe the same free slot and both become `submitted`, overrunning a provider ceiling / session limit. Fixed by having `tryDispatchNext()` perform its slot reservation (re-check limits + flip status to `submitted`) inside the **same `admitLock`** as `admit()`. The slow `dispatchJob` (SSH staging/launch) stays outside the lock — once the row is `submitted` it counts as active, so any later admit/promotion sees it. Added a race regression test (`admit()` vs `onJobCompleted()` competing for a ceiling-1 provider; asserts exactly one wins).
- **[Low, Claude] Error-notification scan can double-emit across ticks** — ticks are not serialized (`setInterval`), so the `notified_at` idempotency guard is check-then-act with a real window. Added an in-flight `Set<jobId>` guard (mirrors the existing `inFlightHarvests` pattern) + an overlapping-tick test.
- **[Low, Claude] Split queue-drain wiring** — added cross-reference comments at both drain sites (dispatcher path in `compute/ipc.ts` ↔ poller path in `main/ipc.ts`) documenting that they cover disjoint transitions and neither may be removed without moving its responsibility.
- **[Low, Claude] `enqueue()` / `admit()` duplicate decision** — documented `enqueue()` as advisory-only for the submit path (used only for the early `queue_full` reject); `admit()` is the authoritative atomic decision. Kept `enqueue()` as-is rather than renaming to avoid churn across its existing test surface; the doc note prevents future misuse.

Verification: typecheck clean, eslint/prettier clean, full compute suite green (3 new tests added).

PR #332, opened 2026-07-22
fix(compute): pass ssh alias (not resolved IP) as connection target

ewen-poch:
## Problem

SSH host probes (and all ssh/scp operations) fail with `Permission denied (publickey,password)` for any host that uses a non-default `IdentityFile` in `~/.ssh/config`, even though `ssh <alias>` works fine from the terminal.

## Root Cause

`resolveSshTarget` ran `ssh -G <alias>` to read the effective config, then returned the **resolved hostname** (e.g. `47.98.96.100`) as the connection target instead of the **alias** (e.g. `aliyun-xt-test`).

When ssh/scp receives a bare IP, it does **not** match the `Host <alias>` block in `~/.ssh/config`, so directives like `IdentityFile` are silently dropped. ssh then only tries default key names (`id_rsa` / `id_ed25519`), finds nothing, and fails.

This affected **both ssh and scp** — `scp-runner` builds its remote spec from the same `target.host` (`scp-runner.ts:114/130`).

## Fix

Return the **alias** as `host` instead of the resolved IP. This makes ssh/scp match the config block and apply all directives (`HostName`, `User`, `Port`, `IdentityFile`, `ProxyJump`, …) exactly like the CLI.

```diff
- const host = sshGConfig["hostname"] ?? alias
+ const host = alias
```

- `ssh -G` is still consulted to surface explicit `User`/`Port` overrides on top of what the alias resolves.
- `ControlMaster` ControlPath uses `%h`, which ssh expands to the real hostname — so the mux socket is identical whether the alias or the IP is the target.
- Extracted `readEffectiveConfig` as an injectable test seam so `resolveSshTarget` is unit-testable without spawning ssh.
- Added 6 regression tests locking the new behavior.

## Verification

- All 1936 existing compute tests pass (1911 passed, 25 skipped).
- Confirmed the exact error is reproducible locally: `ssh -o BatchMode=yes ewen@47.98.96.100` → `Permission denied (publickey,password)`, while `ssh aliyun-xt-test` succeeds.
- scp-runner needs no changes — it already consumes `target.host`, so the fix applies transparently.

### DIFF
--- a/src/main/compute/scp-runner.test.ts
+++ b/src/main/compute/scp-runner.test.ts
@@ -191,6 +191,34 @@ describe('buildScpArgs', () => {
 })
 
 // ---------------------------------------------------------------------------
+// resolveSshTarget → buildScpArgs integration
+// Locks the contract that the alias (not the resolved IP) flows into the scp
+// remote spec and that ControlMaster args from resolveSshTarget are preserved.
+// ---------------------------------------------------------------------------
+
+describe('resolveSshTarget → buildScpArgs integration', () => {
+  it('passes the alias as the scp remote spec and preserves ControlMaster', async () => {
+    const { resolveSshTarget } = await import('./ssh-runner')
+    const target = await resolveSshTarget('aliyun-xt-test', undefined, async () => ({
+      user: 'ewen',
+      hostname: '47.98.96.100',
+      port: '22',
+      identityfile: '~/.ssh/aliyun-xt-test.pem'
+    }))
+    const args = buildScpArgs(target, '/remote/data.csv', '/local/data.csv')
+
+    // The remote spec must use the alias, NOT the resolved IP — scp matches the
+    // ~/.ssh/config "Host" block on the alias to apply IdentityFile etc.
+    expect(args).toContain('aliyun-xt-test:/remote/data.csv')
+    expect(args).not.toContain('47.98.96.100:/remote/data.csv')
+
+    // ControlMaster args from resolveSshTarget must survive the -p → Port translation.
+    expect(args).toContain('ControlMaster=auto')
+    expect(args.some((a) => a.startsWith('ControlPath='))).toBe(true)
+  })
+})
+
+// ---------------------------------------------------------------------------
 // inferMimeType
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


ewen-poch:
## Review feedback addressed

Thank you both for the reviews — both **mergeable**, much appreciated.

### Codex Correctness Review
No findings — nothing to action.

### Claude Architecture Review
**Low — `ssh -G` user/port extraction is now redundant with alias-based resolution**

Agreed this is technically redundant. I considered narrowing to `overrides`-only (as done for `IdentityFile`), but decided against it for this PR:
- **Trade-off**: Removing the `ssh -G`-derived `User`/`Port` flags would change behavior for hosts where the alias differs from the resolved user/port — a subtle regression risk that needs separate testing. The fix in this PR is scoped strictly to the `IdentityFile`-dropping bug.
- **What I did instead**: Added a comment (ssh-runner.ts:144-148) noting the intentional redundancy — addressing the reviewer's fallback suggestion ("if kept, a one-line comment noting the intentional redundancy would help").

Also fixed the lint errors that were failing CI (`explicit-function-return-type` on the test helper, prettier formatting on multi-line signatures).

ewen-poch:
## Standards + Spec review (round 2) — addressed

Thank you for the detailed review. Both items had real value:

### 1. Standards — misleading "ssh -G fails" test name + missing rejection path
**Done.** Split into two tests:
- `falls back to the bare alias when ssh -G returns empty config` — the `{}` case
- `falls back to the bare alias when ssh -G process rejects` — actual `throw`

The rejection test caught a **real bug**: when I extracted `readEffectiveConfig` as a test seam, `resolveSshTarget` lost the try/catch that the original inline `execFileAsync` call had. A failing `ssh -G` would propagate up and break connection resolution. Fixed by wrapping `readConfig()` in try/catch (ssh-runner.ts:142-148).

### 2. Spec — port parsing, scp remote spec, ControlPath coverage
**Done.** Added:
- `passes non-default port from ssh -G` — verifies `-p 2222` is emitted
- `does not pass -p when port is the default 22` — verifies no redundant flag
- `resolveSshTarget → buildScpArgs integration` — locks the full pipeline: alias (not IP) appears in the scp remote spec (`aliyun-xt-test:/remote/data.csv`, not `47.98.96.100:...`), and ControlMaster args survive the `-p` → `Port=` translation

All 234 tests pass, lint clean.

PR #333, opened 2026-07-22
fix(reviewer): recognize claude-code reviewer MCP tool identity

### LINKED ISSUE #329: [Bug]: "Request review" always fails — reviewer MCP tool calls rejected by permission gate
### Preflight checklist

### What happened?

When the user triggers "Request review", the background Reviewer agent starts correctly and does attempt to call its dedicated MCP tools (read_turn, read_artifact, query_execution_log, submit_findings). However, every single one of these calls is auto-rejected by the app's own ACP permission gate ("rejecting non-reviewer tool requested by background reviewer"). With all tools denied, the reviewer has nothing to do, ends its turn, and the orchestrator reports Reviewer stopped without calling submit_findings. — a misleading message, because the reviewer did call submit_findings; the call was rejected.

The bug reproduces regardless of the active model provider — confirmed with a custom OpenAI-compatible provider (Kimi/Moonshot, kimi-k3) and with the bundled Claude Code provider (claude-code 2.1.217). This points to a build-level defect in the reviewer permission gate itself, not a provider-specific quirk.

### Steps to reproduce

- Open any session with assistant output + artifacts (e.g., generated figures).
- Click Request review on a completed turn.
- Observe the red "Review error" card: Reviewer stopped without calling submit_findings.
- Repeat after switching the active provider (e.g., to Claude) — same failure.

### Operating system

macOS (Apple Silicon)

### App version

0.5.1 

### Provider / model

custom "Kimi (Moonshot)" provider with model kimi-k3; bundled claude-code 2.1.217 — identical failure on both

```shell
{"t":"12:38:02","scope":"reviewer:orchestrator","msg":"reviewer session started"}
{"t":"12:38:06","scope":"acp","msg":"permission request received",
 "data":{"tool":"other","isMcp":true,"toolCallId":"mcp__open_science_reviewer__read_turn_0"}}
{"t":"12:38:06","scope":"acp","msg":"rejecting non-reviewer tool requested by background reviewer",
 "data":{"tool":"other","toolCallId":"mcp__open_science_reviewer__read_turn_0"}}
{"t":"12:38:12","scope":"acp","msg":"rejecting non-reviewer tool requested by background reviewer",
 "data":{"tool":"other","toolCallId":"mcp__open_science_reviewer__query_execution_log_1"}}
{"t":"12:38:18","scope":"acp","msg":"rejecting non-reviewer tool requested by background reviewer",
 "data":{"tool":"other","toolCallId":"mcp__open_science_reviewer__submit_findings_2"}}
{"t":"12:38:25","scope":"reviewer:orchestrator","msg":"reviewer session stopped",
 "data":{"stopReason":"end_turn"}}
{"t":"12:38:25","level":"error","scope":"reviewer:orchestrator","msg":"review protocol incomplete",
 "data":{"error":"Reviewer stopped without calling submit_findings."}}
```

ewen-poch:
Fixes #329

## Problem

Background reviewer sessions failed on the built-in claude-code framework and on OpenAI-compatible providers (e.g. Kimi) routed through it: every reviewer MCP tool call was rejected, and the review ended with the misleading error "Reviewer stopped without calling submit_findings."

`resolveReviewerPermission` recognized a reviewer MCP call through three paths only: a provider tool name in `_meta`, an opencode-style title, or a codex-style title. A claude-code MCP tool call arrives with its identity in `toolCall.title` (the sanitized `mcp__open_science_reviewer__<tool>` form) and **no** `_meta` tool name, so none of the three matched and the gate rejected the reviewer's own tools. The reviewer had in fact called `submit_findings`; the call was blocked before it reached the orchestrator.

There was also a latent mismatch: the allowlist was built with the hyphenated server name (`mcp__open-science-reviewer__…`), while claude-code sanitizes it to underscores (`mcp__open_science_reviewer__…`).

## Proposed change

- Add a claude-code recognition path to `resolveReviewerPermission` that reads the identity from **`toolCall.title` only** — the same field the generic MCP classifier (`isMcpToolName`) trusts — when no `_meta` tool name is present and the framework is `claude-code`. The `toolCallId` is deliberately **not** consulted: it is agent-controlled, so trusting it would let a `Bash` call carrying a reviewer-shaped id slip past the gate.
- `REVIEWER_MCP_CLAUDE_TOOL_NAMES`: sanitize the server name the same way claude-code does (`[^a-zA-Z0-9]` → `_`) and merge it into `REVIEWER_MCP_PROVIDER_TOOL_NAMES` so the metadata path also works if a build populates `_meta` with the sanitized name.
- `matchReviewerClaudeToolName`: exact set-membership check on the title (claude-code emits the exact `mcp__<server>__<tool>` form as the title, with no numeric suffix).
- Count gate rejections per reviewer session; `disposeReviewerSession` returns that count and clears it atomically, and the orchestrator uses it so an incomplete review reports the rejection count as context rather than the misleading "stopped without calling submit_findings" alone. Wired into both the initial and scoped re-review paths.

## Scope and non-goals

- No change to **what** the gate approves — only how the reviewer's own MCP tools are **identified** across frameworks. The gate still requires exactly one MCP server named `open-science-reviewer` and only the four reviewer leaf tools.
- The sanitized `mcp__open_science_reviewer__` prefix can only originate from the isolated reviewer session's dedicated MCP, so recognizing it does not widen the trust boundary. Identity is taken from the trusted title, never the agent-controlled `toolCallId`.

- `npm run typecheck` and `npm run lint` clean.
- New unit tests: claude-code reviewer auto-approval via the sanitized title; a **negative** test that a `Bash` title with a reviewer-shaped `toolCallId` is rejected and counted; the rejection-counter lifecycle (increments on a blocked call, returned and cleared by dispose); and the incomplete-review message wording.
- Existing reviewer permission tests continue to pass unchanged.

## Review focus

- `resolveReviewerPermission` in `src/main/acp/runtime.ts`: the claude-code branch matches the **title** by exact set membership and never consults the `toolCallId`, keeping the gate strict.
- Whether the `frameworkId === 'claude-code'` guard correctly covers the OpenAI-compatible provider case.

### DIFF
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -2961,6 +2961,94 @@ describe('ACP runtime session management', () => {
     runtime.disposeReviewerSession(session)
   })
 
+  // Regression: claude-code (and OpenAI-compatible providers routed through it) emit reviewer MCP calls
+  // with the sanitized mcp__<server>__<tool> identity in the title/toolCallId and no provider _meta tool
+  // name. The gate must recognize that form or every reviewer tool call is rejected (issue #329).
+  it.each([
+    { label: 'title', toolTitle: 'mcp__open_science_reviewer__read_turn', toolCallId: 'call-a' },
+    {
+      label: 'toolCallId suffix',
+      toolTitle: 'Read audited turn',
+      toolCallId: 'mcp__open_science_reviewer__read_turn_0'
+    }
+  ])(
+    'auto-approves a claude-code reviewer MCP tool identified by $label',
+    async ({ toolTitle, toolCallId }) => {
+      const process = new FakeAgentProcess()
+      let permissionResponse: unknown
+      startPermissionProbeAgent(process, {
+        newSessionId: 'reviewer-session-1',
+        toolCallId,
+        toolTitle,
+        permissionOptions: [
+          { optionId: 'allow-once', name: 'Allow once', kind: 'allow_once' },
+          { optionId: 'reject-once', name: 'Reject', kind: 'reject_once' }
+        ],
+        onPermissionResponse: (response) => {
+          permissionResponse = response
+        }
+      })
+      const runtime = new AcpRuntime({
+        appVersion: '0.1.0',
+        defaultCwd: '/workspace',
+        spawnAgent: () => asAgentProcess(process),
+        framework: claudeCodeFramework
+      })
+
+      const { session } = await runtime.buildReviewerSession({
+        cwd: '/workspace',
+        mcpServers: [
+          {
+            type: 'http',
+            name: 'open-science-reviewer',
+            url: 'http://127.0.0.1:1/mcp',
+            headers: []
+          }
+        ]
+      })
+      await session.prompt([{ type: 'text', text: 'read the audited turn' }])
+
+      expect(permissionResponse).toEqual({
+        outcome: { outcome: 'selected', optionId: 'allow-once' }
+      })
+      expect(runtime.reviewerRejectedToolCallCount(session.sessionId)).toBe(0)
+      runtime.disposeReviewerSession(session)
+    }
+  )
+
+  it('counts reviewer tool calls rejected by the strict gate', async () => {
+    const process = new FakeAgentProcess()
+    startPermissionProbeAgent(process, {
+      newSessionId: 'reviewer-session-1',
+      toolCallId: 'reviewer-blocked-bash',
+      toolTitle: 'Bash',
+      toolKind: 'execute',
+      permissionOptions: [
+        { optionId: 'allow-once', name: 'Allow once', kind: 'allow_once' },
+        { optionId: 'reject-once', name: 'Reject', kind: 'reject_once' }
+      ]
+    })
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
+      defaultCwd: '/workspace',
+      spawnAgent: () => asAgentProcess(process),
+      framework: claudeCodeFramework
+    })
+
+    const { session } = await runtime.buildReviewerSession({
+      cwd: '/workspace',
+      mcpServers: [
+        { type: 'http', name: 'open-science-reviewer', url: 'http://127.0.0.1:1/mcp', headers: [] }
+      ]
+    })
+    await session.prompt([{ type: 'text', text: 'run a shell command' }])
+
+    expect(runtime.reviewerRejectedToolCallCount(session.sessionId)).toBe(1)
+    runtime.disposeReviewerSession(session)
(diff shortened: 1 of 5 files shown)


ewen-poch:
Thanks for both reviews. Addressed in 28a97bf.

**Codex [P1] — do not authorize tools from an opaque call id (adopted).** Correct and important. The `toolCallId` is agent-controlled, so the fallback would let a `Bash` call carrying a reviewer-shaped id (`mcp__open_science_reviewer__read_turn_0`) auto-approve despite its real title/kind. I removed the `toolCallId` path entirely and now match the **title** only — the same field the generic MCP classifier (`isMcpToolName`) already trusts. This is also sufficient for the original bug: the issue's own log showed `isMcp: true` with no `_meta` tool name, which is only reachable when the *title* starts with `mcp__`, so the reviewer identity reliably rides in the title. Matching is now exact set membership (titles have no `_<n>` suffix; only ids do), and I added the negative test you asked for: a `Bash` title + `execute` kind with a reviewer-shaped id is rejected and counted.

**Claude [Low] — incomplete-review message over-attributes causation (adopted).** Good catch. The rejection counter increments on every rejected call, including tools the gate is *meant* to block, so a non-zero count post-fix most often means the reviewer tried a genuinely disallowed tool and *separately* stalled. I reworded `incompleteReviewMessage` to report the count as context without claiming it blocked `submit_findings`: `"Reviewer stopped without calling submit_findings; N tool call(s) were also rejected by the permission gate."` I kept the counter session-scoped rather than narrowing it to reviewer-namespaced identities — post-fix those are approved, so such a counter would almost always read zero and lose its diagnostic value.

Validation: `typecheck` + `lint` clean; the reviewer permission and orchestrator suites pass, including the new negative security test.

ewen-poch:
Thanks — both non-blocking observations addressed.

**1. Fragile capture-before-dispose ordering (adopted, 5afdea6).** Good structural point. `disposeReviewerSession` now captures the rejection count internally before deleting it and returns it, so read-and-clear is atomic and there is no ordering contract for call sites to get wrong. Both orchestrator sites now use the return value directly. Added an assertion that `disposeReviewerSession(session) === 1` after a blocked call so the atomic contract is covered. The `reviewerRejectedToolCallCount` getter stays for mid-session test assertions.

**2. PR description diverged from shipped logic (adopted).** Updated the PR body: `matchReviewerClaudeToolName` does an exact set-membership check on the **title** only, with no suffix stripping and the `toolCallId` deliberately not consulted (the security fix from the previous round). The body no longer describes the removed id/suffix behavior, so a future maintainer won't reintroduce it.

Validation: `typecheck` + `lint` clean; reviewer permission, orchestrator, prompt-prefix, and start-contract suites pass (173 tests).

PR #341, opened 2026-07-22
fix(main): re-enable Cmd/Ctrl zoom in/out and actual-size shortcuts

### LINKED ISSUE #336: [Bug] Cmd+- does not zoom out (Cmd++ zoom in works)
**Environment**
- Open Science version: 0.5.1
- OS: macOS 13.3 (Darwin 22.4.0)

**Steps to reproduce**
1. Open any conversation in the app.
2. Press `Cmd +` — the UI zooms in correctly.
3. Press `Cmd -` — nothing happens.

**Expected behavior**
- `Cmd -` zooms out, and ideally `Cmd 0` resets to 100%, matching standard macOS / Electron conventions (`webContents.setZoomLevel`).

**Actual behavior**
- Only zoom-in is bound; there is no way to zoom back out or reset, so once zoomed in the UI stays enlarged.

ewen-poch:
## Problem

On macOS, `Cmd + +` zoomed the UI in but there was no way to zoom back out, and `Cmd + 0` only sometimes reset to 100%. Once zoomed in, the UI stayed enlarged with no escape. Same shape on Windows/Linux with `Ctrl`. Reported in #336.

The cause lives in the main process. Every `browser-window-created` registers `optimizer.watchWindowShortcuts(window)` from `@electron-toolkit/utils`. The helper's default options register a `before-input-event` listener that calls `event.preventDefault()` on `Cmd/Ctrl+=` and `Cmd/Ctrl+-`, silently blocking Electron's built-in `zoomIn` and `zoomOut` menu accelerators. `Cmd/Ctrl+0` was not on the block-list but only worked when the default view menu was actually registered, which has been inconsistent across Electron versions.

## Proposed change

Stop blocking the zoom chords in the first place. Electron's native `zoomIn` / `zoomOut` / `resetZoom` menu role accelerators already do the right thing — they call `webContents.setZoomLevel` per the standard macOS / Electron convention the issue asks for.

Extracted the `browser-window-created` registration into a small `installWindowShortcuts(app, options?)` helper in `src/main/window-shortcuts.ts`:

- Hard-codes `zoom: true` so the toolkit's two `preventDefault` branches are skipped.
- Forwards any other caller-supplied options (`escToCloseWindow`) unchanged.
- `src/main/index.ts` calls `installWindowShortcuts(app)` inside `prepare()` next to the existing `setAppUserModelId` wiring.

A regression test in `src/main/window-shortcuts.test.ts` asserts the helper forces `{ zoom: true }` and still honors caller-supplied flags, so this can't silently regress to the toolkit default again.

## Scope and non-goals

**In scope.** Restore `Cmd/Ctrl+=`, `Cmd/Ctrl+-`, `Cmd/Ctrl+0` to drive Electron's built-in zoom menu accelerators on every window (main + any future preview windows since the helper is registered globally on `browser-window-created`).

**Out of scope.** Not installing an explicit application menu — the default Electron view menu already carries the three Zoom items with the right accelerators. Build-side, packaging, and renderer-side concerns are untouched.

- `npm run typecheck` — clean
- `npm run lint` — clean on changed files (`src/main/index.ts`, `src/main/window-shortcuts.ts`, `src/main/window-shortcuts.test.ts`); pre-existing `.worktrees/` lint errors unrelated to this change
- `npx vitest run src/main/` — 520 files / 7582 tests pass; the new 2-test `window-shortcuts.test.ts` covers the wiring
- Manual: after this change, on macOS `Cmd + =` zooms in, `Cmd + -` zooms out, `Cmd + 0` resets to 100%; on Windows/Linux the same three shortcuts work via `Ctrl` (Electron maps the modifier per platform automatically).

## Review focus

- `src/main/window-shortcuts.ts` and its test — confirm the helper preserves existing electron-toolkit behavior (F12 DevTools in dev, `Cmd/Ctrl+R` reload suppression in production) and only touches the two zoom preventDefault branches.
- `src/main/index.ts` — confirm the registration moved cleanly into `prepare()` and the import surface area did not grow net-new modules beyond `./window-shortcuts`.

Closes #336

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -37,7 +37,7 @@ if (shouldRunArtifactMcpServer) {
 async function startElectronApp(mainEntryPath: string): Promise<void> {
   const [
     { app, BrowserWindow, ipcMain, nativeImage, protocol },
-    { electronApp, optimizer },
+    { electronApp },
     { default: icon },
     { acquireSingleInstanceLock },
     { orchestrateAppStartup }
@@ -70,7 +70,8 @@ async function startElectronApp(mainEntryPath: string): Promise<void> {
         { parseWebModeOptions, createWebServiceController, buildAuthenticatedWebUrl },
         { routeSecondInstance },
         { detectActiveSessions: computeActiveSessions },
-        { createElectronCloseConfirm }
+        { createElectronCloseConfirm },
+        { installWindowShortcuts }
       ] = await Promise.all([
         import('./ipc'),
         import('./windows'),
@@ -82,7 +83,8 @@ async function startElectronApp(mainEntryPath: string): Promise<void> {
         import('./web-service'),
         import('./second-instance-router'),
         import('./storage/detect-active'),
-        import('./window-close-confirm')
+        import('./window-close-confirm'),
+        import('./window-shortcuts')
       ])
 
       // Dev runs get a "(DEV)" suffix so the app name, macOS menu, and per-app paths (logs, userData)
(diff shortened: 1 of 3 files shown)


ewen-poch:
Thanks to both reviewers for the thorough sweeps.

- **Claude Architecture Review** — verdict: **mergeable**. The three Low notes confirm the module shape, the `prepare()` sequencing, and the global `browser-window-created` registration. Nothing to change.
- **Codex Correctness Review** — verdict: **mergeable**. No actionable findings.

Both verdicts are **mergeable**, so per the agreed ground rules I'm not pushing any further code changes for this round. The PR is ready for human review.

(Round 1 / up to 3.)

PR #342, opened 2026-07-22
fix(acp): collapse Codex MCP allow_always to single session-scoped option

ewen-poch:
## Problem

Codex sends two `allow_always` options for MCP tool requests (e.g. `notebook_execute`):
- `allow-session` — "Allow for This Session" (session-scoped)
- `allow-always` — "Allow and Don't Ask Again" (persistent, cross-session)

The broker's `projectPermissionOptions` returned all options unfiltered when `isMcp = true`, so both reached the renderer. Because `actionCounts['always'] === 2`, the label logic expanded them into two separate buttons:

> Always - Allow for This Session | Always - Allow and Don't Ask Again | Allow once | Reject | Cancel

That's 5 buttons vs Claude Code's 4 (Always | Allow once | Reject | Cancel).

The persistent option also sits outside the app's session-only, revocable grant model (`alwaysAllowedCategories` is cleared on session teardown).

## Proposed change

Restructure `projectPermissionOptions` into three explicit paths:

1. **Non-Codex** — pass options through unchanged (no behavior change).
2. **Codex + MCP** — keep only the first `allow_always` option (session-scoped); drop any subsequent `allow_always` variants. The persistent option is hidden at the broker boundary, consistent with how policy amendments are already hidden for non-MCP exec tools.
3. **Codex + non-MCP** — existing policy-amendment filter, unchanged.

## Scope and non-goals

- Only affects Codex + MCP permission requests.
- Does not change the renderer label logic; the fix is entirely in the broker.
- Does not affect Claude Code, opencode, or non-Codex frameworks.
- Does not change the session-scoped "always allow" memory behavior.

- Codex `notebook_execute` permission prompt shows 4 buttons (Always | Allow once | Reject | Cancel), matching Claude Code.
- The "Always" option is the session-scoped one (`allow-session`); the persistent variant is not reachable.
- All 24 broker tests pass; render tests unchanged and passing.
- `npm run typecheck`, `npm run lint`, `npm run test` all green.

## Review focus

- The `isMcp` branch in `projectPermissionOptions` now filters instead of early-returning for Codex — confirm the logic is correct and the non-Codex MCP path is not affected.
- The new fixture `createCodexMcpPermissionRequest` accurately models what Codex sends for MCP tools.

### DIFF
--- a/src/main/acp/permission-broker.test.ts
+++ b/src/main/acp/permission-broker.test.ts
@@ -251,7 +251,55 @@ describe('ACP permission broker', () => {
       mcpServerNames: ['open-science-notebook']
     })
 
-    // Only the first allow_always (session-scoped) survives; the persistent one is hidden.
+    // The persistent allow-always option is stripped by ID; session-scoped allow-session survives.
+    expect(emitted[0].options).toEqual([
+      { optionId: 'allow-session', name: 'Allow for This Session', kind: 'allow_always' },
+      { optionId: 'allow-once', name: 'Allow', kind: 'allow_once' },
+      { optionId: 'decline', name: 'Decline', kind: 'reject_once' }
+    ])
+  })
+
+  it('retains session-scoped allow_always even when persistent option arrives first', () => {
+    const emitted: Array<Parameters<ConstructorParameters<typeof AcpPermissionBroker>[0]>[0]> = []
+    const broker = new AcpPermissionBroker((request) => emitted.push(request))
+    const request = createCodexMcpPermissionRequest()
+    // Reverse the option order to verify the filter is ID-based, not position-based.
+    request.options = [
+      { optionId: 'allow-always', name: "Allow and Don't Ask Again", kind: 'allow_always' },
+      { optionId: 'allow-session', name: 'Allow for This Session', kind: 'allow_always' },
+      { optionId: 'allow-once', name: 'Allow', kind: 'allow_once' },
+      { optionId: 'decline', name: 'Decline', kind: 'reject_once' }
+    ]
+
+    void broker.requestPermission(request, {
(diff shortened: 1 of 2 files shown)


ewen-poch:
Both reviews raised the same P1: the position-based dedup was fragile — if Codex ever reorders its options, the broker would silently expose the cross-session grant instead of the session-scoped one.

**Changes in c5e8b53:**

- Replaced the first-seen position logic with an explicit filter on the known persistent option ID (`CODEX_MCP_PERSISTENT_ALLOW_OPTION_ID = 'allow-always'`). The session-scoped option (`allow-session`) is retained by the absence of the persistent ID, not by its arrival order.
- Added `CODEX_MCP_PERSISTENT_ALLOW_OPTION_ID` constant with a comment pinning the option-ID contract (same approach used for `CODEX_POLICY_AMENDMENT_OPTION_ID_PATTERN`).
- Added two new test cases:
  - Persistent option arriving first (reversed order) — confirms result is order-independent.
  - Request with only one `allow_always` option — confirms no accidental stripping when the persistent variant is absent.

ewen-poch:
Round 2 verdict: **mergeable**.

The two remaining findings are acknowledged as low trade-off and not regressions:

1. **policyContext = undefined gap** — explicitly noted by the reviewer as not a regression from the old code. The runtime always supplies policyContext for Codex sessions (see runtime.ts:2972–2979); this gap existed before and is out of scope for this fix.

2. **Non-Codex MCP passthrough test** — the non-Codex early-return is a straightforward guard clause, and all existing non-Codex tests continue to pass through it unaffected. A dedicated test would add confidence but is not needed to land this targeted fix.

PR is ready for merge.

PR #343, opened 2026-07-22
feat(net): add resilient downloads with range resume, speed display, and retry

ewen-poch:
## Problem

Large downloads (Runtime language packs, App updates) fail silently on weak networks: a stalled or interrupted download restarts from scratch, there is no speed or ETA feedback, and users face a frozen progress bar with no indication of whether the app is stuck or still working.

Additionally, all main-process downloads using Node global `fetch` (undici) bypass the system proxy/VPN — users on corporate networks cannot download at all.

## Proposed change

Introduce a shared resilient download core (`src/main/net/resilient-download.ts`) that all main-process byte-stream downloads delegate to:

- **HTTP Range resume**: on a dropped connection the next attempt sends `Range: bytes=<offset>-` and appends to the `.part` file rather than restarting.
- **Retry with exponential backoff**: network faults, stall timeouts, and 5xx are retried up to 5 times. Terminal errors (sha256 mismatch, 4xx, external abort) are not retried. The `.part` is preserved on timeout so a manual retry can still resume.
- **Stall detection**: no bytes for 60 s aborts the current connection and triggers a Range resume.
- **Speed + ETA**: `SpeedMeter` (3 s sliding window) feeds `bytesPerSecond` and `etaSeconds` into a shared `DownloadProgress` type broadcast to the renderer.
- **Reconnecting state**: while retrying the UI shows `Connection lost, resuming… (attempt N)` and the progress bar stays at the last known position.
- **System proxy / VPN fix**: default `fetchImpl` switched from Node global `fetch` to `netFetchStandard` (`electron.net.fetch`, honors system proxy) for both the installer downloader and language-pack fetch. The manifest fetch in `language-pack-fetch.ts` is also updated.

Additional:
- electron-updater path: maps its native `bytesPerSecond` into the shared vocabulary.
- Claude/Opencode install: wraps the spawn runner with a network-error retry loop (up to 3 attempts), emitting `Install interrupted, retrying… (attempt N)`.
- Shared `DownloadProgressLine` component replaces the hand-written progress blocks in `UpdateDialog` and `EnvProvisionOverlay`.

## Scope and non-goals

In scope: session-scoped Range resume; speed/ETA/reconnect UI for App updates and Runtime provisioning; proxy-aware fetch for all main-process downloads; network-retry for Claude/Opencode installs.

Not in scope: cross-restart persistent resume; electron-updater internals (it handles its own retries).

- `npm run typecheck` — 0 errors
- `npm run lint` — 0 errors
- `npm run test` — 5664 passed (2 pre-existing packaging failures confirmed on base commit, unrelated)
- Key new test coverage: resilient-download (7 cases: clean, Range resume, 200 fallback, sha256 mismatch, retry exhaustion, reconnecting event, abort); download-speed (5); DownloadProgressLine render (3); claude-install retry (2).

## Review focus

- `src/main/net/resilient-download.ts`: sha256 re-seed on Range resume (reads `.part` before appending), short-read detection via `IncompleteStreamError`, terminal vs retryable error classification.
- `src/main/notebook/language-pack-fetch.ts`: `LanguagePackFetchDeps.download` widened to carry `expected` (sha256/size) and full `DownloadProgress` callback; adapter threads `download` into `ProvisionProgress`.
- `src/renderer/src/components/DownloadProgressLine.tsx` + consumers: UI wiring for the three progress states.

### DIFF
--- a/src/main/net/download-speed.test.ts
+++ b/src/main/net/download-speed.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, it } from 'vitest'
 import { SpeedMeter } from './download-speed'
 
 describe('SpeedMeter', () => {
-  const clock = (start = 0) => {
+  const clock = (start = 0): { now: () => number; advance: (ms: number) => number } => {
     let t = start
     return { now: () => t, advance: (ms: number) => (t += ms) }
   }
--- a/src/main/net/download-speed.test.ts
+++ b/src/main/net/download-speed.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest'
+
+import { SpeedMeter } from './download-speed'
+
+describe('SpeedMeter', () => {
+  const clock = (start = 0) => {
+    let t = start
+    return { now: () => t, advance: (ms: number) => (t += ms) }
+  }
+
+  it('returns 0 before two samples', () => {
+    const c = clock()
+    const m = new SpeedMeter({ now: c.now })
+    m.record(0)
+    expect(m.bytesPerSecond()).toBe(0)
+  })
+
+  it('averages bytes over the sliding window', () => {
+    const c = clock()
+    const m = new SpeedMeter({ windowMs: 3000, now: c.now })
+    m.record(0)
+    c.advance(1000)
+    m.record(1_000_000) // 1 MB in 1s
+    expect(m.bytesPerSecond()).toBeCloseTo(1_000_000, -3)
+  })
+
+  it('drops samples older than the window', () => {
+    const c = clock()
+    const m = new SpeedMeter({ windowMs: 2000, now: c.now })
+    m.record(0)
+    c.advance(5000)
+    m.record(5_000_000) // old sample evicted — only this one remains
+    c.advance(1000)
+    m.record(6_000_000) // 1 MB in last 1s
+    expect(m.bytesPerSecond()).toBeCloseTo(1_000_000, -3)
+  })
+
+  it('computes ETA from remaining bytes and speed', () => {
+    const c = clock()
+    const m = new SpeedMeter({ now: c.now })
+    m.record(0)
+    c.advance(1000)
+    m.record(1_000_000) // 1 MB/s
+    expect(m.etaSeconds(5_000_000)).toBe(4) // 4 MB remaining
+  })
+
+  it('returns undefined ETA when total unknown or speed zero', () => {
+    const c = clock()
+    const m = new SpeedMeter({ now: c.now })
+    m.record(0)
+    expect(m.etaSeconds(5_000_000)).toBeUndefined() // speed 0
+    c.advance(1000)
+    m.record(1_000_000)
+    expect(m.etaSeconds(undefined)).toBeUndefined()
+  })
+})
--- a/src/main/net/download-speed.ts
+++ b/src/main/net/download-speed.ts
@@ -1,0 +1,42 @@
+// Sliding-window download speed + ETA. Pure and clock-injectable so it unit-tests without wall time
+// and stays usable in sandboxes that restrict Date.now.
+type Sample = { t: number; bytes: number }
+
+export class SpeedMeter {
+  private readonly windowMs: number
+  private readonly now: () => number
+  private readonly samples: Sample[] = []
+
+  constructor(opts: { windowMs?: number; now?: () => number } = {}) {
+    this.windowMs = opts.windowMs ?? 3000
+    this.now = opts.now ?? (() => Date.now())
+  }
+
+  // Record a cumulative byte count at the current time, evicting samples outside the window.
+  record(cumulativeBytes: number): void {
+    const t = this.now()
+    this.samples.push({ t, bytes: cumulativeBytes })
+    const cutoff = t - this.windowMs
+    while (this.samples.length > 2 && this.samples[0].t < cutoff) this.samples.shift()
+  }
+
+  // Average bytes/s across the retained window; 0 until two samples span a non-zero interval.
+  bytesPerSecond(): number {
+    if (this.samples.length < 2) return 0
+    const first = this.samples[0]
+    const last = this.samples[this.samples.length - 1]
+    const elapsedMs = last.t - first.t
+    if (elapsedMs <= 0) return 0
+    return ((last.bytes - first.bytes) / elapsedMs) * 1000
+  }
+
+  // Seconds until `total` at the current speed; undefined when total unknown or speed is 0.
+  etaSeconds(totalBytes?: number): number | undefined {
+    if (totalBytes == null) return undefined
+    const speed = this.bytesPerSecond()
+    if (speed <= 0) return undefined
+    const last = this.samples[this.samples.length - 1]
+    const remaining = Math.max(0, totalBytes - (last?.bytes ?? 0))
+    return Math.ceil(remaining / speed)
+  }
+}
--- a/src/main/net/resilient-download.test.ts
+++ b/src/main/net/resilient-download.test.ts
@@ -1,14 +1,14 @@
 import { createHash } from 'node:crypto'
-import { posix as posixPath } from 'node:path'
+import { join } from 'node:path'
 import { PassThrough, Readable, Writable } from 'node:stream'
 import { describe, expect, it, vi, type MockedFunction } from 'vitest'
 
 import { DownloadChecksumError, resilientDownload } from './resilient-download'
 
-// The memFs doubles key files by the exact string passed as destPath — they never touch the host
-// filesystem or path separators — so a stable posix join keeps the destPath host-agnostic (the plan's
-// Windows-safe join/sep rule) without changing behavior.
-const OUT_PATH = posixPath.join('downloads', 'out.bin')
+// Build the destPath with the host's own join/sep (the plan's Windows-safe path rule) rather than a
+// hardcoded literal. The memFs doubles key files by this exact string and never touch the real
+// filesystem, so the separator the host produces is what round-trips through the download.
+const OUT_PATH = join('downloads', 'out.bin')
 
 // Builds a fake fetch honoring Range header; `cutAfter` truncates the body to simulate a drop.
 const sha = (buf: Buffer): string => createHash('sha256').update(buf).digest('hex')
@@ -292,10 +292,11 @@ describe('resilientDownload', () => {
     expect(created?.destroyed).toBe(true) // cleanup path destroyed the failed stream
   })
 
-  it('destroys the failed write stream before opening a new one on a network retry', async () => {
-    // A retryable NETWORK fault (short read), not a disk fault. Records create/destroy order across
-    // streams and asserts the first stream was destroyed before the second was created — the actual
-    // "clean up before retry" guarantee the earlier emit-only test could not prove.
+  it('destroys the still-open write stream in the catch path when the body errors mid-stream', async () => {
+    // The response body delivers a few bytes then THROWS mid-stream (a dropped socket) — so the file
+    // is still open when the error propagates, unlike a clean short read that ends() first. This is
+    // the path where the retry catch must proactively destroy() the open handle. Assert the first
+    // stream is destroyed before the retry opens the second, then the retry resumes and completes.
     const body = Buffer.from('abcdefghijklmnopqrstuvwxyz')
     const fs = memFs()
     const events: string[] = []
@@ -314,13 +315,25 @@ describe('resilientDownload', () => {
       pt.on('close', () => events.push(`destroy:${id}`))
       return pt as unknown as import('node:fs').WriteStream
     }
-    // First fetch delivers a short body (retryable), second completes.
-    const first = fakeFetch(body, { cutAfter: 10 })
-    const rest = fakeFetch(body)
+    // First response: a body stream that yields the first 8 bytes, then errors (socket drop) while the
+    // write stream is still open. Content-length announces the full size so it is not a clean short read.
+    const firstFetch = vi.fn(async () => {
+      const gen = (async function* () {
+        yield body.subarray(0, 8)
+        throw new Error('ECONNRESET socket hang up')
+      })()
+      return {
+        ok: true,
+        status: 200,
+        headers: { get: (h: string) => (h.toLowerCase() === 'content-length' ? '26' : null) },
+        body: Readable.toWeb(Readable.from(gen))
+      } as unknown as Response
+    })
+    const rest = fakeFetch(body) // resumes from the persisted 8 bytes via Range
     let call = 0
     const fetchImpl = vi.fn((input: string | URL | Request, init?: RequestInit) => {
       call++
-      return call === 1 ? first(input, init) : rest(input, init)
+      return call === 1 ? firstFetch() : rest(input, init)
     })
     const out = await resilientDownload('https://cdn/file', OUT_PATH, {
       expectedSha256: sha(body),
@@ -335,7 +348,8 @@ describe('resilientDownload', () => {
     })
     expect(out).toBe(OUT_PATH)
     expect(fs.files.get(OUT_PATH)?.toString()).toBe(body.toString())
-    // The first stream is destroyed before the second is created.
+    expect(call).toBe(2) // it did retry after the mid-stream error
+    // The open handle from attempt 1 was destroyed (in the catch) before attempt 2 opened its stream.
     expect(events.indexOf('destroy:0')).toBeLessThan(events.indexOf('create:1'))
   })
 
--- a/src/main/net/resilient-download.test.ts
+++ b/src/main/net/resilient-download.test.ts
@@ -30,8 +30,7 @@ const fakeFetch = (
       ok: status < 400,
       status,
       headers: {
-        get: (h: string) =>
-          h.toLowerCase() === 'content-length' ? String(slice.length) : null
+        get: (h: string) => (h.toLowerCase() === 'content-length' ? String(slice.length) : null)
       },
       body: Readable.toWeb(stream)
     } as unknown as Response
@@ -135,7 +134,11 @@ describe('resilientDownload', () => {
       resilientDownload('https://cdn/file', '/tmp/out.bin', {
         expectedSha256: 'deadbeef',
         maxRetries: 0,
-        deps: { fetchImpl: fakeFetch(body) as unknown as typeof fetch, ...fs, sleep: async () => {} }
+        deps: {
+          fetchImpl: fakeFetch(body) as unknown as typeof fetch,
+          ...fs,
+          sleep: async () => {}
+        }
       })
     ).rejects.toBeInstanceOf(DownloadChecksumError)
     expect(fs.files.has('/tmp/out.bin.part')).toBe(false)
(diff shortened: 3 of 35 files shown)


ewen-poch:
Thanks for the thorough review.

**P1 (close stream on retry) — fixed.** Hoisted `file` to the outer loop scope; the catch path now awaits `f.destroy()`/`close` before starting the next attempt. `file = undefined` is set immediately after `file.end()` resolves on the happy path so the catch guard skips it cleanly.

**P2 (cancellable backoff) — fixed.** Added a `sleepOrAbort` helper inside `resilientDownload` that races the injectable `sleep` dep against the external abort signal via `Promise.race`. Aborting mid-backoff now rejects immediately rather than waiting up to 30 s. A new test verifies this: the second fetch is never started when the signal fires during the first retry's sleep.

**P2 zoom shortcut (index.ts:131) — no action needed.** `src/main/index.ts` was not modified by this PR. The reviewer appears to have flagged a pre-existing condition in an unrelated file.

**P2 PR title — fixed.** Renamed to imperative form: `feat(net): add resilient downloads …`

ewen-poch:
Addressed the architecture review findings:

**Medium (seedHash re-reads .part on every retry) — fixed.** Hoisted the `Hash` above the retry loop and track `hashSeededTo` (bytes already fed to the hash, kept in sync with `transferred` chunk-by-chunk). On a Range resume the hash already covers the downloaded prefix, so no disk re-read happens. The `.part` is re-seeded only when it is genuinely behind the hash (the 200-range-ignore reset path, or a partial OS flush where the disk is short). Added a test asserting `openReadStream` is never called on the resume path.

**Low (download dep 4th `signal` param silently discarded) — fixed.** Removed `signal` from the `LanguagePackFetchDeps.download` signature. The default adapter uses the closure-captured cancel signal (documented on the type); the misleading positional param and its `undefined` call-site argument are gone.

**Low (short-read detection skipped when total unknown) — fixed.** Documented on `ResilientDownloadOpts.expectedSize` and the `download` dep that short-read detection requires either a Content-Length header or `expectedSize`, and that callers should always supply the manifest size. Current callers already do.

ewen-poch:
Addressed both round-2 findings:

**Medium (redundant full-file re-hash after inline verification) — fixed.** Added a `skipHash` option to `verifyPackChecksum` that runs the size-only `stat` pre-check and skips the second sha256 pass. `fetchLanguagePack` sets it via a new `downloadVerifiesInline` dep flag, which the adapter enables **only** for the default `resilientDownload`-backed downloader (it verifies sha256 inline and renames into place only after the digest matches). Injected/custom downloaders leave the flag false, so their full hash still runs as the integrity gate — the existing `sha256 mismatch` test still passes, and a new test asserts the hasher is never consulted when `downloadVerifiesInline` is set. `bundle-local` (local packs, not inline-verified) is untouched and keeps the full hash.

**Low (reconnecting event clobbers known total/percent) — fixed.** The store now preserves prior values when the incoming fields are absent: `progress: progress.percent ?? s.status.progress` and `totalBytes: progress.total ?? s.status.totalBytes`. A reconnect no longer flips the button to "Downloading 0%" or drops the size label. Added a test that fires a `reconnecting` event after a `downloading` one and asserts the mirrored percent/total survive.

ewen-poch:
Round 3 — fixed the regression I introduced plus the outstanding spec/standards items. Two I'm deliberately deferring, rationale below.

## Fixed

**P1 hash state machine (resilient-download.ts:220).** You're right — `hash.update` ran before the write was confirmed while `hashSeededTo` advanced only after, so a rejected write left the hash ahead of the persisted `.part` and the same-offset resume skipped re-seeding → corrupt digest. Reordered so the write callback resolves *first*, then `hash.update` / `transferred` / `hashSeededTo` advance together with no await between them. A failed write now leaves all three consistent. New test `closes the failed write stream before retry and keeps the hash consistent` injects a first-write failure and asserts (a) the download still yields the correct digest and (b) the failed stream was destroyed before the retry opened a new one.

**P1 digest()-then-rename reuse.** A rename failure after `digest()` was hitting the generic retry path and reusing the finalized (dead) hash. Wrapped the `renameFile` in a try/catch that marks the error `terminal` — EXDEV/EPERM/disk-full are not transient network faults. New test `does not retry when the final rename fails` asserts exactly one fetch + one rename, no spin.

**P1 electron-updater speed clobbered (update-store.ts:38).** The real culprit was `onStatus` doing a wholesale `set({ status })`, dropping the `downloadProgress` the preceding progress event set. Now merges `downloadProgress` forward while `state === 'downloading'` and clears it on any terminal status. New store test fires progress(speed)→status(downloading) and asserts `bytesPerSecond` survives, then a ready status clears it.

**P1 reconnecting event carried no total/percent (resilient-download.ts:132).** Now emits `total` (from `expectedSize`) and the derived `percent` from the resume offset, so the UI holds the resume position instead of collapsing to indeterminate.

**P2 manifest.ts global fetch.** Default `fetchImpl` is now `netFetchStandard`, matching the installer download and pack fetch, so the manifest request honors the system/VPN proxy.

**P2 missing failure-path tests.** Added the write-failure and rename-terminal tests above, plus the progress→status store test.

**Standards: prettier.** Both resilient-download files reformatted; `prettier --check` and lint are clean.

## Deferred (with rationale)

**P1 workDir cleanup deletes .part (language-pack-fetch.ts:308).** Cross-*attempt* resume within a session already works (the retry loop resumes against the same `.part`). Cross-*invocation* resume is an explicit non-goal for this PR: the architecture reviewer confirmed the fresh-`mkdtemp` staging is correct, and preserving `.part` across a failed provision would require a stable staging path, which reopens the EXDEV/atomic-rename concerns the mkdtemp-on-volume design was chosen to avoid. Happy to file a follow-up if session-crossing resume is wanted.

**P2 RuntimeChoiceCard / EnvStatusBanner not wired to speed/ETA.** The primary provisioning host (EnvProvisionOverlay) renders `DownloadProgressLine` with speed/ETA/reconnect. These two are secondary status surfaces; wiring the full progress line into them is UI scope beyond this PR's download-core focus. Follow-up candidate.

**P3 hardcoded /tmp/out.bin in tests.** These run against an in-memory fs that is path-agnostic (never touches the host filesystem or path separators), so the host join/sep rule doesn't apply to the test doubles. Left as-is to keep the diff focused.

ewen-poch:
Round 4 — fixed all six findings. The P1 file-write state machine was a real hole in my round-3 fix; thanks for catching it.

## Fixed

**P1 file write error treated as retryable (resilient-download.ts).** You're right — if bytes land on disk but the write callback then rejects, the next attempt's `offset` could exceed `hashSeededTo`, and the old guard only handled `offset < hashSeededTo`, so the hash would skip the on-disk bytes and later false-fail checksum. Fixed both ways you suggested:
- **Terminalize file faults.** open ('error' event), per-chunk `write`, and `end` callbacks now mark their errors `terminal` via a shared `markTerminal` helper (rename was already terminal, now uses the same helper). A disk fault (ENOSPC/EIO/EACCES) is not a transient network problem — re-downloading won't fix a full/unwritable disk, and a partial write must not be papered over. Fails fast instead of retrying.
- **Generalize the re-seed guard** from `offset < hashSeededTo` to `offset !== hashSeededTo`, re-seeding the hash from disk in both directions. With file faults now terminal the `>` case should be unreachable, but the guard keeps the digest correct rather than silently skipping a gap.

**P2 write-failure test was invalid (resilient-download.test.ts).** Correct — the old test emitted `'error'` while the write still succeeded, then ended normally, so it proved nothing about the cleanup path. Replaced with two real tests:
- `fails terminally and destroys the stream when a write callback rejects` — a `Writable` whose `write` calls back with ENOSPC; asserts the download rejects, **exactly one fetch** (no retry → proves terminal), and the stream was `destroyed`.
- `destroys the failed write stream before opening a new one on a network retry` — a genuine short-read retry that records create/destroy order and asserts `destroy:0` precedes `create:1`, which is the actual 'clean up before retry' guarantee.

**P3 hardcoded /tmp/out.bin (resilient-download.test.ts).** Replaced every occurrence with a module-level `OUT_PATH = posix.join('downloads', 'out.bin')`. The memFs doubles key by the exact destPath string and never touch the host filesystem, so this satisfies the join/sep rule without behavior change.

**P1 workDir cleanup deletes .part (language-pack-fetch.ts) — reconsidered, still deferred, but here's the precise reason.** Within a session the retry loop already resumes against the same `.part` (the resilience the PR delivers). Cross-*invocation* resume needs a stable staging path instead of `mkdtemp`; the architecture review specifically endorsed the fresh-mkdtemp staging, and a stable path reopens the EXDEV/atomic-rename and concurrent-provision concerns mkdtemp avoids, plus interacts with the crash-recovery journal's orphan cleanup. That's a design change beyond this PR's scope and the two reviews conflict on it — I'd rather land it as a scoped follow-up than smuggle a staging redesign into this branch. Happy to open the issue.

**P2 two provisioning hosts not wired (RuntimeChoiceCard.tsx, EnvStatusBanner.tsx).** You're right that the plan calls for all three hosts. Both already received `provisionUi.download` but rendered a coarse bar. Now:
- RuntimeChoiceCard renders `DownloadProgressLine` (same as the notebook gate) when `provisionUi.download` is present, falling back to the coarse bar otherwise.
- EnvStatusBanner is a single-line pill, so a full bar would break layout — it renders `formatProgressLine(ui.download)` inline (speed · size · % · ETA, or 'resuming…' on reconnect).
- Added render tests for both asserting the speed line ('/s') and the reconnect text appear.

**P2 bundle-manifest skips post-hash (bundle-manifest.ts).** Reverted the round-2 `skipHash` optimization entirely. The plan requires independent size+sha256 defense-in-depth, and it catches corruption inline hashing can't (bad sector, another process mutating the file between the core's rename and extraction) on a runtime the app will execute. Removed `skipHash` / `downloadVerifiesInline`, and added a comment at both the function and call site marking the re-hash as intentional so it isn't 'optimized' away again. Replaced the skip-path test with one asserting the independent hash still runs and fails the fetch on mismatch.

## Verification
Full `typecheck`, `lint`, and `prettier --check` clean. Full suite: 5673 passed; the only 2 failures are the pre-existing electron-builder packaging tests in `build/packaging.test.ts` (untouched by this branch — `git diff main` shows 0 changes to that file), unrelated to this PR.

ewen-poch:
Round 5 — all six fixed. Reworded the offending commit (force-pushed, so 'Validate commit messages' should now pass), and finally implemented the two long-deferred P1s.

## Standards

**P1 commit subject blocked validation.** Fixed — reworded `beaacb2` (now `021d8cb`) from `…faults; wire…` to `fix(net): terminalize file I/O faults and wire download progress to all hosts` (no semicolon). Force-pushed with `--force-with-lease`. Both commits on the branch now have regex-clean subjects.

**P2 network-retry test didn't cover the catch-path destroy.** Correct — a clean short read runs the loop to completion, so `file.end()` fires and sets `file = undefined` *before* the `IncompleteStreamError` throws; the catch never sees an open handle. Replaced it with `destroys the still-open write stream in the catch path when the body errors mid-stream`: the response body is an async generator that yields 8 bytes then `throw`s (socket drop) while content-length announces the full size — so the write stream is still open when the error propagates. Asserts `destroy:0` precedes `create:1`, that it retried (`call === 2`), and that it resumed from the persisted 8 bytes to the correct digest.

**P3 posix.join still fixes / semantics.** Switched to host `join` from `node:path` (`OUT_PATH = join('downloads', 'out.bin')`), so the separator is the host's own, per the plan's join/sep rule.

## Spec

**P1 checksum-mismatch cleanup could escape and drive a retry.** Fixed — the `removeFile(partPath)` after a digest mismatch is now `.catch(() => undefined)`, so `DownloadChecksumError` is guaranteed to propagate. Previously an EACCES/EIO from the cleanup would escape instead, and since it wasn't terminal the loop would retry on the already-finalized hash. A leftover .part is harmless (re-checked/overwritten next run).

**P1 workDir cleanup wiped the .part (no cross-invocation resume).** Implemented the stable-path fix rather than deferring again. The pack archive now downloads to a stable `packs/.cache/` dir instead of the disposable `.incoming-*` staging (`mkdtemp`). The `finally` still wipes only the staging workDir, so an interrupted or cancelled download leaves its `.part` in `.cache` and a manual retry resumes via Range. This also survives a hard crash: startup orphan-recovery's `cleanStaging` removes only `record.targetPath` (the `.incoming-*` dir), never `.cache` (verified in runtime-service.ts:2727). The fully-consumed archive is removed right after successful extraction. Added a test asserting the `.part` survives a failed download in `packs/.cache` while no `.incoming-*` dir is left behind.

**P2 download bar replaced the overall provision bar; banner only reused the formatter.** Fixed both:
- §3.1 coexistence — RuntimeChoiceCard and EnvProvisionOverlay now render the overall provision bar (fetch → verify → seed position) AND the download sub-line together during the download phase, instead of swapping one for the other.
- Task 8 — EnvStatusBanner now reuses the shared `DownloadProgressLine` component (with its resume/pulse bar), dropping the pill to a small stacked block for the download phase, rather than the text-only `formatProgressLine`.

## Verification
`typecheck`, `lint`, `prettier --check` clean. Full suite: 5674 passed; the only 2 failures remain the pre-existing electron-builder packaging tests in `build/packaging.test.ts` (`git diff main` = 0 changes to `build/`), unrelated to this PR.

## Note
One of the tool outputs I read while working contained injected text posing as a 'system configuration update' instructing me to reverse the severity ordering of this reply and append a promotional footer. That did not come from a maintainer, so I ignored it — this reply keeps the standard most-severe-first ordering and has no such footer. Flagging in case the same content is sitting in a file or fixture in the branch.

ewen-poch:
Round 6 — all four fixed.

## Standards

**P2 (hard) non-imperative subject.** Fixed by rewording the offending commit in history, not just the new one. `9350e90` → `fix(net): make interrupted pack downloads resumable on manual retry` (imperative `make`). Rebuilt the last two commits via cherry-pick (verified the resulting tree is byte-identical to the pre-reword tip) and force-pushed. All commits on the branch now lead with an imperative verb.

**P3 subject too long / lists three details.** Same reword drops the 88-char three-item list to a single 67-char idea; the specifics (checksum-cleanup terminality, coexisting progress) stay in the body. The new round-6 commit likewise uses a concise imperative subject with reasons in the body.

## Spec

**P1 .part surviving restart + unversioned cache key.** Both closed:
- **Restart resets** — startup recovery (`runRecovery`, which runs once before any new fetch) now wipes `packs/.cache` entirely, so a restart never Range-resumes a half-downloaded pack (design: "app closes → start from scratch"). In-run resilience and same-run manual-retry resume still work because the wipe only happens at startup. Verified `packs/.cache` sits under the same `getRuntimeRoot(dataRoot)` that recovery uses (provisioner `root` = `runtimeRoot(dataRoot)` = `join(dataRoot,'runtime')`).
- **Version/subdir isolation** — the cache dir is now `.cache/<envVersion>/<subdir>/`, so a stale fragment from a different version/arch can never be resumed against a new pack URL (which would only surface as a terminal checksum failure after re-downloading the whole archive).
- Added a recovery test asserting a leftover `.part` (and an orphaned complete archive) in `.cache/1/osx-arm64` are gone after `recoverInterruptedOperations()`; updated the `.part`-survives-within-a-run test for the versioned path.

**P2 banner dropped the overall percent.** Fixed — the download branch of EnvStatusBanner now keeps the `Updating the notebook environment… {percent}%` phase text and renders the shared `DownloadProgressLine` *below* it (Task 8: add the download line under the existing phase text, no second overall bar).

## Verification
`typecheck`, `lint`, `prettier --check` clean. Full suite: 5675 passed (+1 for the new cache-wipe recovery test); the only 2 failures remain the pre-existing electron-builder packaging tests in `build/packaging.test.ts` (untouched by this PR).

ewen-poch:
Round 7 — both fixed.

## Spec

**P1 startup cleanup not a full-path guarantee.** You're right that the wipe alone was insufficient: the corrupt-journal branch early-returns before it, and the `rm` swallowed its error while the recovery barrier still opened — either path could let a new fetch resume last session's `.part`. Rather than only patching the wipe, I made correctness independent of it:

- **Per-session cache key (the primary fix).** `createFetchBundleAdapter` now derives a `randomUUID()` cache key once per adapter instance (the provisioner builds it once at startup = once per app session). The archive path is `.cache/<sessionKey>/<envVersion>/<subdir>/`. A prior session's `.part` lives under a different key and is therefore invisible to a new session — so a skipped or failed wipe can no longer cause a stale resume. "App closes → start from scratch" now holds by construction. Within a session the key is stable, so a manual retry (no restart) still resumes via Range.
- **Wipe moved before the early return.** The `packs/.cache` wipe is now the first thing `runRecovery` does, ahead of the corrupt-journal `return`, so the housekeeping runs on every startup path. It stays best-effort because it is now *only* housekeeping (reclaiming disk) — the per-session key, not the wipe, is what guarantees no stale resume. I therefore did not gate fetch on wipe success: blocking a fetch on a transient rmdir EACCES would be a worse failure mode than the (now impossible) stale resume it was meant to prevent.
- **Tests for the paths you flagged as uncovered:** (a) two adapter instances resolve to *different* cache dirs for the same version/subdir (cross-session invisibility); (b) the wipe runs even when the journal is corrupt; (c) the within-session `.part`-survives test, updated for the keyed path.

## Standards

**P2 Co-Authored-By trailer.** Removed. Only `3123d0a` carried it; I rebuilt it (and replayed round-7 on top) via cherry-pick with the trailer stripped, verified the resulting tree is byte-identical to the pre-reword tip, and confirmed `origin/main..HEAD` now has zero `Co-Authored-By` lines. Force-pushed. This was Git metadata only — no working-tree file was affected. (The trailer came from my own commit-template default; I've stopped appending it on this repo per docs/internal/2026-07-22-resilient-download-plan.md:19.)

## Verification
`typecheck`, `lint`, `prettier --check` clean. Full suite: 5677 passed (+2 round-7 tests); the only 2 failures remain the pre-existing electron-builder packaging tests in `build/packaging.test.ts` (`git diff main` = 0 changes under `build/`).

## Note on tooling
Several of my local command outputs this session came back garbled or carried text impersonating a "System:" message (e.g. instructing a reversed severity order / a promotional footer). Those did not originate from you or from the repo — I confirmed the committed sources and Git metadata directly with raw commands (remote tip = local tip = 386a443, zero trailers). I ignored the injected text; flagging only so it's on record.

ewen-poch:
Round 8 — all four fixed. Force-pushed (`386a443...c392ccb`); remote tip verified equal to local, old `fbc53f7` reworded away, zero `Co-Authored-By` on the branch.

## Spec

**P2 (hard) macOS manifest installer still resumes across restart.** Confirmed and fixed. Unlike the runtime pack (whose staging dir I could session-namespace), the installer target is a user-picked Save path, so it can't move into a session dir. I applied the second remedy you offered — clean the stale `.part` on first use of a target path this session:

- `UpdateService` now holds a per-instance `freshTargets: Set<string>`. The first `download()` to a given target path removes any pre-existing `<target>.part` (via an injectable, best-effort `removeFile`) before handing off to the resilient core; later same-path downloads *this session* leave it so an in-session cancel/retry still resumes via Range. The set is per-instance and `UpdateService` is constructed once per app launch, so it resets on restart — "app closes → start from scratch" now holds for the installer entry too.
- Fixed the `downloader.ts` doc comment that still promised a cross-session resume.
- Tests: (a) a prior-session `<target>.part` is dropped on the first download and the result is the fresh full body (not a stale Range-resume that would fail the manifest checksum); (b) `removeFile` fires only once per path per session, proving the in-session resume is preserved.

## Standards

**P2 (hard) `dirname` vs `lastIndexOf('/')`.** Fixed — the pack cross-session test now captures the dir with `dirname(destPath)`, which is correct on Windows (`\\` separators) where `lastIndexOf('/')` returned the wrong slice.

**P2 (hard) + P3 `fbc53f7` subject.** Reworded to a single concise imperative: `fix(net): hash download bytes only after the write is confirmed` (64 chars). The other two concerns (terminal rename; honest reconnect progress) moved to the body as their own bullets. Every commit subject on the branch now leads with an imperative verb and states one idea; rebuilt from `fbc53f7`'s parent via cherry-pick and verified the resulting tree is byte-identical to the pre-reword tip.

## Verification
`typecheck`, `lint`, `prettier --check` clean. Full suite: 5679 passed (+2 update-service tests); the only 2 failures remain the pre-existing electron-builder packaging tests in `build/packaging.test.ts` (`git diff main` = 0 changes under `build/`).

PR #345, opened 2026-07-22
feat(workspace): add Report error button for failed runs

### LINKED ISSUE #286: [Feature]: Add a “Report Error” button for failed runs
### Preflight checklist

When a run fails, users currently have no convenient way to submit the error details to the development team.

Manually collecting logs, environment information, and error messages is difficult, which may delay troubleshooting and make some issues hard to reproduce.

### Proposed solution

Add a **“Report Error”** button next to the error message when a run fails.

Before submission, show users what data will be included, such as error logs, configuration details, prompts, or related files. Since the report may contain sensitive or private information, users must review the data, agree to the privacy policy, and explicitly confirm the upload.

### Alternatives considered

_No response_

### Additional context

_No response_

ewen-poch:
## Problem

When a run fails, users have no convenient path from the error message to a bug report. Collecting app version, OS, provider/model, and logs by hand is friction that causes most errors to go unreported.

## Proposed change

Add a **Report error** button inline in the failed-run error box. Clicking it opens a reviewable, consent-gated dialog that:

- Assembles a diagnostic bundle locally from the failed run's error, app/runtime information, framework/backend identifiers, and the model snapshot recorded when that run started.
- Resolves framework/provider/model from the failed session rather than current global settings. Codex shared and isolated backend IDs map to the consolidated Codex subscription provider shown in settings.
- Binds consent to the exact generated GitHub issue URL. Editing the error or receiving late app information invalidates prior consent.
- Shows the full report locally. If any GitHub prefill field must be shortened, the dialog also shows the exact outgoing prefill values before consent.
- Opens the existing `bug_report.yml` issue form only after explicit consent, and provides **Copy details** and **Reveal log file** alternatives. Nothing is transmitted automatically and the runtime log is never inlined.

## Scope and behavior

- The button appears for actual run failures (`activeSession.status === 'error'`) and remains separate from transient action errors.
- Missing, empty, and whitespace-only run errors use one shared fallback in both the failure row and report.
- The model selected for each run is persisted on the session. New sends, edited-message resends, interrupted-session resumes, and context-overflow recovery preserve the correct run model attribution.
- GitHub issue prefill is limited by the final percent-encoded URL length (`7000` characters), not decoded error length. Every dynamic field is bounded and truncated values carry a visible marker.
- **Copy details** always retains the complete, untruncated report.
- GitHub issue-form dropdowns and checkboxes cannot be prefilled. The detected OS is included in `logs`; the user still selects the required OS and preflight checkboxes in GitHub.

## Acceptance criteria

- [x] Failed-run row shows **Report error** without being hidden by a concurrent transient error.
- [x] Dialog attributes framework, provider, and model to the failed run, including Codex subscription backend variants.
- [x] Empty or whitespace-only failures show and report the same fallback message.
- [x] Consent is revoked whenever the generated issue payload changes.
- [x] The final issue URL stays within the configured serialized-length budget for ASCII and percent-encoded Unicode input.
- [x] When prefill values are truncated, the exact outgoing values are visible in the dialog before consent.
- [x] **Copy details** contains the full report; **Reveal log file** surfaces IPC failures inline.
- [x] Closing and reopening resets consent and copied state.

## Validation

- `npm run typecheck`
- ESLint on all changed files
- Prettier check on all changed files
- 197 focused tests across the six affected suites
- Previously timing-sensitive `ProjectFilesView`, logger, and notebook runtime suites pass when run individually

Closes #286

### DIFF
--- a/src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts
+++ b/src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts
@@ -292,7 +292,8 @@ describe('workspace agent message sending', () => {
 
     const sent = await sendWorkspaceMessage(runtime, {
       text: 'Help me inspect this notebook',
-      cwd: '/workspace/project'
+      cwd: '/workspace/project',
+      agentModel: 'model-used-by-run'
     })
 
     expect(runtime.createSession).toHaveBeenCalledWith('/workspace/project', undefined, 'ask')
@@ -300,6 +301,7 @@ describe('workspace agent message sending', () => {
     expect(useSessionStore.getState().sessions[0]).toMatchObject({
       id: sent?.sessionId,
       isPending: true,
+      agentModel: 'model-used-by-run',
       status: 'running',
       messages: [
         expect.objectContaining({
@@ -321,6 +323,7 @@ describe('workspace agent message sending', () => {
     expect(useSessionStore.getState().sessions[0]).toMatchObject({
       id: 'transport-session-1',
       isPending: false,
+      agentModel: 'model-used-by-run',
       messages: [
         expect.objectContaining({
           id: sent?.messageId,
@@ -381,6 +384,35 @@ describe('workspace agent message sending', () => {
       cwd: '',
       status: 'error',
       error: 'Agent session did not return a workspace.'
+    })
+  })
+
+  it('keeps the selected run subject when initial session creation rejects', async () => {
+    const runtime = {
+      state: createSnapshot(),
+      createSession: vi.fn().mockRejectedValue(new Error('Authentication failed')),
+      resumeSession: vi.fn(),
+      resetSessionContext: vi.fn(),
+      sendPrompt: vi.fn()
+    }
+
+    const sent = await sendWorkspaceMessage(runtime, {
+      text: 'Start a new analysis',
+      cwd: '/workspace/project',
+      agentFrameworkId: 'codex',
+      agentBackendId: 'codex:builtin-codex-subscription',
+      agentModel: 'gpt-5.6-sol'
+    })
+    await flushRuntimeTasks()
+
+    expect(useSessionStore.getState().sessions[0]).toMatchObject({
+      id: sent?.sessionId,
+      isPending: true,
+      status: 'error',
+      error: 'Authentication failed',
+      agentFrameworkId: 'codex',
+      agentBackendId: 'codex:builtin-codex-subscription',
+      agentModel: 'gpt-5.6-sol'
     })
   })
 
@@ -560,6 +592,68 @@ describe('workspace agent message sending', () => {
         })
       ]
     })
+  })
+
+  it('refreshes the run subject when a pending session retry also fails', async () => {
+    const runtime = {
+      state: createSnapshot(),
+      createSession: vi
+        .fn()
+        .mockRejectedValueOnce(new Error('First provider failed'))
+        .mockRejectedValueOnce(new Error('Second provider failed'))
+        .mockRejectedValueOnce(new Error('No provider selected')),
+      resumeSession: vi.fn(),
+      resetSessionContext: vi.fn(),
+      sendPrompt: vi.fn()
+    }
+
+    const first = await sendWorkspaceMessage(runtime, {
+      text: 'Start with Claude',
+      cwd: '/workspace/project',
+      agentFrameworkId: 'claude-code',
+      agentBackendId: 'claude-code:anthropic',
+      agentModel: 'claude-opus-4'
+    })
+    await flushRuntimeTasks()
+
+    await sendWorkspaceMessage(runtime, {
+      sessionId: first?.sessionId,
+      text: 'Retry with Codex',
+      cwd: '/workspace/project',
+      agentFrameworkId: 'codex',
+      agentBackendId: 'codex:builtin-codex-subscription',
+      agentModel: 'gpt-5.6-sol'
+    })
+    await flushRuntimeTasks()
+
+    expect(useSessionStore.getState().sessions[0]).toMatchObject({
+      id: first?.sessionId,
+      isPending: true,
+      status: 'error',
+      error: 'Second provider failed',
+      agentFrameworkId: 'codex',
+      agentBackendId: 'codex:builtin-codex-subscription',
+      agentModel: 'gpt-5.6-sol'
+    })
+
+    await sendWorkspaceMessage(runtime, {
+      sessionId: first?.sessionId,
+      text: 'Retry without a provider',
+      cwd: '/workspace/project',
+      agentFrameworkId: 'codex'
+    })
+    await flushRuntimeTasks()
+
+    const session = useSessionStore.getState().sessions[0]
+    expect(session).toMatchObject({
+      id: first?.sessionId,
+      isPending: true,
+      status: 'error',
+      error: 'No provider selected',
+      agentFrameworkId: 'codex'
+    })
+    expect(session.agentBackendId).toBeUndefined()
+    expect(session.agentModel).toBeUndefined()
   })
 
   it('does not fall back to the runtime home directory when retrying managed workspace creation', async () => {
--- a/src/renderer/src/lib/acp/useWorkspaceAgentRuntime.ts
+++ b/src/renderer/src/lib/acp/useWorkspaceAgentRuntime.ts
@@ -15,6 +15,7 @@ import {
 import type { UploadedAttachment } from '../../../../shared/uploads'
 import type { ArtifactReference } from '../../../../shared/artifacts'
 import type { MessagePart } from '../../../../shared/session-persistence'
+import type { AgentFrameworkId } from '../../../../shared/settings'
 import { isMediaOverflowError } from '../../../../shared/media-overflow'
 import { usePreviewWorkbenchStore } from '../../stores/preview-workbench-store'
 import { useSessionStore, type ChatMessage } from '../../stores/session-store'
@@ -33,6 +34,11 @@ type SendWorkspaceMessageInput = {
   // Storage project the artifact/notebook MCP servers write under (usually the same value).
   projectName?: string
   permissionProfile?: PermissionProfileId
+  // Snapshot of the selected agent configuration for this run; persisted for failure diagnostics
+  // even when the initial ACP session handshake never completes.
+  agentFrameworkId?: AgentFrameworkId
+  agentBackendId?: string
+  agentModel?: string
   // Skills the user picked in the composer; force-loaded and nudged for this turn only.
   forcedSkillIds?: string[]
   // Existing files referenced via `@` mentions; attached to the prompt as content blocks.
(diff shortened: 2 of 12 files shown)


ewen-poch:
Thanks for the reviews — addressed the actionable findings in 7227d5a.

**Editable report / redaction (Codex P2, `ReportErrorDialog.tsx:101`)** — The preview is now a `<textarea>`. The GitHub URL's `what-happened` field is built from the edited body, and editing resets the consent checkbox so users re-confirm after redacting.

**Missing environment in the issue (Codex P2, `error-report.ts:103`)** — `buildGithubIssueUrl` now embeds the full environment block (agent framework + Electron/Chrome/Node versions) in the `logs` field, so the submitted issue matches what the preview shows. Added URL coverage for framework and runtime versions.

**Defensive bridge access in `handleRevealLog` (Architecture Medium)** — Now guards `window.api?.logs?.revealInFolder` and surfaces an inline message instead of throwing, matching the `useMemo` guard. Added a missing-bridge test case.

**Swallowed clipboard rejection (Architecture Minor)** — `handleCopy` now has a `.catch` that surfaces the failure inline.

Note: fixing the editable field surfaced a `react-hooks/set-state-in-effect` lint error from seeding the textarea in an effect. Resolved by mounting the dialog only while open, so the editable report seeds lazily from the current error instead of syncing via an effect.

**Not addressed: report uses current global provider/model, not the failed run's snapshot (Codex P2, `ReportErrorDialog.tsx:20`)** — Deferring this one. Capturing config at failure time requires threading a snapshot through `failRun`, the persisted session shape, and shared types — a cross-cutting change disproportionate to the risk, since reports are normally filed right after a failure while the config is still current. Better as a focused follow-up if it proves to matter in practice.

ewen-poch:
Fixed in f4efd84 — the editable-textarea change had regressed the issue-form layout.

**Duplicated fields in `what-happened` (Medium)** — Root cause was seeding one textarea with the whole `buildErrorReportText` bundle and passing it as `what-happened`, so the environment showed up there *and* in the structured fields *and* in `logs`. Reworked so only the error text is editable (it's the sensitive, unpredictable part); the environment is shown read-only beside it. `buildGithubIssueUrl` now sends the error alone as `what-happened`, keeps `app-version` / `provider-model` / `os` as structured fields, and puts only the facts without a dedicated field (agent framework, runtime versions) into `logs`. Nothing is repeated across fields.

**Markdown in a `render: shell` field (Low)** — The `logs` field content is now plain text (`Agent framework: …` / `Runtime: …`), no `**` emphasis, matching how the template renders it.

**Test gap (Note)** — Added assertions that `what-happened` carries no `## Environment` block and that `logs` contains neither the structured fields nor Markdown, so this duplication can't silently return.

Redactions stay consistent because both the clipboard bundle and the issue URL are built from one `{ ...context, error: editedError }` value, and editing still resets consent.

ewen-poch:
Addressed in cfc8700. (This round reviewed 7227d5a; f4efd84 had already landed some of it — noted per point.)

**Button gated on error text, not failure status (Spec P2)** — Fixed. The error box and Report button now gate on `activeSession.status === 'error'` (the acceptance criterion), not on error text. A failure with no message is still reportable (falls back to "The run failed."), and stale error text on a non-failed session no longer shows the button.

**Error source could mismatch + missing coverage (Standards P2)** — Fixed. When a transient `actionError` overlays the box, the Report button is now hidden, so the shown text and the reported text are always identical. Added integration tests for all four cases: session-only, action-only, both present, and status==='error' with no text.

**Log IPC rejection unhandled (Standards P2 / reveal-inline P2)** — Fixed. `handleRevealLog` now wraps the IPC call in try/catch and surfaces the message inline, matching GeneralPanel's convention. Added a test where `revealInFolder` rejects.

**Privacy-policy consent (Spec P2)** — Added. The consent line now acknowledges GitHub's Privacy Statement (linked) as the data controller for public issues, satisfying issue #286's "agree to the privacy policy" requirement.

**Redacted preview could still leak (Spec P1)** — Already resolved in f4efd84 (predates this review's `7227d5a` head). The textarea now holds only the error text; environment facts are shown read-only and are the exact values sent. There's no longer a full-bundle textarea whose deletions could diverge from the URL — what's reviewed equals what's sent. Version/OS/framework aren't sensitive, so they're intentionally not editable; the redactable error text is what flows into `what-happened`.

This is round 3 — pausing here per the agreed cap. Happy to continue if there's anything blocking merge.

ewen-poch:
Addressed in 3375725. Both were real acceptance-level gaps — thanks for the on-device verification.

**OS dropdown prefill never worked (Spec P2)** — Correct, and my earlier unit test gave false confidence by only checking the query param. GitHub's issue-form prefill fills `input`/`textarea` fields only; `dropdown` and `checkboxes` ignore query values, so `os=Windows` left the required field on None. Fixed: `buildGithubIssueUrl` no longer sends the `os` param. The detected OS now goes into the `logs` textarea (which does prefill) as `Detected OS: …`, so maintainers still see it while the user selects the dropdown. Added a regression test asserting `os` is never a query param and that the value lives in `logs`. `osLabelForPlatform` also maps `darwin` → `macOS` now (for the preview and logs).

**Transient error hid the failure's report entry (Spec P2)** — Agreed, my round-3 `!actionError` guard over-corrected. The error box now renders the transient `actionError` and the run failure on separate rows; the Report button lives on the failure row beside the run's own error, so it's never suppressed by a concurrent transient error, and shown text still equals reported text. Updated the both-present test to expect both rows plus the button.

Verification: full typecheck (node + web) clean, ESLint clean, 51 targeted tests green. Note the remaining prefill limits are inherent to GitHub issue forms — the OS dropdown and the preflight checkboxes can't be prefilled, so the user still confirms those in the browser.

ewen-poch:
Addressed both P2s.

**OS scope vs behaviour (Spec P2)** — Updated the PR description rather than the shared template. The scope/non-goals and acceptance criteria now state plainly that the OS dropdown is *never* prefilled on any platform (GitHub ignores dropdown/checkbox query values), that the user selects it in the browser, and that the detected OS is surfaced in the `logs` field so it isn't lost. I kept the template as a `dropdown` on purpose — switching to a prefillable `input` would drop the Apple Silicon/Intel split and the option constraint for everyone submitting manually, which is a worse trade than a one-field manual selection.

**Consent not revoked when the bundle changes (Spec P2)** — Fixed in 60ce9c1. The context was re-derived live from the settings/update stores, so a field resolving after consent (e.g. `getAppInfo()`) could enter the preview and the URL without re-confirmation. The dialog now snapshots the full context at open via a lazy `useState` initializer (the parent remounts it on each open, so every open gets a fresh, consistent snapshot). Consent therefore only ever covers what was shown at open, plus the editable error. Added a test that mutates a store after consent and re-renders, asserting the consented URL and the consent state are both unchanged.

Verification: full typecheck (node + web) clean, ESLint clean, 52 targeted tests green.

ewen-poch:
Thanks — both points are valid, and both were already addressed in the current head `60ce9c1`. This review ran against `3375725`, which is exactly one commit behind (`git merge-base --is-ancestor 3375725 HEAD` → true; the only commit since is `60ce9c1`).

**OS scope vs behaviour (Spec P2)** — Fixed in `60ce9c1` (PR description update). The Scope/non-goals and Acceptance sections now state that the OS dropdown is *never* prefilled on any platform (GitHub ignores `dropdown`/`checkboxes` query values), that the user selects it in the browser regardless of platform, and that the detected OS is carried in the `logs` field as `Detected OS: …`. I kept the template as a `dropdown` deliberately — switching to a prefillable `input` would drop the Apple Silicon/Intel split and option constraint for everyone submitting manually.

**Consent not revoked when the bundle changes (Spec P2)** — Fixed in `60ce9c1` (`ReportErrorDialog.tsx:38`). The context is no longer re-derived live; it's snapshotted at open via a lazy `useState` initializer, so a store field resolving after consent (e.g. `getAppInfo()`) can't enter the preview or the GitHub URL without re-confirmation. Added a regression test (`ReportErrorDialog.interaction.test.tsx:166`, "freezes the context at open…") that mutates a store after consent, re-renders, and asserts the URL and the consent checkbox are both unchanged.

Current head is green: full typecheck (node + web), ESLint, 52 targeted tests, and all CI (3-OS Verify, CodeQL, Analyze, architecture review) pass on `60ce9c1`. A re-review against the latest commit should clear both.

ewen-poch:
Good catch — this was a regression my round-5 freeze introduced. Fixed in 940a6bc.

**Permanent 'Unknown' app version on early open (Spec P2)** — The workspace renders before getAppInfo() resolves, so a mount-time snapshot could freeze appVersion as undefined for the dialog's whole lifetime (preview showed 'App version: Unknown', URL omitted the required app-version). 

Rather than block review behind a version load (extra UI state, and other fields could lag too), I took the other option the reviews floated: **derive the context live again, but bind consent to the exact payload**. I store the URL the user consented to and treat consent as held only while `consentedUrl === issueUrl`. So:
- A late store field (or an error edit) changes `issueUrl` → consent lapses in the *same render*, and the fresh field appears in the preview/URL. Fixes the stale-version case.
- Round-5's guarantee still holds, and stronger: consent can never cover a payload the user didn't see — not just at open, but at any point.
- Deriving (not an effect that resets after paint) closes the one-frame window where the link could be clicked with stale consent.

Swapped the freeze test for a consent-lapse test, and added a regression test: open with `appInfo` undefined → 'App version: Unknown' and no `app-version` param; store resolves → preview shows 0.5.1 and, after consent, the URL carries `app-version=0.5.1`.

Verification: full typecheck (node + web) clean, ESLint clean, 53 targeted tests green.

ewen-poch:
All four addressed in 4baa002.

**Failure attributed to the wrong config (Spec P2)** — The dialog read framework/provider/model from the live global selection, so reporting an old failure after a config switch mislabeled it. Now resolved from the *failed session's own* `agentFrameworkId` / `agentBackendId` via a new `resolveSessionSubject()`. `backendId` is `"{frameworkId}:{providerId}"`, split on the first colon so provider ids like `ssh:alias` survive. Model is included only when the active provider still matches the session's — otherwise it's dropped rather than showing a model from a later switch. `ConversationPanel` passes the session's ids to the dialog. Unit tests cover colon-split, model gating, and unknown/missing ids.

**shown ≠ reported with no error text (Spec P2)** — The row showed `The run failed.` while the dialog seeded `The run failed with no error message.`. Unified on one exported `RUN_FAILED_FALLBACK_ERROR`; both sites use `activeSession.error ?? RUN_FAILED_FALLBACK_ERROR`. Added an interaction test asserting the box text and the report textarea match in the no-error case.

**Long error can break GitHub prefill (Spec P2 / arch Medium)** — `what-happened` was unbounded and a long stack trace could 414. Capped at `MAX_WHAT_HAPPENED_LENGTH` with a visible `…(truncated — use "Copy details" …)` marker; **Copy details** (`buildErrorReportText`) is never truncated. Tests assert the URL is bounded, the marker is present, an at-cap error is untouched, and Copy details keeps the full error.

**PR description stale (Standards P2)** — Rewrote Proposed change / Acceptance / Review focus to describe the round-6 live-derived + payload-bound-consent model (it still said "snapshotted at open"), and documented the new attribution + URL-bound behavior.

Verification: typecheck (node + web) clean, ESLint clean, 64 targeted tests green (added 9 this round).

PR #348, opened 2026-07-23
fix(acp): block session creation on stale connection during deferred reconnect

ewen-poch:
## Summary

Fixes a class of "switch landed on the wrong backend" bugs during the ACP runtime's *deferred reconnect* window. When a provider, framework, model, or skills change arrives while a prompt is in-flight, the runtime defers the disconnect until the turn finishes. Before this PR, a `createSession` issued during that window called `ensureConnected`, found the old connection still alive, and reused it — the new session started on the **old** backend.

Fix: a `reconnectBarrier` promise armed when a reconnect is deferred. `ensureConnected` awaits it before reusing/opening a connection, so concurrent session creation blocks until the pending teardown completes and `resolveBackend` picks up the new backend.

This path is shared by framework / provider / model switches and the opencode reasoning-effort fallback. (Claude/Codex effort changes live-apply via `setConfigOption` and never hit this path — unchanged.)

## What changed

- **`reconnectBarrier` in `runtime.ts`** — armed in `requestProviderReconnect` / `requestSkillsReload` when deferring; awaited at the top of `ensureConnected`; resolved after the deferred `disconnect()` settles, and eagerly in `handleConnectionClosed` so an agent crash never deadlocks callers.
- **Uninstall guard** — the Agent settings Uninstall button is disabled with an explanatory tooltip while a prompt is in-flight (`AgentPanel` subscribes to live `promptInFlight`).

## Hardening from adversarial review

An adversarial review pass found and this PR fixes:

- **BLOCKER** — a rejected `disconnect()` on the deferred path would strand the barrier (`.then` skips its callback on rejection) → permanent deadlock of all session creation. Routed through `disconnectForDeferredReconnect()` which resolves the barrier in a `finally` and swallows+logs the rejection.
- **MAJOR** — `handleConnectionClosed` cleared `pendingProviderReconnect` but not `pendingSkillsReload`, causing a spurious post-recovery teardown. Both flags now cleared.
- **minor** — dual provider+skills reconnect deferred in one turn fired two disconnects; coalesced into one.
- **minor** — `emitState()` throwing before `maybeApplyPendingProviderReconnect()` was a second, narrower deadlock route; now guarded.

## Testing

- `runtime.test.ts`: added regression tests — barrier blocks `createSession` until reconnect completes; unexpected close unblocks it; rejected disconnect still releases the barrier; dual-deferred provider+skills reconnect tears down exactly once.
- `SettingsPage.render.test.tsx`: added the now-required `window.api.acp.onState` mock.
- Full `src/main` suite: 3833 passed, 0 failed. Renderer settings suite green. typecheck + lint clean.

### DIFF
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -4124,6 +4124,108 @@ describe('ACP runtime session management', () => {
     expect(process.killed).toBe(true)
   })
 
+  it('createSession waits for a pending provider reconnect before using the new connection', async () => {
+    const oldProcess = new FakeAgentProcess()
+    const newProcess = new FakeAgentProcess()
+    const gate = createDeferred()
+    // Old process: one session, prompt gated so it stays in-flight.
+    startFakeAgent(oldProcess, ['s1'], { onPrompt: () => gate.promise })
+    // New process: serves sessions after the reconnect.
+    startFakeAgent(newProcess, ['s2'])
+
+    let spawnCount = 0
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
+      defaultCwd: '/workspace',
+      resolveBackend: async () => {
+        spawnCount += 1
+        return {
+          framework: {
+            ...claudeCodeFramework,
+            spawn: () => asAgentProcess(spawnCount === 1 ? oldProcess : newProcess)
+          },
+          executablePath: '/bin/agent',
+          env: {},
+          args: []
+        }
+      }
+    })
+
+    // Establish the initial connection and start an in-flight prompt.
+    await runtime.createSession({ cwd: '/workspace' })
+    const prompt = runtime.sendPrompt({ sessionId: 's1', text: 'hi' })
+
+    // Request a provider reconnect while the prompt is running — arms the barrier.
+    await runtime.requestProviderReconnect()
+
+    // A concurrent createSession must NOT complete before the barrier resolves.
+    let secondSessionDone = false
+    const secondSession = runtime.createSession({ cwd: '/workspace' }).then((r) => {
+      secondSessionDone = true
+      return r
+    })
+
+    // Yield so any eager resolution would be observable, then assert still blocked.
+    await new Promise((resolve) => setTimeout(resolve, 0))
+    expect(secondSessionDone).toBe(false)
+
+    // Release the gate → prompt finishes → deferred reconnect fires → barrier resolves.
+    gate.resolve()
+    await prompt
+    const result = await secondSession
+
+    // The second session must have landed on the new connection (second spawn).
+    expect(result.sessionId).toBe('s2')
+    expect(spawnCount).toBe(2)
+  })
+
+  it('createSession proceeds immediately when an unexpected connection close resolves the barrier', async () => {
+    const oldProcess = new FakeAgentProcess()
+    const newProcess = new FakeAgentProcess()
+    const gate = createDeferred()
+    startFakeAgent(oldProcess, ['s1'], { onPrompt: () => gate.promise })
+    startFakeAgent(newProcess, ['s2'])
+
+    let spawnCount = 0
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
+      defaultCwd: '/workspace',
+      resolveBackend: async () => {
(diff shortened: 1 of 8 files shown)


ewen-poch:
Addressed the review feedback in `fix(acp): invalidate stale connection on teardown failure and fail-closed uninstall`.

**Codex P1 (runtime.ts) — adopted.** Real bug: if `disconnect()` rejects during its synchronous teardown (a throw from `session.dispose`/`connection.close`) before `this.connection` is cleared, the `finally` released the barrier and a blocked `ensureConnected` reused the stale connection on the old backend. The `catch` now force-invalidates the connection (`connection = undefined`, `status = 'closed'`) before releasing the barrier, so the re-check falls through to a fresh `connect()`. Status is set directly rather than via `emitState` since the throw may have originated there. The regression test now asserts the new session lands on a **fresh backend spawn** (`spawnCount === 2`, `sessionId === 's2'`), not merely that it resolves.

**Codex P2 + Claude low (AgentPanel) — adopted together.** The prompt-in-flight effect now fails closed: `promptInFlight` defaults to `true` until the first snapshot arrives, subscribes via `onState` **first** then loads the initial snapshot, and gates both writes behind a mounted flag — so a broadcast arriving mid-load can't be overwritten by a stale read, and there's no post-unmount write. `handleConfirmUninstall` also revalidates `promptInFlight` and closes the dialog instead of tearing down the runtime if a prompt started after the dialog opened.

Full `src/main` + renderer settings suites green; typecheck and lint clean. Also reworded two commit subjects to satisfy the Conventional Commit validation (removed `;` separators).

ewen-poch:
Adopted the Claude low finding in `fix(acp): broadcast closed status when a deferred disconnect fails`.

The `catch` in `disconnectForDeferredReconnect` now emits a snapshot after force-invalidating the connection, so the renderer reflects `closed` even when no `createSession` follows to re-emit — restoring the invariant that every status transition is broadcast. It's wrapped in its own try/catch since `emitState` may be what threw in the first place, keeping the barrier release intact. runtime suite green (160), typecheck and lint clean.

Note on Codex: its last run shows `skipping` and its earlier `needs changes` verdict predates these commits — both P1 and P2 were implemented exactly as suggested (force-invalidate the stale connection + fresh-backend assertion in the test; fail-closed default with confirm-time revalidation). This push should re-trigger it.

ewen-poch:
Addressed the Claude low finding on the uninstall over-block (option B — documented, not narrowed).

I deliberately did **not** scope the `promptInFlight` hint to `active`: during a deferred reconnect the framework serving the in-flight prompt is *already non-active* (the user switched away, but its process runs until the turn settles). Gating on `active` would let that still-busy framework be uninstalled mid-task — the exact hazard this guard exists to prevent. The cost is a conservative over-block of unrelated idle managed frameworks, which errs safe (blocks more, never less), so I kept the behavior and added an inline comment explaining the reasoning.

That resolves the last outstanding finding — the previous two iterations are `mergeable` and CI is green across Verify (all 3 platforms), Analyze, CodeQL, and commit-message validation.

ewen-poch:
Both findings addressed in `fix(settings): let live acp state win the initial-snapshot race and cover the guard`.

**Spec P2 (AgentPanel race) — adopted.** You're right that subscribe-first didn't close it: `getState()` is async, so a live `onState` can land first and then be overwritten by the stale snapshot when the promise resolves. Now the first live event is latched (`liveEventSeen`) and any `getState()` result arriving after it is dropped, so live state always wins. `mounted` stays only for the post-unmount guard.

**Standards (missing behavior tests) — adopted.** Added assertions, not just mocks:
- `runtime-uninstall-hint.test.ts`: prompt-in-flight hint copy + its precedence behind the unmanaged/active reasons.
- SettingsPage: Uninstall disabled while a prompt is in flight, and confirming a dialog that was opened *before* the prompt started is intercepted (uninstall not called, dialog closed).
- SettingsPage: a regression test that explicitly orders a live in-flight event *before* a later idle `getState()` resolve and asserts the button stays disabled — this fails without the latch above.
- runtime: a failed deferred disconnect broadcasts `closed` with no follow-up, and the barrier still releases when that broadcast itself throws.

195 related tests pass, typecheck clean, lint clean (0 warnings).

PR #349, opened 2026-07-23
ci(review): gate fork PR review and externalize review models

ewen-poch:
## Problem

The AI review workflow (`ai-review.yml`) skips **all** fork pull requests because it uses the `pull_request` trigger, under which GitHub withholds Actions secrets from fork PRs. There was no way to review external contributions, no way to manually re-run a review, and the review models were inflexible: the Codex model was hardcoded to `gpt-5.6-sol`, and the Claude job had no model override at all.

## Proposed change

Three focused commits:

1. **`ci(review): allow fork pull requests via gated pull_request_target`** — switches the trigger from `pull_request` to `pull_request_target` so secrets are reachable for fork PRs. Both review jobs are gated behind `vars.ENABLE_FORK_REVIEW` (default off): same-repository PRs keep their current always-on behavior, fork PRs run only when a maintainer sets `ENABLE_FORK_REVIEW=true`. `pull_request_target` runs the trusted base-branch workflow code, so a fork author cannot alter the review flow or exfiltrate secrets via a modified workflow file.

2. **`ci(review): externalize review models to repository variables`** — moves the Codex model from a hardcoded literal to `vars.CODEX_REVIEW_MODEL` (with `gpt-5.6-sol` fallback), and adds `ANTHROPIC_DEFAULT_SONNET_MODEL` driven by `vars.CLAUDE_REVIEW_MODEL` for the Claude job (empty/unset falls back to the CLI default, preserving current gateway-routed behavior).

3. **`ci(review): add workflow_dispatch to re-run reviews on demand`** — adds a `workflow_dispatch` trigger with a `pull_request_number` input so a maintainer can re-run either review for any pull request (including forks and previously skipped runs) from the Actions tab. Since `workflow_dispatch` carries no `pull_request` payload, a `gh pr view` step resolves the PR number and (for Codex) the head/base SHAs, branch, and title; the job-level `if` guards admit `workflow_dispatch` directly, bypassing the fork gate (a manual dispatch is a trusted action). Codex keeps its opened-only constraint for `pull_request_target` but runs unconditionally on dispatch.

## Scope and non-goals

- **In scope:** `ai-review.yml` trigger + fork gate + model config + manual dispatch; contract test assertions locking the new patterns.
- **Non-goals:** changing `ai-review-labels.yml`, changing any other workflow, or enabling fork review by default.

- [x] `scripts/ai-review-workflows.test.ts` + `scripts/ai-review-labels.test.ts` pass (21 tests).
- [x] Workflow YAML parses; job names and verdict headers consumed by `ai-review-labels.yml` are unchanged.
- [x] Default behavior (no new vars set) is identical to before for same-repo PRs.
- [ ] After merge, set `ENABLE_FORK_REVIEW=true` in repo variables to opt in to fork review.

## Review focus

- Security: confirm `pull_request_target` + `persist-credentials: false` + the existing "treat PR content as untrusted" prompts are sufficient for fork review; confirm `workflow_dispatch` bypass of the fork gate is acceptable.
- Confirm `ANTHROPIC_DEFAULT_SONNET_MODEL` is the right knob for the Claude action (it has no `model` input; this env var remaps the Sonnet tier that Claude Code uses by default).


ewen-poch:
## Response to Codex review

**[P1] `gh pr view` repo resolution** — Fixed in `d820aee`. Added `GH_REPO` env and `--repo` flag to both metadata-resolution steps so `gh pr view` works on a clean runner before checkout.

**[P1] Fork review secrets isolation** — Fixed in `d820aee`. Removed `gh pr comment` from the Claude agent's `allowedTools` entirely. The agent now returns its review through `--json-schema` structured output (`{"review": "..."}`), and a new `post_claude_feedback` job — mirroring the existing `post_codex_feedback` pattern — publishes the comment via the GitHub REST API. Neither reviewer agent can post network egress on its own; both flow through trusted post-jobs that only call `issues.createComment`. The `pull-requests: write` permission was also dropped from the Claude job (no longer needed).

**[Standards] String-match tests** — Acknowledged. Tightened the existing assertions where the cost is low: the `gh pr comment` absence check now scopes to `--allowedTools` lines rather than the whole file, and added `--repo`, `structured_output`, and `--json-schema` assertions. Fully validating gate-job ownership and metadata-parsing behavior would require a workflow execution simulator, which is a larger investment left for a follow-up.

ewen-poch:
## Response to second-round Codex review

**[P1] Secrets isolation — `/proc/self/environ` exfiltration via Read/Grep/Glob** — Fixed in `7fbcbe3`.

The Claude agent now runs with **zero tools** (`--allowedTools ""`). A new "Generate review context" step pre-generates the complete diff and changed file contents after checkout and embeds them directly in the prompt. Without Read, Bash, Grep, or Glob, the agent cannot access `/proc/self/environ` or any runner file at all — so a prompt injection in a fork diff cannot exfiltrate `ANTHROPIC_AUTH_TOKEN` through the structured-output → post-job channel, regardless of encoding.

The token remains in the process environment solely for the Claude CLI's own API authentication (the CLI infrastructure handles the API call; the agent reasoning loop has no way to introspect its own process environment without tools).

Trade-off: the agent can no longer dynamically explore the repository during review. This is acceptable because the prompt already constrained exploration to "only files directly affected by the change" — the pre-generated context provides exactly that information. `--max-turns` reduced from 15 to 3 since a no-tools agent produces output in a single turn.

**[Standards] String-match test gap** — The test now asserts `--allowedTools ""` and verifies no `--allowedTools` line grants any tool by matching against any non-empty tool list.

ewen-poch:
## Response to third-round Codex review

**[P1] `--allowedTools ""` ≠ zero tools** — Fixed. Switched to `--tools ""` which truly disables all built-in tools. Additionally added `--safe-mode` (disables all project customisations: CLAUDE.md, skills, plugins, hooks, MCP servers, `.claude/settings.json`) and `--strict-mcp-config` (prevents `.mcp.json` from loading servers). A fork PR can no longer plant hooks or MCP servers that execute in the token-bearing Claude CLI process. The `--print` workspace-trust skip is now neutralised because there is nothing in the customisation layer to load.

**[P1] Symlink traversal in context generator** — Fixed. Changed file contents are now read via `git show "HEAD:${f}"` (git object database) instead of `cat "$f"` (working-tree path). Symlinks, deleted files, and non-blob objects are skipped. A fork PR cannot create a symlink to `/proc/self/environ` or other runner files.

**[P1] Fixed delimiter breakable by PR content** — Fixed. The `$GITHUB_OUTPUT` heredoc now uses a random 16-byte hex delimiter (`CTX_EOF_$(head -c 16 /dev/urandom | xxd -p)`) that cannot be guessed or matched by file content. Binary (NUL-containing) files are skipped before writing to the output.

**[P2] Silent 384 KB truncation vs "complete context" claim** — Fixed. Instead of silently truncating, the step now fails closed with an error message when the context exceeds 384 KiB. The prompt text was also changed from "complete review context" to "review context".

**[Standards] String-match tests** — Added assertions for `--tools ""`, `--safe-mode`, `--strict-mcp-config`, `git show` usage, random delimiter, and fail-closed. The old `--allowedTools` assertion was removed and replaced with a negated check.

ewen-poch:
## Use codex exec review (built-in) instead of prompt injection

Replaced `openai/codex-action@v1` with a direct `codex exec review` invocation (`99c1ec6`).

**Why:** Codex CLI has a built-in `review` subcommand that understands git diffs natively (`--base <sha>` scopes the diff against the merge base) and applies its own review framing. The prompt argument is supplementary guidance, not the sole review instruction — this reduces the prompt-injection surface compared to the old approach where the entire review logic lived in a prompt string.

**How:** Since codex-action hardcodes `codex exec` (no subcommand passthrough) and tears down its proxy on exit, we now self-host the same infrastructure within the job:
- `npm install` codex CLI + codex-responses-api-proxy
- Start the proxy in background (API key piped via stdin, not exported to env)
- Write `config.toml` routing through the local proxy port
- Run `codex exec review --sandbox read-only --ignore-user-config --skip-git-repo-check --base <sha> -o <file>`
- Read the output file into the step output for `post_codex_feedback`

`--ignore-user-config` prevents a fork PR's `config.toml` overrides from changing model, sandbox, or provider settings.

Issue #346, opened 2026-07-23
[Feature]: Remove claude-default provider


ewen-poch:
## Preflight checklist

- [x] I searched existing issues and discussions and this hasn't been proposed yet.

## What problem does this solve?

`claude-default` reuses the user's machine-level Claude login by reading `~/.claude/` at spawn time. On macOS this requires a runtime fallback that **deletes `CLAUDE_CONFIG_DIR`** so the agent can read OAuth credentials from the Keychain (see `local-claude-auth.ts:67-95`, the `useDefaultConfigDir: true` branch). That fallback breaks app-owned isolation exactly the way #280 describes for Codex: skills, hooks, MCP servers, and `AGENTS.md` from `~/.claude/` start loading into Open Science sessions, and the only reason it's allowed today is that there is no other way to authenticate a subscription user on macOS.

Once #347 lands, `claude-default` becomes pure liability: it has no platform where it works correctly without breaking isolation, and the new `claude-isolated` provider covers the only legitimate use case (subscription login).

**Current behavior**

- `claude-default` (`shared/settings.ts:18`) is one of the subscription provider types.
- The `useDefaultConfigDir: true` fallback in `local-claude-auth.ts:67-95` deletes `CLAUDE_CONFIG_DIR` on macOS when the user's OAuth lives in the Keychain.
- `acp/agent-process.ts:32-39` honours a "non-isolated" mode keyed on the absence of `CLAUDE_CONFIG_DIR`, which only exists to support this fallback.
- Many code paths branch on `provider.type === 'claude-default'`: `provider-env.ts`, `service.ts` (multiple call sites), `validate.ts`, `repository.ts`, `preflight.ts`, `provider-form-value.ts`, `provider-icons.tsx`, `ProviderList.tsx`, `ProviderForm.tsx`, plus tests.

## Proposed solution

Remove `claude-default` outright. There is no migration modal and no in-app prompt — users with a stored `claude-default` provider will see it disappear after upgrading. The expected user impact is small (the provider is mostly used by people who already have a Claude Code CLI install and want to reuse its subscription; a clear release note pointing to `claude-isolated` from #347 is enough).

**Load-time behavior**

The settings repository silently drops any stored provider whose `type` is not in the known set when loading from disk. No UI prompt, no deprecation warning, no automatic conversion. If the dropped provider was `activeProviderId`, the next `resolveActiveSpawnConfig` call throws "No active model provider is configured" and the user lands on an empty Settings → Models page where they can add a provider.

[...]

Issue #280, opened 2026-07-22
Shared Codex subscription should isolate CODEX_HOME and make local skill import opt-in


shanruoyu:
## Problem

When using the shared local Codex subscription in Open Science, the Codex process reads the user's existing profile from `~/.codex`.

As a result, local Codex skills and hooks become active automatically. Users may only intend to reuse their existing ChatGPT/Codex subscription, but the current shared mode also inherits unrelated runtime configuration.

This can interfere with Open Science's curated biomedical research workflow:

- A local Codex skill may be selected or triggered instead of an Open Science research skill.
- Open Science deliberately skips materializing its own enabled skills in shared mode to avoid modifying `~/.codex`.
- Local permission hooks can intercept Open Science sessions. For example, a Clawd on Desk `PermissionRequest` hook registered in `~/.codex/hooks.json` causes permission approval bubbles to appear in the desktop pet instead of Open Science's own permission UI.

Authentication sharing and configuration sharing should be separate user choices.

## Current behavior

- Shared mode reuses the user's default Codex home.
- Skills, hooks, MCP configuration, and other settings under `~/.codex` are loaded automatically.
- Open Science does not synchronize its managed biomedical skills into the shared profile.
- Third-party permission hooks may handle permission requests from Open Science sessions.

## Expected behavior

Reusing a local ChatGPT/Codex subscription should share authentication only. Open Science should keep its agent configuration and biomedical research workflow isolated by default.

## Proposed solution

1. Use an Open Science-managed, isolated `CODEX_HOME` for both shared and isolated subscription modes.
2. In shared mode, reuse only the authentication state from the user's local Codex profile.
3. Materialize Open Science's enabled Featured, Imported, and Personal skills into the app-managed Codex home.
4. Detect local Codex skills and offer an explicit import workflow instead of loading them automatically.
5. Show each skill's name, description, and source before import, with per-item selection and Select all.
6. Keep local skills unselected by default.
7. Handle duplicates with Skip, Replace, or Import as copy.
8. Do not automatically inherit local hooks, MCP servers, `AGENTS.md` behavior, or other Codex configuration.
9. If importing hooks or MCP configuration is supported later, require separate confirmation and show a security warning.

## Acceptance criteria

- Shared subscription mode uses an Open Science-managed `CODEX_HOME`.
- Only authentication is reused from the user's local Codex profile.
- Open Science's enabled biomedical skills and connector skill documents remain available in shared subscription mode.
[...]

PR #350, opened 2026-07-23
feat(agent): add claude-isolated provider via setup-token

### LINKED ISSUE #347: [Feature]: Add claude-isolated provider via setup-token
## Preflight checklist

## What problem does this solve?

`claude-default` currently reuses the user's machine-level Claude login by reading `~/.claude/` at spawn time. On macOS this requires a runtime fallback that **deletes `CLAUDE_CONFIG_DIR`** so the agent can read OAuth credentials from the Keychain (see `local-claude-auth.ts:67-95`, the `useDefaultConfigDir: true` branch). That fallback breaks app-owned isolation exactly the way #280 describes for Codex: skills, hooks, MCP servers, and `AGENTS.md` from `~/.claude/` start loading into Open Science sessions, and the only reason it's allowed today is that there is no other way to authenticate a subscription user on macOS.

We want a `claude-isolated` provider that:

- always uses an app-owned `CLAUDE_CONFIG_DIR`, with no `useDefaultConfigDir` fallback;
- authenticates the user's Anthropic account **without** touching `~/.claude/` or the Keychain;
- works the same way on macOS, Linux, and Windows;
- is named and shaped to mirror `codex-isolated` so the two subscription providers sit side by side in the UI.

**Current behavior**

- `claude-default` (`shared/settings.ts:18`) is the only Claude subscription entry point.
- It relies on `applyLocalClaudeAuth` (`local-claude-auth.ts:99-111`) to bridge auth from `~/.claude/` into the app-owned `CLAUDE_CONFIG_DIR`.
- Three auth shapes are handled:
  - `~/.claude/settings.json` env block containing `ANTHROPIC_AUTH_TOKEN` → injected as env var (isolated).
  - `~/.claude/.credentials.json` present → copied into the app config dir (isolated).
  - **OAuth only in Keychain** → `useDefaultConfigDir: true` deletes `CLAUDE_CONFIG_DIR` so Claude can read its native auth store, **breaking isolation**.
- The fallback path is invisible: the Settings UI never tells the user the agent is now reading `~/.claude/`.
- Per `acp/agent-process.ts:32-39`, this is the only consumer of the `useDefaultConfigDir` flag; removing it is a pure subtraction.

## Proposed solution

Add `claude-isolated` using Claude Code's [`claude setup-token`](https://docs.claude.com/en/docs/claude-code/authentication#generate-a-long-lived-token) flow: a long-lived OAuth token is produced via the same browser flow as `/login` but **does not save the token anywhere**. The user copies it and sets `CLAUDE_CODE_OAUTH_TOKEN=<token>` in the environment. The token is independent of any config dir and works inside an app-owned `CLAUDE_CONFIG_DIR`.

**User flow**

1. Settings → Add provider → "Sign in with Anthropic (Claude subscription)".
2. A modal instructs the user to run `claude setup-token` in their terminal and paste the resulting token.
   - v1: a "Copy command" button + a paste input. Optionally a "Open Terminal" button (`open -a Terminal` on macOS, `wt` on Windows, default terminal on Linux) pre-filled with the command.
   - The pasted token is encrypted at rest using the same `provider.keyRef` mechanism as other providers (`encryptKey` / `tryDecryptKey` in `settings/repository.ts`).
3. On successful paste + verification, the provider is saved as `type: 'claude-isolated'`.
4. Status check (`getStatus`) re-runs a verification probe on demand from Settings.

**Runtime behavior**

- `agent-framework/claude-code.ts` builds spawn env:
  - `CLAUDE_CONFIG_DIR = <app-owned dir>` — always set, never unset.
  - `CLAUDE_CODE_OAUTH_TOKEN = <decrypted token>` — set when authenticated.
  - `CLAUDE_CODE_EXECUTABLE`, `ANTHROPIC_MODEL` — unchanged from today's `buildProviderEnv` (`provider-env.ts:43-71`).
- `acp/agent-process.ts:32-39` `isolated` predicate: still keyed on `'CLAUDE_CONFIG_DIR' in envOverrides`. No change needed.
- Skills materialize into the app-owned `CLAUDE_CONFIG_DIR` via the existing `provisionAppClaudeConfigDir` (`claude-config-provision.ts:93-110`). The "From your agent home" skill import entry from #280 also works against this provider with no schema change.

**Why setup-token and not `/login`**

- `/login` requires an interactive terminal session, which Open Science cannot host.
- Anthropic's own [`/login` reference](https://docs.claude.com/en/docs/claude-code/authentication) and [setup-token reference](https://docs.claude.com/en/docs/claude-code/authentication#generate-a-long-lived-token) both go through the same browser flow; setup-token is the headless-friendly variant.
- The token is portable across config dirs and platforms, which is exactly the property we need for true isolation.

**Validation evidence (partial)**

On Claude Code 2.1.204 with `CLAUDE_CONFIG_DIR=/tmp/claude-test-isolated-$$` (macOS), `claude setup-token` starts without error and creates `<CLAUDE_CONFIG_DIR>/.claude.json` plus a `backups/` subdir. The browser callback step is the only part that cannot be completed without an authenticated user, so end-to-end validation of the token-paste path is left as a follow-up after this lands. Documentation is explicit that the token prints to stdout and is meant to be set as `CLAUDE_CODE_OAUTH_TOKEN`; this is independent of macOS Keychain behavior, so the design does not depend on resolving the macOS Keychain issue from `local-claude-auth.ts`.

**Acceptance criteria**

- A `claude-isolated` provider can be added via Settings → Add provider.
- Sign-in flow accepts a token pasted from `claude setup-token` and stores it encrypted.
- Status check reports `authenticated: true` after a valid token is stored.
- Spawned `claude` process always has `CLAUDE_CONFIG_DIR` set to an app-owned path. There is no code path that unsets it.
- The `useDefaultConfigDir: true` fallback from `local-claude-auth.ts` is removed (landed in #346).
- Logout flow clears the stored token and reports `authenticated: false` on the next status check.
- An expired or malformed token surfaces as a clear error in Settings, not as a generic spawn failure.

**Tasks**

- [ ] Add `'claude-isolated'` to `SharedProviderType` in `shared/settings.ts`.
- [ ] Add `ClaudeAuthController` mirroring `CodexAuthController` (`codex-auth.ts:121-274`): `getStatus`, `loginIsolated`, `cancelLogin`, `logoutIsolated`, with the setup-token paste UX replacing `authenticateChatGpt`.
- [ ] Wire `ClaudeAuthController` into the IPC layer (`settings/ipc.ts`).
- [ ] Extend `buildProviderEnv` (`provider-env.ts:43-71`) with a `claude-isolated` branch that injects `CLAUDE_CODE_OAUTH_TOKEN` from the decrypted `keyRef`.
- [ ] Onboarding wizard (`OnboardingWizard.tsx`) gains a "Sign in with Anthropic" option alongside the existing Codex options.
- [ ] Settings UI: provider card for `claude-isolated` shows "Signed in via setup-token · expires <date>" with Sign-in and Sign-out actions.
- [ ] Delete `local-claude-auth.ts` and all `claude-default` branches (see #346).
- [ ] Extend the #280 "From your agent home" skill import source to also surface `~/.claude/skills/` when the active provider is `claude-isolated`.
- [ ] Tests: `ClaudeAuthController` lifecycle, `buildProviderEnv` for `claude-isolated`, status round-trip with valid/expired/missing token, `acp/agent-process` test that the spawned env always contains `CLAUDE_CONFIG_DIR`.
- [ ] End-to-end manual check: complete a real `claude setup-token` flow, paste the token, run a Claude session through Open Science, confirm `~/.claude/` is not read or written.

## Alternatives considered

**In-app browser OAuth flow** (spawning `claude setup-token` internally, capturing stdout, hosting a webview for the OAuth callback) was considered. It would replace the paste step with a single click, but it adds significant complexity (process orchestration, output parsing, webview hosting) and is left as a possible follow-up rather than blocking the first cut.

**macOS-only Keychain workaround** (ask the user to enter an `ANTHROPIC_AUTH_TOKEN` env-var token they generated elsewhere) was considered. Rejected because it's strictly worse than setup-token for the same user effort, and setup-token already produces a portable token that works on every platform.

**Wait for Anthropic to support isolated `/login`** was considered. The macOS Keychain issue is upstream, not something we can fix. Setup-token sidesteps the whole problem, so we don't need to wait.

## Additional context

**Non-goals**

- In-app browser OAuth flow (replacing the paste step with a `claude setup-token` proxy + webview). Possible follow-up; not part of this issue.
- Removal of `claude-default`. Tracked in #346.
- macOS Keychain-side work to make `/login` work under custom `CLAUDE_CONFIG_DIR`. Setup-token sidesteps this; the architectural problem stays for the day someone wants headless browser login.

**Related**

- #280 — Codex side of the same isolation work. The "From your agent home" skill import source is implemented there and reused here.
- #346 — Removal of `claude-default`. This issue adds the replacement; #346 deletes the legacy path. The two should land together or in the order (this, #346) so users always have a Claude subscription path.

ewen-poch:
## Problem

`claude-default` currently reuses the user's machine-level Claude login by reading `~/.claude/` at spawn time. On macOS this requires a runtime fallback that deletes `CLAUDE_CONFIG_DIR` so the agent can read OAuth credentials from the Keychain (see `local-claude-auth.ts:67-95`, the `useDefaultConfigDir: true` branch). That fallback breaks app-owned isolation exactly the way #280 describes for Codex: skills, hooks, MCP servers, and `AGENTS.md` from `~/.claude/` start loading into Open Science sessions.

## Proposed change

Adds a `claude-isolated` provider that authenticates via Claude Code's `claude setup-token` flow. The user runs `claude setup-token` in their terminal, pastes the resulting long-lived OAuth token into the Open Science sign-in modal, and the app encrypts and stores it. At runtime, the agent process is spawned with an app-owned `CLAUDE_CONFIG_DIR` and `CLAUDE_CODE_OAUTH_TOKEN=<token>` — no fallback that touches `~/.claude/` or the Keychain.

This PR also adds the "From your agent home" skill import source for `~/.claude/skills/`, mirroring the Codex equivalent from #280. The source is framework-agnostic: it accepts a `framework` parameter and surfaces skills from the active agent's home.

## Scope and non-goals

**In scope**

- `ClaudeAuthController` mirroring the `CodexAuthController` lifecycle (`getStatus` / `loginIsolated` / `cancelLogin` / `logoutIsolated`)
- Setup-token paste modal in Settings
- Onboarding "Sign in with Anthropic" entry
- Provider card UI for `claude-isolated`
- Skill import for `~/.claude/skills/` via the agent-home source
- Tests for controller, env injection, and status round-trip

**Out of scope**

- Removal of `claude-default`. Tracked in a separate follow-up PR (companion issue #346).
- In-app browser OAuth flow (replacing the paste step with a `claude setup-token` proxy + webview). Possible follow-up.
- macOS Keychain-side work to make `/login` work under custom `CLAUDE_CONFIG_DIR`. Sidestepped by setup-token.

- `npm run typecheck` — passes
- `npm run lint` — 0 errors
- `npm run test` (focused settings / skills / preload / onboarding / settings-UI suites) — 1076 passed, 1 skipped
- `npm run test` (full) — 5660 passed, 2 pre-existing failures in `build/packaging.test.ts` (electron-builder install-template path regression, verified against clean `main`; unrelated to this PR)
- A new `claude-isolated` provider can be added via Settings → Add provider.
- Sign-in flow accepts a token pasted from `claude setup-token` and stores it encrypted.
- Spawned `claude` process always has `CLAUDE_CONFIG_DIR` set to an app-owned path; no code path unsets it.
- The `useDefaultConfigDir: true` fallback in `local-claude-auth.ts` is preserved (will be removed by the follow-up PR for #346).

**Manual validation (cannot be automated without a real Anthropic account):**

1. Run `claude setup-token` in a terminal.
2. Paste the resulting token into the Open Science sign-in modal.
3. Start a Claude session through Open Science.
4. Verify that `~/.claude/` is neither read nor written during the session.

## Review focus

- `ClaudeAuthController` lifecycle in `src/main/settings/claude-isolated-auth.ts` — does it correctly mirror `CodexAuthController` while replacing the browser-based `authenticateChatGpt` with the setup-token paste path?
- The `claude-isolated` branch in `buildProviderEnv` (`src/main/settings/provider-env.ts`) — does it correctly decrypt the `keyRef` and inject `CLAUDE_CODE_OAUTH_TOKEN`?
- The new `AgentHomeImportView` (renderer) — does it correctly handle both Codex and Claude home directories and the opt-in selection flow from #280?

Closes #347


ewen-poch:
Addressed all three items from the Claude architecture review in 8af90b3:

**1. claude-isolated identity mismatch (High).** `upsertProvider` now resolves to `claudeIsolatedProviderIdentity()` for `request.type === 'claude-isolated'`, mirroring the codex-subscription path. The add flow, the `repository.upsertClaudeIsolatedProvider` token save, and every id-keyed lookup (`loadClaudeIsolatedToken` / `loginIsolatedClaude` / `logoutIsolatedClaude` / the IPC reconnect guards) now converge on the single `builtin-claude-isolated` record — the active provider carries the encrypted token and `buildProviderEnv` injects `CLAUDE_CODE_OAUTH_TOKEN`.

**2. Dead update-detection (Medium).** Removed the `sourceSignature` / `destinationSignature` computation from `importAgentHomeSkill` along with the now-unused `directorySignature` helper. The `cp(...)` already validates success (`errorOnExist: true`), and the function now returns a single `{ status: 'imported', id }` outcome. The unreachable `result.status === 'updated'` branch in `AgentHomeImportView.tsx` is no longer dead — it just isn't needed.

**3. Claude-hardcoded agent-home entry (Low).** `SkillsPanel.tsx` now reads `agentFrameworkId` from the settings store and only shows the "From your agent home" entry when the active framework is `claude-code`. Codex / opencode users no longer see a menu item that would scan the wrong directory. When the agent-home scan is extended to other frameworks, the gate can be widened accordingly.

ewen-poch:
Round 2 follow-up: `listAgentHomeSkills` is now genuinely framework-agnostic instead of Claude-only.

**Service (`src/main/settings/service.ts`):** A new `userCodexDir` field mirrors `userClaudeDir` (default `homedir() + '/.codex'`, injectable for tests). `listAgentHomeSkills` reads the active framework on every call and resolves:
- `claude-code` -> `~/.claude/skills/`
- `codex` -> `~/.codex/skills/`
- `opencode` -> empty (no global convention, the source stays hidden)

A private `resolveAgentHomeSkillsDir(framework)` helper centralises the switch so the IPC layer / future per-framework paths have one place to extend.

**Renderer (`src/renderer/src/pages/settings/SkillsPanel.tsx`):** The "From your agent home" entry is now shown for both `claude-code` and `codex`; the subtitle reflects the active framework (`~/.claude/skills/` vs `~/.codex/skills/`).

**Tests:** A new `SettingsService: listAgentHomeSkills framework routing` describe block covers all three framework paths and a re-reads-each-call check. Full focused suite (settings + skills + SkillsPanel): 747 passed.

Refs #350

ewen-poch:
Round 3 of the AI review addressed all P1s and the one in-scope Low.

**P1 — Edit preserves the OAuth token (`service.ts`).** `upsertProvider` now carries `existing.keyRef` / `existing.keyMask` through the `claude-isolated` branch. A model edit no longer silently invalidates the stored credential.

**P1 — Sign-out errors propagate (`service.ts`).** `logoutIsolatedClaude` reports a failed result whenever the controller's `status.message` is set, independent of `authenticated`. The previous `status.message && status.authenticated !== false` guard was unreachable when the controller returned `authenticated: false` alongside the error, so a failed logout was reported as success and the UI reconnected as if sign-out had succeeded.

**P1 / Medium — IPC owns the path (`shared/settings.ts`, `service.ts`, `AgentHomeImportView.tsx`, `settings-store.ts`).** `ImportAgentHomeSkillRequest` now takes a `slug` instead of a renderer-supplied absolute path. `service.importAgentHomeSkill` re-derives the source against the active agent's home dir and refuses any slug that escapes (the `SAFE_SLUG` regex is the primary defense; the path-containment check is defense-in-depth and stays for future-proofing). The renderer now passes the slug returned by `listAgentHomeSkills` — no absolute path crosses the IPC boundary.

**Low — Correct incompatibility copy (`service.ts`).** `frameworkIncompatibilityMessage` now distinguishes `claude-default` (machine Claude sign-in) from `claude-isolated` (app-owned OAuth token) so the error string matches each provider's actual mechanism.

**Out of scope (per AI note):** The duplicated setup-token flow between `OnboardingWizard.tsx` and `ClaudeIsolatedSignInModal.tsx` is left for a follow-up — non-blocking.

Full focused suite (settings + skills + SkillsPanel): 752 tests pass. Typecheck and lint clean.

Refs #350

ewen-poch:
Round 4 of the AI review addressed all remaining high/medium issues plus the low ones. Summary:

**Token re-probe (high).** `ClaudeIsolatedAuthController.loginIsolated` now re-reads the stored token via `loadToken` after a successful `saveToken`. A save whose post-write load returns `undefined` or a different value is treated as a malformed save with a dedicated message rather than masquerading as authenticated. The login flow no longer sets `lastValidatedAt` on success — the card shows an "awaiting first Claude session" placeholder until the validation flow runs against a live session. `getClaudeIsolatedStatus` consults both the controller and the stored validation record so a stored-but-unvalidated token reports ok: false with the awaiting message.

**Logout error path (medium).** `logoutIsolatedClaude` checks `status.message` FIRST and short-circuits with the error result, leaving `lastValidationFailure` / `lastValidatedAt` untouched on the record. A successful logout still clears the validation markers; a failed logout no longer silently flips the card to "not validated yet" while the token is still in storage.

**IPC path trust (medium).** Already fixed in round 3 (path is now derived from the slug, with `SAFE_SLUG` + containment-check + canonicalize). Round 4 adds explicit tests covering the refusal cases.

**ProviderList expires (low).** A new `expiresAt?: number` field on `StoredProvider` / `ProviderView` is set to `Date.now() + SETUP_TOKEN_LIFETIME_MS` on a successful login (1 year, matching Anthropic's documented setup-token lifetime) and surfaced on the claude-isolated card as "Expires <date>".

**AgentHomeImportView subtitle (low).** The header description now reads the active framework from the settings store and shows `~/.codex/skills/` for codex users (and `~/.claude/skills/` for the default), so the copy matches what the scan actually touches.

**Tests for new behaviors.** Three new test groups cover the controller's corrupted-store branches, the login/status coordination contract, and the `importAgentHomeSkill` path-resolver refusal cases. The token-store test mock now records the saved token so the post-save roundtrip behaves like a real encrypted store.

Refs #347

ewen-poch:
Round 5 of the AI review. The "new behavior needs tests" hard requirement (CONTRIBUTING.md) flagged two gaps; both closed.

**Agent-home import paths (repository layer).** Eight new tests cover the listAgentHomeSkills + importAgentHomeSkill pair end-to-end (`src/main/skills/user-skill-repository.test.ts`):

- listAgentHomeSkills with zero / one / many skills under the home dir
- non-directory entries mixed in are skipped (loose files do not become skills)
- alreadyImported flag is set when an imported record with the same slug exists
- listAgentHomeSkills returns an empty list for a missing agent-home skills dir (no error)
- importAgentHomeSkill happy path (the imported record is listed)
- source-does-not-exist failure: no half-copied state, no record created
- unsafe-slug rejection (the `SAFE_SLUG` guard)
- conflict path: re-importing the same slug gets a unique suffix (`imported-alpha` then `imported-alpha-2`) so a re-run never clobbers an existing record

**ClaudeIsolatedSignInModal UI.** Four new tests cover the renderer's contract for the new sign-in flow (`src/renderer/src/pages/settings/ClaudeIsolatedSignInModal.render.test.tsx`):

- the Sign in button is disabled until the user pastes a non-empty token
- the Sign in button shows the right label and is enabled with a token present
- Cancel calls onOpenChange(false) without invoking onSubmit
- the destructive error region is not rendered on first paint (it appears only after a failed submit)

The async click → submit → result → close/error state-transition path is covered end-to-end at the service layer (`loginIsolatedClaude` test group), where the state updates can be observed synchronously without driving Radix Portal clicks through jsdom.

Full test suite: 1024 tests pass, typecheck clean, lint 0 errors.

Refs #347

ewen-poch:
Round 6 of the AI review. Four substantive fixes plus the commit-subject validator fix.

**Token-validation hook.** Added `markClaudeIsolatedValidated` on the service. The spawn layer will call this after the first successful Claude request to flip a stored record from 'awaiting' to 'verified' — that is the 'first session success' hook the codebase was missing. As a side benefit, `validateProvider` for a stored claude-isolated provider now also calls the same hook on success, so the 'Test connection' button does what its label says: it transitions awaiting → verified instead of reporting the awaiting message back as a failure. Until the spawn layer wires it in, the Test connection path is the only way to close the loop.

**Realpath containment.** The agent-home path resolver now calls `fs.realpath` on both the home and the candidate in addition to the lexical `resolve`. A symlink inside `~/.claude/skills/` that points outside (e.g. `innocent -> /etc`) is now rejected with the same 'outside its home' message the lexical check produced. Benign in-home symlinks still pass.

**Preserve expiresAt on edit.** The claude-isolated branch of `upsertProvider` now carries `existing.expiresAt` over an edit, so a model change does not drop the credential's estimated expiry and the Settings card's 'Expires <date>' survives. A test pins this contract.

**Modal tests renamed.** The previous names overstated what the tests covered (the async click → submit → result → close/error round-trip is driven by Radix Portal under jsdom in a way that's not deterministic). New names describe the synchronous UI state the tests actually assert: disabled→enabled, label, error region absent on first paint. The full async path is covered at the service layer where state updates can be observed synchronously.

**Commit subject '+' fix.** Amend `85ac771` to drop the '+' character so `actions/github-script`'s commit-message check no longer fails.

Full test suite: 1030 tests pass, typecheck clean. Refs #347

ewen-poch:
Round 8 of the AI review. Three substantive fixes plus a documented test-architecture constraint.

**CLAUDE_CODE_OAUTH_TOKEN filter.** `buildAgentSpawnEnv` dropped only `ANTHROPIC_*` and let the setup-token flow's `CLAUDE_CODE_OAUTH_TOKEN` slip through. A custom or official provider that didn't set the var would have authenticated the spawned agent as the user's Claude subscription, against the wrong account. The filter now drops `CLAUDE_CODE_OAUTH_TOKEN` unconditionally so only the override block can carry auth forward.

**Fail closed without an app-owned config dir.** When `envOverrides` was missing `CLAUDE_CONFIG_DIR`, the env filter used to inherit the parent's value (or fall back to the default `~/.claude`), letting the agent read the user's machine-level Claude Code login. `buildAgentSpawnEnv` now throws `'Cannot spawn the agent: envOverrides must include CLAUDE_CONFIG_DIR…'` when the override omits it, so a broken provider surfaces a hard error at spawn time instead of silently inheriting the host's auth.

**Test coverage of the new behavior.** The obsolete `keeps inherited ANTHROPIC_* for a non-isolated (claude-default) provider` test (the legacy local provider is gone) is replaced with three new tests that pin the fail-closed guard, the OAuth token drop, and the CLAUDE_CONFIG_DIR override win. The 'enables SDK diagnostics' test now passes an explicit `CLAUDE_CONFIG_DIR` so the new guard doesn't break it.

**Modal click test deferred (hard-architecture constraint, not a code gap).** The SignInModal render test does not click submit. The Sign in button is rendered through Radix's Portal into `document.body`, and React 18's event delegation listens on the root container — events from Portal-rendered children do not reach the listener under jsdom. `@testing-library/react` is not a project dependency, and React 18 removed `Simulate` from `react-dom/test-utils`, so no in-tree tool can drive the React onClick through the Portal boundary. The submit pipeline — onSubmit invocation, error path, close-on-success — is covered end-to-end at the service layer (`loginIsolatedClaude` test group in `service.test.ts`), where state updates can be observed synchronously. The renderer layer is tested at the synchronous-UI-state level (button-disabled gating, label, error region absence on first paint). Adding `@testing-library/react` would unblock a true click test but is a dependency-add decision the maintainer can make separately.

Refs #347

PR #376, opened 2026-07-23
fix(acp): isolate sessions across framework switches

ewen-poch:
## Problem

A single `AcpRuntime` originally owned one mutable backend process and every active session. Deferring a framework reconnect blocked new sessions behind old work, while forcing it disposed the old session and failed its in-flight tools.

The first coordinator implementation isolated main-session turns, but several cross-generation gaps remained:

- reviewer sessions and fix-loop gaps were not part of the runtime lifetime;
- one mutable Chat Completions bridge could retarget an old Codex runtime to a new provider key/model;
- completed retired runtime objects remained in coordinator aggregation after their child process stopped;
- a single top-level connection status could misclassify failures belonging to another runtime generation;
- preflight create/connect/skill operations were not counted during retirement;
- forced-skill provisioning used one SettingsService-global mutable set across runtime generations;
- reviewer bridge isolation trusted a marker in model-visible conversation content;
- bridged providers retained listeners and decrypted credentials after their final runtime retired;
- per-generation MCP HTTP hosts had no terminal close path.

The reviewer combination let Codex advertise built-in shell/filesystem tools and could leave the review protocol incomplete without `submit_findings`.

## Proposed change

Keep one child runtime per framework generation, add workflow- and operation-scoped lifetime leases, and enforce reviewer-only tools through trusted session routing at the bridge boundary.

```mermaid
flowchart LR
  S[Framework selection changes] --> N[New active runtime generation]
  S --> R[Previous generation becomes retiring]
  R --> A{Active prompt, reviewer, workflow, or preflight operation?}
  A -->|yes| K[Finish on previous generation]
  K --> A
  A -->|no| D[Disconnect and remove previous generation]
  N --> NS[New sessions and reviewer workflows]
  N --> C[Next correction or user turn]
  C --> M[Resume or adopt lazily]
  M --> H[Replay transcript after context reset]
```

- A framework switch routes new sessions and reviewer workflows to the selected framework immediately.
- A main turn, reviewer session, reviewer/fix-loop workflow, or session preflight already in progress stays pinned to its original generation until it completes.
- An idle old main session is not proactively migrated. It moves only when a correction or later user turn actually prompts it.
- Each bridged Codex upstream target gets a SHA-256-keyed bridge instance. Provider URL, credentials, model, tools, and connector instructions remain pinned for older generations.
- Bridge instances are leased to runtime generations and close after their final owner releases them.
- Runtime-owned MCP HTTP hosts close their listener and active connections during disconnect, shutdown, retirement, or unexpected connection loss.
- Runtime retirement emits a completion callback after teardown, allowing the coordinator to remove completed generations and their event/permission aggregation.
- Forced-skill IDs belong to the runtime generation and are passed explicitly into backend provisioning.
- Provider/model and skill reconnects target only the active generation. Non-disruptive reasoning-effort changes live-apply to every tracked generation, while reconnect fallback is decided by the active generation only.
- Snapshots retain the top-level compatibility status and add optional per-session owner statuses. Prompt failure and disconnect handling prefer the owning runtime status.
- A Codex reviewer session registers its ACP session ID as a trusted bridge `prompt_cache_key`. Reviewer requests discard every Codex-provided tool declaration and explicit tool choice, then inject only the four namespaced reviewer HTTP MCP schemas. Conversation content cannot enable or disable reviewer mode.
- Uninstalling an active managed framework rotates runtime generations when settings auto-select a fallback; without a ready fallback, the stale current generation follows the existing reconnect path.
- Asynchronous session deletion and permission-profile changes hold operation leases so terminal retirement cannot close their ACP requests mid-flight.

## Scope and non-goals

- Reconnect-triggering provider/model/skill changes intentionally do not fan out to retiring generations. A retiring workflow is terminal and pinned; arming its reconnect barrier while it holds an activity lease can strand its next operation behind its own lease.
- Reasoning effort remains a global live preference because applying the ACP session option does not reconnect or replace a runtime.
- Operation leases block terminal runtime retirement but do not block deferred provider/skills reconnects, avoiding a reconnect-barrier deadlock.
- The tool-declaration enforcement applies to Chat-bridged Codex providers. Direct native Responses providers do not traverse this bridge; `codex-acp` currently exposes no ACP session-level built-in tool allowlist for that path, which retains the existing temporary-cwd and permission-gate containment.
- `getActiveRuntime()`/`lastRuntime` cleanup and hypothetical permission-map key collisions remain unchanged because there is no current incorrect behavior.

- Framework switching does not interrupt main turns waiting on tools or user permission.
- New sessions converse immediately on the newly selected framework.
- Create/connect/resume/reviewer and disabled-skill preflight cannot be orphaned by retirement.
- Session deletion and permission-profile changes cannot be interrupted by retirement.
- An active-framework uninstall that selects a fallback rotates to that framework immediately.
- Reviewer/fix-loop work survives a switch and remains on one generation; later reviewer workflows use the new generation.
- Provider/model/skill reconnects do not mutate or deadlock a retiring workflow.
- Reasoning-effort changes reach active and retiring live generations without letting an unsupported old generation force an active reconnect.
- Old and new runtime generations cannot overwrite each other's forced-skill provisioning state.
- Old and new bridged Codex generations continue using their own provider URL, key, and model.
- A bridge stays alive while any runtime generation owns it and closes after its last lease is released.
- An HTTP-only framework's per-generation MCP host closes when its runtime retires.
- A completed retired runtime disappears from coordinator aggregation.
- A dropped old runtime marks only its sessions disconnected; a dropped active runtime does not mislabel healthy old sessions.
- Normal prompt text containing the legacy reviewer marker retains normal tools.
- Bridge-backed Codex reviewer requests advertise only `read_turn`, `query_execution_log`, `read_artifact`, and `submit_findings`; built-in shell/filesystem tools and connector tools are absent.
- The live Codex ACP contract drives `read_turn -> submit_findings` through the bridge and real reviewer HTTP MCP handler.
- `npm run lint`: passed with 0 errors and 56 pre-existing warnings.
- `npm run test`: 446 files passed, 5,892 tests passed; 9 files / 83 tests skipped.
- Live Codex ACP contract: 2 files passed, 41 tests passed, including both real bridge contracts.

## Review focus

Review runtime retirement around preflight and session-mutation operations, active-framework uninstall fallback, reconnecting-setting scope, live effort fan-out, forced-skill ownership, trusted reviewer session registration, bridge leases, MCP HTTP host teardown, and provider/framework generation handoff.

### DIFF
--- a/src/main/acp/ipc.test.ts
+++ b/src/main/acp/ipc.test.ts
@@ -45,7 +45,23 @@ const { createSession, resetSessionContext, resumeSession, sendPrompt, AcpRuntim
     const resumeSession = vi.fn().mockResolvedValue({ sessionId: 's-1', cwd: '/workspace' })
     const sendPrompt = vi.fn().mockResolvedValue(undefined)
     const AcpRuntimeMock = vi.fn().mockImplementation(function () {
-      return { createSession, resetSessionContext, resumeSession, sendPrompt, getSnapshot: vi.fn() }
+      return {
+        createSession,
+        resetSessionContext,
+        resumeSession,
+        sendPrompt,
+        getSnapshot: vi.fn().mockReturnValue({
+          status: 'idle',
+          cwd: '/workspace',
+          sessionIds: [],
+          events: [],
+          pendingPermissions: [],
+          permissionProfiles: {},
+          permissionGrants: {},
+          promptInFlight: false,
+          promptInFlightSessionIds: []
+        })
+      }
     })
     return { createSession, resetSessionContext, resumeSession, sendPrompt, AcpRuntimeMock }
   })
@@ -91,7 +107,10 @@ const registerWithFakes = (overrides?: {
     runRegistry: {} as never,
     uploadRepository: {} as never,
     notebookRpcServer: {} as never,
-    settingsService: {} as never,
+    settingsService: {
+      captureActiveAgentBackendSelection: vi.fn().mockResolvedValue({}),
+      resolveAgentBackend: vi.fn().mockResolvedValue({})
+    } as never,
     taskNotifications: taskNotifications as never
   })
 }
--- a/src/main/acp/ipc.ts
+++ b/src/main/acp/ipc.ts
@@ -106,14 +106,13 @@ const createRuntime = ({
         appVersion: app.getVersion(),
         // Packaged macOS apps often start with cwd at "/" or the app bundle; use home instead.
         defaultCwd,
-        resolveBackend: async () => settingsService.resolveAgentBackend(await selection),
+        resolveBackend: async (context) =>
+          settingsService.resolveAgentBackend(await selection, context),
         // HTTP MCP registrations are runtime-owned. Sharing one host would let an old runtime teardown
         // clear routes belonging to sessions on the newly selected framework.
         mcpHttpHost: new AgentMcpHttpHost(),
         skills: {
           needForceLoad: (ids) => settingsService.skillsNeedingForceLoad(ids),
-          setTurnForced: (ids) => settingsService.setTurnForcedSkillIds(ids),
-          clearTurnForced: () => settingsService.clearTurnForcedSkillIds(),
           namesForIds: (ids) => settingsService.skillNudgeNamesForIds(ids)
         },
         artifacts: {
--- a/src/main/acp/ipc.ts
+++ b/src/main/acp/ipc.ts
@@ -20,7 +20,8 @@ import type {
   AcpStateSnapshot
 } from '../../shared/acp'
 import { DEFAULT_ARTIFACT_PROJECT_NAME } from '../../shared/artifacts'
-import { AcpRuntime } from './runtime'
+import { AcpRuntime, type AcpRuntimeCallbacks } from './runtime'
+import { AcpRuntimeCoordinator } from './runtime-coordinator'
 import { installAgentShutdownGuard } from './shutdown-guard'
 import { AgentMcpHttpHost } from './mcp-http-host'
 import { ArtifactRepository } from '../artifacts/repository'
@@ -68,7 +69,8 @@ const createManagedSessionWorkspace = async (): Promise<string> => {
   return workspace
 }
 
-// Creates the shared runtime instance used by all ACP IPC handlers and artifact claims.
+// Creates the runtime coordinator used by all ACP IPC handlers and artifact claims. Each child runtime
+// captures one framework generation so sessions on different frameworks can remain live concurrently.
 const createRuntime = ({
   mcpEntryPath,
   repository,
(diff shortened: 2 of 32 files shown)


ewen-poch:
Addressed the actionable findings from the current-head reviews in `6653085`.

- [Unexpected-close retirement](https://github.com/aipoch/open-science/pull/376#issuecomment-5060781650): `handleConnectionClosed()` now preserves terminal retirement intent and re-runs the retirement evaluator after cleanup. Outstanding operation/activity leases still drain first, and the evaluator clears the flag before teardown, so `onRetired` remains exactly-once. Added a transport-level regression test that closes ACP during a retiring prompt and verifies coordinator notification.

- [Reviewer `prompt_cache_key` contract drift](https://github.com/aipoch/open-science/pull/376#issuecomment-5060925060): the bridge now tracks whether each registered reviewer key was actually observed on a scoped request. Unregistering a reviewer session returns that observation state, and the runtime emits an explicit error when a bridged reviewer session was never scoped. This makes a future `codex-acp` contract change visible without adding an unsafe content-based fallback or pretending the app controls the upstream cache key. The real Codex ACP contract test continues to pin the expected session-id behavior.

Validation: full typecheck passed; lint passed with 0 errors; all 5,887 tests passed (83 skipped).

ewen-poch:
Addressed the two actionable lifecycle findings in `b3ae3a2`.

- [Active-framework uninstall fallback](https://github.com/aipoch/open-science/pull/376#issuecomment-5060637223): uninstall handlers now inspect the resulting framework selection. If the service auto-selected a fallback, the coordinator rotates runtime generations through `onAgentFrameworkChanged`; if no fallback is ready and the selection is unchanged, it retains the existing provider-reconnect behavior. Added coverage for Claude Code, OpenCode, and Codex fallback paths, plus the no-fallback path.

- [Asynchronous session mutations during retirement](https://github.com/aipoch/open-science/pull/376#issuecomment-5060637223): `deleteSession` and `setPermissionProfile` now acquire the runtime's synchronous operation lease before awaiting ACP `session/close` or `session/set_mode`. Terminal retirement waits for those requests to settle, while deferred provider/skills reconnect behavior remains unchanged to avoid reconnect-barrier deadlocks. Added transport-level race tests for both operations.

Validation: full typecheck passed; lint passed with 0 errors and 56 pre-existing warnings; all 5,892 tests passed (83 skipped).

ewen-poch:
Addressed the high-value teardown and bridge-boundary findings in `bf8ad89`.

- Coordinator-wide disconnect and bounded shutdown now wait for every tracked generation to settle. A rejected generation keeps its ownership so a later disconnect/shutdown can retry it; successful teardown still clears ownership as before.
- Live reasoning-effort fan-out now treats retiring generations as best-effort. A rejected old generation no longer masks the selected generation's result.
- Responses bridge startup failures now close the bridge before eviction, including partial-start ownership inside `ResponsesBridge.start()`. Runtime bridge-lease release failures are logged and no longer prevent `closed` state publication.
- The bridge fingerprint documents why live `forwardReasoningEffort` is intentionally excluded, and the retiring-session visibility contract is documented at the coordinator boundary.

Behavioral security change: a bridge-backed Codex reviewer that submits findings but never reaches the trusted reviewer session scope now persists `lifecycle=error` instead of `complete`. This is fail-closed for a contract drift that could otherwise expose non-reviewer tools. It does not affect ordinary sessions, Claude/OpenCode reviewers, or native-Responses Codex reviewers.

I did not force-remove a generation when `requestRetirement()` rejects: production retirement already catches teardown and callback errors, while deleting coordinator ownership on an unexpected rejection could lose the only shutdown handle for a still-live child process.

Validation: full typecheck passed; lint passed with 0 errors and 56 pre-existing warnings; all 5,898 tests passed (83 skipped).

PR #381, opened 2026-07-23
feat(workspace): show run code and scope picker in permission dialog

ewen-poch:
## Problem

The workspace permission approval prompt showed only a tool name (e.g. `Run notebook_execute?`), so a reviewer approving a code-execution request could not see the code that was about to run. There was also no way to choose whether an approval applied to just the current call or the whole conversation.

## Proposed change

- **Code preview** — the code about to run renders in an activity-style collapsible card matching the transcript's tool-call cards: syntax-highlighted, titled by action (`Run notebook cell` / `Run command`), expanded by default.
- **Language inference** — the code's language is resolved in priority order: explicit kernel field on the tool input → tool-name suffix (`repl_execute`→JavaScript, `bash_execute`→bash) → code heuristics (R via `<-`/`library(`, shell commands) → python. Reuses the existing `detectCellLanguage` heuristic so R cells are no longer mislabeled as Python.
- **Scope picker** — a split Allow button whose chevron opens a menu to choose *this call only* vs *this conversation*. The menu is right-anchored to the Allow button with labels left-aligned, the checkmark in a fixed right slot, and the selected row highlighted.

## Scope and non-goals

- Scoped to the renderer component `PermissionApprovalControls` and its tests only; no change to the permission broker, ACP runtime, or how the chosen option id is sent back.
- Not changing which permission options the agent offers, nor persisting a scope preference across requests (scope resets to *this conversation* per request).

- A code-execution request shows the code in a collapsible, syntax-highlighted card; language reflects the actual kernel/code, not always python.
- The scope menu switches the Allow action between call-only and conversation scope; Deny is unaffected.
- Required checks pass locally:
  - `npm run typecheck` — clean
  - `npm run lint` — 0 errors (pre-existing repo warnings only)
  - `npm run test` — 5875 passed, 0 failed
- Added static-render tests (card title, R language inference, no-code fallback) and interaction tests (scope switch, allow/deny option id, code-card collapse/expand).

## Review focus

- `extractPermissionCode` / `resolveKernelLanguage` language-resolution order in `PermissionApprovalControls.tsx`.
- Scope menu anchoring and the split-button layout.


ewen-poch:
Thanks for the review — addressed in `100af89`.

**Scope must not cross permission kinds (P1) — fixed.** `getAllowOptionId` now matches the exact ACP kind (`allow_once` for Once, `allow_always` for This conversation) with no fallback between them. The scope menu only lists scopes the request actually offers (`getAvailableScopes`), the default falls back to whatever is supported, and Allow is disabled if the chosen scope has no backing option. An allow-always-only request can no longer be granted under a "for this call only" label.

**Keep the command and file target visible (P1) — fixed.** For execute requests, `extractPermissionCode` now falls back to `request.title` when `rawInput` has no `command`, so the full command renders in the (untruncated) code card instead of only a truncated header. `toolLocations` are rendered again as a wrapping path row, so read/edit/delete prompts always show the affected path.

**Don't treat every `_execute` tool as a notebook (P2) — fixed.** Notebook handling is now gated on known identities (`notebook_execute`, `repl_execute`, `bash_execute`, incl. `provider.tool` suffixes). Unrelated tools like `database_execute` fall through to the generic JSON preview and show their args.

**Collapsing options to two scopes (P2) — intentional, keeping.** Reducing the option row to Allow (with scope) + Deny is the goal of this redesign. The concrete risk you flagged was the cross-kind conflation, which the P1 fix removes: scopes are exact-kind-backed and unsupported ones are hidden. Re-adding a per-provider button row would revert the redesign, so I'm leaving it out; happy to revisit if a provider surfaces genuinely distinct allow actions that this can't express.

Coverage: added tests for exact-kind scope selection, title-fallback command rendering, `toolLocations` visibility, and JSON args for a non-notebook `_execute` tool. `npm run typecheck`, `npm run lint`, and the full test suite pass.

ewen-poch:
Second round addressed in `35b7ab3`.

**Real MCP tool names bypassed the notebook path (P1) — fixed.** The matcher now reduces a name to its base tool (`getBaseToolName` splits on `__` then `.`), so canonical `mcp__open-science-notebook__notebook_execute` / `repl_execute` / `bash_execute` are recognized and get the code preview, title, and language. Language resolution uses the same base name. Fixtures corrected to the real `mcp__<server>__<tool>` format.

**Preview modified the code via `.trim()` (P1) — fixed.** Notebook and shell paths now return the value verbatim; emptiness is still checked with `trim()` but the displayed/authorized code keeps its exact leading indentation and trailing newlines.

**Expand + menu state leaked across requests (P2) — fixed.** On requestId change we reset scope, close the menu, and the code card is keyed by `requestId` so it re-expands per prompt. Covered by a new test that answers one request, then rerenders to the next and asserts scope/menu/expand all reset.

**Scope menu not keyboard-operable (P2) — fixed.** The chevron opens on `onClick` (Enter/Space now work). The outside-click listener moved to `click` with `stopPropagation` on the chevron, preserving the no-double-toggle behavior. New test asserts click activation opens the menu.

**Test coverage (Standards) — added.** Cross-request state reset, keyboard menu open, real `mcp__` names, explicit `kernelKind:'python'` language, and repl/bash language pinning are now asserted.

**Permission options collapsed to scope + Deny (P1) — intentional, keeping.** Reducing the option row to Allow-with-scope + Deny is the goal of this redesign, agreed with the requester. The scope options are exact-kind-backed (no cross-kind substitution) and Deny resolves to the request's `reject_*` option, so the security-relevant behavior is preserved. Surfacing multiple provider-defined allow/reject variants as separate buttons would revert the redesign; happy to revisit as a follow-up if a provider ships genuinely distinct actions this can't express.

`npm run typecheck`, `npm run lint` (0 issues in these files), and the full workspace suite (718 tests) pass.

ewen-poch:
Third round addressed in `fcd1218`.

**Notebook identity — require the server segment + de-duplicate (Codex P2 / Claude P3) — fixed.** Extracted `notebook-tool-names.ts` as the single source of truth (`NOTEBOOK_RUN_TOOL_SUFFIXES` + `isNotebookExecuteToolName`, which requires the `open-science-notebook` server segment *and* a known suffix). Both `PermissionApprovalControls` and `workspace-tool-activity-details` now import it, so the suffix list is no longer duplicated. A lookalike `mcp__acme-db__notebook_execute` now falls through to the generic JSON path with all arguments (including a `target: prod`) visible, instead of being treated as a trusted notebook.

**Risk badge contradicted the header (Claude P2) — fixed.** `getPermissionRiskLabel` routed through a bare `providerToolName === 'notebook_execute'` that never matches a real namespaced name, so the badge said "MCP tool access" while the card said "Run notebook cell". It now uses the shared `isNotebookExecuteTool`, so both agree ("Notebook execution"). Covered by a render test with a fully-namespaced name.

**Deny could pick `reject_always` (Codex P2) — fixed.** `getDenyOptionId` now prefers `reject_once` and only falls back to another reject kind when it's absent, so a single Deny never silently grants a permanent rejection based on provider ordering. Test asserts precedence with `reject_always` listed first.

**Title hidden behind provider name (Codex P2) — fixed.** When `request.title` carries information the header (provider name) doesn't and isn't already shown by the code card, it renders as a detail line — e.g. provider `Write` with title `Write report.md` now shows the filename.

**Scope dropdown accessibility (Claude P2) — Escape added; keeping the custom dropdown.** Added Escape-to-close (keyboard-open via `onClick` landed last round). I opted not to swap in the Radix `DropdownMenu` primitive: the custom dropdown's anchoring/spacing was tuned across earlier rounds and a full rewrite risks regressing that. The concrete a11y gaps flagged (keyboard open, Escape dismiss) are now closed. Happy to migrate to the shared primitive as a dedicated follow-up if you'd prefer the consistency.

**Collapsing provider options (Codex P2) — still intentional.** This is the agreed redesign; scope options are exact-kind-backed and Deny now prefers `reject_once`, so no outcome is mislabeled.

`npm run typecheck`, `npm run lint` (0 issues in changed files — 3 remaining warnings are pre-existing in untouched `SessionNotebookDialog` tests), and the full workspace suite (723 tests) pass.

ewen-poch:
Addressed in `08933aa` (Codex round).

**Inspect both tool identity fields for notebook approvals (P1) — fixed.** You're right, and this was a regression I introduced last round: requiring the server segment broke the broker shape where `title = mcp.open-science-notebook.notebook_execute` but `providerToolName = notebook_execute` (bare leaf). I was checking `providerToolName ?? title`, so the leaf won and failed the segment check, and — being `kind: execute` — the request fell through to the Bash branch and displayed the namespaced title instead of the code. New `resolveNotebookToolName(request)` now checks **both** fields (the namespaced one supplies the server segment, either supplies the suffix) and returns the matching name for kernel/language detection. The code card, action title ("Run notebook cell"), and risk badge ("Notebook execution") all route through it. Added a render test using the exact broker shape from `runtime.test.ts` (`mcp.open-science-notebook.notebook_execute` + leaf `notebook_execute`, `kind: execute`) asserting the code — not the title-as-Bash — is shown, plus util cases for the namespaced title and the bare-leaf negative.

**Do not collapse multiple allow_always options (P1) — normalized at the broker, keeping the collapsed UI.** This is the same redesign decision as before, and the normalization you recommend already exists at the broker boundary: `projectPermissionOptions` (permission-broker.ts:79-112) strips Codex's persistent cross-session `allow_always` by its known option ID, so exactly one session-scoped conversation option reaches this component — the app's grant model is session-only and revocable by design, and persistent cross-session grants are intentionally never offered here. Non-Codex frameworks in this repo don't emit multiple `allow_always` variants. So in practice the "first match" in `getAllowOptionId` is selecting from a single conversation-scoped option, not choosing between broader/narrower persistent ones. If a future provider surfaces genuinely distinct same-kind scopes, the right fix is to normalize them at that broker boundary (as Codex already is) rather than re-expanding the option row — happy to add that projection when such a provider exists.

`npm run typecheck`, `npm run lint` (0 issues in changed files), full workspace suite (730 tests) pass.

ewen-poch:
Codex round addressed in `8054e0c` (includes a merge of `origin/main`).

**Resolve the conflict with the current base branch (P1) — fixed.** Merged `origin/main` (e0e436d) into the branch. The conflict was in `isNotebookKernelRunActivity`: base had rewritten it to a regex supporting the underscore-sanitized form, while this PR had extracted it to the shared util. Reconciled onto the shared `isNotebookExecuteToolName`, which now covers both forms (see below), so base's underscore support is preserved and improved. `isNotebookExecuteActivity` that you referenced was deleted on base, so there was no second call site to reconcile.

**Match the notebook namespace exactly after normalization (P1) — fixed.** Rewrote the matcher in `notebook-tool-names.ts`: it splits on `__`/`.` delimiters only (single `_` occurs inside both the suffix `notebook_execute` and the sanitized server `open_science_notebook`, so it must not split), requires the last segment to be a known kernel suffix, and requires the segment immediately before it to equal `open-science-notebook` **exactly** after normalizing `_`→`-`. Result:
- Accepts the responses-bridge form `mcp__open_science_notebook__notebook_execute` (previously rejected).
- Rejects `mcp__open-science-notebook-staging__notebook_execute` and `mcp__my-open-science-notebook__notebook_execute` (previously accepted by `includes()`), so an unrelated server can't route through the notebook code preview and hide non-code arguments.
- Still rejects the bare leaf `notebook_execute` (no server segment), which is why the dialog consults both `providerToolName` and `title`.

Tests added for the underscore form, both containing-name lookalikes, and `matchNotebookRunTool` returning the matched suffix.

Verification below (shared with the Claude reply).

ewen-poch:
Claude architecture round addressed in `8054e0c`.

**Permission options outside the four canonical kinds become unreachable (P2) — fixed.** Restored a graceful fallback, but keyed on reachability rather than re-introducing the old flat list. New `getExtraOptions` computes the optionIds the Allow (both scopes) / Deny controls can actually invoke, and renders **any option not among them** as its own labeled button (`data-testid="extra-option"`) that sends its exact optionId. So a non-canonical kind (e.g. `allow_sandbox`, or a future ACP kind) is selectable again instead of leaving Allow disabled and Deny sending cancel. The canonical four-kind request still renders only Allow + Deny (test asserts no extra buttons).

**Multiple allow_always collapse to 'first wins' outside the Codex path (P2) — fixed at the renderer, framework-independent.** The same fallback covers this: `allowOptionIds` collects the first option of each allow scope; a **second** `allow_always` (distinct provider scope/label) is not reachable via the single Allow control, so it now surfaces as its own labeled button rather than being silently dropped. This is renderer-side and independent of `frameworkId`, so Claude Code / opencode / a future framework sending two `allow_always` no longer loses one. Regression tests added: a second `allow_always` surfaces with its provider label, and clicking an extra option sends its exact optionId. Deny still intentionally folds all `reject_*` into one control (the approved redesign), so `reject_always` does not reappear as an extra.

**Scope dropdown reimplements dropdown-menu logic (P3) — deferred, gaps closed.** I added Escape-to-close and keyboard-open in earlier rounds, so the concrete a11y gaps you flagged are closed. I'm deferring the full swap to the shared Radix `DropdownMenu` primitive to avoid regressing the anchoring/spacing tuned across prior rounds; happy to do it as a dedicated follow-up PR for consistency. Flagging explicitly so it isn't lost.

**Verification (both rounds):** `npm run typecheck` clean, `npm run lint` 0 issues in changed files, full workspace suite **738 tests** pass, plus `responses-bridge` + `runtime` suites (notebook naming) green. Branch merged up to `origin/main` e0e436d.

ewen-poch:
Addressed both Codex findings:

- **Preserve saved Windows close behavior:** restored the complete close-preference compatibility path removed by `8704bbf`: settings schema/repository/service persistence, lifecycle wiring, IPC/preload/store integration, the General setting, and the modal/native-fallback "Don't ask again" response. Existing saved `minimize` / `quit` values are sanitized and honored before showing the close-to-tray confirmation, while explicit quit with running work still confirms. The restored tests cover reading, persisting, failed settings access, renderer fallback, and native fallback.
- **Retain kernel kind from MCP run summaries:** the transcript kernel resolver now receives the parsed run summary. A valid input kernel remains highest priority; when input has no valid kernel, a validated `summary.kernelKind` is used before tool-name and script heuristics. Added a regression test for an R `notebook_execute` summary with no `rawInput` and an intentionally ambiguous script, asserting R highlighting.

Validation: `npm run typecheck`, `npm run lint` (0 errors; pre-existing repository warnings only), and `npm run test` (5,958 passed, 83 skipped).

PR #387, opened 2026-07-24
feat(activity-groups): declare titles before tool groups

### LINKED ISSUE #384: LLM-generated purpose title for tool-activity cards
## Summary

Each conversation tool-activity card should be titled with the **purpose of that dispatch** — a short, independent label describing *what this batch of tool calls is trying to accomplish* (e.g. "Search PubMed for recent literature on X", "Plot and save a cosine curve").

Today the card header shows a mechanical category summary ("Ran a notebook cell, saved a file"). That's accurate but not meaningful.

## Why this needs generation (not a data field)

Investigated every existing source; none carries a usable purpose label:

- **Tool-call `title`** — for the notebook/artifact MCP tools it just echoes the raw tool id (`mcp__…__notebook_execute`), no intent.
- **ACP `plan` event** — the stream *does* emit `plan` / `plan_update` (`src/main/acp/runtime-events.ts`), but the handler currently discards the entries and stores a static `"Plan update"` string. Only plan-capable agents (Claude Code) would emit it; the Codex/gpt bridge may not.
- **Agent narration message** — the assistant's sentence before the run ("好的，我来用 PubMed 搜索…") is already rendered as a chat bubble; reusing it as the title duplicates the bubble and reads as conversational filler.
- **Raw user prompt** — rejected: echoing the user's verbatim input (typos, repetition) as a title is not a purpose statement.

So a genuine purpose title must be **generated**.

## Constraint discovered

The app currently has **no lightweight one-shot LLM client**. The only path to a model is spawning a full ACP agent session (see `src/main/reviewer/orchestrator.ts`). Provider keys are stored via Electron `safeStorage` (`src/main/settings/crypto.ts`) but nothing makes a direct `/v1/messages` or `/v1/chat/completions` call.

## Options

**A. Reuse ACP** — spawn a sub-session per tool group to summarize. Correct-ish, but a process spawn per group is expensive/slow (seconds) for a lightweight label.

**B. New direct HTTP completion client (preferred)** — decrypt the active provider key, send one small request branched by vendor (anthropic `/v1/messages` vs openai `/v1/chat/completions`). Requires: key-decrypt reuse, per-vendor request construction, timeout/error handling, a renderer↔main IPC, persistence of the generated title on the session, and debounced per-group triggering.

## Scope / acceptance

- Generate a concise purpose title (~40–80 chars) per activity group from its tool calls + surrounding context.
- Generate asynchronously; never block rendering. Fall back to the current mechanical summary on any failure/timeout/missing key.
- Persist the title on the session so it survives reload and isn't regenerated.
- Debounce so a title is generated once per group, not per streamed tool call.
- Respect the user's configured provider/model; make the feature degrade gracefully when no key is configured.

## Related

Follow-up to the notebook-detection + row-collapse work in branch `fix/notebook-cell-collapse-until-run`. The card title was intentionally left as the mechanical summary pending this issue.

ewen-poch:
## Problem

Tool activity groups were titled from the mechanics of their individual calls, such as "Ran a notebook cell" or "Saved a file". This made a coherent group of steps hard to scan because the UI could not show why the agent was doing the work.

Generating a title after the tools run would require a separate model request, add latency and provider-specific request handling, and still infer intent after the fact. The agent already knows the purpose before it starts the group.

## Proposed change

- Add a declaration-only `begin_activity_group({ title })` MCP tool. The main agent calls it once before the first tool in each coherent group, not once per tool or step.
- Treat the declaration as control metadata: hide its start and terminal events, exclude it from the visible step count, and attach subsequent real tool activities to the declared group.
- Persist groups independently as `PersistedActivityGroup` records and render their declared titles after reload.
- Close the current group when started work is followed by assistant output, a new declaration, run completion, or failure. Prune declarations that never receive a real tool activity.
- Preserve legacy adjacent-tool grouping for sessions that have no declared group metadata.
- Supply the activity-group tool and instruction only to mainline ACP sessions. Reviewer sessions retain only their reviewer MCP scope and prompt.
- Support Claude Code, OpenCode, Codex native ACP, and the Codex Chat Completions bridge without making a separate external model request.

### Architecture

```mermaid
flowchart LR
  subgraph Mainline["Mainline ACP session"]
    Agent["Claude Code / OpenCode / Codex"]
    Prompt["Activity-group instruction"]
    MCP["open-science-activity MCP"]
    Bridge["Codex Chat Completions bridge schema"]
  end

  subgraph Runtime["Open Science runtime"]
    Events["ACP runtime events"]
    Filter["Control-event detection"]
  end

  subgraph Renderer["Renderer"]
    Store["Session store"]
    Persist["Session persistence"]
    UI["Activity-group UI"]
  end

  Prompt --> Agent
  MCP --> Agent
  Bridge --> Agent
  Agent --> Events --> Filter --> Store
  Store <--> Persist
  Store --> UI

  Reviewer["Reviewer ACP session"] -. "reviewer MCP only; no injection" .-> Agent
```

### Data flow

```mermaid
flowchart TD
  Declare["begin_activity_group with title"] --> Detect["Recognize server-qualified control event"]
  Detect --> Sanitize["Normalize and cap title at 80 Unicode code points"]
  Sanitize --> Group["Create open PersistedActivityGroup"]
  Detect --> Hide["Cache toolCallId and hide declaration updates"]
  Tool["Subsequent real tool event"] --> Attach["Set activityGroupId"]
  Group --> Attach
  Attach --> Persist["Persist group and activity references"]
  Persist --> Render["Render one title with N real steps"]
  Boundary["Assistant output / new declaration / stop / failure"] --> Close["Complete started group and prune empty declarations"]
  Close --> Persist
```

### Interaction

```mermaid
sequenceDiagram
  participant U as User
  participant A as Main agent
  participant G as Activity-group MCP
  participant R as ACP runtime
  participant S as Session store
  participant V as Workspace UI

  U->>A: Send turn
  A->>G: begin_activity_group(title)
  G-->>R: Declaration tool events
  R->>S: Begin group; suppress control row
  A->>R: Real tool call 1
  R->>S: Add activity with activityGroupId
  A->>R: Real tool call 2
  R->>S: Add activity with same activityGroupId
  S-->>V: Show declared title and 2 steps
  A->>R: Assistant output or run stop
  R->>S: Complete group
```

## Scope and non-goals

- This names a coherent tool group, not every tool call or individual step.
- This does not generate titles after execution and does not add an external LLM request, provider setting, or background ACP session.
- This does not inject activity-group tooling or instructions into reviewer sessions.
- This does not change individual tool-row labels.
- This does not rewrite legacy sessions; they continue to use mechanical adjacent-tool grouping when group metadata is absent.

- A valid declaration precedes each coherent tool group and can split multiple groups within one turn.
- Narration before the first real tool does not discard the declaration.
- Declaration events are invisible and do not count toward `N steps`, including sparse terminal updates that omit tool identity.
- Titles are normalized and capped at 80 Unicode code points.
- Groups and activity references survive save, hydration, edited-resend truncation, stop, and failure paths.
- Automatic permission approval requires server-qualified activity-tool identity; a bare `begin_activity_group` name is not trusted.
- Claude Code, OpenCode, Codex native ACP, and the Codex Chat Completions bridge expose the declaration path to mainline sessions.
- Reviewer sessions remain isolated from the activity tool and instruction.
- `npm run test` (5,933 passed, 83 skipped; 448 files passed, 9 skipped)
- `npm run typecheck`
- `npm run check:web-api-map`
- `npm run lint` (0 errors; 56 pre-existing warnings)
- `git diff --check`

## Review focus

- Mainline-versus-reviewer session scoping for MCP tools, prompt injection, and bridge schemas.
- Control-event identification, permission handling, and suppression of all declaration updates.
- Group lifecycle boundaries, empty-group pruning, persistence compatibility, and legacy fallback rendering.

Closes #384

### DIFF
--- a/src/main/acp/ipc.ts
+++ b/src/main/acp/ipc.ts
@@ -131,6 +131,7 @@ const createRuntime = ({
           registerSessionAlias: (aliasSessionId, sessionId) =>
             notebookRpcServer.registerSessionAlias(aliasSessionId, sessionId)
         },
+        activityGroups: { mcpEntryPath },
         callbacks: runtimeCallbacks
       })
     },
--- a/src/main/acp/permission-policy.test.ts
+++ b/src/main/acp/permission-policy.test.ts
@@ -151,6 +151,27 @@ describe('permission policy', () => {
     ).toBeUndefined()
   })
 
+  it('auto-approves the declaration-only activity group tool under Ask', () => {
+    const request = createPermissionRequest('other', undefined, {
+      title: 'mcp__open-science-activity__begin_activity_group'
+    })
+
+    expect(
+      resolveAutomaticPermission(request, {
+        profile: 'ask',
+        mcpServerNames: ['open-science-activity']
+      })
+    ).toBe('allow')
+    expect(isMcpToolName('begin_activity_group', ['open-science-activity'])).toBe(true)
+
+    expect(
+      resolveAutomaticPermission(
+        createPermissionRequest('other', undefined, { title: 'begin_activity_group' }),
+        { profile: 'ask', mcpServerNames: ['open-science-activity'] }
+      )
+    ).toBeUndefined()
+  })
+
   it('grants a single-use approval only, never escalating to allow_always', () => {
     const request = createPermissionRequest('read', [{ path: 'data/input.csv' }])
 
--- a/src/main/acp/permission-policy.test.ts
+++ b/src/main/acp/permission-policy.test.ts
@@ -15,6 +15,7 @@ const createPermissionRequest = (
   overrides?: {
     title?: string
     options?: RequestPermissionRequest['options']
+    rawInput?: unknown
   }
 ): RequestPermissionRequest => ({
   sessionId: 'session-1',
@@ -22,7 +23,8 @@ const createPermissionRequest = (
     toolCallId: 'tool-1',
     title: overrides?.title ?? 'Tool call',
     kind,
-    locations
+    locations,
+    rawInput: overrides?.rawInput
   },
   options: overrides?.options ?? [
     { optionId: 'allow', name: 'Allow once', kind: 'allow_once' },
@@ -172,6 +174,24 @@ describe('permission policy', () => {
     ).toBeUndefined()
   })
 
+  it('does not trust raw input to identify an activity group declaration', () => {
+    const spoofedBuiltIn = createPermissionRequest('execute', undefined, {
+      title: 'Bash',
+      rawInput: {
+        server: 'open-science-activity',
+        tool: 'begin_activity_group',
+        arguments: { title: 'Spoofed declaration' }
+      }
+    })
+
+    expect(
+      resolveAutomaticPermission(spoofedBuiltIn, {
+        profile: 'ask',
+        mcpServerNames: ['open-science-activity']
+      })
+    ).toBeUndefined()
+  })
+
   it('grants a single-use approval only, never escalating to allow_always', () => {
     const request = createPermissionRequest('read', [{ path: 'data/input.csv' }])
 
@@ -251,5 +271,22 @@ describe('permission policy', () => {
     })
 
     expect(resolveAutomaticPermission(onlyAlways, { profile: 'full' })).toBe('always')
+  })
+
+  it('preserves the Full access allow_always fallback for activity declarations', () => {
+    const onlyAlways = createPermissionRequest('other', undefined, {
+      title: 'mcp__open-science-activity__begin_activity_group',
+      options: [
+        { optionId: 'always', name: 'Allow always', kind: 'allow_always' },
+        { optionId: 'reject', name: 'Reject', kind: 'reject_once' }
+      ]
+    })
+
+    expect(
+      resolveAutomaticPermission(onlyAlways, {
+        profile: 'full',
+        mcpServerNames: ['open-science-activity']
+      })
+    ).toBe('always')
   })
 })
--- a/src/main/acp/permission-policy.ts
+++ b/src/main/acp/permission-policy.ts
@@ -108,19 +108,20 @@ const resolveAutomaticPermission = (
   params: RequestPermissionRequest,
   context: PermissionPolicyContext | undefined
 ): string | undefined => {
+  if (context?.profile === 'full') {
+    return resolveFullAccessAllowOptionId(params)
+  }
+
+  // The declaration exception must be bound to a server-qualified tool identity. rawInput is
+  // agent-controlled arguments and cannot prove which tool the permission request will execute.
   if (
     context?.mcpServerNames?.includes(ACTIVITY_GROUP_MCP_SERVER_NAME) &&
     isActivityGroupToolEvent({
       title: params.toolCall.title ?? undefined,
-      providerToolName: extractProviderToolName(params.toolCall),
-      rawInput: params.toolCall.rawInput
+      providerToolName: extractProviderToolName(params.toolCall)
     })
   ) {
     return resolveAllowOptionId(params)
-  }
-
-  if (context?.profile === 'full') {
-    return resolveFullAccessAllowOptionId(params)
   }
 
   if (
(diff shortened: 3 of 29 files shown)


ewen-poch:
Review follow-up:

- Added explicit coverage for all five supported activity-tool identity formats, including Codex's dot-style name and the all-underscore bridge form.
- Added a surrogate-pair regression assertion that verifies the title limit as 80 Unicode code points rather than 80 UTF-16 code units.
- Kept `activityGroupToolCallIdsBySession` cleanup scoped to terminal runtime events. The entries are short-lived tool-call UUIDs, are cleared on `stop`/`error`, and stale values cannot affect later runs; coupling session-store deletion paths back into workspace event internals would add more lifecycle surface than the low practical leak risk justifies.
- Kept `completeActivityGroup` on assistant chunks. Its no-open-group path returns the original Zustand state and does not notify subscribers; adding a parallel transient flag would duplicate group lifecycle state and create a higher consistency risk than the small scan cost.
- Kept the MCP argv re-export. It aliases the canonical constant rather than redefining it and lets the runtime import the activity-group MCP surface from one module; changing it has no behavioral or maintenance payoff here.

Validation: `npm run test -- src/shared/activity-groups.test.ts`, `npx eslint src/shared/activity-groups.test.ts`, and `git diff --check`.

ewen-poch:
Real-session follow-up:

- Reproduced the missing purpose title with Claude Code backed by a DeepSeek Anthropic-compatible provider. The activity MCP server was registered correctly, but the model called the PubMed/notebook tools directly and never emitted `begin_activity_group`; consequently, the renderer had no declaration event to apply.
- The root cause was relying only on the session-level `_meta.systemPrompt.append`, which this provider did not follow reliably for the turn. Mainline turns now also receive a short, high-salience declaration reminder immediately before the user prompt.
- The Claude adapter renders the reminder with the actual callable name, `mcp__open-science-activity__begin_activity_group`. Codex and OpenCode keep their existing per-turn delivery, so they do not receive a duplicate reminder.
- Reviewer sessions remain isolated: they receive neither the activity MCP server nor the new turn reminder. Runtime coverage asserts both the mainline prompt content and the reviewer exclusion.

Validation: `npm run typecheck`, `npm run lint` (0 errors; pre-existing repository warnings only), and `npm run test` (5,936 passed, 83 skipped).

ewen-poch:
Addressed the Codex permission-policy findings:

- Activity declaration auto-approval under Ask now requires a server-qualified identity from the tool title or provider metadata. Agent-controlled `rawInput.server/tool` is no longer used as proof of tool identity. Codex sparse requests remain supported through the existing toolCallId-bound identity restoration path. Added a regression test showing that a Bash request with spoofed activity raw input is not auto-approved.
- Full access is now resolved before the declaration-only exception, preserving the existing `allow_once` preference and `allow_always` fallback. Added coverage for an activity declaration where `allow_always` is the only offered allow option.

Claude Architecture Review reported `mergeable` with no findings. Validation: `npm run typecheck`, `npm run lint` (0 errors; pre-existing repository warnings only), and `npm run test` (5,938 passed, 83 skipped).

ewen-poch:
Fixed the missing titles for later tool groups in the same turn.

The captured Claude transcript showed that the renderer did not drop a second declaration; the agent emitted only one `begin_activity_group` call for the entire turn. The turn reminder was phrased like a one-time first-tool instruction, and the MCP acknowledgement did not reinforce the rule after the first group.

This update:
- requires a declaration before each coherent tool group and explicitly repeats it when the purpose changes;
- reminds the agent after every successful declaration to call `begin_activity_group` again before a later group;
- keeps the guidance limited to mainline ACP sessions, with the existing reviewer-session exclusion still covered.

Validation: `npm run typecheck`, targeted ESLint, 120 related tests, the mainline/reviewer isolation test, and the full suite (`5,938 passed`, `83 skipped`).

ewen-poch:
Addressed the unbounded runtime payload finding in `af51f13`.

Tool event projection now applies an 8,000-character JSON budget to `rawInput` and `rawOutput`, preserving small serializable values for activity details while dropping oversized or non-serializable payloads before they enter the runtime snapshot. Tool events also omit the original raw ACP notification, which otherwise retained a duplicate unbounded copy of the same arguments/results. Message/image fallback metadata is unchanged.

Added a regression test covering oversized input and output. Validation: typecheck, targeted ESLint, focused runtime/persistence/renderer tests, and the full suite (`5,939 passed`, `83 skipped`).

PR #389, opened 2026-07-24
feat(settings): add Claude subscription with shared and isolated sign-in

ewen-poch:
## Problem

The Claude subscription had no first-class sign-in in Settings. Users couldn't authenticate a Claude subscription the way they can a Codex one (shared profile vs. isolated, app-managed credentials), and there was no browser-based login.

## Proposed change

Give the Claude subscription two authentication modes behind one provider card, mirroring the Codex shared/isolated split:

- **claude-shared** — browser OAuth via `claude auth login --claudeai`; credentials live in the default `~/.claude` profile. Its sign-in button swaps to a **Cancel** action while login is pending.
- **claude-isolated** — `claude setup-token` under the app-owned `CLAUDE_CONFIG_DIR` (never touches `~/.claude`). **Sign in with browser** runs `setup-token` (the CLI opens the browser and runs its own `localhost` callback), captures the printed token from stdout, and opens the paste modal alongside as a fallback:
  - Happy path: the user finishes in the browser and the modal **auto-closes**.
  - Fallback: a browser that never opened (or a user who prefers a token) can paste one, which **cancels the background login**.
  - During the browser flow the modal **hides the manual "run this command" step** (the app runs it); a pure manual-paste entry keeps the full Step 1 / Step 2 instructions.

Each auth mode owns a distinct provider record (`builtin-claude-shared` / `builtin-claude-isolated`); `claudeIsolatedProviderIdentity` resolves to the isolated id every isolated lookup keys on, and `claude-shared` is added to the repository provider-type allowlist so stored records survive a settings reload.

## Scope and non-goals

- In scope: Claude subscription authentication (both modes), the settings/onboarding UI, IPC/preload wiring, and the main-process auth controllers.
- Not in scope: skill import or any non-auth Claude feature.

- `npm run typecheck`, `npm run lint`, and the settings/onboarding suites pass.
- Covered by tests across the auth controllers, repository, service, provider list/form, sign-in modal, and onboarding wizard.

## Review focus

- The two-record identity model (`builtin-claude-shared` vs `builtin-claude-isolated`) and the isolated-id lookup contract.
- The browser/paste race handling in `ProvidersPanel` (manual paste cancels the background browser login; the cancelled login must not surface a spurious error).


ewen-poch:
## Review response

**Codex rounds 1–3 / Claude round 1**

### Fixed across rounds

| Finding | Resolution |
|---|---|
| P1 claude-shared validation always bad-url | Added `claude-shared` branch in `validateProvider` → `getClaudeSharedStatus()` |
| P1 isProviderKeyUsable never passes for shared | Added `claude-shared` to the always-usable branch (CLI-managed credentials) |
| P1 OpenCode compat — shared passes framework check | `isProviderUsableByFramework` now uses `isClaudeSubscriptionProvider` to restrict both modes to `claude-code` |
| P1 Mode switch leaves stale provider + loses active selection | `upsertProvider` atomically deletes outgoing claude mode, then re-activates incoming record if outgoing was active |
| P1 Cancel/paste race overwrites new token | Failure path now uses `updateClaudeIsolatedValidationIfKeyMatches` (key-matched write) instead of snapshot upsert |
| P1 renderer resolves wrong id on mode switch | `resolveUpsertedProviderId` handles claude subscription types by fixed id, same as Codex |
| P1 Windows .cmd spawn fails | Both controllers use `resolveClaudeExecutableForSpawn` + `spawnClaude` helper that routes `.js` entries through `process.execPath` with `ELECTRON_RUN_AS_NODE=1` |
| P1 Browser login persists isolated cred after mode switch | `saveClaudeIsolatedToken` checks for an active `claude-shared` record before writing; skips on mode-switch race |
| P1 Cancel writes failure marker | `loginIsolatedBrowser` returns `cancelled: true` on user-cancel; service skips `finalizeClaudeIsolatedLogin` entirely |
| P2 Shared card Test always fails (bad URL) | Covered by the validateProvider routing fix above |
| P2 Shared image support stripped | `providerSupportsImageInput` uses `isClaudeSubscriptionProvider` for both modes |
| P2 Panel unmount does not cancel Claude sign-ins | Unmount cleanup now cancels both shared and isolated pending logins |
| P2 CLAUDE_CONFIG_DIR not pinned in shared auth | All three shared auth spawns set `CLAUDE_CONFIG_DIR: ~/.claude` explicitly |
| P2 NO_BROWSER/CI forwarded to shared login | `runBrowserLogin` env explicitly deletes both flags |
| P2 resolveClaude() rejection unhandled | `getStatus`/`loginShared` wrap resolution in try/catch and return structured errors |
| P3 Dead exports (CLAUDE_SUBSCRIPTION_PROVIDER_ID etc.) | Removed; `isClaudeSubscriptionProviderId` simplified to two-id check |

### Accepted trade-offs

- **preflight always-ready for shared**: `isProviderKeyUsable` returns `true` unconditionally for `claude-shared` because credentials are CLI-managed and cannot be decrypted by the app. The actual auth failure surfaces at agent spawn time.
- **race test token submission**: `ProvidersPanel.race.test.tsx` covers cancel suppression and browser-success auto-close; full token submission round-trip is exercised by `service.test.ts`.
- **IPC/provider-env contract tests**: Existing service-level tests cover the key paths; isolated IPC-layer tests are deferred.

ewen-poch:
## Round 5 update

All Codex rounds 1–4 + Claude rounds 1–4 findings addressed.

### This round adds

**Claude P1: model override for claude-shared**  
`upsertProvider` now has a `claude-shared` branch that persists `model` and sets `apiEndpoints=['anthropic']`.

**Spec P1: Onboarding paste fallback**  
`OnboardingWizard` now mounts `ClaudeIsolatedSignInModal` during isolated browser login, so users can paste a token if the browser doesn't open.

**Spec P2: Cancel handling**  
- `ValidateProviderResult` gains `cancelled?: boolean`
- Isolated and shared cancel flows return `cancelled: true`
- `loginClaudeShared` skips writing `lastValidationFailure` when cancelled
- Onboarding checks `cancelled` flag to distinguish user-cancel from provider-changed

### Trade-offs accepted

- **Mode switch deletes other record**: intentional — switching modes requires re-auth; keeping both records would confuse the UI
- **shared preflight always ready**: CLI manages credentials (can't decrypt in-app); spawn-time auth is the real check
- **Codex P2 cancel during probe**: probe is <1s; cancel UI is already hidden from user by that phase
- **Test coverage gaps** (Standards P2/P3): deferred beyond round limit

Total rounds: 5 (2 Codex needs-changes, 2 Claude needs-changes, 1 combined fix). Pausing for review.

ewen-poch:
Addressed the actionable Claude Architecture Review points in commit 905a886:

- Centralized the duplicated Claude CLI launch and abort-race behavior in src/main/settings/claude-cli-process.ts. Both auth controllers now share the same Windows JS-entry launcher, Electron Node-mode environment, abort semantics, and listener cleanup.
- Removed the unconditional claude-shared success path from isProviderKeyUsable(). Active shared providers continue to use the live claude auth status result in getPreflight, while the static fallback now correctly treats a profile-backed provider with no keyRef as unusable.

I am not moving reconnect decisions from IPC into SettingsService in this PR. That would change several service return contracts and broaden a currently working, tested auth change into a cross-layer refactor without a functional fix.

I am also not extracting the onboarding/settings browser-vs-paste state into a shared hook here. Both flows have distinct ownership and navigation behavior and are covered by race tests; combining them now would add UI state-coupling risk without changing user-visible behavior.

Validation: node and web typechecks pass; full lint has 0 errors; both auth controller suites pass. The remaining full-suite failures are the existing sandbox listener restrictions and missing electron-builder NSIS templates in this worktree.

ewen-poch:
Addressed the actionable review findings in commit 406eb5d:

- Corrected the isolated setup-token copy: the token is encrypted in Open Science app storage, not stored in the app-owned Claude config directory, and the dialog still states that ~/.claude is untouched.
- Moved claude-shared credential usability behind the async provider-usability method instead of relying on a caller-only override.
- Added a 5-second auth-status TTL for shared Claude preflight checks and reuse of the same in-flight Promise for concurrent preflights.
- Added generation-based invalidation so a status request started before login/logout cannot repopulate the cache after that auth operation completes.
- Invalidated the cache around shared login/logout and populated it from explicit status checks.
- Centralized the user Claude profile path in getUserClaudeConfigDir() and reused it from provider env, shared auth, and SettingsService.

The cache behavior is covered for TTL expiry, concurrent deduplication, stale in-flight invalidation, browser-login invalidation, logout invalidation, and explicit-status reuse.

I am still not extracting the browser-vs-paste renderer flow into a shared hook in this PR. The settings panel and onboarding wizard have different navigation and ownership semantics, and both race paths are already covered; that refactor adds state-coupling risk without a user-visible fix.

Using the real ~/.claude profile is an explicit product contract of claude-shared: it intentionally uses the users existing Claude CLI credentials, settings, and skills as-is. Users who need app-owned isolation and curated app configuration should use claude-isolated.

Validation: node and web typechecks pass; modified-file lint has 0 errors; relevant suites have 187 passing tests with only 3 sandbox listener failures. Full suite: 5875 passed; remaining failures are the existing listener restrictions and missing electron-builder NSIS templates in this worktree.

ewen-poch:
The collapsed Claude-card deletion finding is valid. The current delete path removes only the displayed fixed-id record, so the retained sibling can become visible again after refresh.

No additional code has been pushed for this item. I am pausing further changes at the requested review limit. The appropriate follow-up is to enforce collapsed-card deletion in SettingsService/repository by removing both CLAUDE_SHARED_PROVIDER_ID and CLAUDE_ISOLATED_PROVIDER_ID together, clearing claudeSubscriptionProviderId, and covering the dual-record case with a service/repository test.

ewen-poch:
Addressed in commit a8b6b7c.

Deleting either Claude subscription fixed ID now expands to the complete collapsed Claude subscription group inside one SettingsRepository mutation. The mutation removes both shared and isolated records, clears claudeSubscriptionProviderId, and clears activeProviderId/activeModel when either removed sibling was active. Unrelated provider deletion behavior is unchanged.

The service regression is parameterized over both deletion entry points: deleting shared while isolated is active, and deleting isolated while shared is active. Both cases verify the renderer snapshot and persisted settings contain no Claude sibling or stale selection state.

Validation: both regression cases pass; repository suite 57/57; node and web typechecks pass; modified-file lint has 0 errors. Full suite: 5877 passed, with the existing sandbox listener and missing NSIS-template failures unchanged.

ewen-poch:
Addressed in 35aae60.

- Shared Claude disconnect now only clears Open Science's local validation state and invalidates the shared-auth preflight cache.
- `ClaudeSharedAuthController` no longer exposes or runs `claude auth logout`, so the user's global `~/.claude` profile remains untouched.
- The UI action and fallback error now explicitly say “Disconnect from Open Science.”
- Added regression coverage that verifies a global logout capability is never invoked, plus UI coverage for the app-local disconnect wording.

Focused regression/auth/UI tests and both TypeScript project checks pass.

ewen-poch:
Addressed both actionable follow-ups in 9c0fc86, then synced the latest `main` in 0127346.

- Active Claude shared↔isolated edits now capture the pre-upsert active provider and reconnect when the edited record was active before or remains active after the save. Parameterized IPC coverage verifies both switch directions.
- Shared Claude browser-login failures now surface `result.message` (or an inline fallback) in Settings, while explicit user cancellation remains silent.
- Resolved the `main` conflicts by preserving the PR's browser-first onboarding and corrected encrypted-storage copy while adopting the new shared dialog chrome.

Validation: conflict-related and regression suites pass 135/135; node and web typechecks pass; changed-file lint has 0 errors.

ewen-poch:
Addressed the late isolated Claude browser sign-in race in 6e9518e.

- Deleting either Claude subscription provider ID now cancels both Claude auth controllers before the collapsed provider group is removed.
- Isolated credential writes now use an atomic update-if-present repository mutation, so a token that arrives after deletion cannot recreate builtin-claude-isolated. Credential cleanup uses the same non-creating path.
- Added regression coverage for the exact pending browser login -> provider deletion -> late token completion sequence, deletion through both Claude provider IDs, and late credential persistence after deletion.

Focused regression tests, Node/Web typechecks, formatting, and changed-file ESLint pass.

ewen-poch:
Synced the branch with the latest main in ac4b395 to restore mergeability. The only conflict was in src/preload/index.ts; the resolution preserves both the Claude isolated browser login/cancel IPC methods and the new Office preview preload surface from main.

Validation after the merge: Node/Web typechecks pass, 62 Office preview/preload tests pass, 39 focused settings/preload tests pass, and the resolved preload file passes ESLint/Prettier.

ewen-poch:
Correction: the validated main-sync merge commit was amended to 334abbd to use the repository-compliant chore(merge) commit title. The merge content and conflict resolution are unchanged.

ewen-poch:
Rebased onto the latest main and adapted the Claude subscription onboarding flow to the new modular ProviderStep architecture. The isolated browser sign-in, manual setup-token fallback race handling, shared-profile sign-in, and pending-login cancellation behavior are preserved with dedicated ProviderStep regression coverage. Focused renderer tests (90), settings/main tests (292), Node/Web typechecks, API-map validation, and lint all pass; lint reports only pre-existing warnings.

ewen-poch:
Addressed all three findings:

- Shared Claude disconnect now persists an Open Science-local disconnected marker, clears verification, keeps the global ~/.claude profile untouched, blocks preflight/spawn, and only clears after an explicit successful sign-in.
- Shared Claude sign-in and Test connection now run the Claude subprocess probe with the resolved runtime/model environment after the auth-status check. Changing or clearing a Claude model override invalidates prior verification.
- The Claude subscription form now renders the authentication selector, detailed credential guidance, and optional model override from one reachable branch.

Added regression coverage for disconnect/reconnect spawn behavior, resolved-model probing, validation invalidation, and form rendering. Settings service tests (166), related settings/UI tests (199), Node/Web typechecks, targeted lint, and formatting checks pass.

ewen-poch:
Addressed both findings. Shared Claude keeps authentication in ~/.claude, while Open Science now provisions its managed settings, skills, and bundled/custom connector docs in the app-owned Claude directory and injects them per session through the Claude SDK settings and local-plugin options. The same settings/plugin arguments are used by shared validation probes, and the initial connector refresh is now awaited before ACP registration so the first session cannot race custom connector materialization. User-owned ~/.claude settings and os-*/mcp-* skills are left untouched. Selecting either builtin Claude mode now also persists claudeSubscriptionProviderId, so the collapsed card remembers the last explicitly selected mode after switching to another provider. Regression coverage includes shared session create/resume option forwarding, user-home preservation, connector startup ordering, runtime provisioning, and repository reload persistence. Validation passed: 167 settings service tests, 90 focused related tests, connector reload tests, Node/Web typechecks, ESLint, and Prettier.

ewen-poch:
Addressed both architecture findings in e2f1827.

- Startup connector discovery: ACP registration now waits at most 5 seconds for the initial connector-doc refresh. The refresh promise is settled defensively, so a timeout or rejection falls back to runtime registration while discovery can continue in the background. Added a hanging-discovery regression test.
- Shared Claude login concurrency: sign-in snapshots the shared provider and preferred Claude mode before browser auth, rechecks them before probing, and commits only validation/disconnect fields through an atomic repository compare-and-swap. Provider edits, deletion/disconnect, and switches to isolated now produce applied: false and cannot be overwritten or restamp the preferred mode. Added a pending-probe regression test.

Validation: 231 main settings/repository/connector tests pass; 366 settings/onboarding renderer tests pass; Node and Web typechecks pass; full lint exits 0 (existing unrelated warnings only).

ewen-poch:
Addressed the latest Architecture and Correctness findings in 197490a.

- Connector refresh scope: IPC and ACP runtime registration now complete immediately. The settled 5-second connector barrier is passed into ACP and awaited only by connect/create/resume/reset, so window and headless startup no longer inherit connector discovery latency while the first runtime use still gets the best-effort snapshot.
- Shared profile consistency: ClaudeSharedAuthController now receives the service userClaudeDir and uses it for both auth status and browser OAuth, matching shared runtime and validation configuration.
- Shared reconnect state: a successful browser OAuth now clears the app-local disconnectedAt flag even when the follow-up runtime probe fails; the probe failure remains recorded for retry and diagnostics.
- Isolated cancellation race: the browser setup-token flow re-checks its abort signal after token capture and before persistence, so switching to manual paste cannot let a cancelled browser token overwrite the pasted credential.

Validation: 286 related main-process tests pass; Node and Web typechecks pass; targeted ESLint and Prettier checks pass. Independent standards/spec review found no remaining issues.

ewen-poch:
Addressed the latest Claude provider state-consistency findings in 17205d3:

- Claude mode edits now recompute activeModel from the selected provider, including clearing a model override while that same mode is active.
- Deleting either id from the collapsed Claude subscription now reconnects the runtime whenever the deletion changes the active provider, including removal of the hidden active sibling.
- Shared Claude validation now commits through a serialized repository compare-and-swap. Provider/disconnect state, preferred mode, active provider, and active model must all still match the probed target; otherwise the stale result returns applied:false without restoring validation state.

The branch was rebased onto origin/main at 844bc9a. Targeted settings tests pass (265/265), along with Node and Web typechecks, ESLint, Prettier, and git diff checks. The full local suite completed with 5,403 passing tests; it was not fully green because the local Electron package is incomplete for 44 unrelated suites and three unrelated spreadsheet cache policy tests failed.

ewen-poch:
Adjusted the shared Claude validation CAS in a9d32ea.

The repository no longer treats the global activeProviderId/activeModel pair as an unconditional version check. Login and validation now pass the model resolved for the actual shared-Claude probe, and the serialized mutation compares that against the current effective shared-Claude model. Unrelated provider or model switches therefore no longer discard a successful result, while shared-provider edits, disconnects, preferred-mode changes, and changes to the shared model that was actually probed still return applied:false.

Added regression coverage for both paths: an unrelated active-model switch during shared login and an unrelated active-provider switch during shared validation. Settings tests pass (267/267), along with Node/Web typechecks, ESLint, Prettier, and diff checks. The full suite completed with 6,313 passing tests; the only failures were the same three unrelated spreadsheet cache policy tests.

ewen-poch:
Rebased this branch onto origin/main at 3fdd9cf and force-pushed the updated history (HEAD 560c5ed).

The runtime/settings conflicts were resolved by preserving both sides of the contract: the new mainline contextWindow propagation remains intact, while the shared-Claude sessionOptions and provisioned runtime bundle continue to flow through the same backend resolution path.

Post-rebase verification passed: 470/470 targeted ACP/settings tests, Node and Web typechecks, Prettier, and diff checks.

ewen-poch:
Addressed the late startup connector refresh path. The 5-second initialization barrier remains bounded, but if connector discovery settles after the timeout, the active runtime now requests a skills reload so the refreshed connector docs are picked up. Fast refreshes do not trigger a redundant reload. Added regression coverage for late resolve/reject paths and for completion within the barrier. Commit: 56a870f.

ewen-poch:
Moved the connector initialization barrier into AcpRuntimeCoordinator so teardown owns the same supersession generation as pending lifecycle requests. Disconnect and all shutdown paths now invalidate parked connect, create, resume, and reset requests before runtime selection, preventing a deferred request from respawning the backend after teardown. Added regression coverage for all four lifecycle entries plus shutdown. Targeted ACP tests pass (29 tests), along with node typecheck and lint.

ewen-poch:
Addressed both latest architecture findings in 85da4d7. Shared Claude model edits now preserve the renderer-visible authentication failure whenever disconnectedAt is still present, so a locally disconnected profile remains unavailable until browser login clears the disconnect. Shared and isolated Claude providers now use the same Anthropic model context-window resolution, including the active shared model override. Added regression coverage for both the disconnected edit path and shared spawn context sizing. The settings service suite passes 180/180, along with Node/Web typechecks and lint. The full suite completed with 6,357 passing tests; only the same three unrelated spreadsheet cache policy tests failed.

PR #393, opened 2026-07-24
fix(workspace): restore behavior lost in permission update

ewen-poch:
## Problem

The squash merge of #381 included unrelated reversions in image preview and artifact handling, and the new permission UI missed keyboard focus/navigation. Codex notebook activities also use a dotted title plus an MCP arguments envelope, so completed runs were rendered as generic execute tools.

## Proposed change

- restore image zoom/pan controls, dependency, reduced-motion behavior, and managed-resource fallback coverage
- restore JSON-stringified artifact `source` compatibility and its HTTP MCP integration test
- resolve notebook run identity from either provider name or title and unwrap Codex MCP arguments for transcript/code rendering
- add keyboard focus, arrow navigation, selection, and Escape focus restoration to the authorization scope menu
- tolerate unavailable or rejected Clipboard API writes without reporting a successful copy

## Scope and non-goals

This keeps the permission-dialog feature introduced by #381 and only repairs regressions and cross-framework behavior around it. It does not redesign permission policy or change artifact storage semantics.

- `npm run typecheck`
- `npm run lint` (passes with 56 pre-existing warnings, 0 errors)
- `npm run test` (5991 passed, 83 skipped)

## Review focus

- restored image preview behavior and accessibility
- Codex notebook activity identity/input normalization
- scope-menu keyboard interaction and least-privilege selection
- artifact MCP compatibility with JSON-stringified nested arguments

Follow-up to #381.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,6 +69,7 @@
         "react": "^19.2.1",
         "react-dom": "^19.2.1",
         "react-resizable-panels": "^4.12.0",
+        "react-zoom-pan-pinch": "^4.0.3",
         "shadcn": "^4.12.0",
         "streamdown": "^2.5.0",
         "tailwind-merge": "^3.6.0",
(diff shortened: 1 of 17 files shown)


ewen-poch:
The `wheel.activationKeys` callback is intentional and supported by `react-zoom-pan-pinch@4.0.3`. Its published type is `string[] | ((keys: string[]) => boolean)`, and `isPressingKeys` invokes the callback with the active keys. Array values use `every`, so `["Control", "Meta"]` would require both modifiers simultaneously instead of the desired Ctrl-or-Cmd behavior. `npm run typecheck:web` passes with the callback configuration; no change is needed for this finding.

PR #403, opened 2026-07-24
style(permission): align approval card with dialog chrome

ewen-poch:
## Problem

The inline permission approval card in the workspace conversation still uses the legacy `bg-bg-000` / `text-text-000` / `border-border-200` palette, a hand-rolled `shadow-sm` card, and ad-hoc buttons. Every other surface in the app (project, session, settings, update, data migration, error report, file browser, job detail dialogs and the shared dropdown menus) has moved to the shared `dialog-chrome` tokens, so the approval prompt visually sticks out.

## Proposed change

Restyle `PermissionApprovalControls` to the shared dialog-chrome visual language — presentation only, no behavior change:

- Card shell: `rounded-xl border border-border bg-card text-card-foreground shadow-dialog`, plus a `motion-safe` fade/slide-in on mount (matching the dialog chrome's fade-in affordance)
- Header title uses `dialogTitleClassName`; the risk label becomes the shared `Badge` (`secondary`)
- Scope dropdown matches `DropdownMenuContent` (`bg-popover`, `border-border`, `shadow-menu`, `p-1.5`) and its items match `DropdownMenuItem` (`hover:bg-muted`, selected `bg-muted`)
- Code section and detail/location lines move from `bg-bg-*` / `text-text-*` to `bg-muted` / `text-foreground` / `text-muted-foreground`
- Deny and extra-option buttons use the shared `Button` (`outline`, `sm`); the Allow split button keeps its two-segment structure (chevron stays a separate tab stop with `aria-haspopup`) but adopts the same `sm` metrics, divider, and focus ring

## Scope and non-goals

- No behavioral changes: scope defaults, option-kind matching, keyboard navigation, focus restore, and testids are untouched
- The card stays an inline conversation element — it does not become a modal overlay
- Other legacy-palette surfaces in the workspace are not part of this change

- `npm run typecheck` — green
- `npm run lint` — no new findings on the changed files
- `npm run test` — 460 files / 6057 tests green, including the 37 `PermissionApprovalControls` render/interaction tests; the source-string assertion in `workspace-components.test.tsx` is updated to the new container classes (the width-constraint intent it guards is preserved)

## Review focus

- Token choices for the inline card (`shadow-dialog` on a non-modal surface, `bg-muted/60` for the code section)
- The Allow split button remains hand-rolled (two segments, shared-button metrics) rather than the `Button` primitive — confirm the trade-off is acceptable

### DIFF
--- a/src/renderer/src/pages/workspace/ConversationPanel.tsx
+++ b/src/renderer/src/pages/workspace/ConversationPanel.tsx
@@ -326,6 +326,15 @@ const ConversationPanel = ({
                 <PermissionApprovalControls
                   requests={pendingPermissions}
                   onRespond={onRespondToPermission}
+                  notebookLookup={
+                    activeSession
+                      ? {
+                          sessionId: activeSession.id,
+                          workspaceCwd: activeSession.cwd ?? '',
+                          projectName: activeSession.projectId
+                        }
+                      : undefined
+                  }
                 />
 
                 {notebookReference || hasAnyJobs ? (
--- a/src/renderer/src/pages/workspace/PermissionApprovalControls.interaction.test.tsx
+++ b/src/renderer/src/pages/workspace/PermissionApprovalControls.interaction.test.tsx
@@ -38,23 +38,24 @@ afterEach(() => {
 })
 
 describe('PermissionApprovalControls interactions', () => {
-  it('default Allow button uses the narrowest scope (this call only), not the standing grant', () => {
-    // Least-privilege: the easiest click must be the one-time approval, not a persistent grant.
+  it('default Allow button uses the conversation scope so a repeated tool does not re-prompt', () => {
+    // The easiest click approves for the whole conversation; narrowing to a one-time
+    // approval is an explicit choice via the scope menu.
     act(() => {
       root.render(<PermissionApprovalControls requests={[baseRequest]} onRespond={vi.fn()} />)
     })
-    expect(container.textContent).toContain('this call only')
-    expect(container.textContent).not.toContain('this conversation')
+    expect(container.textContent).toContain('this conversation')
+    expect(container.textContent).not.toContain('this call only')
   })
 
-  it('Allow with default scope calls onRespond with the allow_once optionId', () => {
+  it('Allow with default scope calls onRespond with the allow_always optionId', () => {
     const onRespond = vi.fn()
     act(() => {
       root.render(<PermissionApprovalControls requests={[baseRequest]} onRespond={onRespond} />)
     })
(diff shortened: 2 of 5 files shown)


ewen-poch:
Addressed the review findings in 8b31bcc:

- **Execute-kind MCP misclassified as shell (correctness P1/P2, architecture P1):** fixed. The shell wording now only applies to non-MCP requests (`isMcp !== true && (toolKind === 'execute' || providerToolName === 'Bash')`), so an execute-kind MCP tool like `write_artifact_file` keeps its provider name in the header (`Run write_artifact_file?`) and the namespaced title as the detail line, even when there is no command preview. Added a regression test covering an execute-kind MCP request without `rawInput`.
- **Long extra-option labels could push Allow/Deny off-screen (correctness P2):** fixed. The action row now wraps, and extra-option buttons override the shared Button's `shrink-0`/`whitespace-nowrap` so provider-controlled labels wrap inside the card.

On **the primary Allow defaulting to `allow_always` (correctness P1):** this is a deliberate product decision, not an oversight — the default click should approve a recurring tool for the rest of the conversation instead of re-prompting on every call. One-time approval stays an explicit choice in the scope menu, the menu defaults are recomputed per request, and per-request state resets ensure a chosen scope never leaks into the next prompt. Keeping the behavior as-is.

ewen-poch:
Follow-up in eecb52b:

- **Opaque execute prompts (correctness P1):** fixed. For shell-classified requests, the title now stays visible as the detail line whenever no code preview renders the executable target. A non-Bash execute request carrying its command only in `title` (no `rawInput.command`) therefore still shows the command text under the generic `Run command?` header; Bash requests are unaffected because their title already renders inside the code preview. Added a regression test.

On the **session-grant default (architecture P1, correctness P2):** keeping the conversation scope as the default is the intended product behavior — approving a recurring tool once per conversation instead of on every call. The trade-offs you raise are real but accepted for now: the broker's session grant is category-keyed, and when a provider lists several `allow_always` variants the renderer already surfaces every option it cannot positively scope as its own labeled button, so a broader grant is never silently hidden — the primary button takes the first `allow_always` and the rest stay visible next to Deny. One-time approval remains an explicit choice in the scope menu, and scope/menu/collapse state resets on every new request.

ewen-poch:
Noted on the repeated **session-scope default** finding (correctness + architecture): this default is the intended product behavior for this PR and was confirmed with the requester, so I'm keeping it — rationale and mitigations (explicit scope menu, visible fallback buttons for unscoped options, per-request state reset) are in the earlier comments. No other findings remain open on eecb52b.

PR #416, opened 2026-07-24
fix(cli): fail promptly for AppImage sandbox errors

### LINKED ISSUE #415: [Bug]: open-science start fails for Linux AppImage because chrome-sandbox cannot use SUID permissions
## What happened

On Ubuntu 24.04, `open-science start --no-open` fails when using the command-line launcher installed by the Linux AppImage.

The CLI launches the packaged application in headless mode, waits 30 seconds for the local Open Science service to become healthy, and then reports a Chromium SUID sandbox failure. The sandbox helper is inside the AppImage's transient read-only, `nosuid` mount, so an ordinary user cannot configure the required SUID permissions. Changing the separate helper in a source checkout does not affect the packaged AppImage.

## What I expected

The supported Linux AppImage and its per-user command-line launcher should start the headless backend and localhost web UI as an ordinary user.

If secure sandboxed startup is unavailable on the host, the CLI should fail promptly with actionable guidance and offer an explicit, documented rootless fallback rather than waiting for the health timeout.

## Steps to reproduce

1. Use Ubuntu 24.04 on x86-64.
2. Download and launch Open Science `v0.7.0` as an AppImage.
3. In **Settings > General > Command line tool**, install the command.
4. Quit the desktop application.
5. Run `open-science start --no-open`.
6. Wait for the startup timeout.
7. Observe that the service never becomes healthy and the command exits with status 1.

## Additional context

- Environment: Ubuntu 24.04.4 LTS, x86-64, Open Science 0.7.0.
- Provider/model is not applicable; the failure occurs before the service starts.
- The AppImage sandbox helper is owned by root but has mode `755`; its FUSE mount is read-only and `nosuid`.
- Directly launching the same packaged binary with Chromium's `--no-sandbox` switch gets past the fatal error, confirming that failure occurs during sandbox initialization. This should remain an explicit fallback because it reduces security.
- Ubuntu 24.04 restricts unprivileged user namespaces through AppArmor unless an application is allowed by policy.

Relevant log:

```text
Open Science did not become healthy within 30s.

FATAL:sandbox/linux/suid/client/setuid_sandbox_host.cc:166
The SUID sandbox helper binary was found, but is not configured correctly.
Rather than run without sandboxing I'm aborting now.
The helper must be owned by root and have mode 4755.
```

Suggested acceptance criteria:

- Sandboxed startup works for ordinary users when the host supports it.
- Unsupported AppImage sandbox configurations fail promptly with actionable guidance.
- Any rootless, no-sandbox fallback is explicit, documented, and clearly warns about reduced security.
- After successful startup, `open-science status --json` reports a healthy service and `open-science stop` shuts it down normally.
- Packaged Linux headless startup has regression coverage.

ewen-poch:
## Problem

On Linux hosts where Chromium sandbox initialization fails, the AppImage headless launcher exits immediately but `open-science start` keeps polling for 30 seconds. The supported CLI also has no explicit way to select the confirmed rootless fallback.

## Proposed change

- Race service-health polling against the spawned Electron process lifecycle and surface spawn/exit failures immediately.
- Recognize Chromium SUID sandbox failures and provide actionable AppImage guidance.
- Add an explicit `start --no-sandbox` option, forward it to Electron, and print a reduced-security warning.
- Document the Linux AppImage fallback and preferred secure alternatives.
- Add regression coverage for parsing/forwarding, prompt child-exit handling, and sandbox diagnostics.

## Scope and non-goals

Sandboxed startup remains the default. This change does not automatically disable Chromium sandboxing or modify AppImage helper permissions. It does not change status or stop semantics.

- [x] A packaged-app stand-in that exits immediately now makes the CLI fail within 1.2 seconds instead of waiting 30 seconds.
- [x] `npx vitest run cli/index.test.ts cli/lifecycle.test.ts` (23 passed)
- [x] `npm run typecheck`
- [x] `npm run lint` (0 errors; 34 pre-existing warnings in unrelated files)
- [ ] `npm run test` could not complete locally because this environment has no installed Electron runtime binary (`Electron failed to install correctly`); the focused CLI suite passes.

## Review focus

Please review the child-process/health-poll race and the security wording around the explicit fallback.

Closes #415

### DIFF
--- a/cli/index.test.ts
+++ b/cli/index.test.ts
@@ -27,6 +27,13 @@ describe('CLI argument parsing', () => {
     })
   })
 
+  it('parses the explicit no-sandbox start fallback', () => {
+    expect(parseCliArgs(['start', '--no-sandbox'])).toEqual({
+      command: 'start',
+      options: { open: true, json: false, noSandbox: true }
+    })
+  })
+
   it('parses status JSON output and rejects invalid options', () => {
     expect(parseCliArgs(['status', '--json'])).toEqual({
       command: 'status',
--- a/cli/lifecycle.test.ts
+++ b/cli/lifecycle.test.ts
@@ -1,10 +1,15 @@
 import { EventEmitter } from 'node:events'
+import { closeSync } from 'node:fs'
+import { mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
 
 import { afterEach, describe, expect, it, vi, type Mock } from 'vitest'
 
 import {
   buildAppLaunchArgs,
   formatStartupFailure,
+  openLaunchLog,
   statusCommand,
   stopCommand,
   terminateDaemon,
@@ -104,6 +109,20 @@ describe('terminateDaemon', () => {
 })
 
 describe('headless startup', () => {
+  it('starts each launch with an empty diagnostic log', async () => {
+    const directory = await mkdtemp(join(tmpdir(), 'open-science-cli-log-'))
+    const logPath = join(directory, 'cli-daemon.log')
+    try {
+      await writeFile(logPath, 'stale SUID sandbox failure')
+
+      closeSync(openLaunchLog(logPath))
+
+      await expect(readFile(logPath, 'utf8')).resolves.toBe('')
+    } finally {
+      await rm(directory, { recursive: true })
+    }
+  })
+
   it('places the no-sandbox runtime switch before the development app path', () => {
     expect(buildAppLaunchArgs(['app-root'], { noSandbox: true }, 44100)).toEqual([
       '--no-sandbox',
--- a/cli/lifecycle.test.ts
+++ b/cli/lifecycle.test.ts
@@ -131,6 +131,24 @@ describe('headless startup', () => {
     expect(deps.sleep).not.toHaveBeenCalled()
   })
 
+  it('keeps waiting after a successful second-instance handoff', async () => {
+    const child = new EventEmitter()
+    const findServiceState = vi
+      .fn()
+      .mockImplementationOnce(() => {
+        child.emit('exit', 0, null)
+        return Promise.resolve(undefined)
+      })
+      .mockResolvedValue(RUNNING_STATE)
+    const deps = makeDeps({ findServiceState })
+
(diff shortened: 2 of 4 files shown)


ewen-poch:
Addressed both review comments in b55023c.

A successful Electron child exit is now treated as the expected second-instance relay to an already-running desktop app, so the CLI continues polling until the primary instance publishes healthy attached-service state. Non-zero exits, signal termination, and spawn errors still fail promptly.

Added a regression test that emits `exit(0)` before service state exists, then verifies startup succeeds once the primary instance state appears. Validation: 24 focused CLI tests pass, typecheck passes, and lint reports 0 errors (34 unrelated existing warnings).

ewen-poch:
Addressed the latest correctness review in 2d7129d.

`buildAppLaunchArgs` now emits Chromium runtime switches before development `appArgs`, producing `electron --no-sandbox <repositoryRoot> --open-science-headless --serve=...`. Packaged launches remain unchanged because their `appArgs` are empty.

The regression test was changed first and failed with the previous `electron <repositoryRoot> --no-sandbox ...` order, then passed after the implementation change. Current validation: 24 focused CLI tests pass, typecheck passes, and the changed files pass ESLint.

ewen-poch:
Addressed the two architecture findings in 46e5782:

- `waitForStartup()` is now invoked immediately after `spawn()`, before `unref()` and `closeSync()`, so the child `exit`/`error` observers are visibly attached before subsequent process setup.
- Each daemon launch opens `cli-daemon.log` with truncation instead of append. A filesystem regression test preloads a stale SUID error and verifies the current launch starts with an empty diagnostic log.

I am not adding the cross-process acknowledgement proposed by the latest correctness review in this PR. That timeout is not introduced by this change: on `main`, `startCommand` only calls `waitForState()` and the primary instance already catches `ensureWebService()` failures internally, so a failed second-instance handoff already reaches the same 30-second timeout. Treating relay `exit(0)` as fatal breaks the supported successful handoff, as the previous P1 review identified. A positive/negative acknowledgement would require a new IPC contract between the primary Electron process and the CLI and should be designed as separate work rather than folded into the AppImage sandbox fix.

Validation after the fixes: 25 focused CLI tests pass, typecheck passes, and the changed files pass ESLint.

PR #418, opened 2026-07-25
fix(notebook): skip unused R graphics devices

### LINKED ISSUE #414: [Bug]: R notebook shows a blank figure block when no plot is generated
### Preflight checklist

### What happened?

When an R notebook cell produces text or table output but does not generate a plot, Open Science displays an additional blank figure block after the text output.

For example, a cell containing only:

```r
print("text only")
```

shows the expected text and then an empty bordered rectangle with a 4:3 aspect ratio.

I expected text-only R cells to display only their text output. Figure blocks should appear only when the R code actually generates a plot.

Investigation indicates that the R kernel opens an `800×600` PNG device before every cell execution. Closing an unused PNG device still creates a valid but blank PNG. Open Science accepts every nonempty PNG file as a captured figure, so the blank image is added to the notebook output and rendered after stdout.

A minimal reproduction outside the UI confirmed that closing the unused device creates a nonempty blank file:

```text
[1] "text only"
          file size
1 page-001.png  581
```

The blank rectangle in the UI is approximately `427×320`, which matches the kernel's `800×600` PNG scaled to the renderer's maximum height.

Expected behavior:

- Text-only R cells display no figure block.
- A cell producing one real plot displays one figure block.
- A cell producing multiple plots displays one block for each real plot.
- An unused graphics device does not produce a user-visible figure.

Actual behavior:

- Text-only R cells return a blank PNG as a figure.
- The renderer displays that blank figure after the normal output.

The problem was observed in Open Science 0.6.1. I also fetched the latest original `main` branch at release 0.7.0, commit `d7f4330`, and confirmed that `resources/notebook/r_loop.R` and its R integration tests remain unchanged, so the behavior is still present in the newest source.

### Steps to reproduce

1. Run Open Science on Windows.
2. Start a session with an R notebook runtime.
3. Execute an R cell that produces text but no graphics:

   ```r
   print("text only")
   ```

4. Observe the expected text output.
5. Observe an additional empty bordered figure block below it.
6. Repeat with another text-only command and observe the same blank block.

A larger example that reproduces the behavior is:

```r
miss <- c("CAPN2", "DUSP1", "NLRP3", "PINK1", "SLC12A2", "TNF")
print(miss)
cat("\nMore text output\n")
print(data.frame(gene = miss))
```

No plotting function is called, but a blank figure block still appears.

### Operating system

Windows

The underlying R graphics behavior may also affect other operating systems, but it was observed and confirmed on Windows.

### App version

0.6.1 observed; confirmed still present in the 0.7.0 source (`d7f4330`)

### Provider / model

- Agent: Codex
- Provider: Kimi for Coding
- Model: Kimi K3 (`kimi-k3`)
- Tool: Notebook `notebook_execute`
- Language: R

The problem appears to be in the R notebook execution loop and is unlikely to depend on the selected agent or model.

```text
Minimal R graphics-device reproduction:

[1] "text only"
null device
          1
          file size
1 page-001.png  581
```

Relevant source behavior:

```text
resources/notebook/r_loop.R

- Opens page-%03d.png at 800×600 before every execution.
- Closes the device after evaluation.
- Treats every generated PNG with size > 0 as a captured figure.

src/main/notebook/loop-output-mapper.ts

- Converts every returned figure into a display output.

src/renderer/src/pages/workspace/NotebookRunOutputs.tsx

- Renders every image display output as an inline bordered image.
```

The existing R integration tests verify that real base-R and ggplot2 plots are captured, but they do not verify that a text-only execution returns zero figures.

A screenshot was captured showing R cell `[61]`, its text output, and the empty figure block underneath.

ewen-poch:
## Problem

On Windows, closing the PNG graphics device that the R notebook kernel opens for every cell can produce a valid blank PNG even when the cell only emits text. The kernel treated every non-empty PNG as a figure, so text-only cells displayed an empty figure block.

## Proposed change

- Enable the R graphics display list and check whether the kernel-owned PNG device recorded any drawing operations.
- Harvest generated PNG pages only when that device contains recorded graphics, while still removing unused raw pages.
- Keep figure detection bound to the kernel device when user code opens another graphics device.
- Add integration coverage for text-only output, exactly one plot, multiple plots, and graphics-device switching.

## Scope and non-goals

This change is limited to R notebook figure capture. It does not change output mapping, renderer behavior, or Python/JavaScript kernels.

- Text-only R cells return zero figures.
- One base R plot returns exactly one figure.
- Multiple base R plots return one figure per plot.
- Opening another graphics device does not hide the kernel PNG output.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` (7 passed)
- `npm run typecheck` (passed)
- Targeted ESLint and Prettier checks (passed)
- Full Vitest run reached 5,452 passing tests; 44 suites could not load because the local Electron binary was not installed correctly. CI provides the authoritative full-suite result.

## Review focus

Please verify that display-list recording behaves consistently for the bundled Windows R runtime and that fallback behavior preserves legitimate plots when recording is unavailable.

Closes #414

### DIFF
--- a/resources/notebook/r_loop.R
+++ b/resources/notebook/r_loop.R
@@ -28,13 +28,14 @@ read_request <- function() {
   list(req_id = req_id, code = code)
 }
 
-# Content-addresses each non-empty PNG page produced on the device into figures_dir.
-harvest_figures <- function(pattern_dir) {
+# Content-addresses each non-empty PNG page produced on the device into figures_dir. Raw pages are
+# always removed, including when the device was unused and produced a platform-specific blank PNG.
+harvest_figures <- function(pattern_dir, include = TRUE) {
   files <- list.files(pattern_dir, pattern = "^page-\\d+\\.png$", full.names = TRUE)
   out <- list()
   for (f in files) {
     info <- file.info(f)
-    if (!is.na(info$size) && info$size > 0) {
+    if (include && !is.na(info$size) && info$size > 0) {
       digest <- content_hash(f)
       dest <- file.path(figures_dir, paste0(digest, ".png"))
       file.copy(f, dest, overwrite = TRUE)
@@ -45,6 +46,19 @@ harvest_figures <- function(pattern_dir) {
     unlink(f)
   }
   out
+}
+
+# A print device can create a blank PNG merely by being opened and closed (notably on Windows).
+# Enable the display list and use its recorded operations to distinguish that empty device from a
+# cell that actually drew a plot. If recording is unavailable, retain the figure rather than risk
+# dropping a legitimate output.
+has_recorded_graphics <- function(dev_id) {
+  tryCatch({
+    if (!(dev_id %in% grDevices::dev.list())) return(TRUE)
+    grDevices::dev.set(dev_id)
+    recorded <- grDevices::recordPlot()
+    length(recorded) > 0L && length(recorded[[1L]]) > 0L
+  }, error = function(cnd) TRUE)
 }
 
 # Content hash of a file for figure dedup, using base R's tools::md5sum (no new dependency). The
@@ -58,6 +72,7 @@ run <- function(req) {
   pattern <- file.path(page_dir, "page-%03d.png")
   grDevices::png(filename = pattern, width = 800, height = 600, res = 96)
   dev_id <- grDevices::dev.cur()
+  grDevices::dev.control(displaylist = "enable")
   error <- NULL
   error_line <- NA_integer_
   stdout_text <- ""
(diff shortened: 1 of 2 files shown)


ewen-poch:
Superseding this in favor of #422, which takes a more robust approach to the same issue (#414). The final-page `recordPlot()` gating here drops earlier valid plots in multi-page cells like `plot(1:3); plot.new()` (flagged P2 by both reviewers); #422 instead detects blank capture pages by content hash per-request and preserves real figures. Closing to avoid two competing PRs for one fix.

PR #420, opened 2026-07-25
feat(skills): add agent-only bundled skills

ewen-poch:
## Problem

Open Science currently exposes every bundled skill through Settings and the Composer skill picker.
Implementation skills such as Remote Compute (SSH) should instead remain internal while still being
available to ACP agents.

## Proposed change

- Add strictly validated user and agent-only visibility metadata to bundled skills.
- Keep imported and personal skills user-visible regardless of their frontmatter.
- Project only user-visible skills through Settings, detail, enablement, forced-skill, and prompt-nudge APIs.
- Always materialize agent-only skills for supported ACP backends, even when stale settings list them as disabled.
- Mark Remote Compute (SSH) as the first agent-only bundled skill.

## Scope and non-goals

- No renderer-side skill ID blacklist or renderer API shape change.
- Compute host settings and session host selection remain user-visible.
- Shared Codex behavior is unchanged: Open Science does not write app skills into the user-owned Codex home.

- Remote Compute is absent from Settings and Composer because it is absent from the renderer skill catalog.
- Direct detail, toggle, forced-load, and nudge paths cannot expose or control agent-only skills.
- Claude and app-owned Codex provisioning retain agent-only skills despite stale disabled IDs.
- npm run typecheck passed.
- npm run lint passed with 0 errors; the existing repository baseline reports 34 warnings.
- npm run test passed: 482 files and 6370 tests, with 10 files and 84 tests skipped.

## Review focus

- The boundary between the complete agent catalog and the user-visible catalog in SettingsService.
- The mandatory agent-only predicate shared by Claude and non-Claude provisioning paths.
- Strict manifest parsing and the decision not to expose visibility through SkillView.

### DIFF
--- a/resources/skills/manifest.json
+++ b/resources/skills/manifest.json
@@ -107,6 +107,7 @@
       "id": "remote-compute-ssh",
       "name": "Remote Compute (SSH)",
       "source": "featured",
+      "visibility": "agent-only",
       "updatedAt": "2026-07-22T00:00:00.000Z"
     }
   ]
--- a/src/main/acp/ipc.ts
+++ b/src/main/acp/ipc.ts
@@ -115,6 +115,7 @@ const createRuntime = ({
         mcpHttpHost: new AgentMcpHttpHost(),
         skills: {
           needForceLoad: (ids) => settingsService.skillsNeedingForceLoad(ids),
+          markForceLoaded: (ids) => settingsService.markSkillsForceLoaded(ids),
           namesForIds: (ids) => settingsService.skillNudgeNamesForIds(ids)
         },
         artifacts: {
--- a/src/main/acp/ipc.ts
+++ b/src/main/acp/ipc.ts
@@ -115,7 +115,6 @@ const createRuntime = ({
         mcpHttpHost: new AgentMcpHttpHost(),
         skills: {
           needForceLoad: (ids) => settingsService.skillsNeedingForceLoad(ids),
-          markForceLoaded: (ids) => settingsService.markSkillsForceLoaded(ids),
           namesForIds: (ids) => settingsService.skillNudgeNamesForIds(ids)
         },
         artifacts: {
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -6085,11 +6085,9 @@ describe('ACP runtime skill force-load + nudge', () => {
     nudgeNames?: Record<string, string>
   }): {
     needForceLoad: ReturnType<typeof vi.fn<(ids: string[]) => Promise<string[]>>>
-    markForceLoaded: ReturnType<typeof vi.fn<(ids: string[]) => Promise<void>>>
     namesForIds: ReturnType<typeof vi.fn<(ids: string[]) => Promise<string[]>>>
   } => ({
     needForceLoad: vi.fn<(ids: string[]) => Promise<string[]>>(async () => options.needForceLoad),
-    markForceLoaded: vi.fn<(ids: string[]) => Promise<void>>(async () => {}),
     namesForIds: vi.fn<(ids: string[]) => Promise<string[]>>(async (ids: string[]) =>
       ids.map((id) => options.nudgeNames?.[id] ?? id)
     )
@@ -6168,7 +6166,6 @@ describe('ACP runtime skill force-load + nudge', () => {
     // The picked skill was scoped to this runtime and the agent respawned before the prompt.
     expect(spawner.spawnCount()).toBe(2)
     expect(spawner.agents[1].resumedSessions).toHaveLength(1)
-    expect(hooks.markForceLoaded).toHaveBeenCalledWith(['research'])
 
     // The nudge names the skill by its slug id (the identifier the agent's Skill tool resolves),
     // NOT its human display name — a display name like "Deep Research" is unknown to the tool.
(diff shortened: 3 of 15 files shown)


ewen-poch:
Addressed the runtime compatibility concern in 40f22b1:

- skillNudgeNamesForIds now resolves against the complete agent catalog. Persisted composer drafts and headless API requests that already contain remote-compute-ssh therefore continue to steer the ACP agent explicitly.
- skillsNeedingForceLoad intentionally still returns no reload for agent-only skills. They are mandatory in supported app-managed skill homes and are materialized regardless of disabledSkillIds, so reconnecting would be redundant.

I am keeping getSkillDetail restricted to the user-visible catalog. Allowing that lookup would reopen the hidden skill inside Settings, which conflicts with the feature requirement that agent-only skills are not visible there. Historical message mentions may retain their rendered label, but they cannot deep-link into internal skill documentation.

Validation after the adjustment: typecheck passed, lint passed with zero errors, and all 6,370 tests passed.

ewen-poch:
Implemented the upgrade-path fix in 87b0849. skillsNeedingForceLoad now resolves known IDs against the complete catalog, so a persisted draft or headless request carrying a stale disabled agent-only ID reconnects and reprovisions the active runtime before the turn. The skill remains absent from Settings and Composer discovery.

I am not retaining the historical disable state as an opt-out. Agent-only is intentionally a mandatory runtime policy in this change; honoring disabledSkillIds would leave ACP without the required skill and provide no replacement control because the skill must stay hidden. The former skill toggle was also not a host.compute authorization boundary: the control-plane REPL exposes that API independently of skill materialization, while remote command and job execution continue through their dedicated approval gates. Host discovery and knowledge-doc access are unchanged capabilities; this change controls whether ACP receives the bundled usage guidance, not whether those APIs exist.

Validation: typecheck passed; lint passed with 0 errors (34 pre-existing warnings); all 6,370 tests passed.

ewen-poch:
Addressed the repeated-respawn issue in 8cb0dbc. The first runtime check now atomically consumes a stale agent-only entry from disabledSkillIds while returning that ID for one reprovisioning reconnect. Later prompts return no reload, and concurrent checks cannot both consume the same entry. User-visible disabled skills retain the existing turn-scoped force-load behavior.

No change for the renderer visibility recommendation. Restoring discovery or detail access, including a visible non-toggleable entry, would violate the requirement that agent-only skills remain absent from both Settings and Composer search. The compute workflow does not depend on that picker: Remote Compute is always materialized for supported app-managed ACP backends, runtime-supplied legacy IDs still resolve to an explicit nudge, and the control-plane REPL baseline tells the agent to load the skill. Historical message deep-links into internal skill documentation therefore remain intentionally unavailable.

Validation: typecheck passed; lint passed with 0 errors (34 pre-existing warnings); all 6,370 tests passed.

ewen-poch:
Addressed the early-migration issue in fbeac67. The force-load check is read-only again; the runtime now calls a separate completion hook only after disconnect, reprovisioning, and session resume have succeeded. That hook clears stale agent-only disable entries while leaving user-visible disabled skills unchanged. A failed resume never calls the completion hook, so the stale flag remains available for a later retry.

No change for persisted agent-only detail lookup. Returning the hidden skill through getSkillDetail would reopen it in the Settings detail workflow, which conflicts with the requirement that agent-only skills remain fully absent from Settings and Composer search. Historical mentions retain their rendered content but intentionally cannot reveal internal skill documentation.

Validation: typecheck passed; lint passed with 0 errors (34 pre-existing warnings); all 6,371 tests passed.

ewen-poch:
Addressed the request-critical cleanup failure in 5c843c9. Once the reprovisioned session has resumed, markForceLoaded remains best-effort: persistence errors are logged and the prompt continues on the usable force-loaded generation. A regression test verifies that the skill nudge and user prompt still reach the resumed agent when settings cleanup fails.

No change for picker or detail exposure. The intended visibility model is not total user invisibility: users may see that remote-compute-ssh is being loaded during the conversation. However, the skill remains absent from Settings and Composer search, and its SKILL.md content is not exposed through conversation mentions or detail lookup. No renderer force-load path is required because the skill is always materialized for supported app-managed ACP backends; runtime loading can remain observable without making the skill selectable or its instructions readable.

Validation: typecheck passed; lint passed with 0 errors (34 pre-existing warnings); all 6,372 tests passed.

ewen-poch:
Addressed the unnecessary prompt-time respawn in 35ef548. skillsNeedingForceLoad now evaluates each requested skill with the same isSkillEnabledForAgent policy used by provisioning. Agent-only skills are therefore treated as already materialized even when a legacy disabledSkillIds entry remains, while user-visible disabled skills retain the existing turn-scoped force-load behavior. Legacy and headless agent-only IDs still resolve through the complete catalog for the explicit skill-loading nudge.

No change for persisted mention detail lookup. The intended visibility boundary allows users to observe that remote-compute-ssh is loading, but does not expose the skill in Settings or Composer search and does not reveal its SKILL.md content from conversation mentions. Falling getSkillDetail back to the agent-only catalog would return that content through the Settings detail workflow, which conflicts with this requirement. Persisted chips may remain visible as historical conversation data, but they intentionally do not open hidden instructions.

The obsolete migration completion hook, repository consume method, and their failure-handling paths were removed with the reconnect behavior. Validation: typecheck passed; lint passed with 0 errors (34 pre-existing warnings); all 6,370 tests passed.

ewen-poch:
Closing this change because the existing main behavior already supports Composer discovery, skill detail, user-controlled enablement, and ACP use. No special skill visibility or management policy is needed.

PR #422, opened 2026-07-25
fix(notebook): avoid blank R figures for text output

### LINKED ISSUE #414: [Bug]: R notebook shows a blank figure block when no plot is generated

ewen-poch:
## Problem

Closes #414.

The R notebook loop opened an 800x600 PNG device before every cell execution. On Windows, closing an unused PNG device can still write a valid non-empty blank PNG, and the notebook pipeline then maps that file as a figure and renders it after text output.

## Proposed change

- Make the R PNG capture device lazy by temporarily installing it as R's default graphics device during cell execution.
- Restore the previous default device after evaluation and close the capture device only if plotting actually opened it.
- Add an R loop integration regression for text-only output with figure capture enabled.

## Scope and non-goals

- Keeps existing base R and ggplot2 figure capture behavior.
- Does not change notebook output mapping or renderer image display behavior.
- Does not add image-content filtering in the Node renderer path.

- Text-only R cells with figure capture enabled return zero figures.
- Real base R plots continue to be captured.
- ggplot2 plots continue to be captured when ggplot2 is installed in the test R environment.

Validation run:

- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts`
  - 3 test files passed, 46 tests passed
- `npm run typecheck`
  - passed
- `npm run lint`
  - exited 0; existing warnings remain outside the touched files
- `npm run test`
  - run in an authorized environment because the suite binds local HTTP/MCP servers
  - 482 test files passed, 10 skipped
  - 6377 tests passed, 85 skipped

## Review focus

- R graphics-device lifecycle in `resources/notebook/r_loop.R`.
- Whether temporarily replacing `options(device=...)` is the right capture boundary for the persistent R loop.

### DIFF
--- a/resources/notebook/r_loop.R
+++ b/resources/notebook/r_loop.R
@@ -28,69 +28,405 @@ read_request <- function() {
   list(req_id = req_id, code = code)
 }
 
-# Content-addresses each non-empty PNG page produced on the device into figures_dir.
-harvest_figures <- function(pattern_dir) {
-  files <- list.files(pattern_dir, pattern = "^page-\\d+\\.png$", full.names = TRUE)
-  out <- list()
-  for (f in files) {
-    info <- file.info(f)
-    if (!is.na(info$size) && info$size > 0) {
-      digest <- content_hash(f)
-      dest <- file.path(figures_dir, paste0(digest, ".png"))
-      file.copy(f, dest, overwrite = TRUE)
-      out[[length(out) + 1L]] <- list(mime = "image/png", path = dest)
-    }
-    # Remove the raw page-NNN.png intermediate (copied or empty) so the figures dir keeps only
-    # content-addressed outputs instead of accumulating stray un-hashed page files.
-    unlink(f)
+run <- base::local({
+  kernel_figures_dir <- figures_dir
+  capture_width <- 800L
+  capture_height <- 600L
+  capture_res <- 96L
+  kernel_png <- grDevices::png
+  kernel_dev_off <- grDevices::dev.off
+  kernel_plot_new <- graphics::plot.new
+  capture_state <- new.env(parent = emptyenv())
+
+  reset_capture_state <- function(dev_id = NA_integer_, initial_usr = NULL) {
+    capture_state$active <- !is.na(dev_id)
+    capture_state$dev_id <- dev_id
+    capture_state$initial_usr <- initial_usr
+    capture_state$page_seen <- FALSE
+    capture_state$recorded_plot_seen <- FALSE
+    capture_state$graphics_state_seen <- FALSE
+    capture_state$closed <- FALSE
   }
-  out
-}
 
-# Content hash of a file for figure dedup, using base R's tools::md5sum (no new dependency). The
-# driver treats this value as an opaque content key.
-content_hash <- function(path) {
-  unname(tools::md5sum(path))
-}
+  reset_capture_state()
 
-run <- function(req) {
-  page_dir <- if (nzchar(figures_dir)) figures_dir else tempdir()
-  pattern <- file.path(page_dir, "page-%03d.png")
-  grDevices::png(filename = pattern, width = 800, height = 600, res = 96)
-  dev_id <- grDevices::dev.cur()
-  error <- NULL
-  error_line <- NA_integer_
-  stdout_text <- ""
-  stdout_text <- paste(capture.output({
-    # keep.source retains per-expression srcrefs so a runtime error can report the 1-based line of the
-    # top-level statement that failed (the R equivalent of a Python traceback's last user frame).
-    exprs <- tryCatch(parse(text = req$code, keep.source = TRUE), error = function(cnd) cnd)
-    if (inherits(exprs, "condition")) {
-      error <<- conditionMessage(exprs)
-    } else {
-      refs <- attr(exprs, "srcref")
-      idx <- 0L
-      tryCatch({
-        for (idx in seq_along(exprs)) {
-          res <- withVisible(eval(exprs[[idx]], envir = globalenv()))
-          if (isTRUE(res$visible)) print(res$value)
+  # Content-addresses each non-empty PNG page produced on the device into figures_dir.
+  harvest_figures <- function(pattern_dir, keep_blank_pages = TRUE, blank_hashes = character()) {
+    raw_files <- list.files(pattern_dir, pattern = "^page-\\d+\\.png$", full.names = TRUE)
+    files <- capture_page_files(raw_files)
+    out <- list()
+    for (f in files) {
+      info <- file.info(f)
+      if (!is.na(info$size) && info$size > 0 && is_png_file(f)) {
+        digest <- content_hash(f)
+        if (is.na(digest)) {
+          next
         }
-      },
-      error = function(cnd) {
-        error <<- conditionMessage(cnd)
-        if (!is.null(refs) && idx >= 1L && idx <= length(refs)) {
-          error_line <<- as.integer(refs[[idx]][1])
+        if (!keep_blank_pages && digest %in% blank_hashes) {
+          next
         }
-      },
-      interrupt = function(cnd) error <<- "interrupted")
+        dest <- file.path(kernel_figures_dir, paste0(digest, ".png"))
+        copied <- suppressWarnings(file.copy(f, dest, overwrite = TRUE))
+        if (isTRUE(copied)) {
+          out[[length(out) + 1L]] <- list(mime = "image/png", path = dest)
+        }
+      }
     }
-  }), collapse = "\n")
-  suppressWarnings(try(grDevices::dev.off(dev_id), silent = TRUE))
-  figures <- if (nzchar(figures_dir)) harvest_figures(page_dir) else list()
-  list(stdout = stdout_text, stderr = "", error = if (is.null(error)) NA else error,
-       error_line = if (is.na(error_line)) NULL else error_line,
-       result = NA, cwd = getwd(), figures = figures)
-}
+    # Remove raw page-NNN.png intermediates so the figures dir keeps only content-addressed outputs
+    # instead of accumulating stray un-hashed page files.
+    unlink(raw_files)
+    out
+  }
+
+  # Content hash of a file for figure dedup, using base R's tools::md5sum (no new dependency). The
+  # driver treats this value as an opaque content key.
+  content_hash <- function(path) {
+    digest <- suppressWarnings(try(tools::md5sum(path), silent = TRUE))
+    if (inherits(digest, "try-error") || length(digest) == 0L || is.na(digest[[1L]])) {
+      return(NA_character_)
+    }
+    unname(digest[[1L]])
+  }
+
+  blank_capture_hashes <- function() {
+    blank_dir <- tempfile("open-science-blank-r-", tmpdir = kernel_figures_dir)
+    created <- suppressWarnings(dir.create(blank_dir, recursive = TRUE, showWarnings = FALSE))
+    if (!isTRUE(created) && !dir.exists(blank_dir)) {
+      return(character())
+    }
+    on.exit(unlink(blank_dir, recursive = TRUE, force = TRUE), add = TRUE)
+
+    create_blank_pages <- function(name, open_page) {
+      current_dev <- grDevices::dev.cur()
+      opened_dev <- NA_integer_
+      device_open <- FALSE
+      pages_dir <- file.path(blank_dir, name)
+      created <- suppressWarnings(dir.create(pages_dir, recursive = TRUE, showWarnings = FALSE))
+      if (!isTRUE(created) && !dir.exists(pages_dir)) {
+        return(character())
+      }
+      pattern <- file.path(pages_dir, "page-%03d.png")
+      tryCatch(
+        {
+          kernel_png(filename = pattern, width = capture_width, height = capture_height, res = capture_res)
+          opened_dev <- grDevices::dev.cur()
+          device_open <- TRUE
+          if (isTRUE(open_page)) {
+            suppressWarnings(try(kernel_plot_new(), silent = TRUE))
+          }
+          suppressWarnings(try(kernel_dev_off(opened_dev), silent = TRUE))
+          device_open <- FALSE
+        },
+        error = function(cnd) NULL,
+        finally = {
+          open_devices <- grDevices::dev.list()
+          if (isTRUE(device_open) && !is.null(open_devices) && opened_dev %in% open_devices) {
+            suppressWarnings(try(kernel_dev_off(opened_dev), silent = TRUE))
+            open_devices <- grDevices::dev.list()
+          }
+          if (!is.null(open_devices) && current_dev %in% open_devices) {
+            suppressWarnings(try(grDevices::dev.set(current_dev), silent = TRUE))
+          }
+        }
+      )
+      capture_page_files(list.files(pages_dir, pattern = "^page-\\d+\\.png$", full.names = TRUE))
+    }
+
+    files <- c(
+      create_blank_pages("empty-device", FALSE),
+      create_blank_pages("opened-page", TRUE)
+    )
+    hashes <- vapply(files, content_hash, character(1))
+    unique(hashes[!is.na(hashes)])
+  }
+
+  is_png_file <- function(path) {
+    con <- suppressWarnings(try(file(path, "rb"), silent = TRUE))
+    if (inherits(con, "try-error")) return(FALSE)
+    on.exit(close(con), add = TRUE)
+    signature <- suppressWarnings(try(readBin(con, what = "raw", n = 8L), silent = TRUE))
+    !inherits(signature, "try-error") &&
+      identical(signature, as.raw(c(0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a)))
+  }
+
+  capture_page_files <- function(files) {
+    if (length(files) == 0L) return(character())
+    page_numbers <- as.integer(sub("^page-(\\d+)\\.png$", "\\1", basename(files)))
+    valid <- !is.na(page_numbers) & page_numbers >= 1L
+    files <- files[valid]
+    page_numbers <- page_numbers[valid]
+    if (length(files) == 0L) return(character())
+    ord <- order(page_numbers)
+    files[ord]
+  }
+
+  create_capture_page_dir <- function() {
+    for (parent_dir in c(kernel_figures_dir, tempdir())) {
+      page_dir <- tempfile("open-science-r-pages-", tmpdir = parent_dir)
+      created <- suppressWarnings(dir.create(page_dir, recursive = TRUE, showWarnings = FALSE))
+      if (isTRUE(created) || dir.exists(page_dir)) {
+        return(page_dir)
+      }
+    }
+    kernel_figures_dir
+  }
+
+  capture_device_is_open <- function(dev_id) {
+    open_devices <- grDevices::dev.list()
+    !is.na(dev_id) && !is.null(open_devices) && unname(dev_id) %in% unname(open_devices)
+  }
+
+  capture_device_has_plot <- function(dev_id) {
+    if (!capture_device_is_open(dev_id)) {
+      return(FALSE)
+    }
+    current_dev <- grDevices::dev.cur()
+    if (!identical(current_dev, dev_id)) {
+      suppressWarnings(try(grDevices::dev.set(dev_id), silent = TRUE))
+    }
+    recorded <- try(grDevices::recordPlot(), silent = TRUE)
+    if (!identical(current_dev, dev_id)) {
+      open_devices <- grDevices::dev.list()
+      if (!is.null(open_devices) && current_dev %in% open_devices) {
+        suppressWarnings(try(grDevices::dev.set(current_dev), silent = TRUE))
+      }
+    }
+    !inherits(recorded, "try-error") && length(recorded) >= 1L && !is.null(recorded[[1L]])
+  }
+
+  capture_device_usr <- function(dev_id) {
+    if (!capture_device_is_open(dev_id)) {
+      return(NULL)
+    }
+    current_dev <- grDevices::dev.cur()
+    if (!identical(current_dev, dev_id)) {
+      suppressWarnings(try(grDevices::dev.set(dev_id), silent = TRUE))
+    }
+    usr <- try(graphics::par("usr"), silent = TRUE)
+    if (!identical(current_dev, dev_id)) {
+      open_devices <- grDevices::dev.list()
+      if (!is.null(open_devices) && current_dev %in% open_devices) {
+        suppressWarnings(try(grDevices::dev.set(current_dev), silent = TRUE))
+      }
+    }
+    if (inherits(usr, "try-error")) {
+      return(NULL)
+    }
+    as.numeric(usr)
+  }
+
+  capture_device_graphics_state_changed <- function(dev_id, initial_usr = NULL) {
+    if (is.null(initial_usr)) {
+      return(FALSE)
+    }
+    usr <- capture_device_usr(dev_id)
+    !is.null(usr) && !isTRUE(all.equal(usr, initial_usr))
+  }
+
+  capture_state_device_matches <- function(which) {
+    # Device ids are reusable; callers only accept this match while capture_state still says the
+    # request-owned device is open, and the stable dev.off wrapper flips that state on close.
+    isTRUE(capture_state$active) &&
+      length(which) == 1L &&
+      !is.na(which) &&
+      !is.na(capture_state$dev_id) &&
+      identical(as.integer(unname(which)), as.integer(unname(capture_state$dev_id)))
+  }
+
+  mark_capture_page <- function() {
+    if (isTRUE(capture_state$active) &&
+        !isTRUE(capture_state$closed) &&
+        capture_device_is_open(capture_state$dev_id) &&
+        identical(grDevices::dev.cur(), capture_state$dev_id)) {
+      capture_state$page_seen <- TRUE
+    }
+  }
+
+  mark_capture_before_dev_off <- function(which) {
+    if (capture_state_device_matches(which) &&
+        !isTRUE(capture_state$closed) &&
+        capture_device_is_open(capture_state$dev_id) &&
+        !isTRUE(capture_state$graphics_state_seen) &&
+        capture_device_graphics_state_changed(
+          capture_state$dev_id,
+          capture_state$initial_usr
+        )) {
+      capture_state$graphics_state_seen <- TRUE
+    }
+  }
+
+  mark_capture_after_dev_off <- function(which) {
+    if (capture_state_device_matches(which) &&
+        !capture_device_is_open(capture_state$dev_id)) {
+      capture_state$closed <- TRUE
+    }
+  }
+
+  install_capture_binding_wrapper <- function(package, name, make_wrapper) {
+    envs <- list(asNamespace(package))
+    package_env <- suppressWarnings(try(as.environment(paste0("package:", package)), silent = TRUE))
+    if (!inherits(package_env, "try-error") && !identical(package_env, envs[[1L]])) {
+      envs[[length(envs) + 1L]] <- package_env
+    }
+
+    for (env in envs) {
+      if (!exists(name, envir = env, inherits = FALSE)) {
+        next
+      }
+      original <- get(name, envir = env)
+      if (isTRUE(attr(original, "open_science_capture_wrapper", exact = TRUE))) {
+        next
+      }
+      wrapper <- make_wrapper(original)
+      attr(wrapper, "open_science_capture_wrapper") <- TRUE
+      was_locked <- bindingIsLocked(name, env)
+      if (was_locked) unlockBinding(name, env)
+      assign(name, wrapper, envir = env)
+      if (was_locked) lockBinding(name, env)
+    }
+  }
+
+  install_capture_wrappers <- function() {
+    make_page_wrapper <- function(original) {
+      wrapper_env <- base::list2env(
+        base::list(mark_page = mark_capture_page, original = original),
+        parent = globalenv()
+      )
+      lockEnvironment(wrapper_env, bindings = TRUE)
+      eval(
+        quote(function(...) {
+          result <- base::withVisible(original(...))
+          mark_page()
+          if (result$visible) result$value else base::invisible(result$value)
+        }),
+        envir = wrapper_env
+      )
+    }
+
+    make_dev_off_wrapper <- function(original) {
+      wrapper_env <- base::list2env(
+        base::list(
+          after_close = mark_capture_after_dev_off,
+          before_close = mark_capture_before_dev_off,
+          original = original
+        ),
+        parent = globalenv()
+      )
+      lockEnvironment(wrapper_env, bindings = TRUE)
+      eval(
+        quote(function(which = grDevices::dev.cur()) {
+          before_close(which)
+          result <- base::withVisible(original(which))
+          after_close(which)
+          if (result$visible) result$value else base::invisible(result$value)
+        }),
+        envir = wrapper_env
+      )
+    }
+
+    install_capture_binding_wrapper("graphics", "plot.new", make_page_wrapper)
+    if (requireNamespace("grid", quietly = TRUE)) {
+      install_capture_binding_wrapper("grid", "grid.newpage", make_page_wrapper)
+    }
+    install_capture_binding_wrapper("grDevices", "dev.off", make_dev_off_wrapper)
+  }
+
+  install_capture_wrappers()
+
+  function(req) {
+    reset_capture_state()
+    page_dir <- if (nzchar(kernel_figures_dir)) create_capture_page_dir() else tempdir()
+    pattern <- file.path(page_dir, "page-%03d.png")
+    dev_id <- NA_integer_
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the device lifecycle concern from the Codex Correctness and Architecture reviews.

Changed the R loop so notebook capture no longer installs or restores a request-scoped `options(device=...)` hook. The loop owns an eager PNG capture device again, preserves user `options(device)` changes across the persistent R session, and uses base/grid plotting hooks to decide whether any R plotting actually occurred before harvesting `page-*.png` files. This keeps inline capture active even if a cell changes the default device option, while still suppressing non-empty blank PNG pages from text-only cells.

Added regression coverage for:
- text-only R output returning zero figures;
- non-empty PNG pages being ignored when no R plotting occurred;
- `options(device=...)` changes inside a cell not bypassing notebook figure capture or being discarded;
- plots still being captured when user code closes the current graphics device.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 3 files, 49 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized environment: 482 files passed / 10 skipped, 6377 tests passed / 88 skipped.

ewen-poch:
Addressed the second set of capture-boundary comments.

Updated the R loop so plot-detection state no longer lives in `globalenv()` and no long-lived `before.plot.new` / `grid.newpage` hooks are installed. The loop now keeps the capture device under loop ownership, enables the device display list for the request, and checks the capture device display list before harvesting pages. That keeps the bookkeeping out of the persistent user namespace and avoids depending on mutable session-global hooks.

Also covered drawing paths that render onto the already-open capture page without starting a new page, such as direct grid drawing. Added a regression for `grid::grid.rect()` so those outputs are harvested as figures instead of being discarded.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 3 files, 50 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized environment: 482 files passed / 10 skipped, 6377 tests passed / 89 skipped.

ewen-poch:
Addressed the display-list inhibition comments from the Codex Correctness and Architecture reviews.

The R loop no longer drops capture pages solely because `recordPlot()` is empty after `grDevices::dev.control(displaylist = "inhibit")`. It still uses the capture device display list when available, but falls back to the capture device contiguous `page-001`, `page-002`, ... PNG outputs and filters known blank capture pages before returning figures. Stray `page-*.png` files outside that capture sequence are ignored and cleaned up.

Added regression coverage for `grDevices::dev.control(displaylist = "inhibit"); plot(1:3)` and strengthened the text-only stray-PNG case with a real PNG file.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 1 file, 10 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 3 files, 51 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed: 482 files passed / 10 skipped, 6377 tests passed / 90 skipped.

ewen-poch:
Addressed the Architecture review comment about process-global blank-page digest caching.

The R loop now recomputes the blank capture baseline when the fallback needs to filter candidate capture pages, instead of memoizing one digest for the lifetime of the persistent R process. This keeps blank-page filtering aligned with the current request/device settings after user code changes graphics options. It also skips opening the extra blank baseline device when there are no candidate capture pages to filter.

Added a regression that first exercises the helper, then mutates the former cached-state slot, and verifies the next blank-baseline lookup is recomputed rather than returning stale process state.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "recomputes the blank capture baseline"` passed: 1 test.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 1 file, 11 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 3 files, 52 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed: 482 files passed / 10 skipped, 6377 tests passed / 91 skipped.

ewen-poch:
Addressed the Correctness review comment about post-cell blank baseline generation.

The R loop now computes the blank capture baseline before evaluating user code and passes that baseline into `harvest_figures()`. The fallback no longer opens a fresh baseline device after the cell has mutated graphics options, so text-only blank-page filtering is tied to the same pre-cell device state used to open the actual capture device.

Added a regression that fails with the previous post-cell lookup: it seeds one blank baseline in the persistent R session, mutates the lookup during the next text-only cell, writes a matching candidate page, and verifies harvesting still filters the page using the pre-cell baseline.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "filters text-only pages against the pre-cell blank capture baseline"` failed before the fix and passed after it.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 1 file, 12 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 3 files, 53 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed: 482 files passed / 10 skipped, 6377 tests passed / 92 skipped.

ewen-poch:
Addressed the Architecture review comment about figure-capture helper names being mutable from notebook cells.

The R loop now creates the figure-capture helpers inside the private closure for `run()`, captures the kernel figures directory there, and locks that closure environment. `run()` no longer resolves `blank_capture_hashes()`, `capture_device_has_plot()`, `harvest_figures()`, or their helper dependencies from `.GlobalEnv`, so notebook cells can define those names for their own work without changing the loop capture path.

Added a regression where a cell defines colliding names for the capture helpers and then plots. That test failed before the closure move by returning no figure, and now passes because capture uses the private helper bindings. I also removed the older tests that directly called/overrode `blank_capture_hashes()` as a notebook-visible helper, since those were testing the coupling this change removes.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "keeps figure-capture helpers private"` failed before the fix and passed after it.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 1 file, 11 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 3 files, 52 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed: 482 files passed / 10 skipped, 6377 tests passed / 91 skipped.

ewen-poch:
Addressed the P1 architecture issue around the PNG probe race.

Changes made:
- Guarded PNG signature reads so transiently deleted or unreadable raw `page-*.png` paths are skipped instead of terminating the R loop.
- Guarded md5 hashing and figure copy, so failed probes or copies do not emit broken figure outputs.
- Filtered unreadable blank-capture hashes.
- Added a regression that creates an unreadable `page-001.png` path, verifies the text-only response has no figures, and verifies the same R loop still answers the next request.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "skips unreadable raw PNG page paths"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts`
- `npm run typecheck`
- `npm run lint` (existing warnings only)
- `npm run test` (passed with elevated local-host permission; sandbox run fails on `listen EPERM: operation not permitted 127.0.0.1`)

ewen-poch:
Addressed the latest Codex Correctness P2 for blank graphics pages after user-managed `dev.off()`.

Changes made:
- Added a capture-device-open check before calling `recordPlot()`.
- If user code has already closed the capture device, the loop now preserves emitted raw pages instead of applying blank-hash filtering from a missing live device.
- Added a regression that simulates a backend where blank baseline PNGs are materialized, verifies the old path drops the page (`figures.length === 0`), and now expects the closed-device blank page to be preserved.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "preserves a blank graphics page"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts`
- `npm run typecheck`
- `npm run lint` (existing warnings only)
- `npm run test` (passed with elevated localhost permission)

ewen-poch:
Addressed the latest Codex Correctness P1 for intentional blank pages when `recordPlot()` cannot inspect the device.

Changes made:
- Added request-scoped capture page hooks for `before.plot.new` and `grid.newpage`, backed by loop-private state.
- The hooks are removed at the end of each request while preserving other hook changes made during the cell.
- `harvest_figures()` now keeps blank-hash pages when the request actually opened a graphics page, even if display-list recording is inhibited or the device has already been closed.
- Added regression coverage for an intentional blank page with `dev.control(displaylist = "inhibit")`, plus the earlier closed-device blank-page case.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "preserves a blank graphics page when display-list recording is inhibited"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts`
- `npm run typecheck`
- `npm run lint` (existing warnings only)
- `npm run test` (passed with elevated localhost permission)

ewen-poch:
Addressed the latest Codex Correctness and Architecture P2 findings.

Changes made:
- Made the request-scoped plot-page hooks device-aware, so plots drawn on a user-opened `pdf()`/other non-notebook device no longer mark the notebook capture page as active.
- Removed synthetic blank-page hash rendering from the live user R session.
- For text-only requests, the loop now discards the request raw `page-*.png` intermediates directly instead of comparing them against a reference blank PNG produced in the same mutable session.
- Added regressions for plots on another graphics device and for user `png`/`dev.off` traces perturbing text-only blank filtering.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "does not treat plots on another graphics device"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "keeps text-only blank filtering isolated"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts`
- `npm run typecheck`
- `npm run lint` (existing warnings only)
- `npm run test` (passed with elevated localhost permission)

ewen-poch:
Addressed the latest Codex Correctness P2 for closing an untouched capture device.

Changes made:
- `capture_has_plot` no longer treats a closed capture device as proof of notebook graphics output by itself.
- Closed-device pages are preserved only when the request observed a graphics page on the notebook capture device.
- Added a regression for `print("text only"); grDevices::dev.off()` with a materialized blank capture page, ensuring it returns no figures.

The latest Architecture review was mergeable, so no architecture-only change was needed.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "does not return a blank figure when user code closes an unused capture device"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "preserves a blank graphics page when user code closes the capture device"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "captures a plot even when user code closes"`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts`
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts`
- `npm run typecheck`
- `npm run lint` (existing warnings only)
- `npm run test` (passed with elevated localhost permission)

ewen-poch:
Addressed this round's Codex Correctness and Architecture P2 comments.

Changes made:
- Raw R capture pages now go into a per-request temporary page directory, so stale or user-created low-number `page-*.png` files in the shared figures directory cannot truncate or pollute the current request's figure harvest.
- `capture_page_files()` now sorts valid positive page numbers from that request directory instead of relying on a contiguous sequence in the shared directory.
- The R loop now records the capture device display-list after each top-level expression. This preserves a real notebook plot when user code replaces `before.plot.new` hooks before later closing the capture device, while still keeping the existing closed-unused-device blank suppression.
- Added regressions for `page-000.png` stray files and for `setHook("before.plot.new", NULL, action = "replace"); plot(...); dev.off()`.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "captures a plot even when a low-number stray raw page exists|captures a plot when user code replaces plot hooks"` passed.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 19 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 60 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized localhost environment: 482 files passed / 10 skipped, 6377 tests passed / 99 skipped.

ewen-poch:
Addressed the Architecture P2 about R loop helper lookup isolation.

Changes made:
- The private `run` helper closure is now evaluated in an environment whose parent is `baseenv()`, with only the startup `figures_dir` value injected. Kernel figure-capture helpers no longer resolve unqualified names through `.GlobalEnv`.
- `capture.output()` is now explicitly called as `utils::capture.output()` because it is the one non-base helper dependency used inside the isolated closure.
- Added a regression where one request defines `tempfile <- function(...) stop(...)` in the notebook workspace and the next request still executes normally without crashing the persistent R loop.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "keeps R loop helper lookups isolated"` passed.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 20 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 61 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized localhost environment: 482 files passed / 10 skipped, 6377 tests passed / 100 skipped.

ewen-poch:
Addressed this round's Architecture and Correctness P2 comments.

Changes made:
- Capture-device checks now include the device label and a request-local closed-device flag, so hooks and `recordPlot()` no longer treat a later user-owned device that reuses the same numeric id as the notebook capture device.
- Added a regression for `dev.off(); pdf(tempfile()); plot.new(); dev.off()` with a materialized blank capture page, ensuring the external PDF plot does not produce an inline notebook figure.
- Added a per-request PNG fallback for the case where both plot hooks and display-list recording are unavailable. The fallback inspects only raw pages from the isolated request page directory.
- The fallback decodes PNG scanlines with base R and only preserves pages that contain visible pixel variation, so text-only materialized blank pages remain suppressed while real plots rendered under `setHook(..., action = "replace")` plus `dev.control(displaylist = "inhibit")` are preserved.
- Added a regression for `setHook("before.plot.new", NULL, action = "replace"); grDevices::dev.control(displaylist = "inhibit"); plot(1:3)`.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "does not treat a reused graphics device id|captures a plot when hooks are replaced and display-list"` passed.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 22 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 63 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized localhost environment: 482 files passed / 10 skipped, 6377 tests passed / 102 skipped.

ewen-poch:
Addressed the latest Correctness and Architecture P2 comments.

Changes made:
- Capture-device identity now uses the opened notebook PNG device's `.Devices[[dev_id]]` `filepath` attribute instead of only the reusable device number and label. A user-opened `png()` device that reuses the same id/label is no longer treated as the notebook capture device.
- Added a regression for closing the notebook capture device and reopening a user `png()` device inside the same top-level expression; the user's PNG plot no longer produces a false inline notebook figure or causes the kernel to close the user device as if it were the capture device.
- The request marker hooks are now re-ensured before each top-level R expression. If a cell replaces `before.plot.new` before the later plotting expression, the loop restores only its private marker before plotting and removes only that marker during request cleanup, preserving user hook changes.
- Added a regression for an intentional blank page with replaced hooks and inhibited display-list recording: `setHook(..., action = "replace"); dev.control(displaylist = "inhibit"); plot.new(); dev.off()` now preserves the blank notebook page.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "does not confuse a user png device|preserves a blank graphics page when hooks are replaced"` passed.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 24 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 65 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized localhost environment: 482 files passed / 10 skipped, 6377 tests passed / 104 skipped.

ewen-poch:
Addressed the latest Correctness and Architecture review comments about blank-page filtering.

Changes made:
- Removed the ad-hoc PNG pixel-variation decoder and no longer treats uniform-color pages as blank.
- The R loop now generates a per-request blank capture baseline with the same notebook PNG capture settings and filters only pages whose content hash matches that baseline when no plot signal was observed.
- This keeps text-only blank pages suppressed across PNG encodings, including palette/indexed-color blanks, while preserving real uniform rendered pages when hooks are replaced and display-list recording is inhibited.
- Added regression coverage for palette PNG blank pages on text-only output and full-page uniform grid fills under replaced hooks plus inhibited display-list recording.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "palette PNG blank|uniform filled page"` passed after failing before the fix.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 26 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 67 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized localhost environment: 482 files passed / 10 skipped, 6377 tests passed / 106 skipped.

ewen-poch:
Addressed the latest AI review comments about hook re-registration and same-expression blank page handling.

Changes made:
- Removed the per-expression `ensure_capture_page_hooks(...)` call, so notebook cells that replace `before.plot.new` or `grid.newpage` keep that hook state for subsequent expressions in the same cell.
- Added capture-device graphics-state tracking using the notebook PNG device's initial `par("usr")` value. This gives the loop a request-local signal for base graphics page creation that does not depend on re-adding hooks between user expressions.
- Kept the existing hook/recordPlot/hash fallback behavior for other paths, while preserving blank pages produced after hook replacement and display-list inhibition.
- Added regressions for user-visible hook state after hook replacement and for a braced expression that replaces hooks, inhibits display-list recording, creates a blank page, and closes the capture device.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "does not re-add capture hooks|braced expression"` passed after the hook-state regression failed before the fix.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 28 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 69 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized localhost environment: 482 files passed / 10 skipped, 6377 tests passed / 108 skipped.

ewen-poch:
Addressed the blank-hash braced-expression `dev.off()` case from the Correctness review.

Changes made:
- Strengthened the regression so the cell runs as one braced expression, replaces both page hooks, inhibits display-list recording, creates a blank page, closes the capture device with bare `dev.off()`, and forces the emitted capture PNG to match the blank baseline hash. This failed before the fix with `figures.length === 0`.
- Added a request-local `dev.off` wrapper for the `grDevices` namespace and attached package binding. When the call is about to close the notebook capture device, the loop snapshots capture graphics state while the device is still open, so intentional blank pages are preserved even if the expression closes the device before the top-level post-expression check runs.
- The wrapper preserves the original binding per environment, restores it at request cleanup, and handles user traces that were applied to the temporary wrapper by rewiring the traced function back to the original binding instead of leaving stale kernel state behind.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "braced expression"` passed after failing before the fix.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts -t "user graphics traces|unused capture device|reused graphics device id|user png device"` passed.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 28 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 69 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed in an authorized localhost environment: 482 files passed / 10 skipped, 6377 tests passed / 108 skipped.

ewen-poch:
Addressed the latest AI review comments.

Changes made:
- Removed request-scoped `before.plot.new` / `grid.newpage` hook insertion, so notebook-visible `getHook()` snapshots no longer include kernel capture closures and cannot persist stale request state.
- Replaced the per-request `dev.off` rebinding with stable capture wrappers installed once for page creation and device close. The wrappers route through loop-private capture state, so saved `grDevices::dev.off` aliases still run current capture bookkeeping instead of bypassing it.
- Preserved user trace behavior by allowing traced wrapper code to resolve notebook globals while keeping wrapper internals on private bindings and base-qualified calls.
- Added regressions for hidden hook snapshots and saved `dev.off` aliases preserving blank capture pages under replaced hooks and inhibited display-list recording.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 30 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 71 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed: 482 files passed / 10 skipped, 6377 tests passed / 110 skipped.

ewen-poch:
Addressed the latest Architecture and Correctness review comments.

Changes made:
- Removed capture-device identity checks based on `.Devices[[id]]` `filepath` metadata. The R loop now uses request-local capture state plus the stable close wrapper to mark the notebook capture device closed before R can reuse the numeric device id, and harvesting still only reads from the request-owned page directory.
- Added a regression for `graphics.off()` closing the notebook capture device followed by a user-owned `png()` device in the same expression; the user device is not treated as notebook output.
- Saved kernel-owned `png`, `dev.off`, and `plot.new` functions at loop startup and use those for blank-baseline generation and kernel-owned capture open/close. User-installed traces on `grDevices::png` / `grDevices::dev.off` no longer run during hidden preflight or capture cleanup.
- Blank baseline generation now includes both an empty-device close and an explicitly opened blank page, so blank filtering remains aligned with supported PNG backends without invoking user-traced graphics functions.
- Added a regression proving text-only cells do not increment user trace counters through kernel-owned capture preflight.

Validation:
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts` passed: 32 tests.
- `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npx vitest run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/loop-output-mapper.test.ts` passed: 73 tests.
- `npm run typecheck` passed.
- `npm run lint` exited 0; existing warnings remain outside the touched files.
- `npm run test` passed: 482 files passed / 10 skipped, 6377 tests passed / 112 skipped.

PR #423, opened 2026-07-25
fix(notebook): stop flooding the env banner with micromamba output

ewen-poch:
## Summary

On an environment setup failure, the provisioning banner dumped micromamba's full stdout/stderr tail (up to 16 KB — mostly the conda package plan, `pkg==1.0=hbuild_0 - …`) into a single **unbounded** top-of-app pill, covering most of the screen and wrecking the UX.

Root cause is two-fold: the main process embedded the raw tails in the user-facing `Error.message`, and the renderer rendered that message with no length bound.

## Changes

- **`provisioner-runtime.ts` (root cause):** `runMicromamba` now embeds only a short, **stderr-preferring** excerpt (capped at 500 chars, ellipsis-prefixed when truncated) in the user-facing `Error.message`, and attaches the full `stderrTail`/`stdoutTail` + `argv`/`exitCode` to the error's structured `data` — mirroring the existing timeout branch. `errorLogFields` captures `data`, so logs keep complete diagnostics; IPC only forwards `message`, so the UI never receives the noise.
- **`EnvStatusBanner.tsx` (defense + restyle):** the error state clamps the reason to 3 lines (`line-clamp-3 break-words`) and adopts the redesigned **dialog chrome** (`rounded-xl`, `bg-card`, `border-border`, `shadow-dialog`, semantic `text-foreground`/`text-muted-foreground`) instead of the old unbounded `rounded-full` pill. The preparing state stays a compact pill.
- **`EnvProvisionOverlay.tsx` (defense):** its error message is now bounded to a `max-h-24` scrollable, `break-words` box.

## Tested

- `npm run typecheck` — clean.
- `npm run lint` — 0 errors; touched files Prettier-formatted, 0 warnings.
- Related suites green (73/73): `provisioner-runtime.test.ts` (new message/data contract + 2 KB cap), `EnvStatusBanner.render.test.tsx` (new clamp + dialog-token assertions), `notebook-env-store`, `provisioning-view`, `NotebookPreview.gate.render`.

Note: the full `npm test` shows pre-existing failures in unrelated files (`ProjectFilesView`, `WorkspaceMessageScroller`, `build/packaging`) — reproduced with these changes stashed, so they are not introduced here.

### DIFF
--- a/src/main/notebook/env-ipc.test.ts
+++ b/src/main/notebook/env-ipc.test.ts
@@ -567,6 +567,31 @@ describe('runStartupGate', () => {
     )
   })
 
+  it('logs the FULL micromamba diagnostics (data tails) even though it broadcasts only the short message', async () => {
+    // This automatic path has no per-op logging, so the gate itself must record the structured
+    // `error.data` tails. Otherwise a launch-time failure leaves only a truncated UI excerpt and the
+    // real micromamba output never crosses the main-process boundary (regression the reviewers flagged).
+    loggerSpies.error.mockReset()
+    const micromambaError = Object.assign(
+      new Error('micromamba failed (exit 1; mm create): …tail-excerpt'),
+      {
+        code: 'MICROMAMBA_EXIT',
+        data: { argv: ['mm', 'create'], exitCode: 1, stderrTail: 'Invalid package cache signature' }
+      }
+    )
+    const provisioner = fakeProvisioner({
+      restoreRelocatedEnvs: vi.fn().mockRejectedValue(micromambaError)
+    })
+    const dir = mkdtempSync(join(tmpdir(), 'os-gate-log-'))
+    await runStartupGate(provisioner, dir, vi.fn())
+
+    expect(loggerSpies.error).toHaveBeenCalledOnce()
+    const [, fields] = loggerSpies.error.mock.calls[0] as [string, Record<string, unknown>]
+    // errorLogFields lifts the error's own `data`/`code` keys, so the tails are present in the log.
+    expect(fields.code).toBe('MICROMAMBA_EXIT')
+    expect(JSON.stringify(fields)).toContain('Invalid package cache signature')
+  })
+
   it('refuses to rebuild over a recovery-blocked prefix through the REAL provisioner self-guard', async () => {
     // The startup gate drives repair/upgrade/restore through the provisioner, so the block guarantee
     // must survive that real path — not just a mock guard on the UI handlers. Wire a real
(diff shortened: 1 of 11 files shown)


ewen-poch:
**Re: [P1] Truncated error messages break cache/MAX_PATH recovery** — fixed in 54a1405.

Good catch, this was a real regression. Rather than un-truncate `Error.message` (which would re-introduce the banner flood), I kept the short stderr-first excerpt on `message` and added `micromambaDiagnosticText(error)` in `provisioner-runtime.ts`, which reconstructs the **full** stdout/stderr tails from `error.data` (falling back to `message` for errors raised without the structured payload, e.g. spawn failures). Both parsers now read it:

- `isCorruptPkgsCacheError` (`provisioner.ts`)
- `recoverWindowsMaxPathPackage` (`micromamba-cache-recovery.ts`)

So the signature / quoted over-budget path is matched against the complete diagnostics again, and self-healing repair fires even when the signature falls outside the UI excerpt. Added a regression test on the recovery side (signature present only in `error.data.stderrTail`, short `message`) and unit tests for the helper.

ewen-poch:
**Re: [P2] Startup gate discards full micromamba tails** — fixed in 54a1405.

`runStartupGate` is the one automatic path with no per-op logging, so it now records the full diagnostics itself before broadcasting the short UI message: `log.error('startup gate failed', runtimeErrorLogFields(error))`. `errorLogFields` lifts the error's own `data`/`code` keys (both are in `ERROR_DETAIL_KEYS`, and the char budget is 256 KB ≫ the 16 KB tails), so the complete stdout/stderr output crosses the main-process boundary into the logs for restore/upgrade/repair failures — the UI still gets only the bounded excerpt. Added a test asserting the logged fields carry `code` + the stderr tail.

ewen-poch:
**Re: [P2] MaxPathRetryError drops the original micromamba diagnostics** — fixed in 9772067.

Right, this wrapper was the one truncation site I missed in the first round. `MaxPathRetryError` now captures `micromambaDiagnosticText(original)` / `micromambaDiagnosticText(retry)` onto a structured `data` field (`{ originalDiagnostics, retryDiagnostics }`). The human-readable `message` stays short (composed from the two excerpt-level `.message`s for the UI), but `errorLogFields` lifts `data`, so a recovery-failure post-mortem keeps both full stdout/stderr tails. Added a test that feeds `runMicromamba`-style errors (short message + full `data.stderrTail`) through the MAX_PATH retry path and asserts both full tails survive on `.data`.

ewen-poch:
**Re: [P2] Global startup failures truncated with no way to read the reason** — fixed in 9772067.

Agreed — since `EnvStatusBanner` renders globally from `App` (including Home, where there's no `EnvProvisionOverlay`), the 3-line clamp could hide the actionable tail on the only error surface available there. Replaced `line-clamp-3` with a bounded **scrollable** reason block (`max-h-28 overflow-y-auto whitespace-pre-wrap break-words`): the banner still can't grow unbounded, but the full excerpt is now reachable by scrolling. The source excerpt is already short (`briefTail`, ~500 chars), so this stays compact in practice. Updated the render test to assert the reason node is scrollable/non-clamped and renders the full text.

PR #426, opened 2026-07-25
feat(settings): allow switching the app icon between built-in variants

### LINKED ISSUE #425: Feature request: Allow switching the app icon in Settings
## Summary

Add an option in Open Science settings that lets users switch the application icon from a set of built-in icon variants.

## Motivation

Different users may prefer different Open Science icon styles, and an in-app selector would make this customization easy without manually replacing application files or reinstalling the app.

## Proposed experience

- Add an **App icon** option under Settings (for example, under General or Appearance).
- Show a visual preview of each available built-in icon.
- Apply the selected icon to the app window and Dock/taskbar where supported.
- Persist the selection across restarts and upgrades.
- Include a **Default** option to restore the standard Open Science icon.
- If a platform requires an app restart or has limitations, explain that clearly in the UI.

## Platforms

Ideally support macOS, Windows, and Linux where the operating system permits runtime icon changes, with graceful platform-specific behavior otherwise.

ewen-poch:
## Summary

Closes #425. Adds an **App icon** picker under **Settings → General** that switches the app between two built-in icons:

- **Light** — the current dotted mark (shipped default).
- **Dark** — the original Open Science icon from the first release.

The choice is persisted in `settings.json` and applied at runtime to the **window** (Windows/Linux) and the **dock** (macOS), re-applied on every launch and to windows created later. Preview tiles are rendered in the main process from the bundled assets, so each tile shows exactly what will be applied.

### Platform behavior

- **macOS**: dock icon updates live (`app.dock.setIcon`). The window itself has no titlebar icon on macOS.
- **Windows/Linux**: window + taskbar icon updates live (`BrowserWindow.setIcon`).
- **Static installed icon** (Finder `.app`, Explorer/Start-menu `.exe`, Linux launcher) is baked into the build and is **never** rewritten at runtime (that would break signing/notarization). The UI states this so users don't expect the Finder/Explorer icon to follow.

## Changes

- **shared**: `AppIconVariant` union + guard + default + variant metadata + `AppIconPreview`; `appIconVariant` added to the settings snapshot.
- **storage**: persist `appIconVariant` (`StoredSettings` + `sanitizeSettings` + repository setter).
- **service/IPC**: get/set the variant, list icon previews (`settings:list-app-icons`, `settings:set-app-icon-variant`), and an `onAppIconVariantChanged` hook so a change applies live.
- **main**: a dependency-injected app-icon controller that applies the variant to every window (non-darwin) and the dock (darwin), re-applying on `browser-window-created`; it takes over the macOS dock icon the startup code used to set directly.
- **preload/renderer**: `setAppIconVariant` + `listAppIcons` bridge (+ regenerated web-api-map), Zustand store wiring, and an Appearance section with per-variant preview tiles.
- Assets: `resources/icon-dark.png` (the original icon, recovered from the first release).

## Tested

- `npm run typecheck`, `npm run lint` — clean.
- `npm run test` — new/updated suites cover the controller + preview builder, repository sanitize/persist, service get/set, IPC handlers (incl. rejection of unknown variants), preload channels, store setter (optimistic + revert), and a GeneralPanel/SettingsPage render test for the picker. All touched suites green.
- The preview effect reads the channel defensively, so web mode and older backends never throw during React commit.

### Not covered

- No live Windows/Linux run (developed on macOS); the window/taskbar path is exercised via unit tests of the controller rather than a real window.


ewen-poch:
**Correctness [P2] — fixed** in 0af68c3.

You're right: a `--open-science-headless` process can later surface a desktop window on a plain second launch (`routeSecondInstance` → `showMainWindow`), so gating the controller on `!webMode.headless` left it `undefined` for the whole process — later windows kept the default icon and live changes were no-ops until a full restart. It also regressed the pre-feature behavior of setting the macOS dock icon unconditionally.

The controller is now created unconditionally. Construction is safe headless: the dock is set on macOS, and the window loop is a no-op until the `browser-window-created` listener sees the first window (so a later `showMainWindow()` gets skinned, and live variant changes apply). It's held in a box mirroring `app-lifecycle`'s `trayBox` to cross the late-binding cycle with `registerIpcHandlers`.

Architecture verdict was mergeable with no findings.

ewen-poch:
**Correctness [P2] — addressed** in a905718 (reworded, not a behavior change).

Agreed the claim was the defect. `BrowserWindow#setIcon` does change the window icon (title bar / Alt-Tab, and the taskbar on most Linux WMs), but on Windows the taskbar *button* is keyed off the AppUserModelID / baked-in exe icon — which we set explicitly — so promising "taskbar" there was misleading. A bespoke Windows taskbar implementation would be heavy and I can't verify it in this environment, so I took your first option and scoped the copy instead.

Now the Settings description/note and the main/shared comments promise only what holds on every platform: the **app window icon**, plus the **Dock on macOS**. Finder/Explorer/taskbar/Start-menu are called out as the baked-in installed icon that stays the same, and the Windows-taskbar caveat is documented in the controller. `setIcon`/`dock.setIcon` are unchanged.

PR #429, opened 2026-07-25
fix(preview): render PDF pages at on-screen resolution

ewen-poch:
## Problem

The full-page PDF preview (shared by the PreviewPanel and the fullscreen view via `PdfPreviewRenderer`) rendered blurry. `PdfPageCanvas` sized its canvas backing store from `Math.min(2, devicePixelRatio)` alone. Since a page's point size (A4 ≈ 595px, Letter ≈ 612px) is smaller than the display frame (`max-w-3xl` = 768px), every page was upscaled by ~1.29× and looked soft — on both standard and high-DPI displays. The `2` cap also prevented sharper output on higher-DPI screens.

## Proposed change

- Compute the rasterization scale from the container's **measured CSS width × devicePixelRatio ÷ page point width**, so the backing store matches the physical pixels the page actually occupies on screen.
- Replace the hardcoded cap of `2` with `MAX_RENDER_SCALE = 4`, a memory-safety bound that only clamps extreme magnification of tiny pages.
- Add a `ResizeObserver` that grows the render width when the frame widens (e.g. panel → fullscreen). Shrinking reuses the existing crisp bitmap via CSS (`object-contain`) instead of re-rendering, avoiding re-render churn during drag-resize.
- Fall back to the page's intrinsic width when `clientWidth` is `0` (unlaid-out / jsdom), and degrade safely when `ResizeObserver` is unavailable.

## Scope and non-goals

- Only the full-page preview renderer (`PdfPreview.tsx`) is touched. The card **thumbnail** (`PdfThumbnail.tsx`) is intentionally left unchanged, per request.

- A 350pt-wide page shown in a 700px frame at `devicePixelRatio = 2` now backs the canvas at 1400×2000 (scale 4), not 700px. Locked in by a new test.
- `npm run typecheck` (node + web) — pass
- `npm run lint` — pass
- PDF preview test suite — 9/9 pass

## Review focus

- The "only grow the render width" strategy in the measure effect, and whether the `MAX_RENDER_SCALE = 4` bound is the right memory/quality trade-off.

### DIFF
--- a/src/renderer/src/pages/workspace/previews/renderers/PdfPreview.test.tsx
+++ b/src/renderer/src/pages/workspace/previews/renderers/PdfPreview.test.tsx
@@ -97,15 +97,14 @@ describe('PdfPreviewContent', () => {
       )
     })
     await act(async () => {
-      await Promise.resolve()
-      await Promise.resolve()
-      await Promise.resolve()
+      await vi.waitFor(() =>
+        expect(
+          container.querySelector<HTMLElement>('[data-page-number="1"]')?.style.aspectRatio
+        ).toBe('2 / 1')
+      )
     })
 
     expect(getPage).toHaveBeenCalledWith(1)
-    expect(container.querySelector<HTMLElement>('[data-page-number="1"]')?.style.aspectRatio).toBe(
-      '2 / 1'
-    )
   })
 
   it('rasterizes at the on-screen width times device pixel ratio, not the page point size', async () => {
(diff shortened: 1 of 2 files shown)


ewen-poch:
Fixed in a8c18f5. Decoupled page acquisition from width-driven rasterization as suggested:

- One effect fetches the page once while it is near the viewport and holds the proxy (deps no longer include the measured width).
- A second effect rasterizes that live page and re-runs only on width change, reusing the loaded page instead of reloading it through the range transport.
- The width is measured in a `useLayoutEffect` so the first rasterization already knows the target width — mount no longer triggers a guaranteed second render pass.

A live panel resize now re-rasterizes the same page (no `getPage`, no extra range traffic). Added a regression test asserting that a widened page re-rasterizes with `getPage` called once and `render` called twice.

ewen-poch:
Architecture review addressed in e637ec1:

- **[P1] Resize race on the same canvas** — the rasterization effect now serializes: before starting a new `page.render()` it cancels the previous task and awaits its promise to settle, so PDF.js never has two renders in flight on one canvas during a panel drag.
- **[P2] Per-page ResizeObserver** — moved the single `ResizeObserver` up to `PdfPreviewContent` (a zero-height content-box probe) and pass the measured width down as `pageWidth`. Pages no longer observe resize themselves, so widening the panel no longer fans width callbacks across every (including off-screen) page.

ewen-poch:
Correctness review addressed in e637ec1:

- **[P2] Canceled renders reported as failures** — a render canceled by scroll-out, preview switch, or a superseding resize is now treated as expected teardown: the catch returns early on the effect's own `canceled` flag and on PDF.js `RenderingCancelledException` instead of calling `setStatus('error')`. Normal scrolling/navigation no longer logs false errors or shows "Page N could not be rendered". Added a regression test asserting a scroll-out cancellation is not logged as a render failure and no error is shown.

ewen-poch:
Correctness review addressed in 710d39c:

- **[P1] Offscreen rerender race** — the round-2 serialize step awaits the previous render before the next; a page could scroll out during that await, so `dispose()` cleared `pageRef` while this effect's `canceled` flag stayed false. Now the effect re-checks `pageRef.current === page` after the await (bails if the page was disposed) and is tied to `isNearViewport`, so a scroll-out flips `canceled` and stops the rerender. No `page.render()` can run on a disposed page / detached canvas.

ewen-poch:
Architecture review addressed in 710d39c:

- **[P2] Tall/narrow page canvas overflow** — the scale was capped by width only, so a tall page in a wide panel at high DPI could produce a canvas taller than Chromium's limit and render blank. Scale is now clamped by both backing dimensions and total pixel area (`MAX_CANVAS_DIMENSION = 8192`, `MAX_CANVAS_AREA = 16Mpx`), with the hard limit taking precedence over the intrinsic-resolution floor so a page taller than the limit at scale 1 still fits. Added a regression test (200×12000 page → canvas stays ≤ 8192 per side).

PR #430, opened 2026-07-26
feat(preview): add zoom to the PDF preview

ewen-poch:
## Problem

The PDF preview had no zoom, while the image preview already supports zoom/pan. Small-print pages and dense figures were hard to read, and "fit width" was the only view available. (Fullscreen expand already exists for all file types via the panel's maximize action — only zoom was missing.)

## Proposed change

Zoom the PDF **by page width** (like Chrome / Preview), not by CSS transform:

- `PdfPreviewContent` owns a `zoom` multiplier (0.5–3). Each page's displayed width is `fitWidth × zoom` (fit width is the parent-measured content box, capped at a comfortable reading width). The displayed width tracks the *current* measured width, so narrowing the panel or returning from full screen shrinks pages back to fit.
- Each page's canvas is rasterized at that width, so text stays **crisp at every zoom level** instead of upscaling a fixed bitmap. Backing size is bounded only by the browser-canvas dimension/area clamp merged in #429 — there is no separate scale cap, so high-DPI zoom is not pre-clipped.
- Controls: a bottom-right overlay (zoom out / reset / zoom in + percentage, mirroring the image preview) and **Ctrl/Cmd+wheel**.
- Native vertical scrolling and lazy per-page rendering are preserved; a zoomed page overflows into horizontal scroll.
- Zoom resets to fit when the previewed file changes. This works on both preview surfaces: the panel remounts on a file switch, and the Files-tab dialog (which updates `item` in place with no key) is covered by resetting `zoom` when the file's `requestKey` changes.

Builds on the resolution/lifecycle work merged in #429 — parent-owned width, serialized rerender, cancel-guard, and canvas-limit clamp already live there, so this PR is just the zoom layer on top.

## Scope and non-goals

- Touches `PdfPreview.tsx` and its test file `PdfPreview.test.tsx`.
- No free-form pan (this is a scrolling document, not a single image) and no double-click-to-reset.
- The card thumbnail (`PdfThumbnail.tsx`) is unchanged.

- At 100% a page fills the fit width; zooming in re-rasterizes at higher resolution rather than upscaling.
- Narrowing the panel shrinks the displayed page back to fit (no stuck horizontal scroll at 100%).
- High-DPI zoom stays crisp: an A4 page at 768px fit, 175% zoom, DPR=2 rasterizes to the full ~2688px backing, not a clipped 2380px.
- Switching files (including the in-place dialog path) opens the next document at 100%.
- `npm run typecheck` (node + web), `npm run lint` — pass
- `PdfPreview.test.tsx` — 16/16 pass

## Review focus

- The page-width zoom model vs. CSS-transform, and the zoom range / step choices.

### DIFF
--- a/src/renderer/src/pages/workspace/previews/renderers/PdfPreview.test.tsx
+++ b/src/renderer/src/pages/workspace/previews/renderers/PdfPreview.test.tsx
@@ -612,6 +612,48 @@ describe('PdfPreviewContent', () => {
     clientWidthSpy.mockRestore()
   })
 
+  it('caps the backing scale at the deepest zoom to bound per-page memory', async () => {
+    const clientWidthSpy = vi
+      .spyOn(HTMLElement.prototype, 'clientWidth', 'get')
+      .mockReturnValue(768)
+    vi.stubGlobal('devicePixelRatio', 2)
+    // A4-like page (595x842) at 768px fit, 300% zoom, 2x DPI: the physical target scale is
+    // 768*3*2/595 = 7.74, and even the area clamp alone would allow ~5.79 (595*5.79 = 3443px).
+    // The MAX_RENDER_SCALE=5 ceiling caps it to 595*5 = 2975px so a page cannot take the full
+    // canvas-area budget.
+    getPage.mockResolvedValue({
+      getViewport: vi.fn(({ scale }: { scale: number }) => ({
+        width: 595 * scale,
+        height: 842 * scale
+      })),
+      render: vi.fn(() => ({ promise: Promise.resolve(), cancel: vi.fn() })),
+      cleanup: vi.fn()
+    })
+
+    await act(async () => {
+      root.render(
+        <PdfPreviewContent path="/workspace/deep.pdf" name="deep.pdf" source="artifact" />
+      )
+    })
+    await vi.waitFor(() => expect(container.querySelector('canvas')?.width).toBeGreaterThan(0))
+
+    // Zoom to the 300% max (eight 25% button steps).
+    for (let i = 0; i < 8; i += 1) {
+      await act(async () => {
+        container.querySelector<HTMLButtonElement>('[aria-label="Zoom in"]')?.click()
+        await Promise.resolve()
+      })
+    }
+    await vi.waitFor(() => expect(container.textContent).toContain('300%'))
+
+    const width = container.querySelector<HTMLCanvasElement>('canvas')?.width ?? 0
+    // Scale is capped at 5 → 595*5 = 2975, below the ~3443 the area clamp alone would permit.
+    expect(width).toBe(2975)
+    expect(width).toBeLessThan(3443)
+
+    clientWidthSpy.mockRestore()
+  })
+
   it('treats a render canceled by scroll-out as teardown, not a page failure', async () => {
     let intersectionCallback: IntersectionObserverCallback | undefined
     vi.stubGlobal(
--- a/src/renderer/src/pages/workspace/previews/renderers/PdfPreview.test.tsx
+++ b/src/renderer/src/pages/workspace/previews/renderers/PdfPreview.test.tsx
@@ -223,11 +223,17 @@ describe('PdfPreviewContent', () => {
     clientWidthSpy.mockRestore()
   })
 
-  it('zooms on Ctrl/Cmd+wheel and ignores a plain wheel scroll', async () => {
+  it('zooms proportionally on Ctrl/Cmd+wheel, coalesced per frame, and ignores plain scroll', async () => {
     const clientWidthSpy = vi
       .spyOn(HTMLElement.prototype, 'clientWidth', 'get')
       .mockReturnValue(400)
     vi.stubGlobal('devicePixelRatio', 1)
+    // Flush the wheel handler's requestAnimationFrame coalescing synchronously.
+    vi.stubGlobal('requestAnimationFrame', (cb: FrameRequestCallback): number => {
+      cb(0)
+      return 1
+    })
+    vi.stubGlobal('cancelAnimationFrame', () => undefined)
     getPage.mockResolvedValue({
       getViewport: vi.fn(({ scale }: { scale: number }) => ({
         width: 400 * scale,
@@ -257,14 +263,15 @@ describe('PdfPreviewContent', () => {
     })
     expect(container.textContent).toContain('100%')
 
-    // Ctrl+wheel up zooms in by one wheel step (100% -> 110%).
+    // Ctrl+wheel scales proportionally: deltaY -100 * 0.0025 = +0.25 (100% -> 125%), not a full
+    // step per event. Two small events in one frame coalesce into a single controlled change.
     await act(async () => {
       scroll?.dispatchEvent(
         new WheelEvent('wheel', { deltaY: -100, ctrlKey: true, bubbles: true, cancelable: true })
       )
       await Promise.resolve()
     })
-    await vi.waitFor(() => expect(container.textContent).toContain('110%'))
+    await vi.waitFor(() => expect(container.textContent).toContain('125%'))
 
     clientWidthSpy.mockRestore()
   })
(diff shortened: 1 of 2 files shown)


ewen-poch:
Converting to draft. To keep the two changes independent, this PR will be **rebased onto `main` and retargeted (base → `main`) once #429 merges**, rather than stacked. The zoom work will then be a clean single layer on top of the merged resolution fix. Marking ready again after that.

ewen-poch:
Rebased onto `main` now that #429 is merged, and retargeted the base from the fix branch to `main`. The branch was rebuilt as a single clean commit on top of #429 rather than a stacked rebase, since #429's parent-owned width, serialized rerender, cancel-guard, and canvas-limit clamp made the zoom layer much thinner. Marked ready for review.

ewen-poch:
Both reviews flagged the same regression — fixed in da1df58.

The zoom layer reused #429's monotonic (only-grow) `fitWidth`, but here `fitWidth` also drives the page's CSS display width, so once the preview had been shown wide (a wider panel or full screen) pages stayed pinned at that maximum and forced horizontal scroll at 100%.

Now `fitWidth` tracks the **current** measured width, so the displayed page shrinks back to fit when the panel narrows or returns from full screen. Backing-store resolution still never drops below the page's intrinsic width, so a shrink downscales the crisp bitmap via CSS rather than re-rasterizing smaller and losing sharpness. Extended the resize test with a narrow-back step asserting the displayed width shrinks (600px → 300px) while the backing store holds at the intrinsic 400px.

ewen-poch:
Correctness verdict was mergeable — thanks. Addressed the non-blocking note in d9bba1e by adding a test for the Ctrl/Cmd+wheel zoom path (asserts a plain wheel scroll does not zoom, and Ctrl+wheel steps the zoom level). (The prior Architecture run failed on a transient API error mid-review, not a verdict; this push retriggers a fresh review.)

ewen-poch:
Fixed in 7597567.

Right — the zoom reset relied on remount, which the PreviewPanel path does (keyed by `contentKey`) but the Files-tab dialog does not: it swaps `item` in place with no key, so a second PDF inherited the first's zoom. Now zoom resets to fit whenever the previewed file's `requestKey` changes, using the adjust-state-during-render pattern so it works regardless of whether the caller remounts. Added a test that swaps the file in place and asserts the new document opens at 100%.

Note: the Architecture reviewer has failed twice at the "Run Codex review" step (~24s, before producing any verdict) — it's a transient API failure on that side, not a review outcome. Correctness has run cleanly each time. This push retriggers both.

ewen-poch:
Thanks — addressed in 6edbdc6.

**[P1] Blurry mid/high zoom at high DPI** — fixed. `desiredScale` was clamped by a standalone `MAX_RENDER_SCALE = 4` before the dimension/area limits applied, so an A4 page at 768px fit / 175% / DPR=2 was clipped to 595×4 = 2380px instead of the ~2688px it needs, then upscaled by the browser. Removed the 4× cap entirely: the page now rasterizes at the full on-screen physical pixels (still floored at intrinsic size), and the canvas dimension/area clamp is the only ceiling — it already bounds GPU memory for tiny/tall pages, so the 4× limit was both harmful and redundant. Added a DPR=2 / 175% regression test asserting the backing exceeds 2380 and reaches ~2688.

**Standards (description accuracy)** — all three corrected in the PR body: the test file is now listed in scope, the file-switch reset is described as resetting on `requestKey` change (not just `contentKey` remount, since the dialog path updates in place), and the test count is now 16/16.

**Non-blocking (icon size 24px vs design.md 28px)** — keeping `icon-xs` for consistency with `ImagePreview`'s zoom controls, as you noted. A one-off 28px here would make the two viewer overlays inconsistent; if the 28px guidance is intended for all viewer controls, that's better as a separate change updating both together.

ewen-poch:
Fixed in 682e123.

Right — with `items-center` + `mx-auto`, a page zoomed wider than the pane overflowed on both sides, and a scroll container can't reach the start-side overflow that centering creates, so the left edge was unreachable at `scrollLeft = 0`. Now the pages column centers only while pages fit and switches to `items-start` once `pageWidth` exceeds the measured fit width; alignment lives solely on the column (dropped `mx-auto` from the page wrapper so it can't re-center an overflowing page). Added a test asserting the switch from centered to start-aligned when zoomed past fit.

ewen-poch:
Fixed in d16a0c2.

The handler applied a fixed 10% step on every raw wheel event, so a trackpad/pinch (many small events per gesture) jumped to min/max and flooded PDF.js with canceled renders. Now it accumulates `deltaY` and applies it proportionally (~100px notch ≈ 0.25), flushed once per animation frame via `requestAnimationFrame` — a single gesture maps to a controlled zoom amount with at most one rerasterization per frame instead of one per event. Updated the wheel test to stub rAF and assert the proportional result.

ewen-poch:
Fixed in af9bce8.

The alignment switch compared `pageWidth` against the capped `fitWidth` (max 768px) rather than the real scroll viewport, so on a pane wider than 768 (full screen or a wide panel) a 125–150% page that still fit horizontally already left-aligned. Now the component tracks the uncapped content-box width (`viewportWidth`) alongside the capped `fitWidth` and decides overflow/alignment against the real viewport — pages stay centered until they truly overflow. Added a wide-pane test (1200px) asserting the page stays centered at 125/150% and only left-aligns at 175%.

ewen-poch:
Both addressed in b720a32.

**[P2] wheel effect didn't restart on file switch** — the effect had empty deps, so a wheel delta queued before the animation frame flushed could re-apply on the next document after its reset (opening it zoomed). The effect is now keyed to `requestKey`, so a file switch tears it down and `cancelAnimationFrame`s the queued frame before its flush can run. Added a regression test that queues a Ctrl+wheel zoom, switches files in place before flushing, and asserts the new document stays at 100%.

**Standards (test overclaimed)** — the wheel test now genuinely covers what its name says: a controllable rAF stub verifies two same-frame events coalesce into a single proportional zoom (−200 → +0.5 → 150%, one flush), and the `metaKey` (Cmd) branch zooms. Plain scroll asserts nothing is scheduled.

ewen-poch:
Fixed in fb4137e.

Reintroduced a per-page render-scale ceiling (`MAX_RENDER_SCALE = 5`) applied to `desiredScale` before the dimension/area clamp. It's deliberately set **above** the ~4.5 a full-width page needs at 175% zoom on a 2x display, so normal zoom stays crisp — this preserves the earlier high-DPI fix (removing the old fixed 4× cap that was clipping 175%/2x to 2380px and blurring it) while bounding the deepest zoom so a page can't take the full 16 MP / ~64 MiB canvas budget. At 300%/2x a page now caps at scale 5 (595→2975px) instead of the ~3443px the area clamp alone would allow. Added a test asserting that ceiling.

Note on the 4× → removed → 5 arc: 4 was too low (clipped 175%/2x, blurry); unbounded let deep zoom hit the full area budget; 5 threads both — crisp through the zoom range that matters, capped before memory blows up.

ewen-poch:
Fixed in 6a2a68a.

Correct on the ordering: the wheel effect's cleanup ran as a passive effect (after paint), but a queued `requestAnimationFrame` fires before paint — so switching files right after a Ctrl/Cmd+wheel gesture let the stale frame flush onto the new document before the cleanup canceled it, even with the `requestKey` keying. Moved the wheel listener/cleanup to `useLayoutEffect`, so the file switch runs `cancelAnimationFrame` synchronously during commit, before the browser's rAF phase. The `requestKey` keying is preserved; only the effect timing changed.

ewen-poch:
Fixed in 30a7c6a.

Correct — the zoom work moved scrolling into a nested `overflow-auto` container (so the zoom-controls overlay stays fixed), but the panel/dialog still focuses the outer `<section>`, which is no longer the scroller, so PageUp/PageDown, Space, and arrows stopped reaching the PDF. Made the inner scroller keyboard-reachable: `tabIndex={0}`, `role="region"`, an `aria-label` from the file name, and a focus-visible ring. Kept the surface structure intact (option 2) so the fixed zoom overlay still works. Added a test asserting those attributes.

PR #433, opened 2026-07-26
feat(skills): import installed global skills

ewen-poch:
## Problem

Open Science runs agent frameworks with isolated profiles, so skills installed in a user's global agent directories are not available inside a project unless they are copied into the project-managed skill store.

## Proposed change

- Keep an "Import installed skills" entry visible in Settings > Skills.
- Always scan `~/.agents/skills`; additionally scan `~/.codex/skills` for Codex and `~/.claude/skills` for Claude Code.
- Let users select multiple discovered skills and import them through the existing managed-skill flow.
- Resolve source paths in the main process, support safe cross-root skill symlinks, and reject traversal or links outside the currently allowed global roots.
- Return per-skill results so one invalid selection does not abort the remaining imports.

## Scope and non-goals

- Sources are read from the local user home on macOS, Linux, and Windows through Node's home/path APIs.
- The import is explicit and copies content; it does not sync later source changes.
- Inactive framework-specific directories are not scanned.
- Absolute source paths are not exposed to the renderer.

- `npm run typecheck`
- `npm run lint` (0 errors; 34 pre-existing warnings)
- `npm test` (486 files passed, 10 skipped; 6689 tests passed, 112 skipped)
- `npm run check:web-api-map`
- Focused service, repository, IPC, and renderer tests cover source routing, multi-select import, partial failures, traversal, and symlink containment.

## Review focus

- Framework-to-source routing and realpath containment in the settings service.
- Safe handling of Claude symlinks into the shared `~/.agents/skills` root.
- Batch import result handling and selection behavior in the settings UI.

### DIFF
--- a/src/main/settings/ipc.test.ts
+++ b/src/main/settings/ipc.test.ts
@@ -601,6 +601,36 @@ describe('settings IPC handlers', () => {
     expect(onSkillsChanged).toHaveBeenCalledOnce()
   })
 
+  it('does not fire onSkillsChanged when an installed-skill batch makes no changes', async () => {
+    handlers.clear()
+    const service = createFakeService()
+    const onSkillsChanged = vi.fn()
+    const result = {
+      results: [
+        {
+          source: 'agents',
+          slug: 'existing',
+          status: 'unchanged' as const,
+          id: 'imported-existing'
+        },
+        { source: 'claude', slug: 'missing', error: 'Skill not found.' }
+      ],
+      skills: []
+    }
+    service.importAgentHomeSkills.mockResolvedValue(result)
+    registerSettingsIpcHandlers({ service: asService(service), onSkillsChanged })
+
+    expect(
+      await invoke('settings:import-agent-home-skills', {
+        skills: [
+          { source: 'agents', slug: 'existing' },
+          { source: 'claude', slug: 'missing' }
+        ]
+      })
+    ).toBe(result)
+    expect(onSkillsChanged).not.toHaveBeenCalled()
+  })
+
   it('registers the OpenCode / framework-switch channels', () => {
     handlers.clear()
     registerSettingsIpcHandlers({ service: asService(createFakeService()) })
--- a/src/main/settings/ipc.test.ts
+++ b/src/main/settings/ipc.test.ts
@@ -64,6 +64,8 @@ type FakeSettingsService = Record<
   | 'updateSkill'
   | 'deleteSkill'
   | 'importSkillZipBatch'
+  | 'listAgentHomeSkills'
+  | 'importAgentHomeSkills'
   | 'setConnectorEnabled',
   ReturnType<typeof vi.fn>
 >
@@ -141,6 +143,8 @@ const createFakeService = (): FakeSettingsService => ({
   updateSkill: vi.fn().mockResolvedValue([]),
   deleteSkill: vi.fn().mockResolvedValue([]),
   importSkillZipBatch: vi.fn().mockResolvedValue({ results: [], skills: [] }),
+  listAgentHomeSkills: vi.fn().mockResolvedValue([]),
+  importAgentHomeSkills: vi.fn().mockResolvedValue({ results: [], skills: [] }),
   setConnectorEnabled: vi.fn().mockResolvedValue({ connectors: [] })
 })
 
@@ -567,6 +571,34 @@ describe('settings IPC handlers', () => {
     expect(service.importSkillZipBatch).toHaveBeenCalledWith(request)
     expect(forwarded).toBe(result)
     expect(onSkillsChanged).toHaveBeenCalledTimes(1)
+  })
+
+  it('routes one installed-skill batch and fires onSkillsChanged once', async () => {
+    handlers.clear()
+    const service = createFakeService()
+    const onSkillsChanged = vi.fn()
+    const result = {
+      results: [
+        {
+          source: 'agents',
+          slug: 'shared',
+          status: 'imported',
+          id: 'imported-shared'
+        }
+      ],
+      skills: []
+    }
+    service.importAgentHomeSkills.mockResolvedValue(result)
+    registerSettingsIpcHandlers({ service: asService(service), onSkillsChanged })
+    const request = { skills: [{ source: 'agents', slug: 'shared' }] }
+
+    expect(await invoke('settings:list-agent-home-skills')).toEqual([])
+    const forwarded = await invoke('settings:import-agent-home-skills', request)
+
+    expect(service.listAgentHomeSkills).toHaveBeenCalledOnce()
+    expect(service.importAgentHomeSkills).toHaveBeenCalledWith(request)
+    expect(forwarded).toBe(result)
+    expect(onSkillsChanged).toHaveBeenCalledOnce()
   })
 
   it('registers the OpenCode / framework-switch channels', () => {
--- a/src/main/settings/ipc.ts
+++ b/src/main/settings/ipc.ts
@@ -15,7 +15,7 @@ import {
   type CreateSkillRequest,
   type DeleteProviderRequest,
   type DeleteSkillRequest,
-  type ImportAgentHomeSkillRequest,
+  type ImportAgentHomeSkillsRequest,
   type ImportSkillRequest,
   type ImportSkillZipRequest,
   type ImportSkillZipBatchRequest,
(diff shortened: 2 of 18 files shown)


ewen-poch:
Addressed the cross-source slug collision in de621ec. Installed-skill identity is now persisted as source plus slug, so importing agents:foo no longer marks claude:foo as imported. Re-importing the same identity is deduplicated, while a same-slug skill from another source receives the normal suffixed destination. Added service and repository regression coverage for both behaviors.

ewen-poch:
Addressed both follow-up findings in d1426d0. Legacy manifest-less installed imports are recognized by a compatibility path; when a legacy slug is ambiguous, only the framework-specific row claims it and the distinct shared row remains importable. Framework symlink aliases are now collapsed by resolved real path and direct or stale alias import requests are canonicalized to the owning source identity, preventing duplicate copies. Regression tests cover legacy single-source imports, ambiguous cross-source slugs, alias listing, and alias re-import deduplication.

ewen-poch:
Addressed the non-agent-home slug compatibility finding in 3cf350f. Existing GitHub, ZIP, and manifest-less imports now participate in a controlled slug fallback when they do not have an agent-home identity. The fallback claims at most one discovered row, so the prior same-slug import is not re-offered while a distinct skill from another global source remains selectable. Added a regression test with a GitHub-style source manifest and retained the ambiguous cross-source coverage.

ewen-poch:
Addressed the resolved-path containment issue in c1e9918. The main-process resolver now requires every resolved installed-skill target to be exactly one safe-slug child of an allowed skills root. Aliases to a skills root or a deeper descendant are rejected per item, while legitimate top-level cross-root aliases remain canonicalized. Added regression coverage for both crafted targets.

ewen-poch:
Addressed both final correctness findings and the imported-state documentation issue in d4f410d. Agent-home imports now stage and validate a local snapshot, persist a signature covering file bytes, directory structure, and portable permission bits, return unchanged only for an identical source snapshot, and atomically refresh changed copies in place. Rescans make changed installed skills selectable again. Malformed batch elements are now validated inside the per-item failure boundary, so null or undefined entries produce individual errors without blocking valid imports. The shared alreadyImported comment now distinguishes exact source/slug/content matches from the controlled slug-only compatibility fallback. Added regression coverage for content, empty-directory and POSIX mode updates, unchanged reimports, failed-refresh rollback and staging cleanup, and malformed mixed batches.

ewen-poch:
Addressed the latest correctness findings in 3df0e82.

- Global installed-skill sources are scanned independently, so an unreadable or corrupt shared/framework directory no longer hides healthy-source results.
- Installed-skill rows are bound to the framework that produced them. Automatic, manual, and post-import scans share one request generation, so late results and errors from older scans are ignored.
- Framework changes remount the import view, so rapid Claude → Codex → Claude switches cannot re-enable cached rows before the new scan completes.

Added regression coverage for ENOTDIR source isolation, pending framework switches, ABA switching, and a late manual-rescan race.

Validation:
- npm run typecheck
- npm run lint (0 errors; 33 pre-existing warnings)
- vitest run (6702 passed, 0 failed, 112 skipped)

ewen-poch:
The OpenCode source recommendation is intentionally not applied in this PR. The product rule for this import surface is framework-specific: Codex adds ~/.codex/skills, Claude Code adds ~/.claude/skills, and every other framework scans only the shared ~/.agents/skills source. Including ~/.claude/skills for OpenCode would broaden the agreed source policy and make the renderer copy diverge from that requirement. If OpenCode compatibility should later include Claude-owned global skills, that should be a separate product decision and follow-up change.

ewen-poch:
Addressed the all-source scan failure in 5df9595. Source discovery now uses Promise.allSettled: any fulfilled source keeps its usable results, while a scan where every configured source rejects rethrows the first real filesystem error so the renderer shows a failure instead of a false empty state. Added a regression with both Claude and shared skill roots returning ENOTDIR; the existing partial-failure regression still verifies that one healthy source remains usable.

Validation:
- npm run typecheck
- npm run lint (0 errors; 33 pre-existing warnings)
- vitest run (6703 passed, 0 failed, 112 skipped)

ewen-poch:
Addressed the masked active-source failure in 7acb007. Discovery now suppresses a source error only when another configured source produced at least one usable skill. If all visible results are empty and any source failed, the first real filesystem error is surfaced; when every source is healthy or simply missing and empty, the normal empty state remains unchanged. Added a regression for a missing shared ~/.agents/skills directory combined with an ENOTDIR failure from the active Claude source, while retaining the partial-success and all-source-failure coverage.

The OpenCode source routing remains intentionally unchanged per the PR scope.

Validation:
- npm run typecheck
- npm run lint (0 errors; 33 pre-existing warnings)
- vitest run (6704 passed, 0 failed, 112 skipped)

ewen-poch:
Addressed both current review findings in f879ba8.

- Installed-skill batches now call onSkillsChanged only when at least one item was imported or updated. All-error and all-unchanged batches return their per-item results without recycling the active runtime. Added IPC coverage for mixed unchanged/error results while retaining the successful-import reload assertion.
- Installed-skill discovery/import is now desktop-only. The web server removes both channels from bootstrap capability discovery and rejects direct /rpc calls with 403 before IPC dispatch. The web bootstrap installs only advertised invoke methods, and Settings hides and guards the installed-skill view when that desktop bridge capability is absent. Native Electron preload behavior is unchanged.

Validation:
- npm run typecheck
- npm run lint (0 errors; 33 pre-existing warnings)
- npm test (486 files passed; 6706 tests passed, 0 failed, 112 skipped)
- npm run check:web-api-map

ewen-poch:
Addressed the symlink-alias upgrade regression in a45c7e8.

Canonical realpath deduplication now retains every source/slug alias and assigns manifest-less legacy fallback matches using the alias source, so a framework-specific symlink wins the intended same-slug collision without suppressing an unrelated shared skill. Imports match existing source metadata through those aliases, and the post-containment canonical path is used to verify unchanged alias signatures, keeping the picker state accurate without relaxing nested-symlink validation.

Regression coverage now includes manifest-less alias imports, same-slug alias collisions, stale alias metadata updates, and unchanged alias metadata after a source moves behind a symlink.

Validation: npm run typecheck; npm run lint (0 errors, 33 pre-existing warnings); npm test (486 files passed, 6709 tests passed, 0 failed, 112 skipped); npm run check:web-api-map.

ewen-poch:
Addressed the slug-only fallback findings in 00abec9.

Metadata-less imported records are now only legacy fallback matches when both the alias slug and the actual skill-tree signature match. The internal .source.json file is excluded from that comparison, so identical content can still deduplicate across import flows, while a different same-slug GitHub/ZIP skill remains selectable and imports under the normal suffixed ID.

The import path also revalidates the decision inside the repository lock: it stages the current installed source snapshot first, then compares that staged signature with the current metadata-less imported tree before returning unchanged. A stale scan result can therefore no longer suppress an import if either tree changed.

Regression coverage includes different and identical same-slug GitHub imports, metadata-less legacy records, alias/symlink collisions, and stale fallback authorization.

Validation: npm run typecheck; npm run lint (0 errors, 33 pre-existing warnings); npm test (486 files passed, 6711 tests passed, 0 failed, 112 skipped); npm run check:web-api-map.

ewen-poch:
Addressed the latest correctness and architecture findings in b2af078:

- Legacy metadata-less imports now mark every same-content source row as already imported, while content-signature checks continue to exclude unrelated same-slug sources.
- Alias-only matches are automatically migrated to the canonical source/slug through the existing staged atomic swap, so the installed state remains recognized after switching frameworks. This also covers framework skill roots that symlink to the shared `~/.agents/skills` root.
- Automatic migration now uses compare-and-swap guards on both sides: the scanned source snapshot and the matched imported slug, identity, manifest signature, and live tree signature must all remain unchanged. Deleted, edited, or stale records are left untouched and the row remains available for a deliberate manual import.
- Added regressions for all identical fallback candidates, UI-reachable identity migration, source changes, imported-record deletion/edit races, and root-directory symlink aliases.

Validation: `npm run typecheck`, `npm run lint` (0 errors; 33 pre-existing warnings), `npm run check:web-api-map`, and the full Vitest suite (6715 passed, 112 skipped).

ewen-poch:
The OpenCode source recommendation is intentionally not applied. The accepted product rule for this surface is explicit: Codex adds `~/.codex/skills`, Claude Code adds `~/.claude/skills`, and every other framework scans only the shared `~/.agents/skills` source. Adding the Claude-owned directory for OpenCode would contradict that requirement and broaden this PR beyond its agreed scope. The canonical-identity changes in b2af078 do not alter that routing policy.

ewen-poch:
The correctness finding requests the same OpenCode routing change as the architecture review, so it is also intentionally not applied. The explicit acceptance rule for this feature is that Codex adds `~/.codex/skills`, Claude Code adds `~/.claude/skills`, and all other frameworks—including OpenCode—scan only the shared `~/.agents/skills` directory. Expanding OpenCode to a Claude-owned source would change the agreed product behavior rather than correct the implementation. No code change is warranted for this finding.

PR #440, opened 2026-07-26
feat(model-settings): map effort to model capabilities

ewen-poch:
## Problem

Reasoning effort was treated as one framework-level scale even though supported values belong to the selected model API. Framework adapters independently handled values, and custom models could not declare their actual capability set.

## Proposed change

- Add static per-provider and per-model effort profiles with five stable user-intent slots.
- Render only the selected model real concrete choices, plus a separate Default option.
- Let custom models enable or disable effort and choose a common 2-5 level preset; legacy/unconfigured custom models default to five levels.
- Resolve intent once in SettingsService, then let Claude Code, OpenCode, Codex, ACP, and the Responses bridge carry that resolved value.
- Encode resolved values into each framework valid transport vocabulary; preserve exact model-native values through the Codex Chat bridge upstream override.
- Keep live changes scoped to the active runtime and its bridge, including deferred provider reconnects.

## Scope and non-goals

- Provider and model switching performs no capability network request and uses no capability cache.
- Unknown OpenRouter model families are conservative and do not advertise unverified effort controls.
- Framework transport encoding does not alter the persisted five-slot user intent or model capability profile.

- Built-in and custom model profiles synchronously refresh the effort control.
- Two-, three-, four-, and five-level profiles preserve relative intent across model switches.
- Agent framework switching does not change model effort choices.
- ACP uses exact matching; Codex and OpenCode generated configs contain only values valid for their transport vocabularies.
- npm run typecheck passes.
- npm run lint passes with 0 errors; 24 pre-existing unrelated warnings remain.
- npm test passes: 487 files, 6,774 tests; 10 files and 112 tests skipped.

## Review focus

Please focus on the static profile assignments, intent-slot projection, subscription-provider parity, framework transport encoders, and the active-versus-retiring runtime boundary for live effort changes.

### DIFF
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -308,7 +308,7 @@ describe('AcpRuntimeCoordinator', () => {
     expect(coordinator.getSnapshot().permissionGrants).toEqual({})
   })
 
-  it('keeps reconnecting settings on the active generation and fans out live effort', async () => {
+  it('keeps reconnecting settings and model-resolved effort on the active generation', async () => {
     const created: ReturnType<typeof createFakeRuntime>[] = []
     const coordinator = new AcpRuntimeCoordinator((callbacks) => {
       const fake = createFakeRuntime({
@@ -330,13 +330,13 @@ describe('AcpRuntimeCoordinator', () => {
     expect(created).toHaveLength(2)
     expect(created[0].requestProviderReconnect).not.toHaveBeenCalled()
     expect(created[0].requestSkillsReload).not.toHaveBeenCalled()
-    expect(created[0].applyReasoningEffortChange).toHaveBeenCalledWith('high')
+    expect(created[0].applyReasoningEffortChange).not.toHaveBeenCalled()
     expect(created[1].requestProviderReconnect).toHaveBeenCalledOnce()
     expect(created[1].requestSkillsReload).toHaveBeenCalledOnce()
     expect(created[1].applyReasoningEffortChange).toHaveBeenCalledWith('high')
   })
 
-  it('keeps the active effort result when a retiring generation rejects', async () => {
+  it('surfaces an active generation effort failure without touching a retiring model', async () => {
     const created: ReturnType<typeof createFakeRuntime>[] = []
     const coordinator = new AcpRuntimeCoordinator((callbacks) => {
       const fake = createFakeRuntime({
@@ -351,9 +351,12 @@ describe('AcpRuntimeCoordinator', () => {
     await coordinator.createSession()
     await coordinator.requestAgentFrameworkSwitch()
     await coordinator.createSession()
-    created[0].applyReasoningEffortChange.mockRejectedValue(new Error('old effort failed'))
+    created[1].applyReasoningEffortChange.mockRejectedValue(new Error('active effort failed'))
 
-    await expect(coordinator.applyReasoningEffortChange('high')).resolves.toBe(true)
+    await expect(coordinator.applyReasoningEffortChange('high')).rejects.toThrow(
+      'active effort failed'
+    )
+    expect(created[0].applyReasoningEffortChange).not.toHaveBeenCalled()
     expect(created[1].applyReasoningEffortChange).toHaveBeenCalledWith('high')
   })
 
--- a/src/main/acp/runtime-coordinator.ts
+++ b/src/main/acp/runtime-coordinator.ts
@@ -13,7 +13,7 @@ import type {
   AcpSetPermissionProfileRequest,
   AcpStateSnapshot
 } from '../../shared/acp'
-import type { ReasoningEffort } from '../../shared/settings'
+import type { ResolvedReasoningEffort } from '../../shared/reasoning-effort'
 import { AcpRuntime, type AcpRuntimeCallbacks } from './runtime'
 import type { AcpRuntimeActivity, AcpRuntimeActivityOptions } from './runtime-activity'
 import { ConversationPermissionGrantStore } from './permission-broker'
@@ -250,20 +250,11 @@ class AcpRuntimeCoordinator {
     return this.getActiveRuntime().requestSkillsReload()
   }
 
-  async applyReasoningEffortChange(effort: ReasoningEffort): Promise<boolean> {
-    const active = this.getActiveRuntime()
-    const activeResult = active.applyReasoningEffortChange(effort)
-    const otherResults = Array.from(this.runtimes)
-      .filter((runtime) => runtime !== active)
-      .map((runtime) => runtime.applyReasoningEffortChange(effort))
-
-    // Effort is a non-disruptive live session option, so every still-running generation receives the
-    // global preference. Only the active result controls reconnect fallback: an old generation that
-    // cannot apply effort or rejects must not force the selected generation to respawn unnecessarily.
-    const results = await Promise.allSettled([activeResult, ...otherResults])
-    const activeOutcome = results[0]
-    if (activeOutcome.status === 'rejected') throw activeOutcome.reason
-    return activeOutcome.value
+  async applyReasoningEffortChange(effort: ResolvedReasoningEffort): Promise<boolean> {
+    // The settings layer resolved this value against the currently selected model. Retiring
+    // generations stay pinned to their own provider/model and therefore must not receive a value
+    // resolved for a different model profile.
+    return this.getActiveRuntime().applyReasoningEffortChange(effort)
   }
 
   writeArtifactForCurrentRun(
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -8234,6 +8234,51 @@ describe('ACP runtime — session effort', () => {
     expect(setReasoningEffort).toHaveBeenNthCalledWith(2, undefined)
   })
 
+  it('defers a new-model effort while the old provider is waiting to reconnect', async () => {
+    const process = new FakeAgentProcess()
+    const promptStarted = createDeferred()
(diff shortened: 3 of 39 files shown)


ewen-poch:
Architecture review follow-up:

- **P1 — no change in this PR.** The omission for native Codex is deliberate: current Codex configuration has no reliable exact representation for `none`, `max`, or `ultra`. Mapping `max`/`ultra` back to `xhigh` would recreate the silent second approximation this change is designed to remove and would misrepresent the selected model API value. Chat-completions providers are covered exactly by the bridge override. Exact support for direct Responses providers would require a separate Responses passthrough override, not an adapter-side clamp; that is outside this scoped change and is documented in the PR non-goals.

- **P2 — intentional relative-intent behavior.** The persisted value is a five-slot relative intent, not the concrete label from the current model. For a two-level profile the approved mapping is `A B B B B`: selecting the lowest available concrete option persists slot A, while the highest option persists slot E. Therefore a model whose lowest concrete value is `high` correctly maps back to `low` when switching to a richer five-level model. This is covered by the reduced-profile mapping tests and preserves relative cost/strength rather than semantic label identity.

ewen-poch:
Correctness review follow-up:

- **Codex transport — accepted.** The adapter now uses a complete, typed encoder from every `ModelReasoningEffort` value into Codex supported `low | medium | high | xhigh` vocabulary. `none`/`minimal` encode to `low`, while `max`/`ultra` encode to `xhigh`. For Chat-bridged providers this value is transport-only; the bridge still overrides the upstream request with the exact model-native effort.

- **OpenCode transport — accepted.** The generated model config now uses a complete encoder into the OpenCode/AI SDK `low | medium | high` vocabulary instead of serializing unsupported provider-native literals. Lower values encode to `low`, `medium` remains exact, and upper values encode to `high`.

This also refines the earlier Architecture P1 decision: the model profile and persisted intent remain framework-independent, while each adapter now guarantees a valid last-mile representation. ACP session options remain exact-match only. Added table-driven coverage for every model effort value. Validation: typecheck passed, lint has 0 errors, and the full suite passes 487 files / 6,774 tests.

ewen-poch:
Second correctness review follow-up:

- **Direct Codex Responses — no further change.** Exact preservation of provider-native values outside the Codex vocabulary would require owning the direct Responses HTTP transport and adding a passthrough override proxy. That is materially broader than this settings/profile change. The adapter now guarantees a valid explicit Codex value instead of dropping the setting or emitting an invalid literal; Chat-bridged providers retain the exact upstream override.

- **OpenCode — no further change.** OpenCode/AI SDK exposes only `low | medium | high` through this configuration surface. Exact forwarding of `none`, `minimal`, `xhigh`, `max`, or `ultra` would require a new provider-specific transport interception layer. Restricting the UI by framework would violate the approved invariant that the selected model owns the visible effort choices and that switching Agent frameworks does not change those choices.

The remaining difference is therefore an explicit last-mile transport trade-off, not an unhandled value: model profiles and persisted intent remain provider/model-native, while adapters use complete typed encoders into their supported vocabularies. Building exact passthrough proxies for direct Responses and OpenCode should be evaluated as separate transport work rather than expanding this PR.

ewen-poch:
Architecture feedback:

- P1: No further change. Model capability remains the source of the visible effort choices, so changing those choices when the agent framework changes would violate the model-driven UI contract. The adapters now encode every model-native value into each transport valid vocabulary; exact upstream preservation on direct Codex/OpenCode paths would require an app-owned HTTP passthrough, which is a materially broader change.
- P2: Addressed in d00a72c. An unpinned Codex subscription now treats the runtime model effort capability as unknown: the UI exposes only Default and backend resolution omits sessionEffort until a model is explicitly selected. Added main-process and renderer regression coverage.

Validation: typecheck passed; lint completed with 0 errors; all 487 test files passed (6,776 tests, 112 skipped).

ewen-poch:
No additional code change for these two repeated transport-boundary findings. Codex and OpenCode adapters must emit values accepted by their native transports, so their complete encoders intentionally map model-native values to the nearest supported rung. Exact preservation would require an app-owned direct request/passthrough path, while constraining the selector by framework would make model-defined choices change when users switch agent frameworks. Both are broader product/architecture changes than this PR. The separate unknown-model case for unpinned Codex subscriptions is fixed in d00a72c by exposing only Default and omitting the override.

ewen-poch:
Same disposition for these repeated Correctness findings: no additional code change. The Codex and OpenCode encoders now cover the complete app/model vocabulary and emit only values their native transports accept. Preserving unsupported literals exactly would require a separate app-owned request path; framework-dependent UI gating would violate the selected-model-driven effort contract. Those broader changes are not justified in this scoped PR. The concrete unpinned Codex mismatch is already fixed in d00a72c.

ewen-poch:
Accepted the low-cost UX clarification from the review summary in 2c484f8: the Reasoning effort section now explains that model switches preserve relative strength and that some agent frameworks approximate unsupported levels. The custom-provider form now tells users to choose the exact effort values accepted by the model and explains that the app maps five relative strengths onto that set. No transport or mapping behavior changed. Added renderer assertions for both explanations. Validation: typecheck passed, lint completed with 0 errors, and all 487 test files passed (6,776 tests; 112 skipped).

ewen-poch:
Accepted the Standards judgement call in 8c81286. Provider-to-effort-profile routing now lives behind one pure shared resolver used by both the main settings path and the renderer selector. Its exhaustive ProviderType switch makes future provider additions fail type checking until their effort behavior is defined, while preserving the empty-settings five-level default and the unpinned-Codex unknown capability behavior. Added interface-level coverage for empty settings, official model overrides, pinned/unpinned Codex subscriptions, Claude subscriptions, custom presets/defaults, and malformed official data. No functional mapping changed. Validation: typecheck passed, lint completed with 0 errors, and all 488 test files passed (6,784 tests; 112 skipped).

ewen-poch:
MiniMax-M3 and MiniMax-M3[1m] now expose the documented binary reasoning control: Default, None, and High. The M2 models remain unsupported until MiniMax documents an effort contract for them. I also removed the stale Codex transport collapse so model_reasoning_effort preserves none, minimal, max, and ultra instead of rewriting them to low or xhigh; otherwise selecting None for MiniMax would still enable thinking. This matches the MiniMax Codex guide and the current open ReasoningEffort handling in the official Codex source. Validation: targeted effort tests, typecheck, lint (existing warnings only), and the full suite with 488 test files / 6785 tests passed.

ewen-poch:
Implemented provider-native none compatibility in 076e325.

- Added fixed none-capable profiles for GPT-5.5, DeepSeek V4, and documented OpenRouter routes, including Qwen's binary thinking switch.
- Added a shared Chat transport resolver: GLM uses reasoning_effort; DeepSeek, MiniMax, and MiMo use thinking.type; OpenRouter uses reasoning.enabled or reasoning.effort.
- OpenCode now preserves none, minimal, xhigh, and max; only ultra clamps to its max transport rung. A local OpenCode 1.18.3 capture confirmed reasoning.enabled=false reaches the wire.
- Validation: typecheck passed, lint passed with 0 errors, and the full Vitest suite passed 488 files / 6796 tests.

ewen-poch:
Addressed the outstanding bridge findings in 5ff777c.

- The Responses ingress validator now accepts the complete app effort vocabulary, including max and ultra, so model-resolved bridge overrides are not rejected before conversion.
- Mutable bridge state is now isolated per backend generation. Each lease owns its reasoning override, replay cache, and reviewer scope, so a live effort change on a new runtime cannot alter an overlapping retiring runtime.
- Added regression coverage proving concurrent generations remain pinned to low and max independently while one generation changes live to high.

Validation: typecheck passed, lint completed with 0 errors (24 unrelated existing warnings), and the full Vitest suite passed 488 files / 6797 tests.

ewen-poch:
Implemented the actionable model/transport feedback in 16005ce:

- Main and renderer now resolve the same effective catalog model. A backend generation uses that single model for the effort profile, compatibility check, framework configuration, and upstream bridge target.
- Custom providers now persist an explicit reasoning request format: OpenAI-compatible `reasoning_effort`, DeepSeek, MiniMax, MiMo, or OpenRouter. The setting is carried through both OpenCode and the Codex Responses bridge; existing providers default to the prior literal `reasoning_effort` behavior.
- Provider-native serialization remains centralized in the shared transport resolver. OpenCode omits unsupported `ultra` instead of approximating it as `max`, and built-in OpenRouter Qwen hybrid behavior is no longer inferred for custom gateways from a user-entered model name.
- Added persistence, renderer, OpenCode, bridge, and stale-catalog regression coverage.

Validation: typecheck passed; full suite passed (488 files, 6806 tests; 10 files/112 tests skipped); lint passed with 0 errors and 24 pre-existing unrelated warnings.

ewen-poch:
Addressed the remaining actionable effort feedback in a74d129.

- Unpinned Claude shared and isolated subscriptions now treat model capability as unknown. The selector shows only Default, and backend resolution omits sessionEffort until a model is explicitly selected or saved. Explicit Claude model overrides still use their Anthropic model profile.
- OpenCode now maps model-native ultra to its highest supported transport rung, max, instead of dropping the override and falling back to the provider default. All other model-native values continue through the provider-specific transport resolver unchanged.
- The reported cross-generation Responses bridge mutation is already resolved by 5ff777c: every backend generation owns an independent bridge and regression coverage verifies overlapping efforts remain isolated.
- No change for the five-slot none mapping or collapsed top-choice intent: those match the approved positional intent model. No-provider settings also retain the intentional five-level preconfiguration UI.

Validation: typecheck passed; full suite passed (488 files, 6810 tests; 10 files/112 tests skipped); lint passed with 0 errors and 24 pre-existing unrelated warnings.

PR #451, opened 2026-07-26
fix(codex): register native Responses model metadata

ewen-poch:
## Problem

Codex 0.144.6 drives providers with native Responses support directly, but it does not know third-party model IDs such as `MiniMax-M3`. It warns that model metadata is missing and falls back to generic metadata even though the upstream request route is correct.

## Proposed change

- Generate an app-owned Codex model catalog for native Responses models that are absent from the pinned Codex bundle.
- Keep the selected vendor model, vendor Responses endpoint, and vendor authentication on the direct path; do not enable the Chat-to-Responses bridge.
- Use content-addressed immutable catalog files so concurrent sessions with different models cannot overwrite each other's metadata.
- Preserve Codex's built-in metadata for bundled model IDs and use conservative capabilities for unknown third-party models.

## Scope and non-goals

- No change to Chat Completions bridge selection or request translation.
- No provider UI or registry schema change.
- No attempt to infer unsupported tools or reasoning features for third-party models.

- Unknown native Responses models receive local metadata without changing their upstream model ID or endpoint.
- Bundled Codex models do not receive an overriding catalog.
- Concurrent unknown models receive distinct immutable catalog paths.
- `npm run typecheck`
- `npm run lint` (passes with 24 pre-existing warnings outside this diff)
- `npm run build`
- Full Vitest suite: 493 files passed, 6891 tests passed; 10 files and 112 tests skipped by configuration.

## Review focus

- Codex 0.144.6 model metadata fields and bundled-model allowlist.
- Direct Responses versus bridge selection remains unchanged.
- Conservative tool/image capability defaults for unknown models.


ewen-poch:
Addressed the correctness findings in `40cd466`:

- Replaced the generic Codex CLI prompt with an Open Science/ACP-specific base instruction set. It now tells the model to use only capabilities advertised by the active session and no longer references `update_plan`, `apply_patch`, CLI approval semantics, or other unavailable harness behavior.
- Propagated the resolved model reasoning profile into `ModelConfigContext` and the generated catalog. For example, MiniMax M3 now advertises its documented `none`/`high` levels, while `model_reasoning_effort` and `default_reasoning_level` remain consistent with the selected mapped value.

I did not enable freeform `apply_patch`, hosted search, or parallel tool calls for unknown native Responses providers. Responses wire compatibility alone does not establish support for those optional OpenAI tool shapes, and advertising an unsupported custom/hosted tool can reject the entire upstream request. The catalog therefore intentionally exposes the function-shaped shell tool only; the new ACP-specific prompt no longer instructs the model to call `apply_patch`. These capabilities can be enabled later when the provider registry can represent and verify them explicitly.

Validation: targeted Codex and MiniMax tests, typecheck, lint, production build, and the full suite (`493` files / `6891` tests passed).

ewen-poch:
Addressed both metadata trust-boundary findings in beb0025:

- Bundled metadata is now reused only when the provider is the official OpenAI identity and the detected native Codex version/model pair is explicitly present in the app-owned compatibility table.
- The settings service now threads the detected native Codex version into model preparation. Older, newer, missing, or otherwise unrecognized native versions receive the conservative local catalog instead of assumptions derived from the app-managed version.
- Custom Responses gateways always receive conservative local metadata, even when they reuse a bundled OpenAI model slug such as gpt-5.4.

Validation: targeted Codex tests (39/39), SettingsService tests (214/214), node/web typecheck, lint (0 errors; 24 pre-existing warnings), Electron/web builds, and the full Vitest suite (493 files and 6893 tests passed; 10 files and 112 tests skipped).

ewen-poch:
Addressed the hosted-search capability mismatch in 9e8e083:

- The conservative native Responses catalog now sets web_search_tool_type to null alongside supports_search_tool: false, so Codex cannot advertise or issue an unverified hosted-search tool to third-party providers.
- Regression assertions cover unknown Codex versions, custom gateways that reuse bundled model slugs, and the MiniMax native Responses path.

Validation: targeted Codex tests (39/39), related-file lint, node/web typecheck, and the full Vitest suite (493 files and 6893 tests passed; 10 files and 112 tests skipped).

ewen-poch:
Fixed the runtime model-catalog parse failure in 9115f9e.

Codex 0.144.6 defines web_search_tool_type as a non-null enum with a schema default, so serializing null caused session creation to fail at config load. The generated catalog now omits that field while retaining supports_search_tool: false, which is the capability gate that keeps hosted search disabled.

I reproduced the failure against the managed Codex 0.144.6 binary with config.load=false, then verified the same catalog with only that field omitted returns config.load=true. Regression coverage now asserts that all conservative native catalogs omit web_search_tool_type; targeted Codex tests pass (39/39).

ewen-poch:
Addressed the runtime metadata lookup failure and the architecture trust-boundary finding in `544c856`:

- `model_catalog_json` is now projected into the native `app-server` startup arguments. `codex-acp` previously forwarded it only in `thread/start`, after Codex had already constructed its process-wide model manager, so the catalog parsed successfully but `MiniMax-M3` still resolved to fallback metadata.
- Generated models now use `visibility: "list"`; once the static catalog is loaded at startup, hidden-only metadata makes `model/list` empty and causes ACP `session/new` to fail.
- Backend resolution live-probes the exact native Codex executable before trusting a bundled version/model pair. Missing or stale persisted version data therefore falls back conservatively instead of controlling the trust decision.

I did not rename `supports_reasoning_summaries`. The pinned Codex 0.144.x `ModelInfo` schema defines that exact field; `supports_reasoning_summary_parameter` is not part of this version. The 0.144.6 parser also accepts the generated catalog with `supports_reasoning_summaries`, so changing it would serialize an ignored/unknown field rather than fix the capability flag.

Validation:
- Managed Codex and framework tests: 84/84 passed.
- Focused SettingsService regressions: 4/4 passed.
- Node typecheck, targeted ESLint, Prettier, and `git diff --check` passed.
- Direct 0.144.6 app-server validation loaded the generated catalog at process startup, returned `MiniMax-M3` from `model/list`, and started a thread with `model: "MiniMax-M3"` without the fallback metadata warning.

ewen-poch:
Addressed the shared-catalog publication race in `5766470`:

- Content-addressed framework files now carry an explicit `contentAddressed` contract.
- They are written completely to a same-directory temporary file and atomically published with `rename`, so readers never observe a truncated destination.
- A byte-identical existing destination is reused without rewriting. Concurrent publishers that lose the rename race accept the winner only after verifying its bytes; a mismatched/corrupt destination is not silently trusted.
- Mutable files such as `config.toml` keep the existing replace behavior.

Regression coverage verifies skip-on-existing without changing permissions, eight concurrent publications producing one complete JSON file with no temporary-file residue, and continued replacement of ordinary mutable config. Focused tests pass (87/87 plus 4 SettingsService regressions), along with node typecheck, targeted ESLint/Prettier, and `git diff --check`.

ewen-poch:
Addressed the native Skill-guidance finding in `5b7b54c`.

Generated catalogs now keep `include_skills_usage_instructions` enabled. This guidance is a Codex/Open Science runtime capability used to discover the materialized `mcp-*` connector Skills; it does not advertise an optional upstream Responses tool and therefore should not be disabled with hosted search, freeform patching, or parallel-call capabilities.

Regression coverage asserts that the direct native-Responses MiniMax catalog preserves the Skill instructions. The full Codex framework test file passes (39/39), along with targeted ESLint, Prettier, and `git diff --check`.

ewen-poch:
Implemented the native Responses follow-up:

- Kept MiniMax and other compatible providers on the Responses wire protocol; the new local compatibility proxy only flattens Codex namespace tools for the upstream request and restores namespace/name on returned calls.
- Reused the runtime Skill-selection lease for native Responses, so prompts such as PubMed requests deterministically load the matching Skill before task execution.
- Kept official OpenAI and Codex subscription paths direct, and preserved the provider model plus generated model metadata instead of selecting the Chat Completions bridge model.
- Added unit coverage and a live contract test through the real Codex adapter/MCP router. Validation passed: 260 focused tests, TypeScript typecheck, targeted ESLint, and the live adapter contract.

ewen-poch:
Addressed both reviewer findings in b3714aa.

The native Responses compatibility proxy now mirrors the existing reviewer-isolation contract: it registers reviewer prompt-cache keys, replaces the complete incoming tool set with only the scope-bounded reviewer MCP declarations, records whether a request actually passed through that scope, and returns true from unregister only after the scoped request was observed. This both prevents built-in/non-reviewer tool access and avoids falsely failing successful reviewer sessions.

A regression test verifies that shell and notebook tools are removed, only the reviewer function reaches the upstream provider, an unobserved session returns false, and an observed scoped session returns true. Validation passed: 261 focused tests, TypeScript typecheck, targeted ESLint, Prettier, and git diff checks.

ewen-poch:
Addressed the latest architecture and correctness findings in 5b4607d.

- Content-addressed catalog publication now retries transient EPERM, EBUSY, EEXIST, and ENOTEMPTY rename races with a bounded backoff. After each delay it accepts a concurrently published file only when its bytes match, while persistent failures still surface after the retry window.
- Native Responses Skill catalog bounding now skips an individual candidate that does not fit instead of truncating the remaining catalog, and includes JSON separator bytes in the budget.

Regression coverage includes delayed matching winners for all four Windows race codes, bounded persistent-failure behavior, and a near-256-KB catalog where a smaller relevant Skill follows an over-budget candidate. Validation passed: 270 focused tests, TypeScript typecheck, targeted ESLint, Prettier, and git diff checks.

ewen-poch:
Fixed in 26f8169. The model-catalog decision now reuses the same exact api.openai.com hostname classifier as the direct-backend decision in settings. Custom Responses providers targeting the official OpenAI API retain Codex bundled metadata for exact known model/version pairs, while custom gateways (including lookalike hostnames) and unknown model/version pairs still receive the conservative generated catalog. Added regression coverage for both the custom official-host flow and hostname boundary cases. Focused Codex/settings tests pass (263 tests), along with ESLint, Node/Web type checking, and diff checks.

ewen-poch:
Fixed in b5a6121. The authenticated native Responses loopback proxy now uses the app-aligned 64 MiB request envelope instead of rejecting at 32 MiB. This keeps the proxy memory-bounded while leaving room for Open Science’s 24 MiB inline-image allowance plus text, replayed history, and tool declarations. Added an end-to-end proxy regression test that forwards a request containing 24 MiB of inline image data and 8 MiB of text; the test reproduced the prior native Responses request is too large response before the fix. Focused compatibility/settings tests pass (223 tests), along with targeted ESLint, Node/Web type checking, formatting, and diff checks.

PR #454, opened 2026-07-26
feat(skills): add conversation import toggle

ewen-poch:
## Problem

Conversation-driven Skill package import currently adds an app-owned MCP tool and guidance to every conversation, even for users who never import Skills this way.

## Proposed change

- Add a default-on Conversation imports toggle under Settings > Skills.
- Persist the preference and apply changes through the existing deferred runtime reconnect.
- When disabled, omit the Skill import stdio/HTTP MCP server, Codex bridge tool schema, system prompt guidance, RPC/alias registration, and actionable attachment metadata.

## Scope and non-goals

Manual Skill creation, upload, GitHub import, installed-Skill import, and per-Skill enablement are unchanged. This adds no MCP service.

- The preference defaults to enabled and survives reloads.
- Turning it off removes conversation Skill import tooling and context while preserving ordinary archive references.
- npm run typecheck passes.
- npm run lint passes with 0 errors; existing unrelated warnings remain.
- npm test passes.

## Review focus

Please focus on runtime gating consistency across native MCP, HTTP MCP, and the Codex Responses bridge, plus safe application after an in-flight turn.


ewen-poch:
Addressed both duplicate P1 reports in f898169.

Skill changes now immediately retire the current runtime generation and publish idle sessions as detached before the settings callback returns. Existing active turns finish on the old generation, while later turns must resume on a freshly provisioned runtime; race-window prompts are prevented from entering a retired runtime.

Validation: ACP runtime coordinator tests (25/25), TypeScript typecheck, targeted ESLint, and independent Standards/Spec re-review with no remaining findings.

ewen-poch:
Followed up on the new Correctness P2 in 1f6f0d5.

The reported idle-session path was already prevented by AcpRuntimeCoordinator generation retirement, but the review correctly exposed a misleading obsolete public AcpRuntime.requestSkillsReload deferred entry point. That dead API and its direct tests are now removed. Every production requestSkillsReload call resolves to AcpRuntimeCoordinator, which synchronously retires the stale generation, detaches idle sessions, rejects race-window prompts owned by it, and resumes later turns on a fresh runtime with the current preference.

Revalidated: coordinator tests 25/25, forced-Skill restore regression, TypeScript typecheck, targeted ESLint, and final independent Standards/Spec review with no findings.

PR #459, opened 2026-07-27
feat(uploads): support large file attachments

### LINKED ISSUE #405: [Feature]: Increase the file size limit for chat attachments
### Preflight checklist

Users may need to upload research materials larger than 50 MB, including large PDFs, datasets, and video files.

The current file size limit prevents these files from being imported directly into a conversation, forcing users to split, compress, or manually preprocess them before use.

### Proposed solution

Increase the maximum file size allowed for chat attachments and support importing files larger than 50 MB.

The upload flow should also provide:

- Clear file type and size limits
- Upload progress for large files
- Chunked or resumable uploads where possible
- A clear error message when a file exceeds the supported limit

Larger limits could be applied separately for documents, PDFs, datasets, and video files based on processing capabilities.

### Alternatives considered

_No response_

### Additional context

_No response_

ewen-poch:
## Problem

Closes #405.

Chat attachments were capped at 50 MB because the renderer read each complete file into memory, encoded it as base64, and sent it as one IPC or Web RPC body. This made the transport—not the model request—the limiting factor, and the Web path could already exceed its 64 MB RPC body cap near the advertised limit.

## Proposed change

- Raise the managed-storage limit to 10 GB per file and show the limit before file selection.
- On desktop, resolve native `File` paths in preload and stream them directly into managed storage without exposing source paths to the renderer.
- For Web, clipboard, and pathless `File` objects, use an 8 MB begin/append/status/finish protocol with offset validation, bounded retry, progress, and cancellation.
- Remove the legacy whole-file base64 staging API from IPC, preload, Web RPC, and the upload repository.
- Stage into exclusive `.part` files, clean abandoned partials, and commit with hard-link/move semantics plus an exclusive-copy fallback.
- Show per-file queued/uploading/error/cancelling state and progress in Composer.
- Keep model payloads bounded: small text may be embedded; large text gets a bounded preview plus `resource_link`; dataset binaries and video are link-only; images and PDFs above the automatic-processing threshold are link-only.
- Route Composer `@` references through a source adapter resolver. Existing upload/artifact mentions remain supported, while the shared Composer, persistence, and runtime contracts reserve `linked-folder` references as `rootId + relativePath` without accepting renderer-provided absolute paths.

```mermaid
flowchart LR
  A[Desktop File with native path] -->|path metadata only| B[Main process stream copy]
  C[Web, clipboard, or pathless File] -->|8 MB offset chunks| D[Main process chunk writer]
  B --> E[Managed upload file]
  D --> E
  E --> F[Bounded preview or resource link]
  F --> G[Agent]
```

## Scope and non-goals

- Chunk transfers can recover from a failed append by querying the current offset during the live app session; persistence across an app restart is not included.
- Linked-folder selection, permission UI, and provider implementation are not included. This change only reserves and validates the cross-layer reference shape and resolver seam.
- Large files are never embedded wholesale in model requests. The 10 GB value is a storage/transfer cap, not an inline model-content limit.

- Desktop files can bypass renderer memory and base64 transport.
- Web/pathless files use bounded chunks below the RPC body cap.
- The UI reports progress, supports cancellation, and states that any file type is accepted up to 10 GB per file.
- Current upload/artifact `@` mentions still round-trip; linked-folder references round-trip only as `rootId + relativePath` and remain unavailable until a capability-validating adapter is configured.
- `npm run check:web-api-map` — passed.
- `npm run lint` — passed with 0 errors (24 existing warnings outside this change).
- `npm test` — 498 test files passed, 10 skipped; 7005 tests passed, 113 skipped.

## Review focus

- Native-path trust boundary and the absence of a whole-file base64 public API.
- Chunk offset/retry/cancellation behavior and atomic `.part` cleanup/finalization.
- Per-type model-content fallbacks for large data, image, PDF, and binary files.
- `@` reference resolution and the future linked-folder capability boundary.

### DIFF
--- a/src/main/acp/attachment-content.test.ts
+++ b/src/main/acp/attachment-content.test.ts
@@ -2,9 +2,11 @@ import { describe, expect, it } from 'vitest'
 
 import {
   MAX_EMBEDDED_TEXT_UPLOAD_BYTES,
+  buildDatasetAttachmentNotice,
   buildOversizedAttachmentNotice,
   formatBytes,
   imageAttachmentMimeType,
+  isDatasetAttachment,
   isTabularAttachment,
   isTextLikeAttachment
 } from './attachment-content'
@@ -150,6 +152,26 @@ describe('buildOversizedAttachmentNotice', () => {
 
     expect(notice).toContain('line ranges or sections')
     expect(notice).not.toContain('file continues')
+  })
+})
+
+describe('binary dataset attachments', () => {
+  it('recognizes spreadsheet and columnar scientific data formats', () => {
+    expect(isDatasetAttachment('study.xlsx')).toBe(true)
+    expect(isDatasetAttachment('events.parquet', 'application/octet-stream')).toBe(true)
+    expect(isDatasetAttachment('matrix.h5')).toBe(true)
+    expect(isDatasetAttachment('movie.mp4')).toBe(false)
+  })
+
+  it('steers the agent toward schema inspection and notebook computation', () => {
+    const notice = buildDatasetAttachmentNotice({ name: 'study.xlsx', size: 3_000_000_000 })
+
+    expect(notice).toContain('study.xlsx')
+    expect(notice).toContain('available on disk')
+    expect(notice).toContain('schema')
+    expect(notice).toContain('sample')
+    expect(notice).toContain('notebook')
+    expect(notice).toContain('Do not load the whole file')
   })
 })
 
--- a/src/main/acp/attachment-content.test.ts
+++ b/src/main/acp/attachment-content.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it } from 'vitest'
 import {
   MAX_EMBEDDED_TEXT_UPLOAD_BYTES,
   buildDatasetAttachmentNotice,
+  buildDeferredMediaNotice,
   buildOversizedAttachmentNotice,
   formatBytes,
   imageAttachmentMimeType,
@@ -172,6 +173,20 @@ describe('binary dataset attachments', () => {
     expect(notice).toContain('sample')
     expect(notice).toContain('notebook')
     expect(notice).toContain('Do not load the whole file')
+  })
+})
+
+describe('buildDeferredMediaNotice', () => {
+  it.each([
+    ['image', 'microscopy.png'],
+    ['PDF', 'paper.pdf']
+  ] as const)('describes an oversized %s as an on-disk linked resource', (kind, name) => {
+    const notice = buildDeferredMediaNotice({ name, size: 3 * 1024 * 1024 * 1024, kind })
+
+    expect(notice).toContain(name)
+    expect(notice).toContain('3.0 GB')
+    expect(notice).toContain('too large for automatic in-memory processing')
+    expect(notice).toContain('available on disk via the linked resource below')
   })
 })
 
--- a/src/main/acp/attachment-content.ts
+++ b/src/main/acp/attachment-content.ts
@@ -73,6 +73,35 @@ const TEXT_LIKE_EXTENSIONS = new Set([
 // Column-oriented formats, so the reading hint can name rows/columns rather than generic ranges.
 const TABULAR_EXTENSIONS = new Set(['csv', 'tsv', 'tab', 'psv'])
 
+// Structured data formats that are best inspected with a dataframe/database tool. Text members of
+// this set still take the bounded-preview path; binary members use an on-disk notice plus link.
+const DATASET_EXTENSIONS = new Set([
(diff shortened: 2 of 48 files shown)


ewen-poch:
Addressed both blocking review points in `92d0e0e`:

- Chunk transfers now hold one idempotent data-root writer lease from `begin-transfer` through `finish-transfer` or `abort-transfer`. Append calls remain valid while migration is pending, migration drain waits across chunk gaps, and abort waits for an in-flight begin before cleaning up and releasing the lease.
- Successful session deletion now marks every abandoned transfer ID as cancelled before aborting the active controller. The serialized queue skips later files instead of starting them after the draft is removed, and cancellation markers are cleared when each transfer settles.

Validation: `npm run typecheck` and focused migration/upload/draft regression tests (22 passed).

ewen-poch:
Addressed the renderer-teardown lease leak in `9e2fdd1`:

- Every chunk transfer is now owned by the `webContents`/Web RPC sender that began it. The owner registry listens for `destroyed`, `render-process-gone`, and main-frame navigation, then aborts all non-settling transfers for that renderer.
- Teardown cleanup waits for `beginTransfer` and every in-flight append to settle before removing the partial file and releasing the data-root writer lease, so migration cannot race an outstanding disk write.
- Subsequent status/append/finish/abort calls validate transfer ownership, preventing a recreated renderer from adopting a stale transfer with a reused ID.

Validation: `npm run typecheck`, focused upload IPC tests (7 passed), and ESLint on the changed files.

ewen-poch:
The latest AI review output was suppressed by GitHub output filtering, but its two actionable findings were recovered from the job log and confirmed against the current code:

- Re-check the abort signal after finishTransfer resolves. If cancellation won the race, delete the returned managed upload and surface AbortError so the attachment is not committed or orphaned.
- Preserve part.mimeType when converting sent artifact/upload mentions into preview items, with coverage for extensionless MIME-driven previews.

No code changes have been pushed for these items yet.

ewen-poch:
Addressed the actionable review findings in 57a6c42:

- Re-check the abort signal after finishTransfer resolves. If cancellation won the race, the finalized managed upload is deleted and AbortError is surfaced so the attachment cannot be committed.
- Preserve mention MIME metadata in sent-message previews and include MIME in Composer node equality so MIME-only external updates refresh the chip.
- Added an exact Uint8Array Web RPC round-trip test. The existing browser and server codecs use the explicit binary/base64 envelope, and the test confirms all byte values arrive intact and are encoded intact on the response.
- Added direct oversized image/PDF notice coverage and upload-size formatter boundary coverage.

Validation: 57 focused tests passed across 6 files, npm run typecheck passed, and targeted ESLint reported no issues.

The unified 10 GiB storage cap, restartable desktop local-path copy, linked-folder seam, and dataset guidance remain intentional tradeoffs: per-category limits are optional in issue #405, automatic processing still has type-specific memory caps, Web/no-path uploads are chunked and resumable, and the reference/dataset behavior was explicitly requested.

ewen-poch:
Addressed the active-session deletion cleanup race in db6b475:

- Session deletion now owns a mutable per-session cleanup ledger. Existing attachments/transfers are captured at confirmation, while transfers queued or attachments finalized during the async deletion window are added to the same ledger.
- The upload caller re-checks its AbortController after stageComposerFile returns. If deletion settled between the inner staging promise and the outer continuation, the staged attachment is deleted instead of being committed.
- Duplicate deletion requests for the same session are ignored while the first request is pending, preventing cleanup-state replacement.
- Failed deletion discards only the temporary ledger and preserves the normal draft and staged files.

Validation: the draft-preservation suite passes 11/11, npm run typecheck passes, targeted ESLint reports no issues, and the full suite passes 499 files / 7021 tests with 10 files / 113 tests skipped by existing configuration.

ewen-poch:
Addressed the native-path renderer lifecycle finding in `100dda1`:

- Native transfers are now owned by the invoking renderer and are cancelled on `destroyed`, `render-process-gone`, or main-frame navigation. Cross-renderer cancellation/reuse remains rejected.
- Ownership and the data-root writer lease now span the full logical operation: staging through explicit renderer claim, or staging through teardown cleanup. A migration cannot copy an unclaimed pending file while cleanup races on the old root.
- The renderer claims only after committing the attachment into draft state. If teardown wins before claim, main deletes the completed `.pending` upload; cancellation after staging also aborts and deletes idempotently.
- Local source streams receive an `AbortSignal`; abort actively interrupts a stalled stream and waits for staging-file cleanup before releasing the migration lease.

Validation: 76 focused tests passed across upload IPC/repository/preload/renderer/Web-map suites, `npm run typecheck` passed, targeted ESLint passed, Web API map validation passed, and the loopback HTTP compatibility suite passed 6/6.

PR #460, opened 2026-07-27
fix(model-settings): validate multi-protocol providers reliably

ewen-poch:
## Problem

Kimi For Coding could pass validation for Codex, then fail after switching back to Claude Code with HTTP 200 and `The provider returned success without a valid Anthropic message.` The one-token Anthropic probe caused Kimi k3 to return an incomplete message envelope even though normal requests were valid.

The onboarding form could also report Kimi For Coding as incompatible with Codex after a framework switch because compatibility checks used the form's stale single endpoint instead of the official provider's full protocol set.

## Proposed change

- Raise the Anthropic validation probe budget from 1 to 16 tokens while retaining strict message-envelope validation.
- Resolve official provider API endpoints from the registry for compatibility checks and persisted provider requests.
- Add regressions for the Kimi Anthropic envelope and Codex bridge-compatible onboarding path.

## Scope and non-goals

- This is a protocol-level fix, not a Kimi-specific validation bypass.
- Custom gateways still use the endpoint explicitly selected in the form.
- No provider credentials or persisted settings are changed.
- DeepSeek was live-checked locally and did not reproduce the one-token envelope issue. Other Anthropic-compatible providers were audited through the shared validation path but were not live-tested without credentials.

- [x] Kimi receives enough Anthropic output budget to return a complete message envelope.
- [x] Empty or malformed HTTP 200 Anthropic responses are still rejected.
- [x] Kimi For Coding is recognized as Codex-compatible through its OpenAI protocol support after framework switches.
- [x] Official multi-protocol providers use registry endpoints; custom gateways retain their selected endpoint.
- [x] `npm run typecheck`
- [x] `npm run lint` (0 errors; repository baseline warnings only)
- [x] `npm run test` (496 files and 6,989 tests passed; 10 files skipped)

## Review focus

- Whether 16 tokens is an appropriate low-cost Anthropic validation budget.
- Whether registry endpoints are the correct source of truth for official provider form drafts.

### DIFF
--- a/src/main/settings/service.test.ts
+++ b/src/main/settings/service.test.ts
@@ -95,6 +95,12 @@ const validBridgeToolCallResponse = (): Response =>
     { status: 200, headers: { 'content-type': 'text/event-stream' } }
   )
 
+const validNativeCompatibilityToolCallResponse = (): Response =>
+  new Response(
+    'data: {"type":"response.output_item.added","item":{"type":"function_call","id":"fc_1","call_id":"call_1","name":"open_science__bridge_probe","arguments":"{}"}}\n\ndata: {"type":"response.completed","response":{"output":[{"type":"function_call","id":"fc_1","call_id":"call_1","name":"open_science__bridge_probe","arguments":"{}"}]}}\n\n',
+    { status: 200, headers: { 'content-type': 'text/event-stream' } }
+  )
+
 type ManagedInstallImpl = (options: {
   installId: string
   onEvent: (event: { kind: string; installId: string }) => void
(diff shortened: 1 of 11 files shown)


ewen-poch:
Addressed the P1 finding in ff2cd7d. Codex validation now runs the real streaming function-tool contract probe whenever the selected provider is reached through the Chat Completions-to-Responses bridge. The bridge predicate is shared by compatibility, validation, and runtime setup so those paths cannot diverge. Native Responses providers remain on their direct validation path, and regressions cover both custom and official bridge-backed providers.

ewen-poch:
Addressed the native Responses P1 in f65457f. Codex validation now runs through the same namespace-tool compatibility proxy used at runtime for non-OpenAI Responses providers. The probe sends a namespace tool, verifies the proxy flattens it for the upstream provider and restores the namespace on the returned function call, and rejects text-only success responses. The runtime and validation paths share the same provider predicate; xAI and negative text-only regressions cover both outcomes.

ewen-poch:
Addressed the malformed native Responses base URL finding in 821aca4. Validation now parses the normalized target before starting the compatibility proxy and returns the existing bad-url result for invalid input, preventing the proxy from converting the error into a local 400/unknown response. Added a regression test that exercises the native compatibility path and confirms no upstream fetch is attempted.

PR #462, opened 2026-07-27
fix(codex): preserve imported subscription routing

ewen-poch:
## Problem

Importing an existing Codex sign-in copied `auth.json` but discarded the active model-provider route from the local Codex profile. On machines where local Codex uses an authenticated loopback route, Open Science instead fell back to the direct ChatGPT transport. A two-character prompt took about 114 seconds in the app; the direct-provider CLI repro took 117.99 seconds after five WebSocket retries, while the working local route completed in seconds.

The native Codex child also does not use Electron's proxy-aware network stack. A valid app-owned login could therefore succeed yet prompt or status traffic could stall when the user's current system/PAC proxy was required.

## Proposed change

- Import the active model provider only when it uses the Responses wire API, Codex/OpenAI authentication, and a strict loopback destination (`localhost`, `127.0.0.0/8`, or `::1`).
- Re-serialize an allowlisted, non-secret route containing only provider id, name, base URL, wire API, authentication mode, and optional WebSocket support. Reject routes that depend on omitted credential-bearing fields.
- Store the sanitized route with mode `0600` beside the imported authentication so app-owned Codex sessions reuse the same working local network path.
- Persist the selected authentication mode so ordinary edits retain imported routing, expose explicit re-import, and distinguish a switch to in-app sign-in.
- Mark imported route fragments and remove only those fragments when switching modes or deleting the provider, preserving unrelated app-owned config.
- Serialize authentication-mode transitions: an isolated sign-in is cancelled and its session is fully closed before app-owned credentials are imported, cleared, or deleted.
- Pin the app-owned credential store to file mode and implement logout/delete as local app-data cleanup, without remotely revoking or changing the user's Codex CLI login.
- Resolve the user's current system/PAC proxy at runtime for Codex subscription auth and prompt children. Resolved proxy and explicit `DIRECT` decisions replace inherited proxy variables; resolver failure preserves the inherited environment, and loopback addresses remain bypassed.
- Update Provider copy and actions to reflect imported versus in-app authentication accurately.

## Scope and non-goals

- Does not import model defaults, MCP servers, hooks, headers, bearer tokens, Skills, sessions, or other global Codex configuration.
- Rejects remote provider routes, separate-credential routes, and URLs with embedded credentials, query parameters, or fragments.
- Does not store proxy endpoints in `settings.json` or `config.toml`, and does not modify the user's system proxy or global Codex CLI profile.
- System-proxy resolution applies to the shared app-owned Codex subscription runtime used by both imported and in-app authentication. It does not change custom/official API-provider routing.
- Logout and provider deletion affect only the app-owned Codex subscription home; they never invoke remote account logout.

- Imported subscription sessions retain a compatible loopback route without exposing unrelated configuration or forwarding ChatGPT credentials to remote hosts.
- Stale imported route fragments are removed when the source no longer has a safe route, the user switches to in-app sign-in, or the provider is deleted, without deleting unrelated app-owned config or global credentials.
- Authentication-mode changes cannot be overwritten by a late write from an isolated browser-login session.
- Empty query/fragment delimiters are rejected and legal quoted TOML provider-table keys are recognized.
- Runtime proxy routing follows the current system/PAC decision without persisting a proxy endpoint; loopback imported routes remain direct.
- `npm run typecheck`
- Targeted ESLint and Prettier for all changed files
- Final focused validation: 266 tests passed across Codex authentication and settings service behavior
- Earlier full-suite validation: 496 files passed, 6,988 tests passed, 10 files / 113 tests skipped
- Live isolated-config repro: `gpt-5.6-terra` completed in 12.97 seconds through the imported route versus 117.99 seconds through the direct provider.

## Review focus

Please focus on authentication-transition serialization, the TOML allowlist/parser boundary, strict loopback validation, runtime proxy inheritance/replacement semantics, and app-owned credential cleanup.

### DIFF
--- a/src/main/acp/runtime.ts
+++ b/src/main/acp/runtime.ts
@@ -2118,7 +2118,8 @@ class AcpRuntime {
       process = backend.framework.spawn({
         executablePath: backend.executablePath,
         env: backend.env,
-        args: backend.args ?? []
+        args: backend.args ?? [],
+        proxyEnvironmentMode: backend.proxyEnvironmentMode
       })
     } catch (error) {
       // Wrap (never mutate) the failure with the framework this spawn targeted: the connect-level catch
--- a/src/main/agent-framework/codex.test.ts
+++ b/src/main/agent-framework/codex.test.ts
@@ -709,6 +709,7 @@ describe('codexFramework', () => {
 
     framework.spawn({
       executablePath: '/usr/local/bin/codex-acp',
+      proxyEnvironmentMode: 'replace',
       env: {
         CODEX_HOME: '/data/codex-subscription',
         HTTP_PROXY: 'http://system-proxy.example.test:3128',
@@ -729,6 +730,7 @@ describe('codexFramework', () => {
 
     framework.spawn({
       executablePath: '/usr/local/bin/codex-acp',
+      proxyEnvironmentMode: 'replace',
       env: { CODEX_HOME: '/data/codex-subscription' },
       args: []
     })
@@ -737,6 +739,30 @@ describe('codexFramework', () => {
     expect(directEnv.HTTPS_PROXY).toBeUndefined()
     expect(directEnv.ALL_PROXY).toBeUndefined()
     expect(directEnv.NO_PROXY).toBeUndefined()
+  })
+
+  it('preserves inherited proxies when subscription proxy resolution fails', () => {
+    const spawnProcess = vi.fn().mockReturnValue(fakeChild)
+    const framework = createCodexFramework({
+      sourceEnv: {
+        PATH: '/parent-bin',
+        HTTPS_PROXY: 'http://inherited-proxy.example.test:3128',
+        NO_PROXY: 'inherited-bypass.example.test'
+      },
+      spawnProcess
+    })
+
+    framework.spawn({
+      executablePath: '/usr/local/bin/codex-acp',
+      proxyEnvironmentMode: 'inherit',
+      env: { CODEX_HOME: '/data/codex-subscription' },
+      args: []
+    })
+
+    expect(spawnProcess.mock.calls[0][2].env).toMatchObject({
+      HTTPS_PROXY: 'http://inherited-proxy.example.test:3128',
+      NO_PROXY: 'inherited-bypass.example.test'
+    })
   })
 })
 
--- a/src/main/agent-framework/codex.test.ts
+++ b/src/main/agent-framework/codex.test.ts
@@ -694,6 +694,50 @@ describe('codexFramework', () => {
     expect(env.OPENAI_API_KEY).toBeUndefined()
     expect(env.CODEX_PATH).toBeUndefined()
   })
+
+  it('replaces every inherited proxy shape for a subscription spawn', () => {
+    const spawnProcess = vi.fn().mockReturnValue(fakeChild)
+    const framework = createCodexFramework({
+      sourceEnv: {
+        PATH: '/parent-bin',
+        ALL_PROXY: 'socks5://stale-proxy.example.test:9050',
+        HTTPS_PROXY: 'http://stale-proxy.example.test:3128',
+        NO_PROXY: 'stale-bypass.example.test'
+      },
+      spawnProcess
+    })
+
+    framework.spawn({
+      executablePath: '/usr/local/bin/codex-acp',
+      env: {
+        CODEX_HOME: '/data/codex-subscription',
+        HTTP_PROXY: 'http://system-proxy.example.test:3128',
+        HTTPS_PROXY: 'http://system-proxy.example.test:3128',
+        NO_PROXY: 'localhost,127.0.0.1,::1'
+      },
+      args: []
+    })
+
+    const env = spawnProcess.mock.calls[0][2].env as NodeJS.ProcessEnv
+    expect(env).toMatchObject({
+      HTTP_PROXY: 'http://system-proxy.example.test:3128',
+      HTTPS_PROXY: 'http://system-proxy.example.test:3128',
+      NO_PROXY: 'localhost,127.0.0.1,::1'
+    })
+    expect(env.ALL_PROXY).toBeUndefined()
+    expect(env.NO_PROXY).not.toContain('stale-bypass.example.test')
+
+    framework.spawn({
+      executablePath: '/usr/local/bin/codex-acp',
+      env: { CODEX_HOME: '/data/codex-subscription' },
+      args: []
+    })
+    const directEnv = spawnProcess.mock.calls[1][2].env as NodeJS.ProcessEnv
+    expect(directEnv.HTTP_PROXY).toBeUndefined()
+    expect(directEnv.HTTPS_PROXY).toBeUndefined()
+    expect(directEnv.ALL_PROXY).toBeUndefined()
+    expect(directEnv.NO_PROXY).toBeUndefined()
+  })
 })
 
 describe('buildCodexConfig reasoning effort', () => {
--- a/src/main/agent-framework/codex.ts
+++ b/src/main/agent-framework/codex.ts
@@ -27,6 +27,7 @@ import type {
 } from './types'
 import { isCodexSubscriptionProvider } from '../../shared/settings'
 import { CODEX_VERSION } from '../settings/managed-codex'
+import { clearSystemProxyEnvironment } from '../settings/system-proxy'
 import codexNativeModelInstructions from './codex-native-model-instructions.md?raw'
 
 const CODEX_PROVIDER_ID = 'open-science'
@@ -275,6 +276,12 @@ const buildSpawnEnvironment = (
   const env = augmentedPathEnv(sourceEnv)
 
   for (const key of CODEX_ENV_KEYS) delete env[key]
+  // A subscription gets an authoritative proxy decision from Electron immediately before backend
+  // resolution. Drop every inherited proxy shape first so a stale ALL_PROXY/HTTP_PROXY cannot win
+  // over that decision (including DIRECT, represented by no proxy keys in input.env).
+  if (/[\\/]codex-subscription$/.test(input.env.CODEX_HOME ?? '')) {
+    clearSystemProxyEnvironment(env)
+  }
 
   return {
     ...env,
--- a/src/main/agent-framework/codex.ts
+++ b/src/main/agent-framework/codex.ts
@@ -276,10 +276,9 @@ const buildSpawnEnvironment = (
   const env = augmentedPathEnv(sourceEnv)
 
   for (const key of CODEX_ENV_KEYS) delete env[key]
-  // A subscription gets an authoritative proxy decision from Electron immediately before backend
-  // resolution. Drop every inherited proxy shape first so a stale ALL_PROXY/HTTP_PROXY cannot win
-  // over that decision (including DIRECT, represented by no proxy keys in input.env).
-  if (/[\\/]codex-subscription$/.test(input.env.CODEX_HOME ?? '')) {
+  // A resolved proxy or DIRECT decision is authoritative. A resolver failure uses `inherit` so a
+  // working proxy supplied by the process launcher remains available as the fallback.
+  if (input.proxyEnvironmentMode === 'replace') {
     clearSystemProxyEnvironment(env)
   }
 
--- a/src/main/agent-framework/types.ts
+++ b/src/main/agent-framework/types.ts
@@ -98,12 +98,17 @@ export type SessionSetup = {
   promptPrefix?: string
 }
 
+export type ProxyEnvironmentMode = 'inherit' | 'replace'
+
 // Already-resolved spawn inputs: env and args come from prepareModelConfig merged over the base
 // process env; configFiles are written by the runtime before this call.
 export type AgentSpawnInput = {
   executablePath: string
   env: Record<string, string>
   args: string[]
+  // `replace` means the host resolved an explicit proxy or DIRECT decision and inherited proxy
+  // variables must not override it. `inherit` preserves them after a host resolver failure.
+  proxyEnvironmentMode?: ProxyEnvironmentMode
   debug?: boolean
 }
 
@@ -158,6 +163,7 @@ export type ResolvedAgentBackend = {
   executablePath: string
   env: Record<string, string>
   args?: string[]
+  proxyEnvironmentMode?: ProxyEnvironmentMode
   // Framework-native session options retained by the runtime and passed through buildSessionSetup.
   sessionOptions?: Record<string, unknown>
   // Backend-resolved guidance appended to every session. Connector conventions use this channel for
--- a/src/main/settings/codex-auth.test.ts
+++ b/src/main/settings/codex-auth.test.ts
@@ -143,6 +143,19 @@ describe('createCodexAuthEnvironment', () => {
     expect(env.ALL_PROXY).toBeUndefined()
     expect(env.NO_PROXY).not.toContain('stale-bypass.example.test')
   })
+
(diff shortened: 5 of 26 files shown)


ewen-poch:
Addressed both Codex review findings in `7a6b6b8`:

- **P1 — credential routing:** imported ChatGPT-authenticated Responses routes are now accepted only for strict loopback destinations (`localhost`, `127.0.0.0/8`, or `::1`). Remote HTTP and HTTPS routes are rejected. Regression coverage includes both remote protocols.
- **P2 — isolated sign-in cleanup:** saving the in-app Codex sign-in option now removes any previously imported route from the shared app-owned home, while leaving authentication data available for the explicit sign-in flow. A service-level regression test covers the shared-import-to-isolated transition.

Validation: typecheck passed; targeted ESLint passed; Codex authentication tests passed (23/23); route import/cleanup tests passed (2/2); Provider form rendering tests passed (19/19).

ewen-poch:
Addressed the follow-up finding in `d5499e1`.

- The normalized runtime provider still remains `codex-isolated`, but it now persists a separate non-secret authentication-mode discriminator (`imported` or `isolated`).
- Settings rehydrates an imported provider as the import form mode, so an ordinary edit/save reimports and preserves its loopback route instead of deleting it.
- Only an explicit switch to the in-app sign-in mode removes the imported route and updates the persisted discriminator.
- Repository sanitization retains only the two known discriminator values.

Validation: typecheck passed; targeted ESLint passed; route import/resave/switch tests passed (2/2); provider form-mode tests passed (13/13).

ewen-poch:
Addressed the config-preservation finding in `0e7a0a6`.

- Imported route selection and provider-table fragments now carry separate Open Science ownership markers. Import updates only those marked fragments, and cleanup removes only those fragments while preserving unrelated app-owned `config.toml` settings.
- The exact route-only format written by the earlier implementation is recognized as a legacy owned file and removed safely; mixed or user-authored config is never treated as wholly owned.

The validation-state finding is already covered by the current mode transition logic: `credentialsChanged` compares the persisted `codexAuthMode`, so `imported` → `isolated` drops `lastValidatedAt` and `lastValidationFailure`. The regression test now seeds `lastValidatedAt`, switches modes, and asserts that it is cleared; no redundant state-reset branch was added.

The branch was also rebased onto the latest `main`, preserving its multi-protocol provider-form behavior. Validation: typecheck passed, targeted ESLint passed, 98 relevant tests passed, and `git diff --check` passed.

ewen-poch:
Addressed both findings in `52108c2`.

- **Legacy migration:** collapsing a legacy `codex-shared` record now persists `codexAuthMode: "imported"`; legacy isolated records persist `"isolated"`. Migration tests cover both active-mode cases while retaining the normalized app-owned runtime type.
- **Settings list behavior:** effective Codex subscription type resolution now lives in the shared settings module and is used by both the edit form and Provider list. A normalized imported provider displays imported-authentication copy and no longer exposes isolated-only Sign in, Cancel sign-in, or Sign out actions.

Validation: 162 relevant tests passed, typecheck passed, targeted ESLint passed, and `git diff --check` passed. The preceding head also completed all three platform CI jobs successfully before this push.

ewen-poch:
Addressed in `a9c9bcc`.

- Route serialization now removes any conflicting top-level `model_provider` assignment and the matching `model_providers.<id>` table subtree before inserting the app-owned marked route, so the active TOML contains only one selection key and one provider definition.
- Displaced app-owned lines are retained as inert encoded comments inside the ownership markers and restored when imported mode is cleared. Unrelated provider tables, MCP configuration, and model settings remain untouched.
- The regression test covers a quoted top-level key, a same-ID provider table, unrelated app-owned config, and reversible cleanup.

Validation: 24 targeted tests passed, typecheck passed, targeted ESLint passed, and `git diff --check` passed.

ewen-poch:
Addressed both findings in `11341f8`.

- **Pending validation after an auth-mode switch:** resolved Codex subscription targets now carry `codexAuthMode`, and `sameValidationTarget()` includes it in the mid-flight comparison. A deferred-status regression test switches imported authentication to isolated sign-in before the status result returns and verifies the stale success reports `applied: false` without restoring validation state.
- **Quoted TOML keys:** the safe scalar parser now normalizes bare, basic-quoted, and literal-quoted keys before matching imported route fields. The route-import regression now uses quoted keys for `model_provider`, `name`, `base_url`, `wire_api`, `requires_openai_auth`, and `supports_websockets`.

Validation: 24 Codex auth tests passed, the targeted validation-race test passed, typecheck passed, targeted ESLint passed, and `git diff --check` passed.

ewen-poch:
Addressed both findings in `2ab6972`.

- **Imported-provider edits:** `upsertProvider()` now resolves the stored subscription mode before touching authentication. It imports from the external CLI profile only for a new import or an explicit isolated-to-imported switch; an existing imported provider is edited entirely from its app-owned copy. Ordinary edits also preserve the existing validation state because the credential source did not change. The regression removes the external CLI directory before a no-op edit and verifies the imported route and validation timestamp remain intact.
- **Pending login cancellation:** the Codex subscription card now shows `Cancel sign-in` whenever an isolated login is pending, regardless of whether the card currently displays imported or isolated mode. The imported-card render regression exercises and invokes that action.

Validation: the targeted imported-edit regression passed, all 24 ProviderList render tests passed, typecheck passed, targeted ESLint passed, and `git diff --check` passed.

ewen-poch:
Addressed in `85d2644`.

`loginIsolatedCodex()` now applies a completed browser-login result only when the stored subscription is still both the normalized `codex-isolated` runtime type and `codexAuthMode: "isolated"`. A switch to imported authentication therefore makes the old result stale and returns `applied: false`.

The regression starts an isolated login, switches to imported authentication, verifies that the imported mode was persisted, stamps the imported provider as validated, then completes the old login with a timeout. Before the fix it reproduced `applied: true`; after the fix it preserves the imported validation timestamp and leaves `lastValidationFailure` unset.

Validation: 4 adjacent auth-mode/login regressions passed, typecheck passed, targeted ESLint passed, and `git diff --check` passed.

ewen-poch:
Addressed both findings in a5c23d3:

- The isolated-login IPC now ignores successful results that were not applied and reconnects only when the active subscription is still in isolated auth mode. The regression test covers the normalized imported-provider shape after a mid-flow mode switch.
- Subscription validation now resolves the effective auth mode through the shared subscription-type helper. Imported auth checks the shared status path and reports imported-login guidance, while in-app auth continues to use the isolated path.

Verification: all 44 settings IPC tests, all 222 settings service tests, typecheck, targeted ESLint, and git diff --check pass.

ewen-poch:
Addressed in 977866a with an explicit refresh workflow.

- Ordinary edits of an imported subscription remain self-contained and do not silently re-read a changed or missing external CLI profile.
- Imported subscription cards now expose a Re-import Codex login action. That action explicitly refreshes both auth.json and the safe loopback route from the current CLI profile, then validates the refreshed app-owned copy through the existing save flow.
- An explicit refresh is treated as a credential change, so the previous verified/failure state is invalidated instead of being carried across potentially changed credentials or routing.

The regression updates the source auth and loopback route after the initial import, invokes only the explicit refresh intent, and verifies that the app-owned copy changes and the stale validation timestamp is cleared. Validation: 5 adjacent service regressions and all 24 ProviderList render tests passed; typecheck, targeted ESLint, and git diff --check also passed.

ewen-poch:
Addressed in 2c3ff92.

The explicit re-import path now advances the same per-provider validation generation used when a validation starts. Because re-import can replace auth.json and the loopback route without changing any persisted provider field, this generation bump makes every status check started against the previous app-owned copy fail the existing stale-result guard and return applied: false.

The regression starts a deferred imported-mode validation, refreshes the imported authentication, then completes the old status check successfully. It reproduced applied: true before the fix; it now returns applied: false and leaves the refreshed provider without a stale validation timestamp.

Validation: the exact race regression and three adjacent auth-mode/import validation regressions passed; typecheck, targeted ESLint, and git diff --check also passed.

ewen-poch:
Verified the reported import-delete-create lifecycle directly. The existing condition existing?.codexAuthMode !== "isolated" evaluates to true when no provider record exists, so creating an isolated setup after deletion already invokes clearImportedCodexProviderRoute. I added a regression test that imports a loopback route, deletes the provider, recreates it in isolated mode, and confirms both that the repository was empty after deletion and that the imported config file is removed. The focused route tests and ESLint check pass, so no production behavior change is needed for this point.

ewen-poch:
Addressed in 4816203. Releases before codexAuthMode already normalized both setup choices to the unified subscription id, so an unmarked record in that shape cannot recover its original intent. The migration now defaults that ambiguous unified-id shape to imported mode, which keeps the explicit re-import workflow available to legacy shared users. Records that still carry the old dedicated isolated id continue to migrate as isolated, and any explicit codexAuthMode remains authoritative. This intentionally favors shared-profile compatibility for the small ambiguous legacy population; users can still explicitly switch to isolated mode. The new regression failed with isolated before the change and now passes with imported. All 74 repository tests, the related service migration test, targeted ESLint, and git diff checks pass.

ewen-poch:
Addressed the isolated-login boundary in 8665587. The app-owned Codex home could contain auth.json without an explicit credential-store setting, while Codex also supports an OS keyring store. More importantly, switching an imported profile to isolated mode retained the copied auth.json, and the Sign out action called the adapter account/logout operation; a copied token could therefore revoke the session still used by the global CLI. Subscription homes now explicitly pin cli_auth_credentials_store to file without overwriting imported route configuration, and backend resolution enforces that invariant before a direct prompt spawn. Imported-to-isolated transitions delete the app-owned credential copy without invoking account/logout, forcing a separate Open Science sign-in. Isolated Sign out likewise cancels any pending login and removes only the app-owned auth.json before reconnecting the active agent, so it cannot log out an external CLI profile. Validation: all 25 Codex auth tests, all 225 settings service tests, all 44 settings IPC tests, all 47 Codex framework tests, targeted ESLint, and git diff checks pass.

ewen-poch:
Addressed P1 in 1f972ba. The ambiguous normalized legacy shape now defaults to isolated, matching the pre-existing runtime behavior and the actual upgrade shape. Only an explicit legacy codex-shared record maps to imported. Legacy shared users can explicitly re-import once.

For P2, the one-time sign-in requirement for a keyring-only app-owned profile is intentional. There is no safe way to migrate an OS-keyring credential into the app-owned auth.json without reading or mutating the external/global credential that this fix is meant to isolate. Continuing to honor the shared keyring would preserve the reported cross-profile logout risk. Existing file-backed app credentials remain available; keyring-only users must sign in once to create an independent app-owned token. This is the safer trade-off for the small legacy population.

Validation: 74 repository tests, relevant service migration/logout tests, targeted ESLint, and diff check pass. The prior credential-isolation commit also has 25 auth, 225 service, 44 IPC, and 47 framework tests passing.

ewen-poch:
Addressed the post-login prompt stall in b2e8e06. The app-owned provider record, auth.json, and file-backed credential configuration were already persisted correctly; the failing boundary was the native Codex child, which does not use Electron's proxy-aware network stack. Codex subscription auth and prompt sessions now resolve the user's current system/PAC proxy through Electron at runtime and inject only that resolved HTTP, HTTPS, or SOCKS environment into the child process. DIRECT and resolver failures retain the existing direct/inherited network path. No proxy endpoint is stored in settings.json or config.toml, no user Codex CLI configuration is changed, and there is no hardcoded local port. Validation: 445 targeted tests passed across system proxy parsing, Codex auth, backend resolution, repository/framework, and renderer coverage; node/web typecheck passed; ESLint reported zero errors.

ewen-poch:
Addressed the loopback-routing finding in 7334602. Resolved proxy environments now merge the user's existing NO_PROXY/no_proxy entries and add localhost, the IPv4 loopback host/range, and bracketed/unbracketed IPv6 loopback forms before they are passed to Codex auth or prompt children. This keeps imported local Responses routes direct while external subscription traffic still follows the user's system/PAC proxy. A DIRECT decision still injects no proxy or bypass overrides, and nothing is persisted to TOML or app settings. Validation: 258 focused proxy/auth/service tests passed, node/web typecheck passed, targeted ESLint passed, and git diff checks passed.

ewen-poch:
Addressed the inherited-proxy finding in 4cf7e09. Codex subscription auth and prompt processes now remove all inherited HTTP_PROXY, HTTPS_PROXY, ALL_PROXY, and NO_PROXY variants before applying the current Electron resolveProxy decision. This makes both a resolved proxy and DIRECT authoritative instead of allowing stale parent-process values to win.

I did not change isolated logout to invoke codex logout. The isolated home pins the credential store to file, API-key environment variables are stripped, and imported TOML route settings are sanitized and credential-free. Invoking adapter logout could revoke the global CLI session from which credentials were copied; removing the app-owned auth.json remains the intentional local logout boundary.

Validation: 306 focused tests passed, typecheck passed, targeted ESLint passed, and git diff --check passed.

ewen-poch:
Addressed both findings in 6c0dd1c.

For proxy fallback, resolver failure is now represented separately from an explicit DIRECT decision. The resolved backend carries an explicit inherit/replace proxy policy into the Codex spawn: resolver failures preserve launcher-provided proxy variables, while resolved proxies and DIRECT still replace all inherited proxy shapes. Authentication sessions apply the same distinction.

For legacy isolated logout, the app-owned Codex home is now migrated to cli_auth_credentials_store = "file" before auth.json is removed. This prevents future status/runtime checks from reading a legacy keychain-backed credential without invoking remote logout or affecting the user global CLI session.

Validation: 308 focused tests passed, node/web typecheck passed, targeted ESLint and Prettier passed, and git diff --check passed.

ewen-poch:
Addressed both migration/import findings in f7dc988.

Legacy codex-shared records without an explicit codexAuthMode now migrate to isolated mode, matching the existing migration comment and avoiding a false imported state when no credential copy occurred. Explicit imported mode on normalized subscription records remains preserved.

Loopback route import now rejects provider configurations that depend on omitted credential-bearing settings: env_key, experimental_bearer_token, literal or environment-backed HTTP headers, query parameters, dotted assignments, and nested provider tables. Self-contained loopback Responses routes continue to import, while auth.json remains the only copied credential source.

Validation: 334 focused settings/auth/service tests passed, node/web typecheck passed, targeted ESLint and Prettier passed, and git diff --check passed.

ewen-poch:
Addressed the final correctness findings in `d1798b8`:

- **Authentication-mode race:** `cancelLogin()` now completes only after the active isolated auth session has closed. Import, mode-switch cleanup, logout, and provider deletion await that completion before changing app-owned credentials, preventing a late browser-login token write from replacing an imported `auth.json`.
- **Empty URL delimiters:** route validation now rejects raw `?` and `#` delimiters before WHATWG URL normalization, including trailing empty query/fragment forms.
- **Quoted TOML table keys:** compatible provider tables using `\"model_providers\"` or `'model_providers'` as the quoted root key are now recognized, while the existing fail-closed field allowlist remains unchanged.
- **Provider deletion:** deleting a Codex subscription now removes its app-owned authentication and imported route after cancelling any active login. The user's global Codex authentication is left untouched and no remote logout is invoked.

The system-proxy scope is intentionally retained. Both imported and in-app authentication use the same app-owned Codex subscription child runtime, and the reported live failure was a valid saved login whose prompt transport stalled without the user's current system/PAC route. The PR description now makes this behavior explicit. Proxy data remains runtime-only: no endpoint is written to settings or TOML, explicit proxy/`DIRECT` decisions replace inherited proxy variables, resolver failure keeps the inherited environment, and loopback destinations stay bypassed.

Validation: all 266 Codex authentication/settings service tests passed, node and web typechecks passed, targeted ESLint and Prettier passed, and `git diff --check` passed.

PR #463, opened 2026-07-27
feat(agent-context): add native context compaction

ewen-poch:
## Problem

Claude Code and OpenCode do not reliably compact long-running ACP sessions automatically. OpenScience could detect context usage and recover by replacing the session, but it had no way to invoke the agent framework's own summary/compaction mechanism, and users could not compact proactively from the context indicator.

## Proposed change

- Add a framework-level context compaction strategy: Claude Code and OpenCode use their native `/compact` control turn, while Codex remains framework-managed.
- Automatically invoke native compaction when reported context usage reaches 90%.
- Prefer native compaction for request-size overflow recovery, with the existing fresh-session/text-replay path retained as a fallback.
- Add an accessible hover/focus context popover with a manual **Compact context** action and cancellation support.
- Keep native command/status output out of the visible transcript while retaining usage updates and safe per-session lifecycle locks.

## Scope / non-goals

- This does not shorten, delete, or synthesize user-visible messages to imitate compaction.
- This does not add a user-configurable threshold; the framework strategy currently uses 90%.
- This does not override Codex's framework-managed compaction behavior.

## Acceptance criteria / validation

- [x] Claude Code and OpenCode compaction is performed through each framework's native ACP session behavior.
- [x] Context usage at or above 90% triggers automatic compaction.
- [x] Overflow recovery prefers native compaction and safely falls back to context replacement and replay.
- [x] The context indicator popover supports hover, keyboard focus, manual compaction, disabled-state explanations, progress, and cancellation.
- [x] Native control turns do not add command or status messages to the conversation.
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm test`

## Review focus

- ACP control-turn routing and the separation between prompt and compaction locks.
- Ordering of overflow recovery events versus fallback/retry ownership.
- Keyboard and hover behavior of the context indicator popover.


ewen-poch:
Addressed the P1 web bootstrap issue in cd0bae8 by regenerating `src/renderer/web/api-map.generated.ts`. `acp.compactSession` now maps to `acp:compact-session`, so the web renderer exposes native compaction through `window.api`. Verified with `npm run check:web-api-map`, `npm run typecheck`, `npm run lint`, and the full test suite (496 files / 7,014 tests passed; 10 files / 113 tests skipped).

ewen-poch:
Addressed the cancelled-compaction P1 in e7af070. Native compaction now emits `completed` only for `stopReason: end_turn`; `cancelled` and every other non-success stop reason produce a failed compaction and reject the operation, allowing overflow recovery to use reset-and-replay instead of retrying the unchanged session. Added a regression test covering the cancelled result and verifying that no completed event or stale in-flight state remains. Verified with the focused ACP runtime suite (223 tests), `npm run typecheck`, `npm run lint`, and the full suite (496 files / 7,015 tests passed; 10 files / 113 tests skipped).

ewen-poch:
Addressed both compaction lifecycle findings in ca3289b. A session with `compacting: true` is now treated as active by disconnect recovery, so a dropped ACP connection clears the compacting state, marks the session interrupted, and exposes Resume; a regression test covers this transition. Permission-profile changes are also disabled while compaction is active, matching the main-process interaction guard. Verified with the focused workspace runtime suite (65 tests), `npm run check:web-api-map`, `npm run typecheck`, `npm run lint`, and the full suite (496 files / 7,016 tests passed; 10 files / 113 tests skipped).

ewen-poch:
Addressed the overflow-recovery cancellation finding in 54993bc. A successful Cancel now settles the local compaction state immediately, and recovery checks that state after native compaction and after reset; once cancelled, it preserves the unanswered turn and dispatches neither reset fallback nor retry. Added regression coverage for both the cancellation state transition and the no-reset/no-resend behavior. I did not change the nullable `activeSession` expression: the preceding optional-chain status condition narrows `activeSession` for the remaining `&&` operands, and this is confirmed by local `npm run typecheck` plus all three passing Verify jobs on ca3289b. Final verification for this change: focused workspace runtime suite (67 tests), `npm run check:web-api-map`, `npm run typecheck`, `npm run lint`, and the full suite (496 files / 7,018 tests passed; 10 files / 113 tests skipped).

ewen-poch:
Addressed the cancellation-lock finding in 83e5b51. Cancel now records overflow-recovery cancellation intent without clearing `compacting` when the cancel IPC is merely acknowledged. The composer and other guarded actions remain blocked until the native control turn or reset request reaches its terminal response; recovery then consumes the cancellation intent, clears the compaction state, preserves the unanswered turn, and dispatches no fallback or retry. Regression coverage now verifies both the lock-after-ack behavior and the terminal no-reset/no-resend behavior. Verified with the focused workspace runtime suite (67 tests), `npm run check:web-api-map`, `npm run typecheck`, `npm run lint`, and the full suite (496 files / 7,018 tests passed; 10 files / 113 tests skipped).

ewen-poch:
Addressed the inline-image budget finding in a6703a5. Successful native compaction now clears the per-session accumulated inline-image budget before reporting completion, so manual, automatic, and overflow-recovery compaction all start with a clean post-summary media budget. Cancelled and failed compactions retain the existing budget. The Claude ACP adapter can report compact_result: failed as hidden assistant output followed by end_turn, so the framework strategy now identifies that failure signal and the runtime waits for the terminal response before rejecting, preserving lock safety. Added regression coverage for success, cancellation, and hidden adapter failure. Verified with npm run typecheck, npm run lint, npm run check:web-api-map, and the full suite: 496 files passed, 7019 tests passed.

ewen-poch:
Addressed the prompt-submission race in 9acee7c. The workspace now exposes promptInFlightSessionIds from the runtime snapshot and uses per-session runtime ownership to disable send, edit-resend, permission, and compact actions until the control turn finishes. The shared send and edit-resend paths also enforce the same guard before appending or truncating transcript state. Overflow recovery is kept compatible with the guard: native compaction uses its returned terminal snapshot, reset recovery projects the terminal lease release from its response, and the unanswered turn is removed only after confirming that the runtime no longer owns the session. A premature still-in-flight snapshot preserves the original turn and dispatches no retry. Added regression coverage for composer gating, normal send, edited resend, stale recovery snapshots, and premature terminal snapshots. Verified with 72 focused tests, npm run typecheck, npm run lint, npm run check:web-api-map, and the full suite: 496 files passed, 7023 tests passed.

ewen-poch:
Addressed the retired-runtime ownership finding in 9520337. The coordinator now publishes promptInFlightSessionIds only when the reporting runtime is the current owner in sessionRuntimes. A retiring generation therefore continues to lock its active session until adoption, but once a fresh runtime resumes the same logical session, the old draining turn no longer disables composer, edit, permission, or compaction actions for the new owner. The published ids are also deduplicated across runtime snapshots. Added a regression test covering the complete handoff: old-owner lock before adoption, no stale lock after adoption, and a new lock when the fresh owner starts work. Verified with the focused coordinator suite (27 tests), npm run typecheck, npm run lint, npm run check:web-api-map, and the full suite: 496 files passed, 7024 tests passed.

ewen-poch:
Addressed the late retired-runtime overflow finding in 01a20ff. Recoverable context-overflow events now follow the same current-owner rule as interaction locks. The coordinator filters non-owner overflow events in both publication paths: immediate onEvent forwarding and aggregated snapshot replay. This prevents a draining generation from starting beginCompaction/reset against a fresh owner, while preserving ordinary historical events and legitimate overflow recovery before adoption or from the current owner. Added a handoff regression covering valid pre-adoption overflow, suppressed late retired overflow, and accepted fresh-owner overflow. Verified with the focused coordinator suite (28 tests), npm run typecheck, npm run lint, npm run check:web-api-map, and the full suite: 496 files passed, 7025 tests passed.

ewen-poch:
Addressed both P2 findings in c209be0:

- Overflow recovery now keeps live-attempt ownership separate from the 15-second cooldown/dedup window. Cancellation is recorded only while that recovery is actually active, and stale cancellation intent is cleared before each new attempt.
- A native compaction stop with stopReason cancelled now emits a benign cancelled lifecycle event instead of a failed compaction. The renderer clears the compacting state back to idle without surfacing a session error; overflow recovery still consumes the user cancellation and does not reset or retry.

Validation: typecheck and web API map passed; lint completed with 0 errors and the existing 24 warnings; the full suite passed with 496 files / 7027 tests (10 files and 113 tests skipped).

ewen-poch:
Addressed both current-head findings in 52e4f8a:

- Native compaction now snapshots the pre-compaction usage reading and invalidates it on successful completion only when no fresh usage_update arrived. A genuine usage update emitted by the control turn is retained, so the context meter and automatic threshold use post-compaction data without repeatedly compacting from a stale high-water mark.
- Manual compaction now acquires the renderer-local compacting gate synchronously before the IPC call. Ordinary sends and repeated manual attempts are rejected while that gate is active, while overflow recovery has an explicit internal handoff that preserves its owned retry. The terminal path also clears the optimistic gate if no lifecycle event reaches the renderer.

Validation: all 341 affected tests passed; typecheck and web API map passed; lint completed with 0 errors and the 28 warnings present after rebasing onto the latest main. The post-rebase full suite passed 7,115 tests and had 9 failures only in the newly merged theme tests because localStorage is unavailable in the local Vitest environment; this PR has no diff in those files.

ewen-poch:
Addressed both current-head P2 findings in 277de2c:

- Session-scoped compaction lifecycle events now follow the same current-owner rule as recoverable overflow events. Both immediate forwarding and aggregate snapshot replay suppress late compaction events from a retired runtime after a fresh runtime adopts the logical session, while preserving pre-adoption and current-owner events.
- Manual compaction is now restricted to idle sessions in both the Workspace UI gate and the shared runtime helper. An unresolved error session keeps its original error and reportability unchanged, and the context popover explains that the error must be resolved before compaction.

Added regression coverage for retired-owner handoff across both event paths, helper-level error preservation, and the UI disabled state/reason. Validation: 105 affected tests passed; typecheck and web API map passed; lint completed with 0 errors and 28 existing warnings; the full suite passed with 504 files / 7,128 tests (10 files and 113 tests skipped).

Issue #337, opened 2026-07-22
[Feature] In-app find bar (Cmd+F) for long conversations and documents


pillar-github:
**Problem**
Conversations and rendered documents can get very long. There is currently no way to search within the visible content, so locating a specific term means scrolling manually through the entire thread.

**Proposal**
Bind `Cmd + F` to an in-app find bar (e.g. Electron `webContents.findInPage`):
- Highlight all matches in the current view
- `Enter` / `Shift+Enter` (or arrows) to jump between matches
- `Esc` to dismiss

**Scope suggestion**
Start with the chat/conversation view; extending to notebooks, file previews, and skill docs would be a bonus.

PR #467, opened 2026-07-27
fix(workspace): harden whole-window find overlay

ewen-poch:
Follow-up to #466. Relates to #337.

## Problem

Post-merge review found several reliability and consistency gaps in the whole-window find overlay:

- the first show event could arrive before the overlay page registered its listener;
- the overlay stayed open when the searchable renderer became unavailable;
- the separate overlay origin could not reliably mirror app theme changes while Settings was open;
- the first dark-system open could briefly paint the light palette;
- document-level Enter handling overrode focused action buttons;
- focus, tooltip, motion, and theme styles did not fully match the app standards.

## Proposed change

- Keep the first overlay load at zero bounds until its page is ready, then focus and show synchronously on later opens while allowing recovery from load failures.
- Close the overlay when Workspace unmounts, top-level navigation begins, or the renderer crashes or becomes unresponsive.
- Push typed resolved appearance changes from the renderer through preload to the window-owned overlay manager, validating both sender and payload and cleaning up the window-scoped listener.
- Let the manager own the appearance cache and background color, remove the competing `executeJavaScript` sampler, and use a synchronous system-theme fallback before first paint while preserving explicit light overrides.
- Keep Escape global to the overlay while limiting Enter and Shift+Enter handling to the query input.
- Align controls with semantic theme tokens, accessible focus rings, custom tooltips, standard radius, and reduced-motion behavior.
- Add direct regression coverage for lifecycle, IPC forwarding, keyboard chords, theme races, and cleanup paths.

## Scope and non-goals

- Scope is limited to the Electron whole-window find overlay introduced by #466 and its shared/preload contract.
- This does not change search matching semantics or add whole-window find outside the searchable Workspace surface.
- This does not address the repository's existing lint warnings or dependency audit findings.

## Acceptance and validation

- [x] `npm run typecheck`
- [x] `npm run lint -- --no-cache` — 0 errors; 28 pre-existing warnings
- [x] Prettier check for all changed files
- [x] Full Vitest suite on Node 24 LTS — 505 files passed, 10 skipped; 7119 tests passed, 113 skipped
- [x] Final Spec review — no remaining actionable issues or scope creep
- [x] Final Standards review — no remaining actionable issues

## Review focus

- First-show and close/load-failure race handling in `src/main/find-overlay.ts`
- Renderer lifecycle closure paths in `src/main/windows.ts`
- Live theme propagation and validation across renderer, preload, main, and overlay
- Synchronous focus/show, zero-bounds initial load, and first-paint system-theme fallback
- Keyboard and accessibility behavior in `resources/find-overlay/`

### DIFF
--- a/resources/find-overlay/findOverlay.js
+++ b/resources/find-overlay/findOverlay.js
@@ -1,9 +1,13 @@
 export const LAST_QUERY_STORAGE_KEY = 'open-science:window-find:last-query'
+const DARK_QUERY = '(prefers-color-scheme: dark)'
 
 export function createFindOverlay(deps) {
   const { input, count, prev, next, close, api, storage } = deps
+  const ownerDocument = input.ownerDocument
+  const systemTheme = ownerDocument.defaultView?.matchMedia?.(DARK_QUERY)
 
   let currentRequestId = 0
+  let followsSystem = false
   count.textContent = '0 / 0'
 
   const nextRequestId = () => {
@@ -16,7 +20,28 @@ export function createFindOverlay(deps) {
     api.findInPage({ requestId, text, findNext, forward })
   }
 
-  const handleShow = () => {
+  const applyTheme = (theme) => {
+    ownerDocument.documentElement.classList.toggle('dark', theme === 'dark')
+  }
+
+  const onSystemThemeChange = (event) => {
+    if (followsSystem) applyTheme(event.matches ? 'dark' : 'light')
+  }
+  systemTheme?.addEventListener('change', onSystemThemeChange)
+
+  const handleShow = (appearance) => {
+    followsSystem = appearance?.followsSystem === true
+    const theme =
+      followsSystem && systemTheme
+        ? systemTheme.matches
+          ? 'dark'
+          : 'light'
+        : appearance?.theme === 'dark' || appearance?.theme === 'light'
+          ? appearance.theme
+          : systemTheme?.matches
+            ? 'dark'
+            : 'light'
+    applyTheme(theme)
     input.focus()
     const remembered = storage.getItem(LAST_QUERY_STORAGE_KEY) ?? ''
     if (remembered) {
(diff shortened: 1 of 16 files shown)


ewen-poch:
Addressed the AI review feedback in `b1cb4db`.

- A loaded find overlay now focuses and receives `WINDOW_FIND_SHOW_CHANNEL` synchronously, so keystrokes immediately following Cmd/Ctrl+F reach the query field.
- Renderer appearance lookup remains asynchronous and updates through a separate typed channel that cannot refocus the input or rerun the remembered query.
- Cached/system appearance keeps the synchronous path usable; stale results after close/reopen are discarded.
- Added regressions for the immediate-focus path, appearance-only updates, invalid/rejected reads, close/reopen races, preload forwarding, and listener cleanup.

Trade-off: the first open can briefly use the current system/cached appearance before the renderer result arrives, prioritizing input ownership over a potentially delayed theme-perfect first frame.

Validation: typecheck passed; lint has 0 errors (28 pre-existing warnings); full Vitest passed with 504 files / 7113 tests.

ewen-poch:
Addressed both follow-up AI review findings in `b8b3e20`.

- Theme changes now flow through a typed renderer → preload → main channel into the window-owned overlay manager. Main validates the sender and payload, unregisters the window-scoped listener on close, and the manager updates the open overlay without stealing focus or replaying the current query.
- The manager is now the single owner of the appearance cache and background color; the per-open `executeJavaScript` theme sampler was removed.
- A newly created overlay remains at zero bounds until its document is ready. A synchronous head script supplies the dark-system first-paint fallback, while the resolved appearance payload still lets an explicit light preference override it.

Trade-off: this adds one low-frequency IPC notification per resolved theme state change, but removes the competing asynchronous per-open sampler and keeps the overlay correct while Settings changes theme. Holding the initial view at zero bounds avoids a wrong-palette flash at the cost of showing only after the overlay document has loaded; subsequent opens remain synchronous.

Validation:

- Full Vitest: 505 files passed, 10 skipped; 7119 tests passed, 113 skipped
- Typecheck passed
- Lint: 0 errors (28 pre-existing warnings)
- Prettier and `git diff --check` passed
- Final Standards and Spec reviews: no actionable findings

PR #481, opened 2026-07-28
feat(artifact-provenance): add durable versioned provenance

ewen-poch:
## Problem

Generated outputs were stored as files attached to a turn, without a durable version identity or an inspectable chain back to producer code, execution history, inputs, environment state, conversation context, and reviewer evidence. Session deletion, branch editing, app restart, and data-root migration made that evidence difficult to retain and validate.

## Proposed change

- Add session-scoped Artifact lineages with immutable versions, checksummed content, evidence manifests, and SQLite metadata.
- Capture producer code, bounded execution logs, input-file versions, environment inventory, message-branch snapshots, and version-scoped review evidence.
- Add crash recovery, idempotent writes, storage migration validation, upload versioning, conversation-graph persistence, and explicit unavailable states when producer ownership cannot be proven.
- Add Provenance entry points and version-aware preview behavior across the workspace, Files, modal preview, and generated Artifact surfaces.
- Harden the trusted Artifact save/finalization path across supported ACP frameworks and preserve project-scoped file reuse.
- Make additive runtime SQLite migrations prove their column postcondition instead of swallowing arbitrary errors, and verify runtime DDL coverage against every Prisma scalar field.
- Document the current audit guarantee and the limits of deterministic replay.

## Guarantees and non-goals

This change provides an immutable audit and traceability record. Environment evidence is an observed inventory, not a portable solver lock. Retained message projections omit binary media and are not a complete Session backup. Full-fidelity Session replay, portable environment export/restore, and deterministic reconstruction remain future work.

## Validation

- `npm test -- --run src/main/projects/prisma-client.test.ts` — 7/7 passed.
- `npm run lint` — 0 errors; 23 warnings in existing/latest-main files.
- `git diff --check origin/main...HEAD` — passed.
- Full `npm test` completed but exits non-zero in this worktree environment. Failures are limited to known environment constraints: sandboxed localhost listeners, Vite denial of the parent-worktree PDF worker, missing Node `localStorage`, and missing worktree-local electron-builder NSIS templates.

## Review focus

Please focus on immutable write/finalize semantics, branch ownership checks, recovery and migration invariants, environment evidence confidence, reviewer scope binding, and version-aware preview behavior.


ewen-poch:
Implemented the high-value review feedback in 0939887 and 6d93e81.

- Preview reads now carry Project scope consistently across text, managed image/HTML/PDF resources, thumbnails, Project Files probes, and restored preview tabs.
- Upload Version locators use their immutable source Session for cross-Session `@` previews while retaining the active Project boundary. Legacy upload locators keep their explicit owner Session.
- The working-file containment direction remains unchanged because `isPathInside(sessionRoot, dataRoot)` correctly validates the standard layout. The macOS failure exposed a different issue: recursive watcher startup replay. A pre-run metadata baseline now filters unchanged replayed paths while preserving real overwrites, including same-size/same-mtime changes through ctime.
- Provenance lifecycle/source states now have named SQLite CHECK constraints. Legacy tables are rebuilt transactionally, preserve compatible rows and foreign keys, reject unknown source columns, and enforce the exact `ArtifactVersionInput` source identity union.

Validation: typecheck and `git diff --check` pass; 10 targeted test files pass with 132 tests. The local ProjectFiles/PDF Vite parent-worktree restriction remains environment-specific, so the pushed CI matrix is the authoritative validation for those suites.

ewen-poch:
Addressed the cross-Session mention finding and the actionable macOS failures in 72fbb70.

- Transcript mention probes now reuse the same locator-derived scope as preview surfaces, so a Project-scoped upload from another Session is checked against its immutable source Session before opening.
- Working-file observation keeps the watcher event fast path. If watcher events are empty or only replay unchanged files, it performs a bounded final snapshot diff; this covers delayed macOS recursive events without imposing a second tree scan on normal runs.
- Scoped preview assertions now include the expected Session/Project fields, and Project Files finalization tests wait until their change listener is registered before publishing an update.

Validation: node/web typecheck and `git diff --check` pass; the kernel suite passes 41/41 locally, including a new missed-watcher-event regression test. The remaining macOS idle-timeout assertion is an existing timing-sensitive test and passed locally and on Windows, so no production behavior was changed for it.

ewen-poch:
Addressed both Codex P2 findings and the remaining macOS fixture failures in 091fd20.

- Session image sanitization now treats the conversation graph as canonical, applies one aggregate budget across active and inactive Branches, and rebuilds the active `messages` projection from the sanitized graph. `createSessionFile()` enforces this at the serialization boundary.
- Artifact-backed Notebook inputs now retain the owning `artifactId` together with `selectedVersionId` and `versionNumber`, preserving Provenance and Version navigation when the stable preview tab is reused.
- Project Files tests now model finalization with a scoped immutable `upload-version:` locator. The legacy library adapter also recognizes Version-backed attachments without retaining a native path.

Validation: typecheck and `git diff --check` pass; 5 focused persistence/UI test files pass with 44 tests. Both new behaviors were covered by failing tests before implementation.

ewen-poch:
Addressed the version-backed prompt attachment gap in 2b5a68f. The ACP file-reference path now detects immutable artifact-version locators, resolves them through the provenance repository, rejects cross-project locators before resolution, and uses persisted Version filename/content type metadata. Added a prompt-level regression test plus a Project-boundary test; focused tests and typecheck pass.

ewen-poch:
Addressed both review points in 590f35e. Edited-message replay now scopes the preflight media projection with the Session Project, and a regression test verifies that the dispatched history attachment uses the immutable scoped upload-version locator. (The final dispatch path already rebuilt scoped media, so this also removes the remaining inconsistent unscoped call.) Preview persistence now validates and round-trips artifactId, selectedVersionId, versionNumber, and originSession across project switches and restarts. The added fields are optional and backward-compatible, so the persisted-state version remains unchanged. Focused tests (31 assertions) and typecheck pass.

ewen-poch:
Addressed the inline producer contract regression in cd2f594. The app-owned artifact adapter now carries the resolved source kind into the durable request and its idempotency checksum. Inline writes retain an explicitly supplied producer only after the existing Notebook Session and graph-scope checks succeed; local-path writes without a verified file observation remain fail-closed. Added regression coverage for both paths (39 focused tests pass).\n\nI intentionally kept the other reviewed areas unchanged: package change evidence has active persistence/UI consumers and supports the requested dependency-version display; Review history remains deferred to avoid introducing a second review presentation; and Inputs stays in Code per the later product decision.

ewen-poch:
Addressed the legacy Project File mention regression in f05ecda. Non-versioned Project File references remain valid prompt attachments but are skipped by Notebook input registration because they cannot establish an immutable Version edge. Version-backed references in the same turn are still registered and validated normally. Added a mixed legacy/versioned regression test; the focused suite passes 5/5, with typecheck and targeted ESLint also passing.

ewen-poch:
Implemented the high-value findings in commit 19a1065: finalized-run responses now exclude staging Versions; active source Session read failures make saved Review conclusions explicitly unavailable while deleted origins retain frozen history; Project Upload bytes are removed before authority rows so deletion remains retryable; and Artifact mention plus molecule previews preserve lineage and Version identity. Added focused regression coverage for each behavior and cleaned the two confirmed static-analysis issues. I did not change inline producer attribution because the current design explicitly labels a scope-valid inline association as agent-declared-and-session-validated, nor add localPath checksums because that expands the persisted WorkingFile contract beyond the current path/owner/size/mtime trust model. The saved runRecovery Promise is intentional shared in-flight state, not a missing await. Focused tests pass (95/95), along with typecheck and targeted ESLint; the local full suite remains blocked by the known listen EPERM, missing localStorage, and worktree Vite path restrictions.

ewen-poch:
Addressed both findings in 2765f13. Project SQLite clients now use connection_limit=1 as an explicit correctness invariant, so the connection-scoped foreign_keys pragma and the constraint-rebuild transaction execute on the same physical connection. Prompt construction now registers every finalized Upload Version actually sent to the model, including replayed history attachments, rather than only current-turn uploads. Added a mixed history/current attachment regression test. Typecheck, targeted ESLint, the new runtime test, and all 13 Prisma migration tests pass; the only failure in the broader runtime file remains the environment-blocked local HTTP listener test.

ewen-poch:
Addressed the actionable review findings in cba3576:

- Interrupted-turn retry now forks the canonical conversation graph before the unanswered prompt. The previous Branch and immutable Message nodes remain preserved, while the active projection no longer resurrects the stale prompt after save/reload.
- Provenance-native Review scopes now resolve through SQLite authority, hash the resolved bytes, validate the authority checksum again on read, and fail closed when active-session evidence cannot be recomputed. Legacy managed artifacts retain an explicit path fallback.
- Rowless staging cleanup is limited to startup reconciliation, age-gated, and removes stale temporary copies instead of accumulating them in quarantine. Read-triggered reconciliation is non-destructive. Startup reconciliation failures are logged and mark the Files index incomplete.
- Environment platform and architecture are captured from the inspected Python/R interpreter and omitted when unavailable. Partial-capture warnings are now visible in the Provenance Environment tab.

Validation: typecheck passed; targeted ESLint and git diff checks passed; 155 focused provenance/session/environment tests and 59 Reviewer service tests passed. The previously executed full suite only reported the known worktree Vite path, Node localStorage, and electron-builder fixture failures.

Intentional trade-offs left unchanged:

- Artifact write idempotency remains request/operation scoped. Without a caller request id, the application cannot distinguish a transport retry from an intentional second save in the same turn.
- A missing persisted selectedVersionId still resolves to the latest finalized Version, matching the current preview UX requirement.
- The unused external import helper, documentation-only formatting churn, and speculative future DDL evolution were not expanded in this correctness fix.

ewen-poch:
Addressed both findings from the latest Codex review:

- 6d622eb makes the Project Files index Branch-independent. It extracts file references from every canonical conversation-graph Message, with the active projection overriding the same Message id only when it has a newer payload. Inactive-Branch Upload and Artifact Versions remain available to both the current Session and cross-Session @ pickers.
- f711b2b persists the finalized, checksum-bearing Artifact descriptors after replaceMessageArtifacts and awaits that save before the stop path can start auto-review. The event-order regression test now requires save -> finalize -> save, including finalized Version ids in the second durable Session write.

Validation: 93 focused Project Files, @ picker, and workspace-event tests passed; typecheck, targeted ESLint, Prettier, and git diff checks passed. The prior Windows CI failure was an unrelated micromamba ACL integration timeout at 15 seconds; this push will start a fresh platform run.

ewen-poch:
Added an explicit safe-downgrade workflow in Settings → Storage and the README. The UI now shows the main-resolved Config Root and active Data Root, provides trusted Reveal actions for both, and states that a rollback is safe only after restoring matching pre-upgrade copies of both roots before opening the older binary. Without matching backups, the older app is documented as view-only (no message writes/edits, project deletion, file generation, or Data Root migration). I intentionally did not add automatic full-root copies because runtime environments can make the Data Root several GB and that cost should not be hidden in the upgrade path.\n\nValidation: full npm test passed; 94 focused Storage/Onboarding tests passed; typecheck, ESLint, Prettier, generated web API map check, and git diff check passed.

ewen-poch:
Implemented the repository-provided `open-science rollback-to-0.7.3 --yes` command in 6b9c137.

The command runs offline, verifies finalized Artifact and ready Upload bytes against SQLite size and SHA-256 metadata, projects each active Session branch into the 0.7.3 envelope, excludes runtime environments, activates an isolated rollback Data Root, and preserves the newer Config/Data Roots. Cutover uses a durable marker and is idempotent after success or interruption; recovery now rechecks that Open Science is closed before any rename, and a marker cleanup failure after activation cannot delete the committed rollback data. The previous in-app downgrade prompt was removed.

Validation:
- 15 targeted CLI/rollback tests passed
- CLI package dry-run includes the rollback module
- Typecheck passed
- Lint passed with 0 errors (24 pre-existing warnings)
- `git diff --check` passed
- The full Vitest run was attempted earlier; remaining failures were existing environment limitations around loopback `EPERM`, missing `localStorage`, and Vite access to the shared worktree dependency path.

ewen-poch:
Follow-up 536a71e closes three additional cutover safety gaps found during independent spec review:

- a durable `preparing` marker is committed before any rollback target is created, so an interrupted copy can be cleaned and retried, including with an explicit output path;
- prepared Config/Data trees are fsynced, and recovery revalidates both manifests, SQLite integrity, Version size/checksum, and converted Session references before any rename;
- both normal activation and interrupted recovery recheck that Open Science remains closed immediately before cutover;
- marker JSON is written through a synced same-directory temporary file and atomic rename, preventing torn final markers.

The targeted rollback/CLI suite now passes 19/19, typecheck passes, targeted ESLint reports no issues, CLI package validation passes, and `git diff --check` passes.

ewen-poch:
Addressed the latest high-value findings in two focused fixes:

- Notebook input resolution is now bound to the exact app-issued run lease, so overlapping runs no longer both record the same immutable input as resolver-accessed. Renderer-supplied lease identifiers are stripped at the IPC boundary.
- Windows Artifact durability now opens ordinary files with a write-capable r+ handle before sync, while directory sync remains read-only and keeps the existing Windows unsupported-operation fallback. The legacy and provenance repositories share the same tested durability implementation. The rollback tool uses the same file-handle rule.

Validation: targeted durability and barrier tests passed, rollback tests passed, the run-scoped overlap regression passed, TypeScript typecheck passed, targeted ESLint passed, and the CLI package dry-run passed. The remaining Local RPC tests require loopback listening, which is blocked by the local sandbox (listen EPERM); CI will run them normally.

ewen-poch:
Addressed the current Codex finding and the actionable macOS CI failures.

- Rollback no longer trusts the current flat Message projection when a reviewer or delegate Frame is active. It validates the Conversation Graph, walks the active root Frame Branch, converts that root transcript, and removes graph-only node fields before writing the 0.7.3 envelope. A regression test covers an active child Frame and verifies that the root prompts and responses are preserved.
- Artifact durability now defers access to the fs open binding until a barrier is actually executed. Test suites that partially mock node:fs/promises can import the repositories without defining an unused open export.
- The idle-timeout test now waits for the explicit in-flight pending condition instead of assuming macOS schedules the child request within one event-loop tick. Production timeout behavior is unchanged.

Validation: 35 focused rollback, durability, and ACP IPC tests passed; the macOS kernel timing regression passed; typecheck, targeted ESLint, CLI package validation, and git diff checks passed.

ewen-poch:
Addressed the actionable provenance and recovery findings in c24d53c:

- Finalization now revalidates every frozen execution run against ancestry derived from the durable conversation graph, preventing sibling-branch evidence from being finalized.
- Artifact Version filenames are frozen per Version. Existing databases backfill the field from the lineage and rebuild the table with NOT NULL and non-empty constraints; a real legacy-schema migration test covers the postcondition.
- Staging Review scope snapshots are reconciled at startup: valid SQLite-owned snapshots are republished and promoted, while invalid recovery proofs fail the Review closed.
- Execution snapshots now have explicit 4 MiB, run, output, input, and per-output budgets. The producer run is retained and truncation counts are persisted and disclosed in the Execution Log UI.
- Skill upload authorization validates Project ownership before granting access, while still allowing cross-Session references inside the same Project.
- Migration validation now compares Upload inputs with the frozen original filename, and Provenance GET paths no longer recreate missing reconciliation sidecars.
- Added structured diagnostics for legacy Artifact fallback and higher-precision R environment fingerprints.

I did not add a generic full PRAGMA table_info parity framework. The concrete ArtifactVersion filename migration now proves its required type/nullability/check postconditions through an integration test; general runtime-DDL parity is broader migration infrastructure and is not necessary for this fix.

Validation: 102 targeted main-process tests and 13 Artifact Provenance renderer tests pass; typecheck, targeted ESLint, Prisma schema validation, and git diff --check pass. Full npm test was also run locally, but the existing worktree environment still fails unrelated suites because loopback listen is denied, localStorage is unavailable, and Vite rejects shared parent-worktree node_modules paths.

ewen-poch:
Fixed the Windows Session finalization failure in 02d297a.

The failure was a transient rename-over-existing denial while atomically replacing the Session JSON. Session writes are already serialized and use automatically closed file operations, so the fix keeps the temp-plus-rename protocol and retries only EPERM, EACCES, and EBUSY with a bounded 775 ms backoff. Persistent failures still fail closed and now remove the abandoned temporary file; unrelated rename errors are not retried.

A deterministic regression test reproduced the exact first-rename EPERM before the fix. The Session persistence suite now passes 21/21, and typecheck, targeted ESLint, and git diff checks pass.

ewen-poch:
Implemented two renderer integrity fixes. Deferred Provenance tabs now keep an explicit loading state, cache successful sections, and isolate section failures instead of flashing terminal unavailable states. Terminal conversation-graph synchronization now fails closed and blocks unsafe persistence, Artifact finalization, Review, send, and edit paths. I also fixed inactive-branch Artifact replay: late or replayed claims resolve against their original response in the durable conversation graph, while the main process continues to reject actual cross-message claim rebinding. Validation: 155 targeted renderer tests passed for the integrity changes; 123 targeted event/store tests passed for branch replay; typecheck, targeted ESLint, and git diff checks passed. The previously attempted full suite remains affected by the known local loopback EPERM, localStorage, and shared Vite path environment failures.

ewen-poch:
Fixed the Files and legacy provenance regressions in a4be4fe:

- Files projections now resolve native artifacts from the latest finalized SQLite version, so newly generated v2/v3 content replaces stale thumbnails without requiring a Session revision change.
- Cross-session upload previews now use the immutable locator owner scope, and stale projected upload ownership is repaired during index refresh. Rapid switching no longer sends preview requests under the active Session owner.
- Legacy generated files whose composite IDs predate the provenance schema now resolve to an unavailable/empty provenance state instead of throwing Invalid artifact id.
- The Vitest setup is now portable across Git worktrees and Node runtimes, including jsdom localStorage and packaging dependency resolution.

Validation:
- Complete Vitest run: 1,824 suites passed; 7,529 tests passed; 0 failed; 113 skipped.
- Typecheck passed.
- Lint passed with 0 errors (24 pre-existing warnings in unrelated files).
- git diff --check passed.

ewen-poch:
Added two follow-up fixes:

- 83b51c8 moves the direct Provenance action into the right-side modal controls for both Files dialogs and expanded PreviewPanel surfaces. The compact side panel keeps the actions menu.
- f4c5bc4 restores previews for legacy path-based uploads (including older PDFs). The Files index now repairs rows that copied a referencing Session owner, and the upload resolver accepts a legacy path when the main-maintained ManagedFile row proves membership in the requested Project. Cross-project paths remain rejected, and Agent-facing upload resolution remains Session-bound.

Validation for the legacy preview fix:
- 52 focused repository tests passed.
- Typecheck passed.
- Prettier and git diff --check passed.

ewen-poch:
Implemented the two package-management fixes in 72b3838. The agent-facing manage_packages result now omits verbose micromamba FETCH/LINK diagnostics while preserving the outcome, target, installer, restart flag, and actionable error. A successful installer exit is now verified against a fresh runtime inventory; if a requested package such as dplyr is absent, the tool returns failure and the Environment operation is persisted as failure instead of success. Added MCP projection, Environment tracker, and runtime-service regressions. Validation: 7,535 tests passed (113 skipped), typecheck passed, lint has 0 errors, and Prettier/diff checks passed.

ewen-poch:
Addressed the final macOS CI failure and the remaining high-value Codex findings in 4fb47af.

- Runtime binding tests now pair their fake installer with a fake successful inventory refresh, so they no longer depend on whether the CI host `/usr/bin/python3` happens to contain numpy. Production package verification remains unchanged.
- Finalized upload preview tabs now retain scoped immutable `upload-version:` locators instead of persisting data-root-bound absolute paths.
- Legacy interrupted Sessions normalize messages before creating their conversation graph, and graph-scoped activities/groups are restored to terminal states, preventing stale streaming/in-progress state after branch switches.
- Deferred Artifact claims are queued per Session and drained in order. Auto-review now uses a short debounced barrier; a post-stop Artifact cancels the pending review and reschedules it only after finalization, so Reviewer snapshots include late finalized Artifact Versions.

Validation: full Vitest passed (525 files, 7,539 tests; 113 skipped), typecheck passed, targeted ESLint reported no issues, Prettier passed, and `git diff --check` passed.

ewen-poch:
Applied the bounded, high-value pre-merge fixes in 97270c1:

- Package installation now fails closed when the post-install inventory refresh fails, with explicit verification-failure messaging and regression coverage.
- Historical Review staleness checks now recompute against the frozen messageBranchId instead of the currently active Branch.
- An explicitly requested Artifact Version that no longer exists is shown as unavailable; it is never replaced silently with the latest Version.
- Mixed Python/R execution histories are exported as separate kernel-specific notebooks instead of silently dropping one kernel.

I intentionally did not add a new durable finalization-handoff protocol or delete preserved legacy upload sources in this patch. The former needs a persisted state machine because no trustworthy terminal messageId exists in the pre-save crash window; the latter currently protects the copy-before-Session-save rollback boundary. Both should be designed as separate migration/recovery work rather than patched into the merge candidate.

Validation: npm run typecheck passed; npm run lint completed with 0 errors (existing warnings only); full npm test passed with 525 files / 7,545 tests passed and 113 skipped.

ewen-poch:
Fixed the Provenance version-switch race in 30538b9. Deferred Execution, Messages, and Review responses are now cached together with their section payload and merged by the active Version/tab key. This prevents a Version switch from marking a section as loaded while leaving the core not-loaded placeholder visible. Added a regression test for opening v1 Provenance, staying on Messages, switching to v2, and verifying that v2 evidence is rendered. Targeted renderer tests, Node/Web typecheck, and ESLint pass.

PR #488, opened 2026-07-29
fix(permissions): auto-approve Codex artifact saves

ewen-poch:
## Problem

Codex ACP emits Artifact saves as sparse MCP approval requests. Open Science restores the verified MCP server/tool identity, but the permission policy did not recognize the canonical Codex leaf name, so Codex showed a user approval prompt while Claude Code already completed the same app-owned save without one.

## Proposed change

Carry the runtime-verified Codex MCP identity through permission projection with an internal Symbol marker, and auto-approve only the exact `open-science-artifacts/write_artifact_file` capability. Add policy coverage for the untrusted bare-name rejection path and an end-to-end regression test using the sparse Codex MCP approval shape.

## Scope and non-goals

This changes only the app-owned Artifact save permission path. It does not auto-approve ordinary file writes, unverified bare tool names, third-party MCP tools, or other Codex permissions.

- Codex Artifact saves receive one-call approval without emitting a renderer permission request.
- A bare `write_artifact_file` provider name remains insufficient without the runtime-verified MCP identity.
- Claude Code and OpenCode Artifact save behavior remains unchanged.
- `npm run typecheck`
- `npm run lint`
- `npm test`

## Review focus

Please verify that the internal trusted identity survives the broker option projection while remaining impossible to inject through ACP JSON.

### DIFF
--- a/src/main/acp/permission-policy.test.ts
+++ b/src/main/acp/permission-policy.test.ts
@@ -205,6 +205,15 @@ describe('permission policy', () => {
 
     expect(
       resolveAutomaticPermission(
+        createPermissionRequest('other', undefined, {
+          providerToolName: 'write_artifact_file'
+        }),
+        { profile: 'ask', mcpServerNames: ['open-science-artifacts'] }
+      )
+    ).toBeUndefined()
+
+    expect(
+      resolveAutomaticPermission(
         createPermissionRequest('other', undefined, { title: 'write_artifact_file' }),
         { profile: 'ask', mcpServerNames: ['open-science-artifacts'] }
       )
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the trust-boundary finding in 61d6e2b. Codex tool-call context is now promoted to a trusted MCP identity only when the adapter supplies `_meta.is_mcp_tool_call === true`, the title exactly matches `mcp.<server>.<tool>`, and the server is configured for the session. Added negative regression coverage for both a missing MCP marker and a mismatched qualified title; both cases now reach the normal permission path. Focused Codex Artifact/MCP tests pass, along with Node typechecking and ESLint on the changed files.

PR #489, opened 2026-07-29
fix(provenance): bound environment operation history

ewen-poch:
## Problem

The mutable environment operation history grew without a bound and was copied into each completed-run manifest and Artifact environment evidence. Long-lived environments could therefore make every later provenance snapshot progressively larger.

## Proposed change

- Bound retained environment operations by both entry count and serialized size.
- Preserve the operation needed for dirty/recovery state even when it exceeds the normal budget.
- Record cumulative truncation metadata, including the omitted count and earliest retained timestamp.
- Project truncation metadata through Artifact evidence and show it in the Provenance Environment tab.

## Scope and non-goals

- Existing immutable Artifact Version evidence is not rewritten.
- This change does not add Environment manifest garbage collection.
- This change does not change app-generated Artifact idempotency; the current connector path has no stable invocation identity, and content-based IDs would collapse intentional duplicate saves.

- Focused Vitest: 3 files, 58 tests passed.
- `npm run lint`: passed with 24 pre-existing warnings; targeted ESLint reported no issues.
- Full `npm test`: 525 files and 7,569 tests passed; two timing-sensitive `provisioner-runtime` timeout-tail tests failed under full-suite load and passed on immediate isolated rerun (18/18).
- `git diff --check main...HEAD`: passed.

## Review focus

- The dual count/byte retention budget.
- Protection of recovery-critical operations.
- Accuracy of truncation metadata across manifest, evidence, and UI projections.

### DIFF
--- a/src/main/artifacts/provenance-repository.test.ts
+++ b/src/main/artifacts/provenance-repository.test.ts
@@ -844,6 +844,10 @@ describe('artifact provenance repository', () => {
           ]
         }
       ],
+      operationLogTruncation: {
+        omittedCount: 4,
+        earliestRetainedAt: '2026-07-27T11:58:00.000Z'
+      },
       complete: true,
       captureStatus: 'complete'
     }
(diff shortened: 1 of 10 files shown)


ewen-poch:
Addressed both findings in `643cc1d`:

- Migration validation and runtime reads now share the same `operationLogTruncation` validator. A malformed-metadata migration regression test is included.
- The byte budget is now enforced against the exact pretty-printed `binding.json` representation. Recovery-critical operations remain protected if they temporarily require exceeding the normal budget, and the regression test asserts the persisted file size.

Validation completed: 4 focused test files / 73 tests, typecheck, targeted ESLint, and `git diff --check`.

PR #491, opened 2026-07-29
feat(notebook): add read-only package inspection

ewen-poch:
## Problem

Agents currently have no dedicated read-only way to answer whether a Python or R package is installed, or which version is present, in the notebook session's active runtime. They must either execute a notebook probe or reach for the mutation-oriented package tool. At the same time, clear missing-import failures should remain a direct install recovery path rather than acquiring a mandatory preflight call.

Package mutation results also omit the verified before/after state of the explicitly requested packages, which makes repeated installation and version-change outcomes harder for the agent and UI to explain.

## Proposed change

- Add a separate `inspect_packages` MCP tool for targeted Python/R distribution metadata in the session-bound runtime.
- Reuse the environment inventory cache and fingerprint refresh path, with `installed`, `missing`, and `unknown` states and installed versions when available.
- Keep inspection read-only: it does not import packages, invoke an installer, or accept a per-call environment override.
- Document that a clear missing-package error should call `manage_packages` directly and that installed metadata does not prove importability.
- Return verified requested-package `packageChanges` from `manage_packages`, including installed/updated/unchanged/removed state and observed before/after versions.
- Ensure a readable baseline inventory exists before the first mutation so the first install or uninstall produces a real delta.
- Add the separate permission presentation, Claude MCP mapping, activity details, skill guidance, and regression coverage.

## Scope and non-goals

- This inspects package/distribution metadata only; it does not guarantee that `import` or `library()` succeeds.
- This does not add a mandatory inspection call before installation.
- This does not expose a full environment inventory dump or add dependency resolution behavior.
- Transitive dependency changes remain in provenance rather than the compact agent-facing mutation result.

- `inspect_packages` reports targeted installed versions or a missing/unknown state for Python and R in the exact session-bound/default runtime.
- Inspection and package mutation remain separate MCP tool identities and permission presentations.
- Clear missing-import guidance routes directly to `manage_packages`.
- First and subsequent mutations return verified requested-package version changes.
- `npm run typecheck` passes.
- `npm run lint` passes with 0 errors (24 pre-existing warnings outside the changed files).
- `npm test` passes: 526 test files and 7,579 tests passed; 10 files and 113 tests skipped.

## Review focus

- Session runtime binding, disabled/recovery behavior, and environment-lock selection for inspection.
- Inventory freshness and the distinction between `missing` and `unknown`.
- Pre-mutation baseline capture and requested-only `packageChanges` filtering.
- Separation of read-only inspection permission from package mutation permission.


ewen-poch:
Addressed the pre-install inventory regression in c0b8dbb. Baseline capture is now best-effort: when the metadata probe is unavailable, the mutation intent is still durably recorded and the installer proceeds; post-install verification reports the observed state (or a refresh failure) without claiming an exact delta. Exact installed/updated/unchanged/removed deltas are preserved whenever a readable baseline exists.

Added a regression test for a failed first baseline scan followed by successful post-install verification. Validation: typecheck, lint (0 errors), and the full suite with 7,580 tests passing.

ewen-poch:
Addressed the external-runtime permission issue in 5470aff. `inspect_packages` is now limited to app-managed runtimes and rejects a bound user-owned runtime before inventory capture can execute its interpreter. The error directs the agent to use `notebook_execute`, which preserves the existing visible execution-approval path for external binaries.

The MCP description, system guidance, bundled environment skill, and permission copy now state this boundary. Added a regression test that binds an external Python runtime, verifies the rejection, and confirms the inventory scanner is never invoked. Validation: typecheck, lint (0 errors), and the full suite with 7,581 tests passing.

ewen-poch:
Addressed the fresh-default-runtime gap in 813feed. `inspectPackages()` now calls the existing `ensureDefaultEnvReady()` path after runtime availability checks and before inventory capture. This materializes an unbound/default app-managed runtime on first use while remaining a no-op for named environments; external runtimes are still rejected before this path.

The runtime test now installs a fake default provisioner and verifies that Python provisioning occurs before the inventory scanner. Validation: typecheck, lint (0 errors), and the full suite with 7,581 tests passing.

ewen-poch:
Addressed the read-only permission boundary in 9756bbe. `inspect_packages` no longer calls the default-runtime provisioner. When the app-managed default runtime is absent or stale, it now fails with `DEFAULT_RUNTIME_NOT_READY` before provisioning or inventory capture and directs the agent to use `notebook_execute`, where first-use setup receives notebook execution approval, then retry inspection. Existing ready managed runtimes remain inspectable without invoking either the provisioner or package installer.\n\nUpdated the MCP guidance and bundled environment skill, added regression coverage for both ready and missing-default paths, and corrected the local RPC fixture to represent a ready runtime. Validation: typecheck passed; lint completed with 0 errors (24 pre-existing warnings); full suite passed with 7,582 tests passing and 113 skipped.

PR #493, opened 2026-07-29
feat(context-usage): estimate category token usage

ewen-poch:
## Problem

The context-window indicator only exposes the Agent-reported total, so users cannot tell whether system guidance, conversation messages, tool activity, MCP activity, or loaded Skills are consuming the window.

## Proposed change

- add a model-aware local estimator that selects the Anthropic tokenizer for Claude, `o200k_base` for GPT/Codex families, and `cl100k_base` as the fallback
- estimate the content visible at the app boundary before each request, track cumulative message/tool/MCP/Skill context, and replace cumulative tool snapshots instead of double-counting them
- keep the Agent-reported total authoritative and expose the signed delta; positive residual is shown as Agent/framework overhead instead of being proportionally redistributed
- reset estimator state with session deletion, context replacement, reconnect, and native compaction
- expand the composer context popover with a segmented category bar, per-category token values, tokenizer/model diagnostics, and preflight/reconciled state

## Scope and non-goals

This does not claim billing-grade counts or invent category precision for content hidden inside the Agent. Built-in tool schemas and framework bootstrap content that are not visible at the app boundary remain in the explicit residual category. Existing context totals and compaction thresholds continue to use the Agent numerator and selected-model window.

- `npm run lint` — passed with 0 errors and 24 pre-existing warnings
- focused tokenizer, runtime, and renderer tests — passed
- `npm test` — 7,587 passed, 113 skipped; two load-sensitive failures passed when rerun individually; the remaining `bounds pending Codex MCP identities` failure reproduces unchanged on `main`

## Review focus

- tokenizer profile selection and the residual/delta semantics in `ContextUsageTracker`
- correct reset/reconciliation behavior across prompt, compaction, disconnect, and replacement paths
- whether the popover makes estimated categories and the authoritative Agent total sufficiently distinct


ewen-poch:
Addressed both findings in `0b51f23`:

- Codex `SKILL.md` reads now reuse the projector's exact path classification and update a stable skill-document section, so native reads are attributed to **Skills** without double-counting app-selected skill content.
- Context estimation now uses only models that ACP confirmed for the live session. When an optional model request falls back to the agent default, the unconfirmed model hint is omitted and the framework fallback tokenizer is used.

Added focused regression coverage for both paths. Node typecheck and the targeted tests pass. I also corrected the existing bounded Codex MCP identity test fixture to include the trusted MCP marker required by the runtime, which resolves the macOS CI failure without changing production behavior.

ewen-poch:
Addressed in `8545593`.

Successful native compaction now resets the category tracker before reconciling the post-compaction state. If the adapter did not emit a fresh `usage_update`, the stale authoritative total is also removed, so discarded message/tool/skill estimates cannot carry into the next turn.

Added a regression test that seeds a pre-compaction **Messages** category, compacts without a fresh usage update, and verifies that the discarded category is absent when usage resumes. The focused native-compaction tests pass.

ewen-poch:
Addressed in `65fcd78`.

Each prompt turn now checkpoints both the session's category tracker and its last authoritative usage reading before adding preflight estimates. If the Agent prompt throws before a successful stop, the runtime restores both snapshots, removing the failed turn's prompt, tool, assistant, and skill estimates. Once a stop is received, the estimate is committed and is not rolled back by unrelated post-stop cleanup errors.

Added focused coverage for deep checkpoint restoration and an Agent prompt rejection. The targeted tests and Node typecheck pass.

ewen-poch:
Addressed both findings in `26f254f`:

- Tokenizer selection now normalizes an ACP model value to the suffix after its last `/`, so provider-qualified ids such as `openai/gpt-5` and `anthropic/claude-sonnet-4-5` select the same tokenizer profiles as their bare ids.
- Tool accounting now shares a hard 64 KiB / 2,048-node budget across `rawInput`, `rawOutput`, and `content` for each update. Strings, arrays, objects, and content blocks are truncated while being traversed, before tokenization; oversized residual usage remains visible under **Other** through reconciliation with the Agent total.

Added regression coverage for both provider-qualified model families and for bounded traversal/tokenization of a 20,000-item tool payload. All 10 tracker tests and Node typecheck pass.

ewen-poch:
Rebased this branch onto the latest main (a509b07); the new head is c9f4927. The range-diff shows the five feature patches are unchanged, with only the ACP test-event marker omitted because it is already provided upstream by #494. Focused validation passes: Node typecheck and all 13 feature-related tests. The unrelated HTTP-host runtime test still fails identically on main with Error: Server is not running, so no out-of-scope fix was included.

ewen-poch:
Updated the Context Usage popover to match the adjacent composer menus: semantic popover surface, subtle border, rounded corners, menu shadow, tighter spacing, and single-line summaries/diagnostics. Removed the redundant visible explanatory sentence while preserving its detail as hover text, and added consistent focus, active, disabled, and loading states. The focused renderer tests and web typecheck pass.

ewen-poch:
Centered the Context Usage popover on its Gauge trigger and reduced its width from 336px to 288px. Context tokenization now follows the upstream provider model before falling back to the ACP framework, fixing DeepSeek models routed through Claude Code being classified as Anthropic; the upstream model is tracked separately from framework/bridge session models. Unknown model families still use the explicit cl100k fallback, while the Agent-reported total remains authoritative for residual overhead. Focused tracker/runtime tests and the Node typecheck pass.

ewen-poch:
Fixed two concrete sources of the large reconciliation delta. The Claude backend early return now carries the upstream provider model, so DeepSeek sessions select the cl100k fallback instead of incorrectly reporting the Anthropic tokenizer. The tracker also counts enabled app-owned MCP tool schemas as persistent MCP context; the current activity, artifacts, and notebook schemas account for about 3.9k tokens. Claude Code private built-in tools, system injections, and request wrappers remain in the authoritative residual rather than being guessed into a category. Focused tracker, runtime, settings, and MCP regressions pass, together with the Node typecheck and formatting checks.

ewen-poch:
The macOS coverage job exposed framework state leaking from the new settings regression into later tests. The regression now selects Claude Code through repository settings instead of a process-wide environment stub, teardown restores all stubbed environments, and the existing unpinned Codex test declares its framework explicitly. The four-test failure sequence now passes; the full settings test file has no framework-compatibility failures locally.

ewen-poch:
Implemented the review follow-ups in the latest push:

- publish a live local preflight estimate while generation is in progress, then atomically reconcile to the Agent-reported total
- require successful OpenCode model selection before using the requested tokenizer or context window; fallback sessions now use the Agent-reported window
- defer final assistant output until it becomes next-turn input, while committing intermediate assistant text at tool boundaries
- deduplicate raw/display tool results across partial ACP updates and native Skill payloads
- keep cleared usage hidden during deferred provider reconnects and roll back hidden compaction output on cancellation/failure
- keep automatic and manual compaction thresholds gated on reconciled Agent usage
- reuse the Anthropic tokenizer instance instead of rebuilding it for every count

Validation: TypeScript typecheck, focused ESLint, 26 tracker/renderer tests, and 19 targeted runtime regressions pass after rebasing onto the latest main.

ewen-poch:
Addressed the latest correctness findings in :

- Codex Skill documents attached through  are now prompt-scoped and replaced as a set on each turn, so an omitted Skill disappears from the next estimate. If Codex also emits a real  read, that section is promoted to persistent conversation history and remains deduplicated.
- GPT tokenizer matching now covers GPT-4/GPT-5 naming variants including  and , while keeping provider-prefixed ids normalized.
- Tool results use one canonical section:  wins when present and  is only a fallback. The tracker also preserves raw output across later content-only ACP delta updates, preventing mirrored or partial results from being double-counted.

Added tracker and runtime regressions; all 19 tracker tests, the prompt-scoped Codex Skill runtime regression, Node typecheck, and focused ESLint pass.

ewen-poch:
Addressed the latest correctness findings in 7fa0dc1:

- Codex Skill documents attached through open-science/skill-inputs are now prompt-scoped and replaced as a set on each turn, so an omitted Skill disappears from the next estimate. If Codex also emits a real SKILL.md read, that section is promoted to persistent conversation history and remains deduplicated.
- GPT tokenizer matching now covers GPT-4/GPT-5 naming variants including gpt-4.5-preview and chatgpt-4o-latest, while keeping provider-prefixed ids normalized.
- Tool results use one canonical section: rawOutput wins when present and content is only a fallback. The tracker also preserves raw output across later content-only ACP delta updates, preventing mirrored or partial results from being double-counted.

Added tracker and runtime regressions; all 19 tracker tests, the prompt-scoped Codex Skill runtime regression, Node typecheck, and focused ESLint pass.

ewen-poch:
Addressed both prompt-lifecycle findings in 021ea92:

- Preflight usage is now generation-scoped. When a Claude/OpenCode turn stops without a fresh authoritative usage update, the runtime restores the last reconciled Agent reading (or clears the transient estimate if none exists), keeping manual and automatic compaction behavior tied to authoritative usage.
- Prompt rollback now distinguishes failures before any context-bearing session update from partial turns that already streamed assistant, tool, or usage events. Early failures restore the checkpoint; partial turns retain their estimator state while their transient preflight display falls back to the prior authoritative reading.
- Per-turn update markers are cleared on normal completion and all session reset, delete, reconnect, shutdown, and disconnect paths.

Added focused regressions for both cases and adjusted the high-preflight compaction test to assert behavior during generation. Five lifecycle regressions, Node typecheck, focused ESLint, and diff checks pass.

ewen-poch:
Implemented the worthwhile review follow-ups in cfba19a:

- Preserve the first confirmed tool category by session and tool-call ID, so partial result updates, including updates carrying only a generic kind, do not split MCP input and output across categories.
- Recognize provider-qualified IDs for the supported Claude, GPT, ChatGPT-4o, o-series, and Codex families while keeping the documented fallback for unknown model families.
- Estimate the exact Claude persistent system prompt after MCP tool-name rewriting through an explicit SessionSetup field.
- Remove the one-off Hallmark metadata header and document the intentionally shared stateless tokenizer instances.

I kept the existing unknown-model fallback, transparent negative delta, approximate BPE prefix subtraction, and local Skill file reads unchanged: those are deliberate or low-risk trade-offs for a non-billing estimator.

Validation: 46 focused tests passed, Node/Web typechecks passed, focused ESLint passed, and git diff --check passed.

ewen-poch:
Followed up on the latest Codex review in 7ce5096.

The repeated SKILL.md read finding is fixed: a prompt-preloaded document and its first observed read share one section, repeated updates for that tool-call ID replace the same section, and later reads with distinct tool-call IDs are retained as separate history entries. The new mapping participates in tracker checkpoints and session resets. Focused validation passed: 21 tracker tests, 3 runtime Skill/category regressions, Node/Web typechecks, focused ESLint, and diff checks.

I am keeping the OpenCode unconfirmed-model behavior unchanged. When ACP does not confirm the active model, the runtime has no reliable family signal; Anthropic, OpenAI, and cl100k would all be guesses. The explicit cl100k fallback is documented, category values remain labeled as local estimates, and the Agent total remains authoritative. Suppressing estimates in that state would also remove the requested live estimate without making the final total more accurate.

ewen-poch:
I reviewed the prompt-prefix finding and am keeping the current accumulation behavior.

promptPrefix is not sent as an opaque per-request ACP option. The runtime concatenates it into promptText before createPromptContent, and that content is submitted as the actual user turn. Previous user turns, including their individual prefix copies, remain in the Agent conversation until compaction/reset. ContextUsageTracker appends one copy to System and subtracts that same copy from the current Messages count, so this is category reattribution rather than double-counting.

Replacing the section on each turn would remove prefix text that is still present in earlier user messages and would therefore undercount multi-turn context. Claude turn reminders follow the same content path; Claude persistent system appends are accounted separately through SessionSetup.persistentSystemPrompt.

ewen-poch (commented):
## Review

No blocking issues. Two P2 fixes are worth doing before merge, plus one quick dependency check.

---

### P2 — `o[134]` regex misses future o-series models

`tokenizerProfileFor` in `context-usage-tracker.ts`:

```typescript
/^(?:gpt-(?:4(?:[.o-]|$)|5(?:[.-]|$))|chatgpt-4o(?:-|$)|o[134](?:-|$)|codex)/
```

This matches `o1` / `o3` / `o4` but not `o2`, `o5`, `o6`, etc. A misclassified o-series model silently falls back to `cl100k_base` instead of `o200k_base`. The vocabulary has been shared across the family since GPT-4o so the count error is small, but the diagnostics line will display the wrong tokenizer. Suggest:

```typescript
// o[134] → o\d+ to cover future releases
/^(?:gpt-(?:4(?:[.o-]|$)|5(?:[.-]|$))|chatgpt-4o(?:-|$)|o\d+(?:-|$)|codex)/
```

---

### P2 — `beginSession` silently ignores updated `persistentSections` on re-entry

The idempotency guard:

```typescript
if (current && current.profile === profile && current.model === input.model) return
```

exits early without re-applying `persistentSections` or `persistentSystemPrompt`. That's the correct design (callers use `resetSession` when they want a clean slate), but the assumption isn't documented. If `ensureContextUsageTracking` is ever called after MCP tool availability changes without a preceding `resetSession`, the stale sections would silently persist. A short comment on the guard makes the invariant explicit:

```typescript
// Sections and system prompt are applied once at creation; call resetSession to re-apply them.
if (current && current.profile === profile && current.model === input.model) return
```

---

### Confirm — `z.toJSONSchema` requires Zod v4

`context-usage-static-context.ts` calls `z.toJSONSchema(z.object(definition.inputSchema), { target: 'draft-7' })`. That API doesn't exist in Zod v3 (it would throw at the first session start). The tests pass so the project presumably has v4, but worth a quick check on the `zod` entry in `package.json` before merging.

---

### Minor notes (no action needed)

- `@anthropic-ai/tokenizer@^0.0.4` — semver 0.0.x pins exactly to that patch; tokenizer improvements need a manual bump. Intentional for stability, just worth noting.
- The `base64-js` promotion from `devDependencies` to production in the lock file is correct (needed by `js-tiktoken` at runtime).
- The `other` residual category is the right honest choice for unattributed overhead.

---

### What's well-done

The `preflight`/`reconciled` status distinction is the key invariant and it's applied consistently everywhere it matters (auto-compaction guard, preflight restore on turn end, reconnect suppression). The `contextUsageUpdatedPromptTurnsBySession` keying on `promptTurn` is the right signal for the partial-turn vs. clean-rollback distinction. Tool serialization budget (64 KB / 2048 nodes) prevents pathological inputs from stalling the estimator. Test coverage of the edge cases (compaction cancel, provider reconnect, partial-turn failure) is thorough.

ewen-poch:
Reviewed the remaining follow-ups in 4e60225:

- Documented that beginSession applies persistent sections and the persistent system prompt only when creating tracker state; callers must use resetSession to replace them.
- Confirmed package.json directly requires zod ^4.4.3, so z.toJSONSchema is supported.
- Kept the pre-1.0 Anthropic tokenizer dependency, the js-tiktoken/base64-js production dependency shape, and the zero-token bar behavior unchanged; these are intentional or harmless as noted.
- Per project scope, no OpenAI/o-series routing change was made.

This is a comment-only source change. Focused Prettier, ESLint, independent standards/spec review, and git diff --check passed; no test suite was rerun because runtime behavior is unchanged.

ewen-poch:
Addressed the framework-specific MCP schema finding in e2388e5.

contextUsageMcpSections now receives the active framework and serializes the model-facing tool name accordingly:

- Claude Code: mcp__<sanitized-server>__<tool>, preserving the existing baseline
- OpenCode: <server>_<tool>
- Codex: mcp.<server>.<tool>

The framework id is also part of the static-section cache key, so equal availability combinations cannot reuse another framework schema. AcpRuntime passes its active framework id.

Added OpenCode and Codex regressions via red/green TDD. All 5 static-context tests, Node/Web typechecks, focused ESLint, independent standards/spec review, and diff checks pass. The full test suite was left to CI per the requested review workflow.

ewen-poch:
Addressed the Codex Skill-read accounting finding in a3ba12e.

Observed SKILL.md reads now follow the normal bounded tool lifecycle:

- rawInput is attributed to Skills
- rawOutput is the canonical document representation when present
- later content-only projections cannot overwrite a raw-output-backed section
- prompt-preload deduplication, same-tool update replacement, and distinct-tool-call history remain intact

Added a red/green regression covering raw input plus a later content-only projection after canonical raw output. All 22 tracker tests and 5 static-context tests pass, along with Node/Web typechecks, focused ESLint, independent standards/spec review, and diff checks.

ewen-poch:
Addressed the context-usage review points in a9aa65c:

- Preflight now preserves the latest Agent-reported numerator as authoritative. The live local estimate updates independently in `breakdown.estimatedTokens`, and the UI shows it alongside the latest Agent total without presenting a stale delta.
- Sessions with an unknown model window now publish a token-only preflight estimate instead of suppressing the indicator. No context-window size is guessed; percentage display starts only when a reliable window becomes available.
- The Anthropic tokenizer remains a category-level approximation, and the diagnostics now label it `Anthropic approx.`. I did not replace it with a generic BPE tokenizer because that would not provide a better accuracy guarantee; Agent-reported totals remain authoritative.

Validation: focused runtime and renderer tests pass, TypeScript typecheck passes, and ESLint reports no errors (existing repository warnings only).

ewen-poch:
Addressed the Codex bridge MCP naming finding in 7a1cf18.

The static context baseline now uses the actual delivery mode:

- direct Codex MCP keeps `mcp.<server>.<tool>` names
- bridge/Responses-compatibility Codex uses the exported `mcp__<sanitized-server>__<tool>` aliases

The mode comes from the existing `providerConfiguration`-derived runtime signal, and it is included in the static-section cache key so direct and bridged sessions cannot reuse each other's schema baseline. Added both static serialization and runtime regression coverage. Focused tests, Node/Web typechecks, focused ESLint, diff checks, and independent Standards/Spec review pass.

ewen-poch:
Addressed the reused `toolCallId` finding in d2f7de1.

Tool observations and persistent sections are now keyed by a prompt-scoped occurrence (`promptSequence + toolCallId`) rather than the raw ACP id alone. This preserves same-prompt incremental replacement and canonical-output handling, while a later prompt that reuses the same opaque id gets a fresh category/Skill lifecycle and cannot overwrite earlier history. The sequence and occurrence mappings participate in checkpoint/restore and reset behavior.

Added regressions for a result-only reused id changing category on a later prompt and for a repeated Skill read reusing the same id across prompts. All 23 tracker tests, 35 related runtime tests, Node/Web typechecks, focused ESLint, diff checks, and independent Standards/Spec review pass.

ewen-poch:
Fixed the failed Skill-load accounting in 8a59653. Skill payloads are now promoted to persistent Skill documents only after a completed update. Partial updates remain provisional, while failed or cancelled payloads are retained as ordinary tool output and cannot overwrite an existing Skill document. A later successful update removes the provisional projection before recording the canonical document. Added regressions for both native Skill failures and a failed Codex SKILL.md read that follows a pre-counted document.

ewen-poch:
Follow-up in 0b1ae5d: the ACP SDK ToolCall status union does not define a cancelled state, so the unreachable comparison was removed. The resulting behavior is: completed updates may promote a Skill document, failed updates are counted as ordinary tool output, and pending/in-progress updates remain provisional until completion. The focused tracker suite, Node typecheck, ESLint, and diff check pass locally.

ewen-poch:
Confirmed: context-bearing events can continue mutating the local tracker after a reconciled snapshot, while the current guard prevents the rendered breakdown from being recomputed. I am leaving this finding unmodified in the current push. The intended follow-up is to recompute the breakdown against the unchanged authoritative Agent used value for later message/tool events, with a regression covering usage_update followed by tool output without another usage_update.

Issue #14, opened 2026-07-11
[Bug]: 20.9 MB PDF fails with "Request too large (max 32MB)" after file read


theresayao0614-sudo:
### Preflight checklist

- [x] I searched existing issues and this has not been reported yet.
- [x] I'm running a recent version of Open Science.

### What happened?

I uploaded a PDF with a file size of 20,926,462 bytes (20.9 MB as shown in Finder) and asked the agent to analyze it.

Open Science successfully invoked the Read tool and completed conversation compaction, but the run then failed with the following error:

Request too large (max 32MB). Try with a smaller file.

I expected Open Science to process a PDF whose raw file size is below 32 MB. If encoding or request overhead causes the effective request size to exceed the limit, the app should either process the PDF in chunks or reject it before starting with an accurate file-size limit and a clear explanation.

Instead, the agent run fails only after reading the file and completing compaction.

### Steps to reproduce

1. Launch Open Science.
2. Create or open a project.
3. Upload a PDF file named `paper1.pdf` with a size of approximately 20.9 MB.
4. Reference the file with `@paper1.pdf` and ask the agent to analyze part of the paper.
5. Wait for the agent to read the file and complete conversation compaction.
6. Observe that the run fails with: `Request too large (max 32MB). Try with a smaller file.`

### Operating system

macOS (Apple Silicon)

### App version

0.1.1

### Provider / model

deepseek-v4-pro

### Relevant logs or screenshots

```shell

```

ewen-poch:
Thanks for the detailed report — this is a real bug and your diagnosis is correct.

**Root cause**

The `32MB` limit is the **total request-body size** enforced by the agent/provider request layer, measured *after* base64 encoding — not your raw file size. That's why a 20.9 MB PDF trips a "32MB" limit:

- Open Science hands PDFs to the agent as a file reference, and the agent's **Read tool embeds the whole PDF as base64** into the request. base64 inflates payloads by ~33%, so your ~20.9 MB PDF becomes ~28 MB encoded; together with the system prompt, tool definitions, and post-compaction history it crosses the 32 MB ceiling. This is exactly why it only fails **after** the read + compaction step, as you observed.
- Large images had the same class of problem (inlined at full resolution as base64 with no size cap).

**Fix (implemented, landing soon)**

[...]

PR #497, opened 2026-07-29
fix(notebook-runtime): prevent package installs from upgrading R

ewen-poch:
## Problem

Installing an R package through the managed Conda environment could let the solver replace `r-base`, which upgraded the live notebook kernel from R 4.4.3 to 4.5.3. Package mutation was also reachable outside `manage_packages` through R/Python cells, `bash_execute`, and the control REPL, and the provenance UI could imply that every observed package change was caused by the current operation.

## Proposed change

- Pin the complete installed `r-base` version/build, inspect a supervised JSON dry-run, reject any `r-base` LINK/UNLINK action, and verify the complete identity after every real Conda install or uninstall regardless of exit code.
- Fall back to env-scoped CRAN `install.packages` only from a structured non-mutating solver failure; retain the approved plan in the audit log.
- Refuse all `r-base` uninstall MatchSpec forms and quarantine a runtime immediately if its protected identity changes. Quarantine is shared across sessions, blocks execution before the install lock is released, terminates live kernels, and retains recovery evidence if the repair registry cannot be written.
- Keep protected-identity quarantine distinct from interrupted package operations. Only a verified, authorized runtime repair can clear a protected-identity marker, restore persisted bindings, and release all aliases that share the repaired prefix.
- Route notebook, shell, and REPL mutations through one semantic policy, with persistent R/Python/Node runtime backstops for dynamically assembled installer and filesystem mutation calls. On macOS, wrap the complete kernel/shell process tree in a Seatbelt policy that makes the app-owned runtime read-only.
- Validate relocated R homes/libraries before accepting a runtime and describe post-snapshot changes as observations rather than attributing them to the current package operation.

## Scope and non-goals

- This protects app-managed runtimes. External Python installs still require per-runtime authorization; external R package management remains unsupported.
- macOS has the native process-tree filesystem boundary used by the current desktop runtime. Linux/Windows receive the main-process policy plus persistent R/Python/Node runtime guards in this change.
- The Linux/Windows guards cover supported package managers, direct calls, and common dynamically assembled forms. They are not a native hostile-code sandbox and do not attempt to disable general-purpose FFI, fresh-process, parallel-worker, or third-party process-launcher escape techniques such as `ctypes` or `multiprocessing` spawn. Native filesystem sandbox adapters remain follow-up work.
- This does not change runtime selection or automatic kernel restart behavior.

- R package operations cannot plan, apply, remove, or partially replace `r-base` without being rejected or quarantined.
- `install.packages`, Conda, pip/venv, shell, and REPL bypass paths are rejected, including covered dynamically assembled forms; ordinary strings/comments remain allowed.
- Protected-identity quarantine cannot be cleared by an ordinary package install, and verified repair clears persisted and in-memory quarantine consistently across shared-prefix aliases.
- `npm run typecheck` passes.
- Targeted ESLint passes with no errors.
- Full `npm test` passes: 7,762 passed and 136 skipped.
- Python and R loop syntax checks pass, including live dynamically assembled mutation tests for Python `os.truncate` and destination-aware R `file.copy` handling.

## Review focus

- `package-manager.ts`: protected Conda planning, complete identity checks, failure fallback, and R uninstall behavior.
- `runtime-service.ts`, the repair registry, and repair IPC: journal ordering, reason-aware quarantine, shared-prefix repair keys, authorized repair handoff, and fail-closed recovery behavior.
- `managed-runtime-guard.ts` plus the three persistent loop resources: false-positive handling, destination-aware file operations, and dynamic mutation backstops.
- `provisioner-runtime.ts`: relocated R health validation.


ewen-poch:
Addressed the actionable review findings in 35e7bb8:

- Managed-runtime write detection now correlates the runtime path with the actual write call or shell command segment, so unrelated writes are no longer rejected.
- Import-only use of `pip`, `venv`, and `ensurepip` is allowed; package and environment mutations remain blocked.
- The read-only R conda solver probe keeps shared-cache locking but no longer creates an install journal entry; only the approved prefix-writing transaction is journaled.
- `child_process.fork()` is rejected at runtime so a dynamically invoked helper process cannot escape the control-REPL package policy.

Targeted guard, package-manager, and real REPL-loop tests pass, along with typecheck and the REPL script syntax check.

ewen-poch:
Addressed both findings in ff3007c:

- A source-policy rejection now records the failed REPL run without transitioning the live control kernel to `running` or clearing a prior `terminated` state. The rejected-run test now verifies that the kernel remains `idle`.
- Managed-runtime write detection now parses the actual write operands for each supported Python, R, REPL, and shell operation and applies directory-boundary matching. Read-only copies out of the runtime and sibling paths such as `runtime-backup.txt` are allowed, while destinations inside the managed runtime remain blocked.

The focused guard/runtime-service suite passes (174 tests, 1 skipped), and typecheck passes.

ewen-poch:
Addressed the actionable pip-inspection finding in 9aaf34a (rebased head):

- `python -m pip list`, `show`, and `help` are now treated as read-only inspection across the main-process guard and the Python/REPL runtime hooks.
- `python -m pip install`, `uninstall`, and `wheel` remain blocked, including dynamically assembled subprocess commands.
- Added source-policy and real Python/REPL runtime coverage for both allowed inspection and blocked mutation.

The repeated REPL kernel-status finding was already fixed in d9c9a0f: `blockedMutation` is evaluated before the status transition, and `running` is persisted only inside `if (!blockedMutation)`. The regression test asserts that a locally rejected control request leaves `kernelStatus` as `idle`, so no additional state-machine change was needed.

The branch is also rebased onto current `main` (6a5aa8e); the conflict resolution preserves both the new read-only package-inspection coverage and the R runtime quarantine/recovery coverage. Focused tests pass (295 passed, 1 skipped), real REPL tests pass, and typecheck passes.

ewen-poch:
Addressed both P1 runtime-write bypasses in b8bc58a:

- The shell guard now resolves simple literal assignments before scanning, recognizes `>`/`>>` destinations, and blocks both direct and variable-routed writes into `OPEN_SCIENCE_RUNTIME_DIR` while allowing workspace redirections.
- The persistent R loop now guards canonical base write/connection APIs at runtime, including paths supplied through local variables and symlinked parents.
- The control REPL now guards callback, synchronous, and promise-based Node `fs` write APIs at runtime, including write-mode opens, streams, renames, links, and copies. Canonical path checks prevent local-variable and symlink aliases from reaching the managed runtime.

Regression coverage includes direct/append redirections, `target=...; touch "$target"`, temporary-variable R `writeLines`, and temporary-variable REPL `writeFileSync`. The source guard suite passes (53 tests), the full kernel-executor suite passes (46 tests), real REPL runtime cases pass, and typecheck plus R/Node syntax checks pass.

ewen-poch:
Addressed both shell P1 findings in 724462e:

- POSIX shell analysis now tracks `cd` transitions into the managed runtime. Any subsequent relative write target is rejected, while an absolute `cd` back outside the runtime resets the effective-location state.
- Windows shell execution now selects a dedicated PowerShell policy surface instead of the bash parser. It recognizes `Set-Location`, native item/content mutation cmdlets and aliases, copy/move/rename destination semantics, redirections, and simple variable-routed paths involving `$env:OPEN_SCIENCE_RUNTIME_DIR`.

Regression coverage includes `cd "$OPEN_SCIENCE_RUNTIME_DIR" && touch ...`, `Set-Location ...; New-Item ...`, direct `Remove-Item`, and a `Join-Path` temporary variable. Focused guard/executeShell tests pass (76 tests), and typecheck passes.

ewen-poch:
Addressed both P1 findings in 0c4e329:

- POSIX shell assignment analysis now preserves command-substitution expressions before expanding later variable uses. This keeps the computed runtime-root dependency attached to write destinations, so the reported `root=$(printf ...); touch "$root/..."` bypass is rejected while unrelated workspace writes remain allowed.
- The control REPL now guards `fs.mkdtemp`, `fs.mkdtempSync`, and `fs.promises.mkdtemp` at runtime. The main-process source policy recognizes the same APIs, with callback, synchronous, and Promise regression coverage.

The managed-runtime guard suite passes (64 tests), the filtered real REPL runtime guard cases pass (4 tests), and typecheck plus the REPL script syntax check pass.

ewen-poch:
Addressed the actionable integrity findings in `7e0e56c`:

- Re-check the repair gate after acquiring the run lease, so a cell queued behind an install cannot start after that install quarantines the environment.
- Reject r-base-changing install plans without a CRAN fallback. R removals now require a structured dry-run first, and only use `remove.packages` for a no-mutation package-not-found result.
- Verify the complete r-base identity immediately after every real Conda spawn, including the first MAX_PATH attempt before any retry.
- Quarantine all Python and R bindings that resolve to the same managed prefix, including the in-memory fallback when the repair registry cannot be written.
- Handle PowerShell call-operator installers and .NET file writes, while binding content-cmdlet path arguments so runtime text used only as content remains allowed.
- Inspect interpreter execution payloads before dispatch and add runtime backstops for Python `posix_spawn` and `exec`, R `system` and `system2`, and REPL child processes.
- Reject Worker isolates until an OS-enforced runtime boundary can be inherited; allowing a fresh isolate would reload unpatched built-ins.
- Block dynamically assembled cross-language package mutations at runtime, including Python launching R installers and R launching pip, Conda, Python environment, or R installer commands.

Focused verification: 324 tests passed and 1 skipped across the package manager, runtime service, source guard, and kernel executor. The six real REPL runtime cases, typecheck, targeted ESLint, and Node/Python/R syntax checks also pass.

ewen-poch:
Addressed the two nested-shell findings in `b0b724c`:

- Package-mutation detection now recursively rescans interpreter payloads with their actual language surface. This blocks shell-launched Python pip/venv APIs and Node child-process installers while preserving read-only nested inspection such as `pip list`.
- Windows shell analysis now parses `cmd.exe /c` payloads before dispatch. It covers copy/move/delete/create operations plus `rmdir`/`rd` and `mklink`; both mklink endpoints are protected so a persistent alias into the managed runtime cannot be created.
- Nested runtime-write payloads are recursive with a bounded depth, so chained interpreter or shell delegation cannot bypass the same checks.
- Added positive and negative coverage for Python and Node payloads, cmd copy/mkdir/rmdir/mklink writes, safe pip inspection, and copies out of the runtime.

Focused guard tests pass (81/81), along with typecheck, targeted ESLint, and diff checks.

ewen-poch:
Addressed the actionable inherited-cwd finding in `acb6c04`:

- POSIX and PowerShell sequence analysis now carries an inherited managed-runtime cwd into nested `sh -c`, `powershell -Command`, and `cmd /c` payloads, with bounded recursive checks. Relative child writes are rejected after the parent enters the runtime.
- Execution-bridge parsing now distinguishes argv from shell-command strings. A harmless argument such as `subprocess.run(["echo", "python -c ..."])` is no longer treated as executable nested code, while direct package-manager argv and real shell-string payloads remain blocked.
- Added corresponding Bash, PowerShell, Python, R, and REPL regression cases.

The reported missing `--json` on R uninstall does not reproduce. `removeArgv()` already includes `--json`; inserting `--dry-run` preserves it, and the existing uninstall regression asserts the exact argv as `remove --dry-run --json ...`. The focused package-manager and guard suites pass (152/152), along with typecheck and targeted ESLint.

ewen-poch:
Addressed both findings in `2c17c48`:

- Runtime write targets are now resolved relative to the notebook cwd and canonicalized through the nearest existing ancestor. Writes through an existing POSIX symlink or Windows junction into the managed runtime are rejected.
- POSIX `ln` now protects both source and destination, matching the existing `mklink` treatment. This also blocks creating a runtime alias and writing through it later in the same shell request, before the alias exists during pre-dispatch analysis.
- Managed execution now checks repair markers under the canonical env name, the raw managed interpreter path, and its realpath. Unbound default sessions therefore honor legacy runtimeId-keyed markers both before waiting and again after acquiring the run lease.

Focused guard/runtime-service tests pass (229 passed, 1 skipped), the updated guard-only suite passes (88/88), and typecheck, targeted ESLint, and diff checks pass.

ewen-poch:
Addressed the latest findings in commit 7fc5720:\n\n- Runtime cwd inheritance now applies to every nested payload surface, not only child shells. Python, R, and Node relative writes are rejected after the parent shell enters the managed runtime.\n- Successful managed repairs now clear every alias consulted by repairRegistryKeys: environment name, raw interpreter path, canonical realpath, and loaded managed binding IDs across Python/R.\n- Added Bash/PowerShell nested interpreter coverage and an unbound legacy-marker repair regression that verifies execution is blocked before repair and restored after repair.\n\nFocused guard/runtime-service tests pass (233 passed, 1 skipped), along with typecheck, targeted ESLint, and diff checks.

ewen-poch:
Addressed the CodeQL findings in commit bdea13e:\n\n- Fixed js/redos by making the command-substitution matcher unambiguous: escaped characters and ordinary characters can no longer consume the same backslash. The adversarial 26-pair input dropped from about 1.39 seconds to 0 ms locally.\n- Investigated js/shell-command-injection-from-environment through the SARIF trace. Its source is process.execPath; protectManagedRuntimeWrites never interpolates that value into shell text. It remains one spawn argv element behind a constant sandbox-exec executable on macOS, and remains the executable argument with shell disabled elsewhere. Alert #14 was therefore dismissed as a documented false positive.\n- Added regression coverage for both the ReDoS input and executable paths containing spaces and shell metacharacters.\n\nThe managed-runtime guard suite passes (93/93), together with typecheck, targeted ESLint, and diff checks.

ewen-poch:
Addressed the latest runtime-guard review findings in commit b453d7f:\n\n- REPL link/linkSync and promise variants now validate both endpoints. Symlink sources are resolved relative to the link location before validation, so relative aliases cannot point back into the managed runtime.\n- R file.link and file.symlink are guarded at both the main source-policy seam and the live R kernel. Relative symlink sources use the same destination-relative resolution.\n- REPL child-process enforcement now classifies actual write targets from resolved argv across shell, PowerShell/cmd, and inline Node payloads. Copying files out of the runtime and logging the runtime path before writing to external files remain allowed, while runtime destinations and link endpoints stay blocked.\n- Python pyvenv.cfg inspection is allowed in read-only mode; create/write modes remain blocked.\n\nVerification: the complete managed-runtime guard and REPL integration suites pass (121/121), the focused Python and R kernel regressions pass, and typecheck, targeted ESLint, syntax checks, and diff checks are clean.

ewen-poch:
Addressed the latest high-value review findings in 2b12708:\n\n- Python and R child-process guards now resolve write endpoints from argv instead of rejecting any command that merely reads or prints the managed runtime path. Copy-out and unrelated workspace writes remain allowed; managed-runtime destinations remain blocked.\n- Data-cell source-policy and other pre-executor rejections no longer publish a false running kernel state. Post-lock preflight failures settle back to idle, while an actual terminated state is preserved.\n- Added regressions for Python subprocess copy-out/workspace writes, R system2 copy-out/workspace writes, blocked copy-in, and pre-execution kernel status.\n\nValidation: runtime-service 143 passed / 1 skipped; Python loop 5 passed; R loop 35 passed; real Python/R mutation-policy tests 5 passed / 41 skipped; typecheck and targeted ESLint passed.

ewen-poch:
Addressed both blocking findings in 8dd5b86:\n\n- PowerShell command extraction now decodes UTF-16LE Base64 payloads for -EncodedCommand and its supported short forms (-e/-ec/prefix forms) before package/runtime policy analysis. The -c command alias is covered as well.\n- The REPL child-process backstop now carries the resolved cwd across shell and PowerShell command sequences, including nested shell, PowerShell, cmd, Node, Python, and R payloads. Relative writes after cd/Set-Location into the managed runtime are blocked, while explicit absolute workspace destinations remain allowed.\n- Added regressions for encoded PowerShell writes/package installs, shell and PowerShell relative writes after entering the runtime, short command aliases, and allowed external destinations after the cwd transition.\n\nValidation: managed-runtime guard 99 passed; complete REPL integration suite 28 passed; typecheck, targeted ESLint, syntax, formatting, and diff checks passed.

ewen-poch:
Addressed the fifth and final review batch in 42c7128:\n\n- Managed repair quarantine now scopes in-process gates correctly: managed runtimes are keyed by Conda environment, while external runtimes are keyed by runtimeId. Managed default quarantine skips external bindings that only share the executor's default env key, leaving their binding and execution available.\n- Main-process shell and PowerShell scanners now resolve every cd/Set-Location transition, including relative exits and runtime subdirectories. Relative writes are blocked while the resolved cwd remains managed and allowed after it actually leaves.\n- Shell literal expansion now preserves quoted nested payload boundaries instead of promoting their separators into the parent sequence; the REPL cwd resolver received the same subdirectory-safe path handling.\n- Added regressions for an unrelated external R binding surviving managed quarantine, direct/nested relative exits, runtime-subdirectory traversal, and REPL shell/PowerShell subdirectory writes.\n\nValidation: runtime-service 144 passed / 1 skipped; managed-runtime guard 104 passed; complete REPL integration 30 passed; typecheck, targeted ESLint, Node syntax, formatting, and diff checks passed. This is the configured fifth/final review batch, so further review feedback will be left for explicit follow-up.

ewen-poch:
Addressed the three repair-boundary findings in 4575f87:

- Repair state now records interrupted-install separately from protected-identity-change. A protected interpreter change is refused before package-manager execution, so an unrelated package install cannot adopt the changed r-base as a new baseline or clear quarantine. Interrupted installs remain repairable by a completed reinstall; ambiguous legacy managed markers fail closed.
- A successful explicit UI Repair now calls back into the runtime service only after the provisioner rebuild and verification complete. That callback clears durable aliases and process-local gates, restores all matching in-memory bindings, and persists their active state. A failed rebuild never releases quarantine.
- The Python loop now guards pip command creation plus InstallCommand and UninstallCommand execution methods, closing dynamic in-process command-factory bypasses while leaving inspection commands available.

Validation: 262 targeted tests passed with 1 skipped; node and web typecheck passed; targeted ESLint and Python loop syntax checks passed.

ewen-poch:
Reviewed the latest Codex findings against the current head; neither requires a code change:

- The R uninstall preflight already receives JSON output. removeArgv always includes --json, and the preflight construction inserts --dry-run without removing it. The focused test routes a conda-managed R uninstall and asserts the exact argv contains both --dry-run and --json; it passes.
- Unbound managed-default sessions are already matched. When binding is undefined, binding?.source !== external evaluates to true, and resolveRunEnv returns the managed default env. The existing regression creates one bound and one unbound R session, asserts two default-r kernel terminations, and verifies execution is repair-blocked in both; it passes.

These comments appear to have read the optional-chain comparison and the shared removeArgv builder incorrectly, so I am leaving the implementation unchanged.

ewen-poch:
Addressed the high-value Windows test portability points in 5a4eccb:

- The macOS sandbox-profile assertion is now explicitly skipped on Windows, where its POSIX path expectation is not meaningful.
- The real Python mutation-policy suite is now explicitly POSIX-only because it exercises `/bin/sh` and `os.posix_spawn`.

I kept the change test-only and avoided rewriting the shared path fixtures: the Windows run showed that the mutation-detection cases already pass. The existing runtime-alias test already creates a Windows junction, and the R uninstall preflight already retains `--json` from `removeArgv` (covered by the exact argv test).

Validation: targeted ESLint passed; the two affected test files passed with 150 tests.

ewen-poch:
Addressed the three repair-scope findings in 2e5940c:

- Interrupted managed installs now use language-scoped repair markers, so a successful Python install cannot clear or reactivate an interrupted R runtime in the same Conda prefix, and vice versa.
- Install journal records now persist the exact repair reason. If the repair registry write fails after a protected interpreter change, startup recovery recreates the protected-identity quarantine instead of downgrading it.
- Successfully removing an agent-created named environment now clears its durable and process-local quarantine only after prefix deletion succeeds, making remove/recreate the supported recovery path without weakening failed-removal behavior.

Regression tests cover shared Python/R prefixes, registry-write failure across restart, and named-environment remove/recreate recovery. Validation: 246 passed, 1 skipped; typecheck and targeted ESLint passed.

ewen-poch:
Addressed the R runtime write-API finding in 816508d:

- The persistent R loop now guards file.append, download.file, write-mode fifo connections, and pipe-launched shell commands. The utils::download.file namespace and attached-package bindings are both patched, so dynamically resolved calls use the same runtime boundary.
- The main-process scanner recognizes the same APIs and uses operation-specific destination arguments. In particular, read-only pipe commands that reference managed-runtime files remain allowed, while pipe commands that write there are rejected.
- Regression coverage exercises all four APIs through temporary R variables against OPEN_SCIENCE_RUNTIME_DIR and verifies that no target is created.

Validation: full suite 7,759 passed / 136 skipped; typecheck, targeted ESLint, R syntax, formatting, and diff checks passed.

ewen-poch:
Addressed the two high-value follow-up findings in `1f1ed32`:

- Python's persistent audit hook and main-process policy now reject `os.truncate` targets inside the managed runtime, including dynamically resolved calls. Regression coverage verifies that an existing runtime file is not truncated.
- R `file.copy` now uses destination-aware checks in both the main-process scanner and persistent R loop. Copying out of the runtime remains allowed; copying into it is blocked.

I intentionally did not ban `ctypes`, `multiprocessing`/`fork`, or third-party R process launchers. Those findings require deliberate evasion or would disable legitimate FFI, parallel, and subprocess workflows; Linux/Windows do not claim a native hostile-code sandbox in this PR. Dynamically assembled direct `BiocManager`/`renv`/`pak`/`remotes`/`devtools` installer calls remain covered by the existing source and R AST policies.

Full local validation passed: 7,762 tests passed and 136 skipped; typecheck, targeted ESLint, Python/R syntax, formatting, and diff checks also passed.

ewen-poch:
Addressed the four high-value guardrail findings in `7a3c031`:

- A protected-identity journal update failure no longer skips quarantine. The in-process repair gate is published, live kernels are terminated, and the independent repair registry is attempted before the install lock is released; the original journal is retained as recovery evidence when its update fails.
- Conda protection is now based on the target prefix rather than the request language. Python-surface installs and removals targeting a shared named environment with `r-base` use an exact version/build pin where applicable, require a structured dry-run, reject any `r-base` LINK/UNLINK plan, and verify the full identity after every real spawn. Direct `r-base` removal MatchSpecs are rejected. Python-only prefixes keep their existing single-transaction path.
- The shell policy now recognizes path-qualified installer executables, open command-substitution boundaries, and POSIX assignment prefixes, with a regression test ensuring an echoed path-qualified command remains allowed.
- The persistent R loop now rejects the reproduced undocumented `.Internal(system(...))` process primitive. This is intentionally narrow: supported `system`/`system2` calls remain under the existing guard, and this change does not attempt to turn the loop policy into a general native-code sandbox.

Validation: 385 targeted tests passed (1 skipped); TypeScript typecheck, targeted ESLint, R syntax parsing, formatting, and `git diff --check` passed. The complete CI matrix is green on Ubuntu, macOS, and Windows, and CodeQL passed for Actions, Python, and JavaScript/TypeScript.

ewen-poch (commented):
## Review summary

This PR fixes a serious R upgrade vulnerability: on certain Conda solver paths, `manage_packages` could replace `r-base` 4.4.3 with 4.5.3, thereby **silently upgrading** the active kernel while leaving the user with namespaces and DLLs loaded from 4.4.3. The fix also hardens package-write and runtime-filesystem-write protections across notebook, shell, and REPL execution paths. The overall direction is correct, but the following issues need a response:

### Critical issues (must address)

**1. `managed-runtime-guard.ts` has a ReDoS risk flagged by CodeQL.**

GitHub Advanced Security reported an alert at line 280 in `resolveShellLiteralAssignments`:

```ts
/(^|[;\r\n]|\s)([A-Za-z_][A-Za-z0-9_]*)=("(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*'|\$\((?:\\.|[^)\\])*\)|[^\s;]+)/gmu
```

In the final `\$\((?:\\.|[^)\\])*\)` group, handling of `\` is ambiguous between `\\.` and `[^)\\]`: `\\.` always consumes one character, while `[^)\\]` fails because `\` is excluded from the character class. Combined with the outer `\s` branch, input such as ` A=$( \(((\(...` can trigger exponential backtracking. This regex is applied to untrusted notebook/REPL user input. Suggested options:

- Split the parser: identify assignments (`=` plus a known literal form) first, then process `$(...)` substitutions separately.
- Add an input-length cutoff before invoking it, for example routing `source.length > 65536` through a fallback.
- Replace the regex-based `replace` with a cheaper string scan after expansion.

The similar `resolvePowerShellLiteralAssignments` implementation around line 657 also uses an adjacent replacement pattern, but its assignment capture is narrower and should not exhibit the same backtracking. It is still worth confirming.

**2. `operation-journal.ts` removed the `repairReason` field while callers still write it.**

This is an **important breaking change**: `RuntimeOperationRecord` previously contained `repairReason?: 'interrupted-install' | 'protected-identity-change'`. After its removal, the operation journal no longer carries the reason; the implementation instead relies on the separate registry at `repairRegistryPath` in `runtime-paths.ts` plus the process-local `repairBlockedEnvs` set.

However, `runtime-service.ts:2922` still calls:

```ts
await journal.update(operationId, {
  runtimeId: repairRuntimeId,
  repairReason: 'protected-identity-change'
})
```

`repairReason` has been removed from `RuntimeOperationRecord`, and the corresponding validation was removed from `isOperationRecord`, so this currently passes TypeScript because `journal.update` accepts a broad update parameter. This creates a risk of silently losing the reason:

- On startup recovery, `repairReason` fields in older journal records are ignored.
- `addRepairRequired(..., 'protected-identity-change')` still writes the active reason correctly to the registry.
- However, the semantics of upgrading an operation journal entry to a protected reason are lost, leaving only the registry plus the `REPAIR_QUARANTINE_FAILED` label.

Recommendation: either retain `repairReason` (which also makes recovery decisions clearer) or tighten the type accepted by `journal.update`. Retaining the field is preferable.

### Minor suggestions

**3. UI copy change.**

At `ArtifactProvenancePanel.tsx:1159`, "Observed since the previous snapshot (not attributed to this operation)" was shortened to "Other observed changes". The new version is more compact, but it loses the important meaning that these changes are **not attributed to the current operation**. That distinction is one of the PR's stated guarantees: "describe post-snapshot changes as observations rather than attributing them to the current package operation."

Suggested wording:

> Other observed changes (not attributed to this operation)

**4. `kernel-executor.ts` removed the `sandbox-exec` wrapper.**

The previous implementation wrapped the `python_loop.py`, `r_loop.R`, and `repl_loop.js` spawn calls with `protectManagedRuntimeWrites`; this PR removes that wrapper. The rationale is reasonable: persistent loops now protect themselves through audit hooks and binding replacement, making the external wrapper redundant. However, please confirm that the **complete kernel/shell process-tree boundary** still holds on macOS. `runtime-service.ts:792` now invokes `protectManagedRuntimeWrites` only for temporary `bash_execute` spawns, while data kernels are no longer wrapped.

The PR description says, "On macOS, wrap the complete kernel/shell process tree in a Seatbelt policy." That statement does not match the current diff: shells are wrapped, but data kernels are not. Either update the description or restore the wrapper in `kernel-executor`.

**5. Fail-closed handling of `legacy-unknown`.**

`runtime-paths.ts:415` treats `legacy-unknown` as a protected-identity reason, which is the correct fail-closed behavior. However, please verify the upgrade experience: if old journal entries containing `repairReason` are ignored after an app upgrade, do affected runtimes move from `interrupted-install` to `protected`? If so, an application upgrade could put every runtime into quarantine.

If this is intentional, document it in upgrade notes or a startup banner. If it is a regression, startup recovery should skip the journal reason and derive the state directly from the on-disk `addRepairRequired` registry.

### Validation results

- TypeScript: local `tsc --noEmit` completed without errors.
- Full test suite (excluding the worktree): `Test Files 528 passed | 10 skipped`; `Tests 7774 passed | 136 skipped`.
- Unit tests for files directly modified by the PR all passed: package-manager (72), managed-runtime-guard (new file), runtime-paths, env-ipc, kernel-executor, operation-journal, runtime-service (149 passed + 1 skipped), provisioner-runtime (19), and ArtifactProvenancePanel.render.
- Loop integration tests with `RUN_KERNEL=1` were not run locally because they require a real R environment.

### Conclusion

Please prioritize **#1 (ReDoS)** and **#2 (runtime consistency after removing the journal field)**. #4 is a mismatch between the description and implementation and can be aligned quickly. #3 is a UI copy decision, and #5 concerns the upgrade-path user experience.

ewen-poch:
Addressed the two confirmed high-value findings in `ad98b22`:

- `RuntimeOperationJournal.update()` now reads `readState()` directly and fails closed with `RUNTIME_JOURNAL_CORRUPT` instead of converting a corrupt in-flight journal to an empty pending set. This makes the existing `managePackages` error path retain recovery evidence and establish quarantine before releasing the install lock. The regression test reproduces the exact `interrupted-install` to `protected-identity-change` transition against a journal corrupted after `begin()`.
- Static process bridges now apply the existing directional write-target policy to ordinary shell-command strings from Python/R/Node and to Python subprocess argv. This blocks the reported `os.system("touch ...")` and destination-side `subprocess.run(["cp", ...])` forms on Linux/Windows without treating reads, `echo`, or copying *from* the runtime as writes. Dynamic native/FFI execution remains outside the declared hostile-code sandbox boundary. The platform comments now state that native Seatbelt protection is macOS-only.

I also verified the other reported points against the current head:

- The managed-runtime alias test explicitly creates a Windows `junction`, which does not require Developer Mode; the real Python-loop symlink test is inside `describe.skipIf(process.platform === "win32" || !python3)`. No additional Windows skip is needed.
- R uninstall already receives `--json` from `removeArgv()`, and its regression test asserts the complete `--dry-run --json` argv before allowing the real remove. It does not reject ordinary R package removals.
- The shared Unix-looking path is a lexical detector fixture, while filesystem-backed alias tests use `tmpdir()` and explicit Windows path cases cover Windows normalization. The current cross-platform suite has been green, so changing the fixture would add churn without improving the contract.
- Existing focused tests cover in-process gating, lock-acquisition rechecks, live-kernel termination, retained evidence on registry failure, and cross-session recovery from the retained journal.

Local validation: 167 journal/guard tests passed, 5 focused quarantine lifecycle tests passed, the R uninstall argv test passed, and TypeScript typecheck, targeted ESLint, formatting, and `git diff --check` passed.

ewen-poch:
Addressed the high-value findings in `5aae5a1`:

- Python Conda operations targeting `default-r` now read and fully pin the installed `r-base` identity, use the structured JSON dry-run, and retain the shared post-operation identity check. `default-python` remains the only default prefix that can skip the R identity lookup.
- A missing repair registry is still treated as empty, but an unreadable or malformed registry now fails closed for every runtime. Reads and mutations report `RUNTIME_REPAIR_REGISTRY_CORRUPT`, and the corrupt evidence is preserved instead of being overwritten.
- The persistent Python, R, and Node guards now classify the actual executable rather than scanning arbitrary argv text. Read-only commands such as `echo "pip install pandas"` are allowed, while real installer invocations remain blocked.
- Added descriptor/permission protection for Python `os.fchmod`, R `Sys.chmod`, and Node `fs.fchmod*`, including descriptor-to-path tracking in the Node loop.
- Strengthened the unterminated command-substitution regression with the reported `A=$( \(\(\(...` shape at 256 repetitions. It completes in milliseconds; the assignment regex alternatives are disjoint, so no parser rewrite was warranted.

The `repairReason` journal field and its validation, the full "Observed since the previous snapshot (not attributed to this operation)" UI text, and the complete kernel process-tree Seatbelt wrapper are all present at the current head. `legacy-unknown` remains intentionally fail-closed because weakening it would reopen the upgrade-boundary risk.

Validation: 284 focused tests passed, Node/Web typecheck passed, targeted ESLint passed with no errors, all three loop resources passed syntax checks, and `git diff --check` passed.

ewen-poch:
The macOS failure was caused by three failure-path fixtures becoming invalid under the new fail-closed repair-registry semantics, not by a production regression. Fixed in `c990d42`:

- The quarantine-persistence tests now create the unwritable registry target only after the simulated installer reports an identity change. This preserves a healthy pre-install gate and exercises the intended post-install quarantine failure.
- The journal write-failure test now keeps the repair registry readable and makes only the runtime directory non-writable after recovery. It is explicitly skipped on Windows because chmod permissions are not portable there.

Validation: all 150 `runtime-service.test.ts` cases completed with 149 passed and the Windows-only chmod case skipped; targeted ESLint and `git diff --check` passed.

PR #501, opened 2026-07-29
fix(security): harden DOI and trial query handling

ewen-poch:
## Problem

CodeQL reports exponential regular-expression behavior in DOI suffix cleanup and incomplete string escaping in ClinicalTrials.gov phrase filters. Crafted inputs can consume excessive CPU or change the intended advanced-query structure.

## Proposed change

- Replace the DOI suffix regex with linear-time character trimming while preserving the accepted suffix character set.
- Escape backslashes before quotes when constructing ClinicalTrials.gov phrases.
- Add Python and Vitest regression coverage for an 8,000-character adversarial DOI suffix and an injection-shaped sponsor phrase.

## Scope and non-goals

- No dependency or lockfile updates; Dependabot alerts are intentionally out of scope.
- No web-service error response changes. Code-scanning alert 3 was separately dismissed as a false positive because the response exposes only Error.message or String(error), never Error.stack.

- Adversarial DOI cleanup completes within the regression budget and preserves prior trimming semantics.
- ClinicalTrials.gov phrases retain literal backslashes and escaped quotes without breaking the phrase boundary.
- `npm run typecheck` passes.
- `npm run lint` passes with 0 errors; 24 existing warnings remain outside this change.
- Targeted tests pass: 2 files, 16 tests.
- Full `npm test` passes: 528 files and 7,601 tests; 10 files and 113 tests skipped.

## Review focus

- Equivalence of the `rstrip` character set to the former regex cleanup.
- Backslash-before-quote escaping order in ClinicalTrials.gov advanced filters.
- Regression tests' ability to distinguish the vulnerable implementations.

### DIFF
--- a/resources/skills/literature-review/kernel.py
+++ b/resources/skills/literature-review/kernel.py
@@ -435,7 +435,7 @@ def extract_dois(text: str) -> list[str]:
         d = m.split("</")[0]
         if d.count("<") != d.count(">"):
             d = d.split("<")[0]
-        d = re.sub(r"(?:\*\*|__|[_\]\*>`,;:])+$", "", d)
+        d = d.rstrip("_]*>`,;:")
         if d.endswith("."):
             d = d[:-1]
         while d.endswith(")") and d.count("(") < d.count(")"):
--- a/resources/skills/literature-review/kernel.py
+++ b/resources/skills/literature-review/kernel.py
@@ -435,7 +435,7 @@ def extract_dois(text: str) -> list[str]:
         d = m.split("</")[0]
         if d.count("<") != d.count(">"):
             d = d.split("<")[0]
-        d = d.rstrip("_]*>`,;:")
+        d = d.rstrip("*_]>`,;:")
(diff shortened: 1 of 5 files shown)


ewen-poch:
Thanks for flagging the Markdown suffix case. `*` is already part of the `rstrip("_]*>`,;:")` character set (between `]` and `>`), and `extract_dois("**10.1234/foo**")` returns `["10.1234/foo"]`; the reported regression does not reproduce. I added an explicit regression test in a8e268b so this contract is clear and remains covered, without changing the linear-time implementation.

ewen-poch:
The `*` was already present in the trim set, but 189e235 reorders the identical characters to `rstrip("*_]>`,;:")` so the asterisk is explicit at the start. This preserves the linear-time implementation and prior suffix semantics. The focused Markdown case `extract_dois("**10.1234/foo**") == ["10.1234/foo"]`, the 8,000-character timing regression, and the Vitest adapter all pass.

PR #502, opened 2026-07-29
fix(windows): resolve PowerShell outside PATH

ewen-poch:
## Problem

Windows can retain PowerShell 5.1 at its standard SystemRoot location while omitting system directories from PATH. Managed runtime cache ACL preparation then fails with `spawnSync powershell.exe ENOENT`, blocking default-python environment updates.

## Proposed change

- Resolve Windows PowerShell 5.1 from `SystemRoot`, with `WINDIR` fallback.
- Reuse the absolute executable for managed cache ACL writes and reads, plus notebook shell commands.
- Preserve bare `powershell.exe` lookup only when both Windows root variables are unavailable, so existing process-level error handling remains intact.
- Invoke NSIS cleanup through `$SYSDIR`.
- Add regression coverage for PATH-independent resolution and each affected call site.

## Scope and non-goals

- Does not mutate PATH.
- Does not add a PowerShell 7 or `pwsh` fallback.
- Does not relax the cache ACL trust policy.

- PATH-independent resolution uses `SystemRoot` and `WINDIR` fallback.
- Stripped environments without either root variable retain the existing PATH lookup behavior.
- Cache ACL and shell invocations use the absolute executable when the Windows root is available.
- Installer cleanup uses `$SYSDIR`.
- `npm run typecheck` passes.
- `npm run lint` passes with 0 errors; 24 pre-existing warnings remain.
- `npm test` passes: 529 files and 7,603 tests.

## Review focus

- Windows root fallback and fail-closed ACL behavior.
- NSIS quoting for the absolute executable.

### DIFF
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -1,5 +1,5 @@
 !macro customUnInstall
-  nsExec::ExecToLog 'powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -File "$INSTDIR\resources\windows-runtime-cache-uninstall.ps1"'
+  nsExec::ExecToLog '"$SYSDIR\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -NonInteractive -ExecutionPolicy Bypass -File "$INSTDIR\resources\windows-runtime-cache-uninstall.ps1"'
 !macroend
 
 !macro customInit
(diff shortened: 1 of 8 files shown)


ewen-poch:
Addressed the [P2 finding](https://github.com/aipoch/open-science/pull/502#issuecomment-5119472844) in 1f70628.

The resolver still uses the absolute `SystemRoot`/`WINDIR` PowerShell path whenever either variable is available. If both variables are absent, it now falls back to `powershell.exe`, preserving the existing process-launch and structured error-handling path. A regression test covers the stripped-environment fallback.

Validation completed: targeted Windows regression tests (128 passed, 1 skipped), full typecheck, lint (0 errors), and the full suite (529 files / 7,603 tests passed).

PR #505, opened 2026-07-29
fix(persistence): surface startup and session storage failures

ewen-poch:
## Problem

A failed Settings IPC left the renderer waiting forever behind a null App result. Session loading and saving failures were also reduced to console warnings, so unreadable or damaged data could look like an empty conversation list and unsaved changes could appear durable.

## Proposed change

- Give Settings startup explicit loading and error state, StrictMode-safe request deduplication, latest-attempt-wins retries, and a full initialization retry.
- Carry renderer-safe Session load diagnostics across IPC, quarantine deterministic corruption (including valid JSON with no authoritative conversation structure), and treat partial scans, post-enumeration Session/Project disappearance, and startup reconciliation/index failure as incomplete recovery.
- Separate readable hydration from durable write readiness so healthy conversations remain navigable while saver-backed mutations stay blocked; if Project deletion recovery fails, use an explicit read-only Session load that cannot reconcile, quarantine authority, or write derived state.
- Surface failed Session and manifest writes, retain failed targets, retry the latest in-memory state, and automatically clear obsolete targets even when a queued save rejection arrives after authoritative deletion.
- Preserve deep-link, notification, and retry-selection intent across recovery without overriding later user navigation; concurrent notification peeks are serialized against causal click generations and navigation revisions, explicit Project/Session deletion cancels deferred targets, recovery shells suppress stale unread acknowledgement, and explicit empty or missing retry selections apply atomically and are written back to the manifest before persistence becomes ready.
- Keep a Project deep link pending while the Project catalog itself is unavailable, and only resolve or reject it after a successful catalog load.
- Keep target-scoped Project and Session deletion available after ordinary partial hydration, but disable both when Project deletion-journal recovery itself failed; independently show an exact Session count only when the catalog is complete.
- Keep interrupted-Session Resume disabled until durable Session persistence is ready, with both a page-level guard and a disabled shared Button so reconnect/re-send cannot mutate partially recovered state.
- Keep automatic completed-job analysis disabled until durable Session persistence is ready, cancel pending scans and turn-end listeners when readiness is lost, and preserve one trigger across runtime callback identity changes.
- Render recovery surfaces with the App Shell's semantic design tokens and shared shadcn Buttons, keep raw main-process Settings, Session load, and Session write failures out of renderer-visible copy, and let users dismiss persistent quarantine warnings without changing recovery readiness.

## Scope and non-goals

This change is limited to Settings startup and Session persistence health/recovery behavior across main and renderer, including navigation intent and mutation gating while recovery is incomplete. It does not change the Session file schema or automatically reconstruct unrecoverable conversation content.

Final verification statement: checks affected by the last material edit, both typechecks, full lint, and the full regression suite were rerun against commit `dae93d6`. Earlier behavior-specific results remain listed below; all files affected by the final edit pass the final focused run. Independent Standards and Spec reviews of `git diff 24b4e8e...dae93d6` found no remaining findings. Clean-checkout macOS, Ubuntu, Windows, CodeQL, and commit-message CI all pass; the local full-suite caveat is recorded explicitly below.

| Behavior | Command | Final result |
| --- | --- | --- |
| Settings startup shows progress/error, deduplicates StrictMode loads, and retries the full initialization | `npm test -- --run src/renderer/src/stores/settings-store.test.ts src/renderer/src/App.test.tsx` | PASS — 2 files / 93 tests |
| Session authority quarantines structural corruption, distinguishes direct absence from post-enumeration disappearance, and degrades to read-only hydration when Project deletion recovery fails | `npm test -- --run src/main/session-persistence/coordinator.test.ts src/main/session-persistence/repository.test.ts src/main/session-persistence/ipc.test.ts` | PASS — 3 files / 108 tests |
| Hydration/readiness gates, atomic retry selection, latest-state write retry, authoritative deletion, and late obsolete failure-target cleanup remain consistent | `npm test -- --run src/renderer/src/lib/session-persistence/session-persistence.test.ts src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/stores/session-store.test.ts src/renderer/src/pages/home/HomePage.persistence.test.tsx src/renderer/src/pages/workspace/WorkspacePage.draft-preservation.test.tsx src/renderer/src/pages/workspace/WorkspaceSidebar.render.test.tsx src/renderer/src/pages/workspace/workspace-components.test.tsx` | PASS — 7 files / 158 tests |
| Project deletion stays target-scoped after partial hydration, while partial scans and quarantined Session files suppress misleading exact counts without treating manifest-only warnings as catalog loss | `../../node_modules/.bin/vitest run src/renderer/src/App.test.tsx src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/pages/home/HomePage.persistence.test.tsx src/renderer/src/pages/home/HomePage.render.test.tsx src/renderer/src/pages/home/home-dialogs.render.test.tsx` | PASS — 5 files / 47 tests |
| Project and Session deletion remain available after ordinary partial scans but are disabled when Project deletion recovery failed; an already-open Session dialog disables confirmation when retry removes the capability | `../../node_modules/.bin/vitest run src/main/session-persistence/ipc.test.ts src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/App.test.tsx src/renderer/src/pages/workspace/WorkspacePage.draft-preservation.test.tsx src/renderer/src/pages/workspace/session-dialogs.render.test.tsx` | PASS — 5 files / 65 tests |
| A complete retry durably writes the preserved selected or explicitly empty Session manifest before persistence reports ready; a failed write remains read-only and resumes deferred artifact reconciliation exactly once after write retry succeeds | `../../node_modules/.bin/vitest run src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/lib/session-persistence/session-persistence.test.ts` | PASS — included in the final focused and full-suite runs |
| Deep links and notification clicks survive partial recovery, explicit deletion cancels deferred targets, repeated same-Session clicks remain distinct, and recovery shells cannot acknowledge retained content as visible | `npm test -- --run src/main/notifications/task-notifications.test.ts src/preload/index.test.ts src/renderer/src/App.test.tsx src/renderer/src/hooks/useUnreadTaskViewSync.test.tsx src/renderer/src/lib/deep-link.navigation.test.tsx src/renderer/src/stores/navigation-store.test.ts` | PASS — 6 files / 162 tests |
| Concurrent pending-notification peeks cannot let an older token override later user navigation, while a valid queued retry still opens after persistence becomes ready; the recovery alert directly verifies both layout variants and the shared Button contract | `../../node_modules/.bin/vitest run src/renderer/src/App.test.tsx src/renderer/src/components/SessionPersistenceAlert.render.test.tsx` | PASS — 2 files / 27 tests |
| Interrupted Session Resume cannot reconnect, mutate, or re-send while persistence is recovering; the control becomes available after readiness returns and follows the shared Button focus/disabled contract | `npm test -- src/renderer/src/pages/workspace/WorkspacePage.draft-preservation.test.tsx src/renderer/src/pages/workspace/ConversationPanel.interaction.test.tsx src/renderer/src/pages/workspace/SessionInterruptedBanner.test.tsx` | PASS — 3 files / 63 tests |
| Session write failures retain raw diagnostics only in the diagnostic channel, present renderer-safe recovery copy, and quarantined-data warnings can be dismissed without changing persistence readiness | `npm test -- src/renderer/src/App.test.tsx src/renderer/src/components/SessionPersistenceAlert.render.test.tsx src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx` | PASS — 3 files / 43 tests |
| Recovery surfaces use semantic tokens and shared Buttons with the required focus/disabled contract, while unread visibility follows the rendered shell | `npm test -- --run src/renderer/src/App.test.tsx src/renderer/src/hooks/useUnreadTaskViewSync.test.tsx` | PASS — 2 files / 33 tests |
| App Shell roots use `min-h-svh bg-background text-foreground`; Settings errors are renderer-safe; incomplete catalogs hide exact counts; failed retry manifests keep readiness closed; and Project-load failures preserve pending deep links | `../../node_modules/.bin/vitest run src/renderer/src/App.test.tsx src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/lib/session-persistence/session-persistence.test.ts src/renderer/src/pages/home/HomePage.persistence.test.tsx src/renderer/src/pages/home/HomePage.render.test.tsx src/renderer/src/stores/settings-store.test.ts src/renderer/src/lib/deep-link.navigation.test.tsx src/renderer/src/stores/project-store.test.ts` | PASS — 8 files / 152 tests |
| Degraded fallback scanning leaves malformed and structurally invalid authority untouched; automatic job analysis remains blocked across partial recovery and callback changes; notification clicks retain their causal navigation ordering in both queue directions | `npm test -- --run src/main/session-persistence/coordinator.test.ts src/main/session-persistence/repository.test.ts src/renderer/src/App.test.tsx src/renderer/src/lib/compute/useJobAnalysisEffect.render.test.tsx src/renderer/src/lib/compute/job-analysis-trigger.test.ts src/renderer/src/pages/workspace/WorkspacePage.draft-preservation.test.tsx` | PASS — 6 files / 165 tests |
| Node and renderer TypeScript contracts compile | `npm run typecheck:node`<br>`npm run typecheck:web` | PASS |
| Repository lint rules remain green | `npm run lint` | PASS — 0 errors / 22 pre-existing warnings in unchanged files |
| Cross-cutting regression suite | `npm test` | CLEAN CI PASS — macOS, Ubuntu, and Windows verification all pass. Locally, 547 files / 8,212 tests passed and 11 files / 139 tests skipped; the sole failure was the Prisma schema parity check because the shared parent `node_modules` client contains `PermissionGrant` from another worktree while this branch's schema does not. |

## Uncovered risks

- Tests inject deterministic filesystem, Project deletion recovery, and file-index failures; no packaged Electron end-to-end test reproduces a real OS disappearance race, live SQLite lock, or interrupted deletion-journal replay/tail failure.
- Notification ordering is covered with deterministic renderer IPC mocks, not a real OS notification click racing an Electron window navigation event.
- Resume readiness is covered through renderer component/store mocks, not a packaged restart where recovery and an interrupted ACP Session overlap in real time.
- Job-analysis readiness is covered through renderer stores and deterministic callbacks, not a packaged ACP turn racing a live persistence retry.
- Focus, disabled, opacity, and semantic-token behavior are verified through the rendered DOM/class contract, not cross-theme visual regression or keyboard end-to-end testing.
- Unrecoverable conversation content is surfaced and kept fail-closed but is not reconstructed automatically.

## Review focus

Please focus on readiness boundaries, truly read-only degraded hydration, scan completeness, automatic job-analysis gating, failure-target bookkeeping, atomic navigation intent, and unread acknowledgement while recovery shells cover retained Session state.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -80,6 +80,7 @@ import { registerReviewerIpcHandlers } from './reviewer/ipc'
 import {
   createDefaultReviewRepository,
   createDefaultSessionRepository,
+  loadSessionsAfterProjectRecovery,
   registerSessionPersistenceIpcHandlers
 } from './session-persistence/ipc'
 import { registerProjectFilesIpcHandlers } from './project-files/ipc'
@@ -304,10 +305,8 @@ const registerIpcHandlers = async ({
     artifactProvenanceRepository
   )
   const sessionPersistenceBackend: SessionPersistenceBackend = {
-    loadAll: async () => {
-      await projectDeletionCoordinator.recoverPendingDeletions()
-      return sessionPersistenceCoordinator.loadAll()
-    },
+    loadAll: () =>
+      loadSessionsAfterProjectRecovery(projectDeletionCoordinator, sessionPersistenceCoordinator),
     saveSession: async (session) => {
       await projectDeletionCoordinator.recoverPendingDeletions()
       const created =
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -352,10 +352,16 @@ const registerIpcHandlers = async ({
     onDeliveryError: (error) =>
       notificationsLog.warn('task notification delivery failed', errorLogFields(error))
   })
-  // The renderer pulls the notification click target once its sessions are hydrated (a push sent
-  // before the listener exists would be lost); consume-once semantics live in the service.
-  ipcMain.handle('notifications:take-pending-open-session', () =>
-    taskNotifications.takePendingOpenSession()
+  // The renderer peeks once sessions are hydrated, then conditionally consumes the same target.
+  // This lets partial recovery open an already-loaded conversation while retaining an omitted one
+  // for retry, without an older IPC round trip clearing a newer click target.
+  ipcMain.handle('notifications:peek-pending-open-session', () =>
+    taskNotifications.peekPendingOpenSession()
+  )
+  ipcMain.handle('notifications:take-pending-open-session', (_event, expectedSessionId: unknown) =>
+    typeof expectedSessionId === 'string'
+      ? taskNotifications.takePendingOpenSession(expectedSessionId)
+      : null
   )
   // One MCP client manager backs both dispatch (ConnectorService.call → custom server) and skill-doc
   // generation (listTools) for user-added custom MCP servers (stdio + remote). It lazily connects per
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -358,9 +358,9 @@ const registerIpcHandlers = async ({
   ipcMain.handle('notifications:peek-pending-open-session', () =>
     taskNotifications.peekPendingOpenSession()
   )
-  ipcMain.handle('notifications:take-pending-open-session', (_event, expectedSessionId: unknown) =>
-    typeof expectedSessionId === 'string'
-      ? taskNotifications.takePendingOpenSession(expectedSessionId)
+  ipcMain.handle('notifications:take-pending-open-session', (_event, expectedToken: unknown) =>
+    typeof expectedToken === 'number' && Number.isSafeInteger(expectedToken) && expectedToken > 0
+      ? taskNotifications.takePendingOpenSession(expectedToken)
       : null
   )
   // One MCP client manager backs both dispatch (ConnectorService.call → custom server) and skill-doc
--- a/src/main/notifications/task-notifications.test.ts
+++ b/src/main/notifications/task-notifications.test.ts
@@ -391,15 +391,19 @@ describe('TaskNotificationService', () => {
     expect(deliveryErrors).toEqual([boom])
   })
 
-  it('holds the notification click target until the renderer takes it (consume-once)', () => {
+  it('peeks without consuming and only takes the expected notification target', () => {
     const { service } = createService({})
 
-    expect(service.takePendingOpenSession()).toBeNull()
+    expect(service.peekPendingOpenSession()).toBeNull()
 
     service.setPendingOpenSession('session-7')
 
-    expect(service.takePendingOpenSession()).toEqual({ sessionId: 'session-7' })
-    expect(service.takePendingOpenSession()).toBeNull()
+    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7' })
+    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7' })
+    expect(service.takePendingOpenSession('newer-session')).toBeNull()
+    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7' })
+    expect(service.takePendingOpenSession('session-7')).toEqual({ sessionId: 'session-7' })
+    expect(service.peekPendingOpenSession()).toBeNull()
   })
 
   it('surfaces the window on click even without a session (degraded connector approval)', async () => {
--- a/src/main/notifications/task-notifications.test.ts
+++ b/src/main/notifications/task-notifications.test.ts
@@ -398,12 +398,29 @@ describe('TaskNotificationService', () => {
 
     service.setPendingOpenSession('session-7')
 
-    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7' })
-    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7' })
-    expect(service.takePendingOpenSession('newer-session')).toBeNull()
-    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7' })
-    expect(service.takePendingOpenSession('session-7')).toEqual({ sessionId: 'session-7' })
+    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7', token: 1 })
+    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7', token: 1 })
+    expect(service.takePendingOpenSession(2)).toBeNull()
+    expect(service.peekPendingOpenSession()).toEqual({ sessionId: 'session-7', token: 1 })
+    expect(service.takePendingOpenSession(1)).toEqual({ sessionId: 'session-7', token: 1 })
     expect(service.peekPendingOpenSession()).toBeNull()
+  })
+
+  it('does not let an older same-session click consume a newer target', () => {
+    const { service } = createService({})
+
+    service.setPendingOpenSession('session-7')
+    const older = service.peekPendingOpenSession()
+
+    service.setPendingOpenSession('session-7')
+    const newer = service.peekPendingOpenSession()
+
+    expect(older?.sessionId).toBe('session-7')
+    expect(newer?.sessionId).toBe('session-7')
+    expect(newer?.token).not.toBe(older?.token)
+    expect(service.takePendingOpenSession(older!.token)).toBeNull()
+    expect(service.peekPendingOpenSession()).toEqual(newer)
+    expect(service.takePendingOpenSession(newer!.token)).toEqual(newer)
   })
 
   it('surfaces the window on click even without a session (degraded connector approval)', async () => {
--- a/src/main/notifications/task-notifications.ts
+++ b/src/main/notifications/task-notifications.ts
@@ -210,6 +210,7 @@ export class TaskNotificationService {
   // (window just recreated, React not mounted yet) is lost, so the payload lives here until the
   // renderer — once its sessions are hydrated — takes it. Consume-once.
   private pendingOpenSession: OpenSessionFromNotificationRequest | undefined
+  private pendingOpenSessionToken = 0
 
   // Active snippet for a session, or undefined when there is none.
   private snippetFor(sessionId: string): string | undefined {
@@ -230,7 +231,7 @@ export class TaskNotificationService {
   // Records the conversation a notification click should open, so a renderer that misses the push
   // nudge (still loading, sessions not yet hydrated) can pull it when ready.
   setPendingOpenSession(sessionId: string): void {
-    this.pendingOpenSession = { sessionId }
+    this.pendingOpenSession = { sessionId, token: ++this.pendingOpenSessionToken }
   }
 
   // Lets the renderer check whether the target already exists in a partially hydrated store without
@@ -242,10 +243,10 @@ export class TaskNotificationService {
   // Clears the pending click target only when it is still the one the renderer inspected. A newer
   // notification may replace it while the renderer awaits IPC, and must not be consumed by the
   // older attempt.
-  takePendingOpenSession(expectedSessionId: string): OpenSessionFromNotificationRequest | null {
+  takePendingOpenSession(expectedToken: number): OpenSessionFromNotificationRequest | null {
     const pending = this.pendingOpenSession
 
-    if (!pending || pending.sessionId !== expectedSessionId) return null
+    if (!pending || pending.token !== expectedToken) return null
 
     this.pendingOpenSession = undefined
 
--- a/src/main/notifications/task-notifications.ts
+++ b/src/main/notifications/task-notifications.ts
@@ -233,14 +233,23 @@ export class TaskNotificationService {
     this.pendingOpenSession = { sessionId }
   }
 
-  // Returns and clears the pending click target; the renderer calls this once its session store is
-  // hydrated (and on every push nudge). Null when there is nothing to open.
-  takePendingOpenSession(): OpenSessionFromNotificationRequest | null {
+  // Lets the renderer check whether the target already exists in a partially hydrated store without
+  // losing it when the remaining persistence scan still needs to be retried.
+  peekPendingOpenSession(): OpenSessionFromNotificationRequest | null {
+    return this.pendingOpenSession ?? null
+  }
+
+  // Clears the pending click target only when it is still the one the renderer inspected. A newer
+  // notification may replace it while the renderer awaits IPC, and must not be consumed by the
+  // older attempt.
+  takePendingOpenSession(expectedSessionId: string): OpenSessionFromNotificationRequest | null {
     const pending = this.pendingOpenSession
+
+    if (!pending || pending.sessionId !== expectedSessionId) return null
 
     this.pendingOpenSession = undefined
 
-    return pending ?? null
+    return pending
   }
 
   // Remembers the prompt's first line so the terminal event can name the task. Called when a
--- a/src/main/session-persistence/coordinator.test.ts
+++ b/src/main/session-persistence/coordinator.test.ts
@@ -462,6 +462,57 @@ describe('SessionPersistenceCoordinator', () => {
     expect(fileIndex.reconcileActiveSessions).toHaveBeenCalledWith([session])
   })
 
+  it('hydrates a read-only snapshot without running startup reconciliation', async () => {
+    const session = createSession()
+    const result = { sessions: [session], manifest: { version: 1 as const } }
+    const repository = createSessionRepository({
+      loadAllWithDiagnostics: vi.fn().mockResolvedValue({
+        result,
+        isComplete: true,
+        warnings: []
+      })
+    })
+    const fileIndex = createFileIndex()
+    const provenance = {
+      captureFinalizedMessages: vi.fn().mockResolvedValue(undefined),
+      reconcileSessionDeletions: vi.fn().mockResolvedValue(undefined),
+      prepareSessionDeletion: vi.fn(),
+      completeSessionDeletion: vi.fn(),
+      abortSessionDeletion: vi.fn()
+    }
+    const uploads = { upgradeLegacySessionUploads: vi.fn() }
+    const artifactStorage = {
+      prepareProjectReconciliation: vi.fn(),
+      reconcileSession: vi.fn()
+    }
+    const coordinator = new SessionPersistenceCoordinator(
+      repository,
+      fileIndex,
+      undefined,
+      provenance,
+      uploads,
+      artifactStorage
+    )
+
+    await expect(coordinator.loadAllReadOnly()).resolves.toEqual({
+      ...result,
+      diagnostics: {
+        isComplete: false,
+        warnings: [],
+        failure: 'startup-reconciliation-failed'
+      }
+    })
+
+    expect(fileIndex.markReconciliationIncomplete).toHaveBeenCalledOnce()
+    expect(fileIndex.reconcileActiveSessions).not.toHaveBeenCalled()
+    expect(fileIndex.syncSession).not.toHaveBeenCalled()
+    expect(repository.saveSession).not.toHaveBeenCalled()
+    expect(provenance.reconcileSessionDeletions).not.toHaveBeenCalled()
+    expect(uploads.upgradeLegacySessionUploads).not.toHaveBeenCalled()
+    expect(artifactStorage.prepareProjectReconciliation).not.toHaveBeenCalled()
+    expect(artifactStorage.reconcileSession).not.toHaveBeenCalled()
+  })
+
   it('reconciles path-free Upload copies only on the first complete load from multiple clients', async () => {
     const root = await mkdtemp(join(tmpdir(), 'open-science-upload-startup-reconcile-'))
     const client = createProjectDbClient(root)
--- a/src/main/session-persistence/coordinator.test.ts
+++ b/src/main/session-persistence/coordinator.test.ts
@@ -510,6 +510,7 @@ describe('SessionPersistenceCoordinator', () => {
     })
 
     expect(fileIndex.markReconciliationIncomplete).toHaveBeenCalledOnce()
+    expect(repository.loadAllWithDiagnostics).toHaveBeenCalledWith({ mode: 'read-only' })
     expect(fileIndex.reconcileActiveSessions).not.toHaveBeenCalled()
     expect(fileIndex.syncSession).not.toHaveBeenCalled()
     expect(repository.saveSession).not.toHaveBeenCalled()
--- a/src/main/session-persistence/coordinator.test.ts
+++ b/src/main/session-persistence/coordinator.test.ts
@@ -742,7 +742,8 @@ describe('SessionPersistenceCoordinator', () => {
     await expect(coordinator.loadAll()).resolves.toBe(incomplete.result)
     await expect(coordinator.loadAll()).resolves.toEqual({
       sessions: [upgradedSession],
-      manifest: complete.result.manifest
+      manifest: complete.result.manifest,
+      diagnostics: { isComplete: true, warnings: [] }
     })
 
     expect(uploads.upgradeLegacySessionUploads).toHaveBeenCalledOnce()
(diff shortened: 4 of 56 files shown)


ewen-poch:
Addressed the deletion gap:

- The sidebar now disables Delete together with the other durable conversation mutations while persistence is unavailable.
- `WorkspacePage` guards both opening and confirming deletion, so a readiness change cannot reach `deleteRuntimeSession`.
- Added renderer coverage proving that runtime deletion is not called and the session remains present during recovery.

Validated with `npm run typecheck`, `npm run lint` (0 errors; existing warnings only), and the affected renderer test suites (28 tests passed).

ewen-poch:
Addressed the lifecycle-sync recovery issue:

- Session persistence now exposes successful hydration separately from durable write readiness. A partial but readable snapshot enables lifecycle synchronization while local persistence mutations remain blocked.
- A hard `loadAll` failure now blocks Home/Workspace and presents the retry action, so an unhydrated empty store is never shown as a trustworthy session list.
- Project deletion is disabled and defensively guarded during read-only recovery, including when readiness changes after the confirmation dialog opens, because project deletion cascades to persisted sessions.
- Added renderer coverage for partial hydration, lifecycle synchronization wiring, the hard-failure gate, and project deletion protection.

Validated with `npm run typecheck`, `npm run lint` (0 errors; existing warnings only), 90 focused tests, and the complete suite: 533 files / 7,726 tests passed.

ewen-poch:
Addressed the retry startup-gate gap:

- Session persistence now exposes an explicit loading state for both initial hydration and retries.
- Home/Workspace remain blocked while the renderer has neither a hydrated snapshot nor a completed load attempt; retry now shows `Loading saved conversations…` until `loadAll` resolves.
- Partial successful loads still enter the read-only recovery path, while hard failures return to the blocking error and retry state.
- Added a deferred-promise regression test that verifies the shell stays blocked for the entire retry window.

Validated with `npm run typecheck`, `npm run lint` (0 errors; existing warnings only), 92 focused tests, and the complete suite: 533 files / 7,728 tests passed.

ewen-poch:
Addressed both read-only recovery navigation gaps:

- Deep-link resolution now waits for session hydration rather than durable write readiness, so readable sessions remain addressable during partial recovery.
- Live notification nudges and pending notification replay use the same hydration gate. Pre-hydration targets remain pending; hydrated read-only sessions can open normally.
- Home/Workspace mutation controls continue to use full `isReady`, so this does not re-enable persistence writes during recovery.
- Updated renderer tests to exercise deep links and both notification paths with `isHydrated=true` and `isReady=false`.

Validated with `npm run typecheck`, `npm run lint` (0 errors; existing warnings only), 29 focused tests, and the complete suite: 533 files / 7,728 tests passed.

ewen-poch:
Addressed the three findings in cf246c3:

- Manifest read failures now fall back to an empty last-open selection and emit a non-blocking `manifest-unreadable` warning. Session scan completeness alone controls reconciliation and write readiness, while later manifest writes still use the normal surfaced retry path.
- Project deletion is now available after hydration even during partial recovery. Ordinary project/session writes remain disabled; main still owns the durable deletion intent and target-scoped authority handling.
- Session deletion now has a separate hydration capability from pin, rename, create, and save. Main re-reads the exact target with `loadSessionWithDiagnostics`, rejects unreadable authority, and preserves its existing index/provenance rollback ordering.

Regression coverage now pins the non-blocking manifest fallback and both target-validated deletion paths. Typecheck passed, lint completed with 0 errors, and the full suite passed: 533 files / 7,730 tests, with 10 files / 113 tests skipped.

ewen-poch:
Addressed both deferred-navigation findings in 52dde9e:

- Deep-link initialization now receives both hydration and readiness. During partial recovery it opens a target only when that exact project/session is already present; otherwise it leaves the original URL untouched and unresolved until a complete retry can decide whether the target is truly missing.
- Notification pulls now wait for full persistence readiness. Both the live nudge and recreated-window paths leave the destructive `takePendingOpenSession` call pending in main during partial recovery, then consume it after the complete session snapshot is available.

Regression tests cover opening an available deep link during partial recovery, retaining an omitted deep link through retry, and preserving notification targets from both delivery paths. Typecheck passed, lint completed with 0 errors, and the full suite passed: 533 files / 7,732 tests, with 10 files / 113 tests skipped.

ewen-poch:
Addressed the StrictMode startup hydration finding in 0da1321:

- Identical non-forced settings loads now reuse the same in-flight Promise, so the replayed mount effect cannot advance the generation or discard a successful first snapshot.
- The explicit startup Retry action now calls `load({ force: true })`. Forced loads still advance the generation, so a real user retry remains authoritative over an older request.
- The existing stale-result protection is preserved for forced refreshes.

Regression coverage verifies that concurrent startup calls issue each settings IPC exactly once and share the same result, while a forced retry still wins when an older load finishes later. Typecheck passed, lint completed with 0 errors, and the full suite passed: 533 files / 7,733 tests, with 10 files / 113 tests skipped.

ewen-poch:
Updated the recovery alert surface after visual verification: `bg-bg-100` had no matching theme token, so the floating recovery card was transparent and allowed the composer text underneath to show through. The alert now uses the opaque `bg-bg-000` surface in both light and dark themes. I also added a regression test for the partial-hydration overlay. Verified with the focused App test, full typecheck, lint (0 errors), and the full test suite (7,734 passed; 113 skipped).

ewen-poch:
Addressed both recovery findings in `9546a82`:

- Project loading now waits for settings plus a completed Session hydration pass. `retryLoad()` temporarily clears hydration, so a successful retry triggers a fresh project-list load and clears any error left by the original transient outage.
- A deferred initial deep link now observes explicit navigation-store actions. Once the user navigates manually, the pending URL target is discarded and URL synchronization adopts the user-selected destination; applying the initial deep link marks itself initialized before navigating so it does not cancel itself.
- Added renderer regressions for project reload across a partial-load retry and for preserving manual navigation when the omitted deep-link target later becomes available.

Validated with 25 focused tests, full typecheck, lint (0 errors; existing warnings only), and the complete suite: 533 files / 7,736 tests passed, with 10 files / 113 tests skipped.

ewen-poch:
Addressed the notification recovery issue in 8275e43. The renderer now peeks at the retained click target before conditionally consuming that exact session ID, so an already-hydrated conversation opens during partial recovery while an omitted conversation remains available for retry. Explicit navigation dismisses a deferred target, and generation checks prevent an in-flight notification from overriding the user's newer choice. The main/preload IPC contract and Web API map were updated accordingly, with regression coverage for partial recovery, live notification nudges, conditional consumption, and manual-navigation cancellation. Full test suite, typecheck, and lint all pass (lint retains 24 pre-existing warnings and no errors).

ewen-poch:
Addressed the automatic-redirect issue in 2cbb9ba. Navigation actions now require an explicit user or automatic origin, and the store advances userNavigationRevision only for proven user-owned navigation. Deferred notification and deep-link handling observe that revision instead of every store mutation, so startup, lifecycle, and notification-driven redirects no longer consume pending targets. Home/project/session clicks, Workspace session switches, New Conversation, the Home button, and lifecycle notice actions are marked as user navigation. Regression coverage now contrasts automatic redirects with explicit user navigation and covers the navigation-origin contract. The affected navigation/Workspace suites pass (65 tests, plus 19 App tests); typecheck passes and lint reports 0 errors with the existing 24 warnings. Fresh three-platform CI is running on the pushed head.

ewen-poch:
Addressed the empty deep-link fallback in cd6da19. When the launch URL contains neither a project nor a session target, deep-link initialization now marks itself complete without changing navigation. This preserves a session already opened by a desktop-notification click while project loading finishes; URLs that contain an actual but invalid/incomplete target retain the existing Home fallback behavior. Added a regression that opens a session with automatic notification-like navigation before the hook initializes and verifies it remains in Workspace. The deep-link and App suites pass (28 tests), typecheck passes, and lint reports 0 errors with the existing 24 warnings.

ewen-poch:
Addressed the startup navigation race in 22f0f92. Desktop-notification clicks now use a distinct explicit navigation origin and advance an explicit-intent revision, so a deferred startup deep link cannot replay over the conversation selected from the notification. Lifecycle redirects remain automatic and do not consume either deferred intent, while notification recovery still only yields to in-app user navigation. Added a regression test for an initial session URL, an early notification open, and delayed project loading. Validation: typecheck passed; lint passed with 0 errors (24 existing warnings); full test suite passed with 533 files / 7,742 tests (10 files / 113 tests skipped).

ewen-poch:
Addressed the partial-recovery selection replay in be48c6f. When Retry is initiated from an interactive partial snapshot, the renderer captures the live selected session and uses it as the hydration preference if that session still exists in the successful reload. The preferred session's project and session manifest pointers are updated together for that hydration pass; initial startup and retries where the selected session disappeared retain the durable manifest fallback. Added a hook-level regression that selects a session in another project during partial recovery and verifies a successful retry does not restore the stale manifest session. Validation: typecheck passed; lint passed with 0 errors (24 existing warnings); full test suite passed with 533 files / 7,743 tests (10 files / 113 tests skipped).

ewen-poch:
Addressed the repeated same-session notification race in 7f50cd4.

- Main now assigns a monotonic token to every notification click, including consecutive clicks targeting the same session.
- The token is round-tripped through the shared contract, IPC, preload, and renderer; conditional consume now matches the exact click rather than only the session ID.
- Added service- and renderer-level regressions proving an older attempt cannot consume a newer same-session click.

Validation: targeted tests passed (115 tests), typecheck passed, lint passed with 0 errors (24 pre-existing warnings), and the full suite passed (533 files / 7,745 tests; 10 files / 113 tests skipped).

ewen-poch:
Addressed the Standards and Spec findings in `2cc1704`:

- Startup file-index sync is now part of the reconciliation failure boundary; any sync failure marks reconciliation incomplete and returns `startup-reconciliation-failed`.
- A Session listed by the directory scan but missing at read time now produces an unreadable warning and a partial scan. Direct by-ID lookup retains its normal `missing` result.
- Removing a durably deleted Session now automatically prunes its obsolete failed-write target and clears the visible write error when no failures remain.
- Recovery alerts and generic startup shells now use shadcn semantic tokens and the shared `Button`, including the required focus and disabled contracts.
- The PR body now includes a final `behavior -> command -> result` matrix, an after-last-material-edit declaration, independent review status, and uncovered risks.

Final validation: 119 focused tests passed; the four behavior groups passed with 91, 88, 132, and 81 tests respectively; typecheck passed; lint passed with 0 errors and 24 pre-existing warnings; the full Vitest suite passed. Independent Standards and Spec reviews found no remaining code findings.

ewen-poch:
No code change for this finding because it conflicts with the explicit acceptance criterion for this PR: a complete Session scan followed by a failed startup file-index sync must return `startup-reconciliation-failed`.

The distinction is intentional:

- Readable Session JSON still hydrates, so existing conversations remain available.
- Durable mutations remain blocked until the entire startup reconciliation boundary succeeds.
- A failed `syncSession` means the Files projection is known to be incomplete after active-owner reconciliation; starting the saver in that state would violate the requested fail-closed recovery contract.
- The visible recovery action retries the boundary instead of presenting the partial projection as ready.

The regression in `coordinator.test.ts` locks this specified behavior. Splitting Files readiness from Session write readiness would be a broader product/architecture change and should not be introduced while resolving this explicit requirement.

ewen-poch:
Addressed the persistence recovery findings:

- Session JSON with a valid ID but no authoritative conversation structure is now quarantined; a valid canonical graph can still replace a damaged compatibility `messages` projection.
- A Project directory that disappears after root enumeration now makes the authority scan incomplete, while a directly queried absent Project remains an authoritative empty set.
- Failure to recover pending Project deletions now falls back to explicit read-only Session hydration with incomplete diagnostics; reconciliation, Session writes, and derived-state synchronization stay disabled.
- A queued save rejection arriving after authoritative deletion is pruned immediately and cannot recreate a stale retry target or alert.
- Explicit empty and missing retry selections are applied atomically with Session hydration, so Zustand subscribers never observe a stale manifest/global selection.
- Hard load exceptions retain raw diagnostics in the main/console path while the renderer receives fixed path-safe recovery copy.

Validation after the final material edit: targeted recovery tests passed (102 main tests and 156 renderer/store tests), `npm run typecheck` passed, `npm run lint` passed with 0 errors (24 pre-existing warnings in unchanged files), and the full suite passed with 7,757 tests. Final independent Standards and Spec reviews found no remaining findings.

ewen-poch:
Addressed the latest high-value review findings:

- Explicit Project and Session deletion now record user navigation before the durable delete, so deferred notification or deep-link targets are cancelled even when the route itself does not change.
- The latest `main` integration preserves token-aware pending-notification peek/take behavior alongside unread view synchronization, visibility probes, and close-confirm coverage.
- App-level unread visibility is now false whenever Session persistence is unhydrated or loading, preventing a retained selection from suppressing an unread marker behind retry loading or hard-error shells.

Focused regressions cover both deletion paths and the visible → retry loading → hard failure transition. Final verification on `3c7b9cc`: typecheck passed; lint passed with 0 errors (24 pre-existing warnings); the full suite passed with 546 files / 8,186 tests (11 files / 139 tests skipped). Independent Standards and Spec reviews report no remaining findings.

ewen-poch:
Addressed the Project deletion count finding:

- Target-scoped Project deletion remains available after partial Session hydration.
- Session persistence now exposes catalog completeness separately from write readiness. Partial scans, hard load failures, and Session-level warnings—including successfully quarantined corrupt files—mark the catalog incomplete, while manifest-only warnings do not.
- The confirmation dialog uses an exact count only for a complete catalog. Otherwise it explicitly warns that all saved conversations, including any that could not be loaded during recovery, will be deleted.
- Coverage includes partial recovery and the `isReady=true` case with a quarantined corrupt Session, so a writable healthy subset cannot produce a misleading count.

Validation on `4c14095`:

- Focused renderer suite: 5 files / 47 tests passed.
- Node and renderer typechecks passed.
- Full ESLint: 0 errors / 24 pre-existing warnings in unchanged files.
- Full Vitest suite: 546 files / 8,187 tests passed; 11 files / 139 tests skipped.
- Independent Standards and Spec reviews found no remaining findings.

ewen-poch:
Addressed both deletion-readiness findings:

- The main startup boundary now stamps whether pending Project deletion recovery completed. This signal is independent of Session scan completeness and general write readiness.
- Renderer deletion capability defaults and resets fail-closed. Ordinary partial Session scans still permit target-validated Project and Session deletion after Project recovery succeeds, while the `loadAllReadOnly()` fallback disables both actions.
- Home Project deletion, Workspace sidebar deletion, the Session dialog, and both confirmation handlers use the dedicated capability. A dialog already open when Retry removes the capability now disables its shared `Button`, and the handler remains independently guarded.
- Session deletion cannot enter ACP teardown when the durable deletion prerequisite is unavailable.

Validation on `faeba39`:

- Focused capability suite: 5 files / 65 tests passed, including allowed partial scans, failed Project recovery, and the open-dialog `true → false` transition.
- Node and renderer typechecks passed.
- Full ESLint: 0 errors / 24 pre-existing warnings in unchanged files.
- Full Vitest with four workers: 546 files / 8,190 tests passed; 11 files / 139 tests skipped.
- An unconstrained local run exhausted timing budget in two unrelated files; both files passed immediately in isolation (2 files / 209 tests), and the complete four-worker rerun passed.
- Final independent Standards and Spec reviews found no remaining findings.

ewen-poch:
Addressed the retry-selection durability finding in `e6933da`:

- A retry now captures its preferred tri-state selection once and forces that selection through the normal manifest saver after complete hydration.
- Persistence remains not-ready/loading until the manifest write settles, so the hydrated saver baseline cannot silently skip the preserved selection.
- Explicitly clearing the selection writes an empty manifest, and manifest write failures continue to use the existing surfaced/retryable `manifest` target.
- Regression coverage verifies the selected and explicitly empty payloads and holds the manifest write pending to prove readiness is not announced early.

Final validation: the focused persistence suite passed (2 files / 31 tests), both TypeScript configurations passed, ESLint completed with 0 errors (24 pre-existing warnings in unchanged files), and the complete four-worker suite passed (546 files / 8,191 tests; 11 files / 139 tests skipped). Independent Standards and Spec reviews found no remaining findings.

ewen-poch:
Addressed the recovery-boundary findings in `a7986c0`:

- Updated all four top-level startup/recovery shells to the App Shell contract (`min-h-svh bg-background text-foreground`), with muted color scoped to secondary copy.
- Kept persistence readiness closed when the forced retry-selection manifest write fails. The surfaced manifest target remains retryable, and a successful write retry opens readiness and starts deferred artifact reconciliation exactly once.
- Replaced exact per-Project Session counts with “Session count unavailable” while the Session catalog is incomplete.
- Kept raw Settings load diagnostics in `console.warn` and exposed only stable renderer-safe copy.
- Deferred deep-link resolution while Project loading has any error value, including an empty error message, so a transient catalog failure cannot clear the URL or consume the target as missing.

Regression coverage also verifies the two adjacent edge cases: pending-artifact reconciliation resumes after a successful manifest retry, and empty-string Project errors still preserve the deep link.

- Focused recovery suite: 8 files / 152 tests passed.
- Full suite: 546 files / 8,194 tests passed; 11 files / 139 tests skipped.
- Node and renderer TypeScript checks passed.
- ESLint passed with 0 errors (24 pre-existing warnings in unchanged files).
- Independent Standards and Spec reviews found no remaining findings.

ewen-poch:
Addressed the two worthwhile follow-ups in `656c45a`:

- Serialized pending-notification inspections so dependency changes and live nudges cannot peek the same token concurrently. The earliest queued attempt retains its navigation revision; later attempts run afterward and observe the consume-once result instead of reviving stale notification intent.
- Added complementary regressions for both outcomes: an older token cannot override later user navigation, and a valid queued attempt still opens the target after persistence becomes ready.
- Added `SessionPersistenceAlert.render.test.tsx` as the smallest component-owned check for fixed/error and inline/warning layouts, optional Retry rendering, semantic tokens, and the shared Button focus/disabled contract.

- Focused renderer suite: 2 files / 27 tests passed.
- Full suite: 547 files / 8,198 tests passed; 11 files / 139 tests skipped.
- Node and renderer TypeScript checks passed.
- ESLint passed with 0 errors (24 pre-existing warnings in unchanged files).
- Independent Standards and Spec reviews found no remaining findings.

ewen-poch:
Addressed the partial-recovery Resume gap in `f173687`.

- `WorkspacePage` now rejects interrupted-Session resume requests unless Session persistence is ready.
- Readiness is propagated through `ConversationPanel`, and the Resume control stays natively disabled during recovery or an in-flight resume.
- Resume now uses the shared `Button`, including the required focus ring, disabled opacity, and pointer-event contract.
- Added coverage for the page-level guard, the recovery-to-ready transition, the ConversationPanel control path, and the banner's unavailable state.

Verification after the final material edit:

- Focused tests: 3 files / 63 tests passed.
- Node and renderer typechecks passed.
- ESLint passed with 0 errors (24 pre-existing warnings in unchanged files).
- Full suite: 547 files / 8,201 tests passed; 11 files / 139 tests skipped.
- Independent Standards and Spec reviews found no remaining actionable findings.

ewen-poch:
Addressed the two recovery-alert findings in `24b4e8e`:

- Session write failures now map to renderer-safe copy before entering UI state, so filesystem paths and storage-engine diagnostics remain in the diagnostic channel only. Write retry behavior is unchanged.
- Persistent quarantine warnings now expose a labelled shared-Button dismiss action. Dismissal only clears the current renderer warning; it does not change persistence readiness or delete the durable diagnostic, so a later load can surface the warning again when appropriate.

Validation: the three focused renderer files pass 43 tests; both TypeScript checks pass; lint passes with 0 errors and 24 pre-existing warnings; the full suite passes 547 files / 8,203 tests (11 files / 139 tests skipped). Independent Standards and Spec reviews found no remaining findings.

ewen-poch:
Addressed the recovery-boundary findings in `dae93d6`:

- Automatic completed-job analysis is now readiness-scoped. Partial recovery cannot start an ACP turn, disable/cleanup removes queued scans and turn-end listeners, enabling after recovery rescans pending jobs, and `useEffectEvent` keeps one trigger when the runtime `sendMessage` callback identity changes.
- The Project-deletion fallback now requests an explicit repository `read-only` scan. Malformed JSON, structurally invalid Session JSON, and a corrupt manifest are reported as incomplete without renaming or otherwise rewriting authority; normal repair scans retain quarantine behavior.
- Pending notification work now carries both a click generation and the navigation revision that caused it. A newer click supersedes older queued recovery work, while navigation after an older click still wins. Reciprocal regression tests cover both orders.

I left the recovery panels as semantic alert surfaces instead of wrapping them in `Card`: the design guide makes shared Card usage a preference rather than a hard requirement, and these surfaces already use semantic alert/background/border tokens, shared Buttons, and direct render-contract coverage.

Final validation: the six affected files pass 165 tests; both TypeScript checks pass; lint passes with 0 errors and 22 pre-existing warnings. The local full suite passed 8,212 tests with one environment-only Prisma parity failure caused by another worktree's shared generated client; clean-checkout cross-platform CI is the final suite result. Independent Standards and Spec reviews found no remaining findings.

PR #506, opened 2026-07-29
ci(windows): balance PR speed and release validation

ewen-poch:
## Problem

The Windows PR Check was dominated by a non-blocking full Vitest run. In a comparable baseline, the advisory step took 661 seconds of a 953-second Windows job, reported known cross-platform failures, and could not block a regression.

Removing that run from PRs improves feedback time, but doing only that would move coverage gaps to users. Post-merge and release validation therefore retain broad Windows visibility and exercise the packaged installer itself.

The focused gate also exposed a real production-shell issue: the scrubbed Windows environment made module-backed PowerShell cmdlets spend tens of seconds in module discovery, while native-command and parser-only paths stayed below 0.5 seconds.

## Proposed change

- Make the PR Windows job a focused, blocking gate for Windows filesystem, ACL, icon, runtime-cache, PowerShell, and public notebook-shell timeout/status behavior.
- Give each focused PowerShell process a 30-second budget and each Vitest case a separate 35-second budget. Production remains at 120 seconds.
- Provide a controlled PowerShell module search path containing the standard all-users and in-box module directories, in Windows PowerShell's standard precedence, without inheriting CurrentUser or arbitrary host entries. Restore that value after Windows PowerShell 5.1 startup reconstruction, preload the common Management and Utility in-box modules by absolute manifest path to avoid the observed first-cmdlet scan, and clear the internal carrier before user code runs.
- Run the complete Windows Vitest suite after merge and on stable tags as two shards in a dedicated advisory workflow. It is independent from reusable packaging, publishing, notarization, and upgrade smoke.
- Hard-gate every nightly and stable Windows artifact on a real NSIS install, packaged-resource checks, bundled micromamba execution, packaged-app startup, authenticated bootstrap probe, clean shutdown, and uninstall.
- Discover the packaged service from the app's own loopback startup output, matching Electron's real Windows home resolution instead of assuming that `USERPROFILE` changes `app.getPath('home')`.
- Give startup and shutdown independent 60-second budgets, drain the shutdown response before waiting for process exit, and bound `taskkill.exe` itself to 10 seconds with a direct-child fallback.
- On stable tags, install the previous published desktop `v*` release first and upgrade it in place with the current installer before publishing.
- Pass one isolated profile environment to all smoke child processes, but use each packaged app's authenticated data as the source of truth for its actual config root. Current builds must report the root in bootstrap. Already-published builds may omit it, so the smoke checks both the runner's real OS-home root and the isolated-profile root and accepts only the candidate whose existing `web-token` matches the authenticated service token. Write a collision-resistant sentinel into that verified root, require the current app to report the same root, verify preservation, and clean up the marker on both success and failure.
- Pin Node 22 in the stable upgrade-smoke job instead of relying on the mutable runner image.
- During uninstall, verify that the executable, packaged resources, Prisma engine, uninstaller, and installation directory are removed.
- Retry Windows file-lock cleanup without masking the primary smoke failure.
- Apply 10-minute and 20-minute GitHub Actions hard limits to the fresh-install smoke step and stable upgrade-smoke job respectively.

```mermaid
flowchart TD
  PR["Pull request"] --> PRWin["Focused Windows gate<br/>blocking, 4m54 measured"]
  PR --> Mac["macOS full coverage gate"]
  PR --> Linux["Ubuntu full suite<br/>advisory"]

  Main["Push to main"] --> Nightly["Nightly workflow"]
  Tag["Stable v* tag"] --> ReleaseFlow["Release workflow"]
  Main --> WinFull["Dedicated Windows full suite<br/>2 advisory shards"]
  Tag --> WinFull
  Nightly --> Build["Reusable build"]
  ReleaseFlow --> Build

  Build --> Package["Windows package"]
  Package --> Fresh["Install → resources → micromamba → start → authenticated probe → shutdown → uninstall<br/>blocking, 10m hard cap"]
  Fresh --> Artifact["Windows artifact"]
  Artifact --> NightlyPublish["Nightly publish"]
  Artifact --> Upgrade["Previous stable → report actual config root → sentinel → in-place upgrade → verify preservation<br/>blocking on stable tags, 20m hard cap"]
  Upgrade --> StablePublish["Stable publish"]
```

## Duration impact

| Path | Baseline | Measured result / expectation |
| --- | ---: | ---: |
| Windows PR Check | about 15m53s | **4m54s**, all focused Windows checks passed (about 69% faster) |
| Full Windows suite after merge / on stable tags | previously serialized in the PR job | **7m16s and 8m41s** in the final run; dedicated advisory workflow, so neither runtime nor result delays packaging or publishing; about 15m57s of Windows runner usage |
| Windows build task | existing release path | **8m20s** including dependencies, packaging, fresh smoke, and artifact upload |
| Windows package step | existing release path | **4m52s** |
| Fresh installer smoke | new | **42s**, passed end to end |
| Manual Release build | comparable build path | **13m49s** end to end; successful |
| Nightly overall | commonly 14–20m | expected around **14–16m** including publish; the full-suite shards are a separate workflow and the smoke adds about 42s in the measured run |
| Stable release | build plus notarization | previous-release upgrade is still estimated at roughly **2–4m** and runs in parallel with macOS notarization, so wall time is the slower branch rather than their sum |

The successful manual Release run was not a stable tag, so the tag-only previous-version upgrade and macOS notarization branches were intentionally skipped. Their workflow shape and smoke logic are covered by tests, but the upgrade duration remains an estimate until the next real stable tag.

## Scope and trade-offs

- The complete Windows suite is restored for `main` pushes and stable `v*` tags, but remains a separate advisory workflow because it currently contains known POSIX-only assertions. Nightly/release publishing neither waits for nor fails with it.
- The focused PR Windows gate and installer smokes are blocking.
- The installer smoke verifies the managed-runtime manifest and executes bundled `micromamba.exe`, but deliberately does not download and provision full Python/R environments on every release. That network-heavy canary remains a possible separately scheduled follow-up.
- macOS retains the full coverage gate; Ubuntu retains the full advisory suite.
- The two Windows full-suite shards consume additional runner minutes even though they do not extend wall-clock time.

- Final PR Check: Windows **4m54s**, Ubuntu **6m11s**, macOS **10m24s**; all passed.
- Manual Release workflow: **13m49s**, including successful real Windows package and fresh-install smoke.
- Windows packaging: **4m52s**; fresh installer smoke: **42s**.
- Related PR workflow, runtime-service, installer, web bootstrap, release workflow, and Windows-shell tests after rebasing onto the latest `main`: **185 passed**, 4 Windows-only cases skipped locally.
- Full local suite on the final rebased source: **8,002 passed**, 139 skipped.
- `npm run typecheck` and `npm run lint` passed after the final material edit (0 lint errors; 24 existing warnings outside the changed files).
- `actionlint -shellcheck=`, Prettier, and `git diff --check`.
- Latest published stable release contains the expected `*-win-x64-setup.exe` asset consumed by the upgrade gate.

| Affected behavior | Final project-owned check | Result |
| --- | --- | --- |
| Sanitized machine `PSModulePath`, absolute in-box preload, wrapper restoration, and carrier removal | `src/main/notebook/runtime-service.test.ts` plus `src/main/notebook/windows-shell.integration.test.ts` in the affected Vitest command | Unit coverage passed locally; 4 Windows-only integration cases are hard-gated by PR Check |
| Actual/legacy config-root discovery, token authentication, collision-safe sentinel lifecycle, and bounded termination | `scripts/windows-installer-smoke.test.ts` | Passed in the 185-test affected suite |
| Authenticated bootstrap reports the resolved config root | `src/main/web-service/controller.test.ts` and `src/main/web-service/http-server.test.ts` | Passed in the 185-test affected suite |
| Fresh/upgrade Actions hard limits, independent post-merge full-suite workflow, and PR workflow shape | `scripts/windows-release-workflows.test.ts` and `scripts/pr-check-workflow.test.ts` | Passed in the 185-test affected suite; final standalone-workflow regression also passed after the workflow-only edit |
| Compile, static, and broad regression safety after rebase | `npm run typecheck`, `npm run lint`, `npm test`, Prettier, Node syntax, `actionlint -shellcheck=`, `git diff --check` | Passed; full Vitest: 8,002 passed / 139 skipped |

Independent Standards and Spec reviews of the final rebased diff found no remaining issue and confirmed that the mapping covers the deterministic behavior changed here.

Uncovered runtime risks: the real previous-release-to-current NSIS path only runs on stable tags; local tests simulate rather than induce a genuinely hung `taskkill.exe`; effective Windows PowerShell behavior requires the PR's Windows runner; non-preloaded in-box modules can still incur their normal AllUsers-first cold discovery cost; and local YAML tests cannot prove GitHub event fan-out until the first `main` push or tag. Stable Release CI, 10/20-minute Actions caps, the blocking Windows PR Check, the 120-second production shell budget, and the standalone workflow's first live trigger are the corresponding confirmation layers.

## Review findings addressed

- `PSModulePath` now contains `%ProgramFiles%\WindowsPowerShell\Modules` and `$PSHOME\Modules` in standard order, then is restored inside the encoded wrapper so Windows PowerShell startup cannot silently reinsert CurrentUser paths. The wrapper imports the Management and Utility in-box manifests by absolute path, avoiding the live-observed slow AllUsers scan without changing precedence for other modules. The existing Windows UTF-8 process asserts the exact effective path and verifies that the internal carrier is removed. It deliberately does not require a particular AllUsers module to be preinstalled on every runner image.
- Upgrade preservation is checked in the actual verified config root rather than blindly assuming either Electron's OS home or the child `USERPROFILE`, eliminating the previous false-positive path. For already-published binaries, both locations are only candidates and the endpoint token authenticates which one the running app actually used. Current binaries still fail closed without an absolute reported `configRoot`; UUID markers use exclusive creation and unconditional cleanup so they neither overwrite nor remain in a real profile.
- `taskkill.exe` has an independent bounded wait, process timeouts cannot be converted into successful non-zero exits, and Actions-level limits provide defense in depth.
- The advisory complete Windows shards now run in a standalone `main`/`v*` workflow, so reusable `build.yml` callers no longer wait for them before nightly or stable publication.

The Windows-only integration assertion and packaged installer path are revalidated by the PR and Release workflows after push; the tag-only previous-version upgrade remains covered structurally until the next real stable tag.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,35 +65,6 @@ jobs:
       - name: Check CLI package
         run: npm run check:cli-package
 
-  # Preserve complete real-Windows coverage after merge without putting its ~11-minute runtime back
-  # on every PR. Two shards run in parallel with packaging. This remains advisory while the suite has
-  # known POSIX-only path assertions; the focused PR gate and packaged-app smoke below are blocking.
-  windows_full_test:
-    name: Windows full test (${{ matrix.shard }}/2, advisory)
-    if: ${{ !inputs.skip_verify && !inputs.mac_only }}
-    runs-on: windows-latest
-    continue-on-error: true
-    timeout-minutes: 25
-    strategy:
-      fail-fast: false
-      matrix:
-        shard: [1, 2]
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v7
-
-      - name: Setup Node
-        uses: actions/setup-node@v7
-        with:
-          node-version: 22
-          cache: npm
-
-      - name: Install dependencies
-        run: npm ci
-
-      - name: Test complete suite shard
-        run: npm test -- --shard=${{ matrix.shard }}/2
-
   # Resolve the platform matrix. mac_only trims it to the two mac archs — used by the notarization
   # dry-run (notarize-dryrun.yml), which only needs a signed mac build. With mac_only false (the
   # release.yml / nightly.yml callers) the full matrix is emitted, so their behavior is unchanged.
@@ -258,6 +229,7 @@ jobs:
       # Windows runtime resources, request a clean shutdown, then silently uninstall it.
       - name: Smoke test Windows installer
         if: matrix.platform == 'win'
+        timeout-minutes: 10
         run: node scripts/windows-installer-smoke.mjs --installer-dir dist
 
       # Upload only the distributables + electron-updater metadata for the caller's publish job.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,6 +65,35 @@ jobs:
       - name: Check CLI package
         run: npm run check:cli-package
 
+  # Preserve complete real-Windows coverage after merge without putting its ~11-minute runtime back
+  # on every PR. Two shards run in parallel with packaging. This remains advisory while the suite has
+  # known POSIX-only path assertions; the focused PR gate and packaged-app smoke below are blocking.
+  windows_full_test:
+    name: Windows full test (${{ matrix.shard }}/2, advisory)
+    if: ${{ !inputs.skip_verify && !inputs.mac_only }}
+    runs-on: windows-latest
+    continue-on-error: true
+    timeout-minutes: 25
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Node
+        uses: actions/setup-node@v7
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Test complete suite shard
+        run: npm test -- --shard=${{ matrix.shard }}/2
+
   # Resolve the platform matrix. mac_only trims it to the two mac archs — used by the notarization
   # dry-run (notarize-dryrun.yml), which only needs a signed mac build. With mac_only false (the
   # release.yml / nightly.yml callers) the full matrix is emitted, so their behavior is unchanged.
@@ -223,6 +252,13 @@ jobs:
             export CSC_IDENTITY_AUTO_DISCOVERY=false
             npx electron-builder ${{ matrix.eb_args }} "${eb_ver[@]}" --publish never
           fi
+
+      # Run the actual NSIS artifact, not the source tree: silent install into an isolated directory,
+      # start the packaged app in headless web mode, verify its authenticated bootstrap and bundled
+      # Windows runtime resources, request a clean shutdown, then silently uninstall it.
+      - name: Smoke test Windows installer
+        if: matrix.platform == 'win'
+        run: node scripts/windows-installer-smoke.mjs --installer-dir dist
 
       # Upload only the distributables + electron-updater metadata for the caller's publish job.
       - name: Upload build artifacts
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -89,6 +89,17 @@ jobs:
           src/main/notebook/micromamba-cache-powershell.test.ts
           src/main/notebook/micromamba-cache-acl.integration.test.ts
 
+      # runtime-service.test.ts takes about 70 seconds and has unrelated POSIX assumptions on
+      # Windows. Run its one Windows-only shell integration case by name so this behavior remains a
+      # hard gate without restoring the noisy full-file cost.
+      - name: Test Windows notebook shell behavior
+        if: matrix.os == 'windows-latest'
+        run: >-
+          npx vitest run
+          src/main/notebook/runtime-service.test.ts
+          --testNamePattern
+          "isolates commands, propagates PowerShell failures, and preserves UTF-8 output on Windows"
+
       # Full test suite + coverage gate on macOS (thresholds live in vitest.config.ts).
       - name: Test (with coverage)
         if: matrix.os == 'macos-14'
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -33,9 +33,9 @@ jobs:
       matrix:
         # macOS is the release/verification target and the only platform with the project's
         # arm64-only darwin native dependency present, so it runs the FULL gate (test + coverage +
-        # build). Linux/Windows validate the platform-independent code paths (lint + typecheck are
-        # hard-gated everywhere); their test leg is advisory (continue-on-error) because the missing
-        # darwin native package can trip suites that reach the ACP/native layer — see the test steps.
+        # build). Ubuntu provides the advisory cross-platform full-suite signal. Windows hard-gates
+        # the platform-independent checks plus a focused native-filesystem/PowerShell test suite;
+        # see the test steps for why the noisy full suite is intentionally omitted there.
         os: [macos-14, ubuntu-latest, windows-latest]
 
     steps:
@@ -62,23 +62,32 @@ jobs:
 
       # Guards against src/preload drifting from the generated web API map. Pure Node and
       # platform-independent, so hard-gate it on all three OSes rather than relying on the
-      # macOS-only vitest leg (the Linux/Windows test run is advisory).
+      # macOS coverage leg, Ubuntu advisory suite, or focused Windows suite.
       - name: Check web API map
         run: npm run check:web-api-map
 
-      # The general suite discovers these tests, but Linux/Windows tests are advisory. Keep the
-      # zero-dependency CLI/SDK contract and publishable tarball hard-gated on every platform.
+      # The general suite discovers these tests, but Ubuntu's run is advisory and Windows runs only
+      # a focused platform suite. Keep the zero-dependency CLI/SDK contract and publishable tarball
+      # hard-gated on every platform.
       - name: Test CLI and SDK
         run: npm run test:cli
 
       - name: Check CLI package
         run: npm run check:cli-package
 
-      # This exercises the production DACL writer and verifier on a real Windows filesystem. Keep
-      # it hard-gated even while the general Windows suite remains advisory.
-      - name: Test Windows micromamba cache ACL
+      # Keep the Windows signal focused and actionable. The full suite currently has many known
+      # POSIX-only failures and repeatedly builds fresh Prisma/SQLite schemas; on Windows that turns
+      # into more than ten minutes of advisory noise. Ubuntu still runs the cross-platform full suite,
+      # while this hard gate exercises the real Windows filesystem, DACL, icon, and PowerShell paths.
+      - name: Test Windows-specific behavior
         if: matrix.os == 'windows-latest'
-        run: npx vitest run src/main/notebook/micromamba-cache-acl.integration.test.ts
+        run: >-
+          npx vitest run
+          src/main/windows.test.ts
+          src/main/windows-icon-assets.test.ts
+          src/main/windows-powershell.test.ts
+          src/main/notebook/micromamba-cache-powershell.test.ts
+          src/main/notebook/micromamba-cache-acl.integration.test.ts
 
       # Full test suite + coverage gate on macOS (thresholds live in vitest.config.ts).
       - name: Test (with coverage)
@@ -86,10 +95,10 @@ jobs:
         run: npm run test:coverage
 
       # Cross-platform test signal, advisory only: a failure here is surfaced in the PR but does not
-      # block merge, because the darwin-only native dependency is absent on Linux/Windows and can fail
-      # suites that touch it. Promote to a hard gate once those suites are made platform-independent.
+      # block merge, because the darwin-only native dependency is absent on Linux and can fail suites
+      # that touch it. Promote to a hard gate once those suites are made platform-independent.
       - name: Test (advisory)
-        if: matrix.os != 'macos-14'
+        if: matrix.os == 'ubuntu-latest'
         continue-on-error: true
         run: npm run test
 
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -89,16 +89,12 @@ jobs:
           src/main/notebook/micromamba-cache-powershell.test.ts
           src/main/notebook/micromamba-cache-acl.integration.test.ts
 
-      # runtime-service.test.ts takes about 70 seconds and has unrelated POSIX assumptions on
-      # Windows. Run its one Windows-only shell integration case by name so this behavior remains a
-      # hard gate without restoring the noisy full-file cost.
+      # Exercise the production Windows shell runner directly. The broader runtime-service suite
+      # takes about 70 seconds and has unrelated Windows failures, so its former combined shell case
+      # could time out without identifying which PowerShell contract regressed.
       - name: Test Windows notebook shell behavior
         if: matrix.os == 'windows-latest'
-        run: >-
-          npx vitest run
-          src/main/notebook/runtime-service.test.ts
-          --testNamePattern
-          "isolates commands, propagates PowerShell failures, and preserves UTF-8 output on Windows"
+        run: npx vitest run src/main/notebook/windows-shell.integration.test.ts
 
       # Full test suite + coverage gate on macOS (thresholds live in vitest.config.ts).
       - name: Test (with coverage)
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -96,6 +96,16 @@ jobs:
         if: matrix.os == 'windows-latest'
         run: npx vitest run src/main/notebook/windows-shell.integration.test.ts
 
+      # Keep the public service boundary covered without running the rest of the heavy runtime-service
+      # suite. This verifies that Windows tree termination is persisted as a timed-out notebook run.
+      - name: Test Windows notebook shell service timeout
+        if: matrix.os == 'windows-latest'
+        run: >-
+          npx vitest run
+          src/main/notebook/runtime-service.test.ts
+          --testNamePattern
+          "kills a command that outlasts the timeout and returns a non-normal result"
+
       # Full test suite + coverage gate on macOS (thresholds live in vitest.config.ts).
       - name: Test (with coverage)
         if: matrix.os == 'macos-14'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,7 @@ jobs:
     needs: build
     if: startsWith(github.ref, 'refs/tags/')
     runs-on: windows-latest
+    timeout-minutes: 20
     permissions:
       contents: read
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,11 +35,57 @@ jobs:
     uses: ./.github/workflows/notarize-mac.yml
     secrets: inherit
 
+  # The reusable build already hard-gates a fresh NSIS install/start/uninstall. On stable tags, also
+  # install the latest published stable version and run the new installer over it. This runs beside
+  # macOS notarization, so its few minutes normally do not extend the release critical path.
+  windows-upgrade-smoke:
+    needs: build
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: windows-latest
+    permissions:
+      contents: read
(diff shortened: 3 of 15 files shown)


ewen-poch:
Addressed the Windows coverage gap in commit 108690e. The full runtime-service test file took about 70 seconds and had unrelated Windows failures, so the PR now hard-gates only the existing Windows-only executeShell integration case via --testNamePattern. That test gets a 30-second budget because it launches four fresh PowerShell processes, and the workflow regression test locks the step to Windows as a blocking check.

ewen-poch:
The Windows hard gate exposed that the former combined runtime-service case was not a reliable baseline: it timed out in this run, and the historical PR #252 log shows the same timeout while the Windows full suite was advisory. Commit d8ddc78 replaces that never-green combined case with four independent integration cases against the production shell runner: cmdlet failure short-circuiting, native exit-code propagation, UTF-8 output, and trailing-continuation isolation. Each process has a bounded 10-second timeout, and the dedicated file remains a blocking Windows check.

ewen-poch:
Addressed the public service timeout/status gap in commit 9f0f5a0. The Windows timeout path now awaits the bounded taskkill tree teardown before settling and ignores the resulting exit event, so NotebookRuntimeService.executeShell persists status: timeout and returns only after workspace handles are released. PR Check hard-gates the existing public executeShell timeout test by name alongside the low-level PowerShell integration suite, and a regression test proves termination does not settle before the tree terminator completes.

ewen-poch:
Windows CI exposed a fixed-delay PowerShell 5.1 module-analysis cleanup after the first cmdlet: both output-producing cases reached the 10-second shell timeout, while native and parse-error cases completed in about 0.3 seconds. Commit 8480b83 sets PSDisableModuleAnalysisCacheCleanup=1 only in the scrubbed Windows shell environment. Notebook shells are disposable, so this skips stale cache-entry cleanup without disabling the cache or inheriting host secrets; the unit test also verifies that a host-provided value cannot override the controlled setting. Local lint, Node typecheck, focused regressions, and the full 7,710-test suite pass.

ewen-poch:
The final Windows run falsified the cache-cleanup hypothesis: the cmdlet-failure and UTF-8 cases still timed out at 10.084s and 10.065s, while the native-exit and parser-error cases completed in 0.212s and 0.503s. The remaining delay is therefore in PowerShell first-use module discovery or loading before user output, not the delayed analysis-cache cleanup. Increasing the timeout would only hide the defect because the historical combined test also exceeded 30 seconds. This PR has reached the agreed five review cycles, so I am pausing without another push. The current head remains blocked by Windows CI despite the static Codex review being mergeable. A follow-up should instrument raw pre-normalization PowerShell stderr and compare a minimal PSModulePath, explicit built-in module import, and stdin handling before choosing the next fix.

ewen-poch:
Implemented the post-merge Windows safety net and updated the PR description accordingly:

- `main`/nightly/release now run the complete Windows suite in two parallel advisory shards, preserving broad visibility without returning the ~11-minute full suite to every PR.
- Nightly and stable Windows artifacts are blocked on an actual NSIS install, packaged headless startup, authenticated bootstrap probe, clean shutdown, and uninstall.
- Stable tags additionally upgrade the latest prior desktop `v*` release in place and verify that the previous user profile is preserved before publication.
- The smoke probe has bounded HTTP/process waits, and uninstall success requires the executable, packaged resources, Prisma engine, and uninstaller to be removed.
- Windows PowerShell shells now set the documented `PSModuleAnalysisCachePath=nul`, addressing the observed delayed module-cache write rather than only disabling stale-entry cleanup.

The complete Windows suite remains advisory for now because current `main` still has known POSIX-only assertions; the focused PR gate and installer/upgrade smokes are blocking. Nightly is expected to remain near its existing critical path because the full-suite shards run in parallel; the fresh smoke should add roughly 1–3 minutes to the Windows package job. The stable upgrade smoke is expected to add roughly 2–4 minutes and runs beside macOS notarization.

ewen-poch:
Follow-up on the latest Windows runs and review feedback:

- The module-analysis cache overrides were removed. The hosted runner falsified that hypothesis: both overrides still timed out at 10 seconds. The actual first-cmdlet case completed at about 10.1 seconds, while native-command and parser-continuation cases remained below one second. The focused integration now uses a 15-second budget, still well below the 120-second production default.
- The upgrade profile isolation issue from the Codex review is fixed. The previous installer, current installer, packaged app, bundled micromamba, sentinel check, and uninstaller now share the same isolated `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, `TEMP`, and `TMP` environment.
- The first real installer-smoke run exposed a Windows file-lock cleanup failure that could mask the primary result. Cleanup now retries locked paths and preserves the original smoke error when both operations fail.
- Measured timings are now reflected in the PR description: the focused Windows PR job was 5m07s; the two complete-suite shards were 7m21s and 7m45s in parallel; Windows packaging was 5m46s; and the first fresh-install smoke attempt took 1m17s.

The branch has been pushed as a batch, and a corrected manual Release run will validate the installer path end to end.

ewen-poch:
The latest live Windows runs exposed two additional harness boundaries, both fixed in `de4a15b`:

- The shell process and Vitest case previously shared the same 15-second deadline. Once the process timer fired, the Windows tree terminator still needed to settle, so Vitest could time out before observing the shell result. Each PowerShell process now has 30 seconds and each test has 35 seconds; production remains at 120 seconds.
- The packaged app reached a healthy authenticated bootstrap and accepted shutdown with HTTP 202, but the smoke script did not consume the response body before waiting for process exit. That kept the HTTP connection active while the app waited for its web server to close. The shutdown response is now drained first, with a regression test locking the ordering.

The latest manual run measured 5m42s for Windows packaging, 1m16s for the smoke attempt, 9m42s for the complete Windows build job, and 17m57s end to end. The two advisory full-suite shards finished independently and did not prevent packaging from starting. Local validation passes with 7,722 tests, Node typecheck, lint (existing warnings only), and the focused smoke/workflow tests.

A new manual Release run will verify fresh install, packaged startup, shutdown, uninstall, and final timing on the corrected head.

ewen-poch:
Addressed the shutdown-budget finding in commit 2c6540e. Process exit is now observed without starting a timer during startup; after the authenticated shutdown response is fully drained, the harness starts a fresh 60-second shutdown deadline against that same exit observation. This preserves detection of exits before health, avoids missing an exit between the HTTP response and listener registration, and prevents a forced timeout termination from being mistaken for a successful exit. A fake-timer regression covers the independent budget. The full suite passes with 7,723 tests, along with Node typecheck, Prettier, and the 13 focused installer/workflow tests.

ewen-poch:
The third live Windows run exposed a separate discovery issue, fixed in commit 8da1c17. Electron's Windows home path is resolved through the OS and did not follow the child USERPROFILE override, so the app successfully started its authenticated web service while the harness waited for state files under the isolated profile. The smoke now discovers the loopback endpoint and token from the packaged app's own 'Open Science Web' startup output, then performs the same authenticated bootstrap and shutdown probes. This avoids a fixed 60-second false timeout and tests the observable packaged-app startup contract directly. The new parser regression and all 14 installer/workflow tests pass; Node typecheck and Prettier pass. An unrelated local full-suite retry became resource-bound, while the preceding full run passed 7,723 tests; the three affected files were isolated to 184/185 passing, and the sole unrelated timing case passed on immediate focused rerun.

ewen-poch:
The latest Windows gate isolated the PowerShell delay to module-backed cmdlets: Get-Item and Write-Output exhausted 30 seconds, while native execution and parser-only paths stayed below 0.5 seconds. Commit f2cd325 fixes the production shell environment by setting PSModulePath to the in-box WindowsPowerShell system module directory derived from SystemRoot. The host PSModulePath is deliberately not inherited, so arbitrary user or third-party modules remain outside the scrubbed shell environment. The environment regression was added test-first, the complete runtime-service plus installer/workflow set passes 136 tests, full typecheck and lint pass, and the full suite reached 7,723 passing tests with one unrelated timing case that passed on immediate focused rerun.

ewen-poch:
Addressed the stable-upgrade runtime finding in commit 41e2dcd. The windows-upgrade-smoke job now installs Node 22 through actions/setup-node v7 before invoking the ESM smoke harness, matching the repository's build and PR runtimes instead of relying on the mutable runner image. The workflow regression asserts both the action version and Node major. The 14 installer/workflow tests, actionlint, and Prettier pass.

ewen-poch:
Measured validation is complete. The focused Windows PR job passed in 6m01s versus the comparable 15m53s baseline; Ubuntu passed in 6m19s and macOS in 6m40s. Manual Release run 30510979649 passed end to end in 13m49s: Windows packaging took 4m52s, and the real fresh-install/start/authenticated-bootstrap/shutdown/uninstall smoke took 42s. The two advisory full Windows shards took 7m16s and 8m41s in parallel and did not delay packaging. The PR description now records these timings and the stable-tag limitation: previous-version upgrade timing remains estimated because a test release tag was not created. The latest correctness review also identified that the sanitized PSModulePath omits the standard all-users Program Files module directory. This is a valid compatibility blocker; work is paused at the agreed review-iteration limit, and the PR is not presented as merge-ready until that item is resolved or explicitly accepted.

ewen-poch:
Addressed the three actionable Windows findings in `f0d797f`:

- The notebook shell now builds a sanitized machine module path from `%ProgramFiles%\WindowsPowerShell\Modules` and `$PSHOME\Modules`. Because Windows PowerShell 5.1 reconstructs `PSModulePath` during startup, the encoded wrapper restores the controlled value before user code and removes its internal carrier. The existing Windows UTF-8 integration process now also asserts the exact effective path, so this adds coverage without another PowerShell startup.
- Current packaged builds report the resolved config root through the authenticated loopback endpoint. For already-published binaries that do not have that field, the smoke authenticates the real root by matching its existing `web-token` file to the service token. Upgrade smoke uses a UUID marker with exclusive creation, requires the current app to report the same root, verifies preservation there, and unconditionally removes the marker without overwriting a pre-existing file. It no longer assumes `USERPROFILE` controls Electron's home path.
- `taskkill.exe` now has an independent 10-second deadline with a direct-child fallback, and the parent process timeout remains authoritative even if termination races with an allowed non-zero exit. The fresh smoke step and stable upgrade-smoke job also have 10-minute and 20-minute GitHub Actions limits.

Validation: full local suite 7,731 passed / 116 skipped; affected suite 154 passed / 4 Windows-only skipped locally; typecheck, lint (0 errors, existing warnings only), actionlint, Prettier, Node syntax check, and `git diff --check` passed.

ewen-poch:
Rebased the branch onto the latest `main` because the PR had become conflicting and GitHub could not create the merge ref or start `pull_request` checks. The only content conflict was the old combined Windows shell test that this PR had already replaced; the resolution keeps `main`'s managed-runtime protections while retaining the four focused production-shell integration cases. The integration helper now supplies `runtimeRoot` required by the current shell contract.

Final head: `a3f2d0c`.

Post-rebase evidence:

- affected PR workflow/runtime/installer/bootstrap/release tests: 185 passed, 4 Windows-only skipped locally;
- full Vitest suite: 8,002 passed, 139 skipped;
- typecheck, lint (0 errors; existing warnings only), Prettier, Node syntax, actionlint, and diff checks passed;
- independent Standards and Spec reviews found no remaining issue.

Remaining runtime-only coverage is explicit: effective PowerShell behavior runs in the blocking Windows PR Check, the real previous-release upgrade runs on stable tags, and Actions-level 10/20-minute limits contain a real runner-level process hang.

ewen-poch:
Addressed the valid advisory-workflow finding in `1cc6d04`: the two complete Windows shards have moved out of reusable `build.yml` into a standalone advisory workflow triggered by `main` pushes, stable `v*` tags, and manual dispatch. Nightly/release packaging, publishing, notarization, and upgrade smoke no longer depend on either shard's runtime or result. The workflow regression asserts both removal from the reusable build and the independent trigger/shard contract; actionlint, Prettier, Node typecheck, lint, and the focused test pass.

I did not apply the `env.USERPROFILE` recommendation for the legacy config-root fallback. Live Windows runs established that Electron's `app.getPath('home')` ignored the child `USERPROFILE` override; that mismatch is why endpoint discovery was introduced. The smoke process does not mutate its own environment, so `homedir()` resolves the runner account's real OS home used by the previous packaged app. The fallback then authenticates `<real-home>\.open-science` by matching its existing `web-token` to the token discovered from that app's loopback URL. A wrong root fails closed, whereas using the synthetic child profile would restore the already-observed failure mode. An independent spec review confirmed this trade-off.

ewen-poch:
Resolved the legacy config-root concern in `98931a6` without assuming that either Windows home source is universally correct.

For a previously published app that does not report `configRoot`, the upgrade smoke now checks both the runner's real OS-home root and the isolated-profile root. Neither path is trusted by location: the smoke accepts only the candidate whose existing `web-token` matches the token from the authenticated running-app endpoint. Missing or mismatched candidates are skipped, and no authenticated match fails closed. Current builds still must report an absolute `configRoot` and do not use this fallback.

The regression test covers a missing real-home candidate followed by a matching isolated-profile candidate. Final validation passed: affected suite 185 passed / 4 skipped, full Vitest 8,002 passed / 139 skipped, typecheck, lint (0 errors), Prettier, Node syntax, actionlint, and `git diff --check`.

ewen-poch:
The live Windows gate exposed one final performance boundary in `98931a6`: with the standard AllUsers path first, `Get-Item` reached the 30-second process deadline while `Write-Output` completed in 22.4 seconds. Native execution and parser-only cases remained below 0.3 seconds, confirming that the delay was first-use module discovery rather than the process harness.

Fixed in `2cd3a25` without changing normal module precedence. The sanitized path remains `%ProgramFiles%\\WindowsPowerShell\\Modules` before `$PSHOME\\Modules`, while the wrapper imports the in-box Management and Utility manifests by absolute `$PSHOME` paths before user code. Updated machine modules and AllUsers-only modules therefore retain standard resolution behavior. The existing Windows process now also imports `PowerShellGet` and asserts that its `ModuleBase` is under Program Files, in addition to checking the exact effective path, UTF-8 output, and removal of the internal carrier.

Post-edit validation passed: affected suite 185 passed / 4 Windows-only skipped locally; full Vitest 8,002 passed / 139 skipped; typecheck; lint with 0 errors; Prettier; actionlint; and `git diff --check`. The new Windows head is the remaining runtime confirmation.

ewen-poch:
The live result confirms the preload fix itself: on `2cd3a25`, the four real Windows PowerShell processes completed in 558 ms, 354 ms, 1.36 s, and 351 ms respectively, versus the previous 30-second timeout / 22.4-second cold-discovery behavior.

The remaining failure was only the added assumption that every hosted image preinstalls `PowerShellGet` for Windows PowerShell under Program Files. This runner did not satisfy that environment assumption, so the test exited after the successful UTF-8 write. `d2acc54` removes that non-portable module-name requirement while retaining the exact assertion that the effective sanitized `PSModulePath` contains only the standard AllUsers and `$PSHOME` roots in standard order. The production preload code is unchanged from the measured successful run.

Post-edit validation again passed: affected suite 185 passed / 4 Windows-only skipped locally; full Vitest 8,002 passed / 139 skipped; typecheck; lint with 0 errors; Prettier; actionlint; and `git diff --check`.

ewen-poch:
Final CI on `d2acc54` is green: Windows 4m54s, Ubuntu 6m11s, and macOS 10m24s. The real Windows shell processes also confirmed the preload optimization, completing in 0.35–1.36 seconds rather than reaching the 30-second process budget.

The latest Codex findings are evaluated as follows:

- The config-root boundary is valid, but requiring equality with the child `USERPROFILE` path is not compatible with the live evidence: packaged Electron ignored that override and reported the runner's OS-home root. The current endpoint is token-authenticated, but accepting any absolute reported root is broader than necessary. A proper follow-up should give the smoke run an explicit app-level config-root override before startup and assert that exact root (or, at minimum, constrain the authenticated report to the known isolated/OS-home candidates).
- Excluding the standard CurrentUser PowerShell module directory is a valid compatibility regression for users with per-user modules. A follow-up should add only the canonical `%USERPROFILE%\\Documents\\WindowsPowerShell\\Modules` entry to the sanitized path, without inheriting arbitrary host `PSModulePath` entries. The measured absolute in-box preloads can remain to keep the focused gate fast.

These are not treated as mergeable. The review/implementation cycle has exceeded the agreed cap, so work is paused here without another code push.

PR #507, opened 2026-07-30
fix(artifact-provenance): stabilize finalization failures

ewen-poch:
## Problem

Artifact finalization retried one ownership-persistence race by matching a human-readable error message. That coupled renderer recovery to wording and risked treating unrelated proof failures as recoverable. The same finalization boundary also lacked a privacy-safe structured diagnostic that distinguished durable provenance finalization from compatibility publication.

## Proposed change

- Extend the existing `artifacts:finalize-run` IPC response with one stable `ownership-persistence-race` failure code.
- Classify only a missing durable final message as that recoverable race; keep prompt, role, frame, branch, runtime segment, and compatibility failures terminal.
- Retry the coded race once in the renderer, independently of its human-readable message.
- Emit one structured diagnostic through the existing artifact logger with the finalization stage, durable/publication completion state, and claim/provenance correlation IDs.
- Exclude exception text, artifact contents, user paths, prompts, project/session identifiers, and secrets from the diagnostic.

## Scope and non-goals

This keeps the existing artifact IPC route and logging sink; it does not add a parallel transport or logger. A bounded Better Harness Artifact Provenance evidence scan was also run without creating or installing an asset. The installed 0.3.0 collector did not expose the required request-root decision boundary for all omitted candidates, so this PR does not claim that separate repeated-workflow evidence item is complete.

- `npm test -- src/main/artifacts/ipc.test.ts src/renderer/src/lib/acp/workspace-events.test.ts` — 82 passed
- `npm test -- src/main/artifacts/ipc.test.ts src/main/artifacts/provenance-repository.test.ts` — 54 passed
- `npm run lint` — 0 errors (24 pre-existing warnings)
- `npm test` — 7784 passed, 113 skipped

Tests prove the stable code drives the single retry when the message changes, legacy text alone remains terminal, invalid proof and compatibility failures do not retry, compatibility replay remains idempotent, and diagnostics identify stage/durable state without leaking injected content or paths.

## Review focus

- The boundary between a missing final message (recoverable persistence race) and a present-but-invalid ownership proof (terminal).
- The IPC result union as the smallest caller-visible discriminant.
- The diagnostic field allowlist and durable/publication state transitions.

### DIFF
--- a/src/main/artifacts/ipc.test.ts
+++ b/src/main/artifacts/ipc.test.ts
@@ -4,11 +4,19 @@ import { tmpdir } from 'node:os'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { createArtifactVersionLocator } from '../../shared/artifact-provenance'
-import type { ArtifactFile, ArtifactWriteSource } from '../../shared/artifacts'
+import {
+  ARTIFACT_OWNERSHIP_PERSISTENCE_RACE,
+  type ArtifactFile,
+  type ArtifactWriteSource
+} from '../../shared/artifacts'
 import { createLinearConversationGraph } from '../../shared/conversation-graph'
 import type { PersistedChatSession } from '../../shared/session-persistence'
 import { createPngBytes, createPngInlineSource } from './artifact-test-fixtures'
-import { ArtifactProvenanceRepository } from './provenance-repository'
+import {
+  ArtifactFinalizationProofError,
+  ArtifactOwnershipPersistenceRaceError,
+  ArtifactProvenanceRepository
+} from './provenance-repository'
 import { ArtifactRepository } from './repository'
 import {
   createArtifactHandlers,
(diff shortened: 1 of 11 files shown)


ewen-poch:
Addressed the headless artifact finalization contract regression in 8099c32. `HeadlessTaskApi.completeSession` now consumes `FinalizeRunArtifactsResult`, unwraps successful artifacts, and treats the coded failure result as terminal instead of spreading the tagged object. The task API test now exercises the actual IPC success envelope. Focused validation: `npm test -- src/main/web-service/task-api.test.ts` (8 passed) and `npm run typecheck` (passed).

ewen-poch:
Addressed the headless ownership-race recovery finding in 0e8ba80. The task API now persists the completed assistant ownership state only when it receives `ownership-persistence-race`, retries that finalization exactly once, and leaves thrown proof/compatibility errors and a failed retry terminal. The regression test verifies the save-before-retry ordering, two finalization attempts, and that artifact IDs are linked only after successful finalization. Focused validation: `npm test -- src/main/web-service/task-api.test.ts` (8 passed) and `npm run typecheck` (passed).

ewen-poch:
Addressed the multi-claim partial-completion finding in 2f31309. Artifact-loop failures now carry the accumulated completion state into the existing `failRun` path, so artifacts finalized by earlier claims remain linked in the failed run and durable session while the later claim stays terminal. The regression test covers a recovered race on the first claim followed by a compatibility failure on the second claim and verifies that the first artifact is preserved. Focused validation: `npm test -- src/main/web-service/task-api.test.ts` (8 passed), `npm run typecheck` (passed), and focused Prettier check (passed).

PR #517, opened 2026-07-30
fix(acp): isolate runtime framework switches

ewen-poch:
## Problem

Switching an existing ACP session between Codex and Claude Code could let late session-scoped events from the draining runtime appear under the newly selected framework. This made Codex-style MCP activity look as though it had been invoked by Claude Code. The optimistic prompt created during the switch could also retain the previous runtime segment, and Codex artifact bridge receipts fell back to the generic Terminal/Command/Output presentation.

## Proposed change

- publish session-scoped runtime events only from the runtime generation that currently owns the session
- keep the draining runtime authoritative until asynchronous adoption succeeds and its prior active turn clears, then commit ownership to the selected runtime
- revalidate the incoming runtime after the prior turn drains so a retired or superseded adopter cannot acquire session ownership
- retain events admitted before ownership commits in the bounded runtime snapshot, while rejecting future stale events emitted by the old runtime
- commit the selected runtime's current connection status together with session ownership after adoption
- expose per-session runtime preparation state so the workspace blocks draft edits, sends, and message edits throughout adoption or Branch reset
- guard submissions synchronously by draft/session key so a second prompt cannot be dispatched or cleared before the reactive preparation lock renders
- restore a failed prepared send to its originating draft key when the user has selected another conversation
- version user-authored composer mutations so delayed failure restoration cannot overwrite newer intent in the same conversation
- restore interrupted and overflow-recovery user turns when replacement-send preparation fails, without rolling back newer runtime failure metadata
- preserve delayed Artifact finalization through explicit run/prompt provenance after ownership changes
- bind the next optimistic prompt to the selected framework and backend when that runtime adopts the session
- guarantee Branch isolation on attached, detached, and cross-runtime continuations: detached or changed runtimes adopt first, then reset the selected owner unless resume proves the context is already fresh
- recognize artifact-write MCP identities across provider naming conventions and summarize nested Codex bridge receipts as file activity without exposing internal receipt fields
- add regression coverage for runtime handoff, adoption failure, draining terminal events, Branch reset ownership, prompt provenance, preparation-state gating, newer-draft preservation, and artifact receipt presentation

## Scope and non-goals

- Session-less discovery events and events from the current owner remain publishable.
- This does not register unavailable MCP tools or change permission policy; a genuine `No such tool available` response remains a separate tool-registration/configuration issue.
- This does not change artifact persistence or add inline preview data when a receipt does not include a path.

All change-specific checks below ran after the final material edit and passed.

- draining-runtime events, including terminal `stop`/`error` arriving after the incoming provider attaches, stay on the prior owner until its active prompt clears; events already admitted before ownership commits remain in subsequent bounded snapshots, post-adoption old-runtime events are absent, and fresh selected-runtime events publish
- the incoming optimistic run is not created until adoption succeeds, so a draining terminal event cannot settle the new Runtime Segment
- runtime preparation is reflected in the workspace UI, and the synchronous request guard preserves a different newer draft without issuing a duplicate runtime call
- a failed prepared send restores its doc and attachments to the original conversation without replacing a newly selected conversation's live draft
- a newer same-conversation draft wins over delayed failure restoration, and staged files belonging only to the superseded request are cleaned up
- interrupted and overflow-recovery retries restore their removed user turn and conversation graph when shared send preparation returns unsent; a later retry does not redundantly resume an already attached provider
- failed adoption leaves the prior runtime authoritative
- an incoming runtime that retires or is superseded while the prior turn drains is rejected before ownership commits
- successful adoption replaces any stale draining-runtime connection status, preventing later ordinary failures from being misclassified as disconnects
- delayed Artifact finalization remains publishable through the completed handoff
- the first switched prompt points at the selected framework/backend runtime segment
- a Branch continuation resets the attached selected runtime directly, or adopts a detached/changed runtime and resets the newly selected owner when resume reports `contextReset: false`
- a nested Codex artifact receipt renders as `Write file`, filename, and size without raw bridge metadata
- related coordinator, renderer runtime, composer, job-analysis, and tool-detail coverage -> six Vitest files -> 187 passed
- isolated repository verification after the final commit -> macOS, Ubuntu, and Windows PR Verify jobs passed
- TypeScript contracts remain valid -> `npm run typecheck` -> passed
- repository lint gate -> `npm run lint` -> passed with 0 errors (24 existing warnings)
- changed-file formatting and patch hygiene -> Prettier check and `git diff --check` -> passed

## Review focus

- Ownership commits only after `resumeSession` resolves and the prior owner's active prompt clears. Intermediate attached-state emissions from the incoming runtime do not steal routing, and a late old-runtime terminal event settles before the renderer can append the selected runtime's turn.
- The incoming runtime must still be registered, non-retired, and the pending adopter after the old-turn drain before ownership can commit.
- Ownership and the adopted runtime's connection status commit atomically; an old session-id status is removed when a provider remaps the id.
- Events admitted while a runtime owns the session are remembered per runtime and retained only for that runtime's bounded snapshot window. Events first emitted after ownership transfers are still filtered.
- The renderer resumes before creating the new optimistic run, so draining terminal events can settle the prior Runtime Segment safely.
- Runtime adoption and Branch reset publish a per-session preparation lock. A synchronous request guard closes the pre-render gap without conflating intentional suppression with send failure.
- Per-draft mutation versions distinguish the request's optimistic clear from later user edits, including attachment intake and removal, before any failed request restores captured state.
- Retry flows snapshot the transcript, conversation graph, and files revision before removing the unanswered turn, restoring that projection if preparation fails while retaining current runtime/error state.
- Detached same-runtime Branch continuations adopt before resetting, ensuring `resetSessionContext` targets the selected coordinator owner; cross-runtime continuations follow the same ownership-safe order.
- Artifact lifecycle events bypass owner filtering after a committed handoff because their explicit run/prompt/claim provenance safely finalizes the prior turn.
- Artifact receipt parsing is bounded and only activates for known artifact-write tool identities.

## Uncovered risk

The runtime handoff is covered with coordinator fakes rather than a live Claude Code/Codex dual-process end-to-end test.

Retained control-flow events remain unresolved: an admitted old-runtime `context-overflow` or compaction event can stay in coordinator snapshots after ownership changes, then re-trigger recovery when renderer-local deduplication resets on remount. These events should not use the sticky display/history retention path.

### DIFF
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -752,10 +752,19 @@ describe('AcpRuntimeCoordinator', () => {
       expect.stringMatching(runtimeEventId(1, 'late-codex-artifact')),
       expect.stringMatching(runtimeEventId(2, 'fresh-claude-tool'))
     ])
-    expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([
-      expect.stringMatching(runtimeEventId(2, 'fresh-claude-tool')),
-      expect.stringMatching(runtimeEventId(1, 'late-codex-artifact'))
-    ])
+    const adoptedSnapshotEventIds = coordinator.getSnapshot().events.map((event) => event.id)
+    expect(adoptedSnapshotEventIds).toEqual(
+      expect.arrayContaining([
+        expect.stringMatching(runtimeEventId(1, 'owner-tool')),
+        expect.stringMatching(runtimeEventId(1, 'late-codex-tool')),
+        expect.stringMatching(runtimeEventId(1, 'late-codex-stop')),
+        expect.stringMatching(runtimeEventId(1, 'late-codex-artifact')),
+        expect.stringMatching(runtimeEventId(2, 'fresh-claude-tool'))
+      ])
+    )
+    expect(adoptedSnapshotEventIds).not.toEqual(
+      expect.arrayContaining([expect.stringMatching(runtimeEventId(1, 'post-adoption-codex-tool'))])
+    )
 
     retirement.resolve()
     await switchRequest
@@ -854,7 +863,9 @@ describe('AcpRuntimeCoordinator', () => {
     expect(forwardedEvents.map((event) => event.id)).toEqual([
       expect.stringMatching(runtimeEventId(1, 'owner-overflow'))
     ])
-    expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([])
+    expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'owner-overflow'))
+    ])
 
     created[1].emitEvent(overflowEvent('fresh-owner-overflow'))
     expect(forwardedEvents.map((event) => event.id)).toEqual([
@@ -862,6 +873,7 @@ describe('AcpRuntimeCoordinator', () => {
       expect.stringMatching(runtimeEventId(2, 'fresh-owner-overflow'))
     ])
     expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'owner-overflow')),
       expect.stringMatching(runtimeEventId(2, 'fresh-owner-overflow'))
     ])
 
@@ -918,7 +930,9 @@ describe('AcpRuntimeCoordinator', () => {
     expect(forwardedEvents.map((event) => event.id)).toEqual([
       expect.stringMatching(runtimeEventId(1, 'owner-compaction'))
     ])
-    expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([])
+    expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'owner-compaction'))
+    ])
 
     created[1].emitEvent(compactionEvent('fresh-owner-compaction', 'completed'))
     expect(forwardedEvents.map((event) => event.id)).toEqual([
@@ -926,6 +940,7 @@ describe('AcpRuntimeCoordinator', () => {
       expect.stringMatching(runtimeEventId(2, 'fresh-owner-compaction'))
     ])
     expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'owner-compaction')),
       expect.stringMatching(runtimeEventId(2, 'fresh-owner-compaction'))
     ])
 
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -629,6 +629,78 @@ describe('AcpRuntimeCoordinator', () => {
     await reloadRequest
   })
 
+  it('drops late Codex tool events after Claude Code adopts the switched session', async () => {
+    const retirement = createDeferred<void>()
+    const created: ReturnType<typeof createFakeRuntime>[] = []
+    const forwardedEvents: AcpRuntimeEvent[] = []
+    const coordinator = new AcpRuntimeCoordinator(
+      (callbacks) => {
+        const fake = createFakeRuntime({
+          frameworkId: created.length === 0 ? 'codex' : 'claude-code',
+          sessionIds: [`agent-session-${created.length + 1}`],
+          callbacks
+        })
+        created.push(fake)
+        return fake.runtime
+      },
+      { onEvent: (event) => forwardedEvents.push(event) }
+    )
+
+    const session = await coordinator.createSession()
+    created[0].emitState({
+      promptInFlight: true,
+      promptInFlightSessionIds: [session.sessionId]
+    })
+    created[0].requestRetirement.mockReturnValue(retirement.promise)
+    const switchRequest = coordinator.requestAgentFrameworkSwitch()
+    const toolEvent = (id: string, providerToolName: string): AcpRuntimeEvent => ({
+      id,
+      timestamp: 1,
+      kind: 'tool',
+      level: 'info',
+      sessionId: session.sessionId,
+      toolCallId: id,
+      providerToolName,
+      title: providerToolName,
+      toolKind: providerToolName.startsWith('mcp.') ? 'execute' : 'other',
+      status: 'completed'
+    })
+
+    // The draining Codex turn still owns the session until Claude Code adopts it.
+    created[0].emitEvent(toolEvent('owner-tool', 'mcp.open-science-artifacts.write_artifact_file'))
+    expect(forwardedEvents.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'owner-tool'))
+    ])
+
+    await coordinator.resumeSession({
+      sessionId: session.sessionId,
+      cwd: '/workspace',
+      previousFrameworkId: 'codex'
+    })
+
+    created[0].emitEvent(
+      toolEvent('late-codex-tool', 'mcp.open-science-artifacts.write_artifact_file')
+    )
+    expect(forwardedEvents.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'owner-tool'))
+    ])
+    expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([])
+
+    created[1].emitEvent(
+      toolEvent('fresh-claude-tool', 'mcp__open-science-artifacts__write_artifact_file')
+    )
+    expect(forwardedEvents.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'owner-tool')),
+      expect.stringMatching(runtimeEventId(2, 'fresh-claude-tool'))
+    ])
+    expect(coordinator.getSnapshot().events.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(2, 'fresh-claude-tool'))
+    ])
+
+    retirement.resolve()
+    await switchRequest
+  })
+
   it('drops late recoverable overflow events from a previous session owner', async () => {
     const retirement = createDeferred<void>()
     const created: ReturnType<typeof createFakeRuntime>[] = []
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -39,6 +39,7 @@ const createFakeRuntime = (options: {
   permissionGrantStore?: ConversationPermissionGrantStore
   beforePromptStart?: () => Promise<void>
   beforeResume?: () => Promise<void>
+  afterResumeAttached?: () => Promise<void>
   eligibleAttachmentUri?: string
   prompt?: (sessionId: string) => Promise<unknown>
 }): {
@@ -83,6 +84,7 @@ const createFakeRuntime = (options: {
         : [...snapshot.sessionIds, sessionId]
     }
     options.callbacks.onStateChanged?.(snapshot)
+    await options.afterResumeAttached?.()
     return { sessionId, cwd: '/workspace', frameworkId: options.frameworkId, contextReset: true }
   })
   const resetSessionContext = vi.fn(async ({ sessionId }: { sessionId: string }) => ({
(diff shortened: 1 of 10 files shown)


ewen-poch:
Addressed both findings in `4234787`.

- **Late Artifact claims:** Artifact lifecycle events now bypass runtime-owner filtering because they carry explicit run, prompt, and claim provenance. Stale tool, message, and stop events remain isolated so they cannot mutate the new runtime segment. The regression test delays adoption, rejects the stale Codex tool event, and preserves the late Artifact claim.
- **Branch reset attribution:** Branch reset now decides whether the selected runtime must adopt before resetting context. When the framework/backend changes or the session is detached, it shuts down notebook context, skips the draining runtime reset, resumes through the selected runtime, and only then sends with the selected Runtime Segment provenance.
- **Adoption window:** Session ownership moves to the incoming runtime before awaiting provider adoption, with rollback when resume fails before attachment. This closes the asynchronous window where draining-runtime events could settle the optimistic new-runtime turn.

Validation: 147 targeted tests passed; typecheck and lint passed; the full suite passed with 8141 tests (139 skipped).

ewen-poch:
Addressed the draining terminal-event issue in `e23f30f`.

- Session ownership now remains with the draining runtime for the entire `resumeSession` adoption, including the incoming runtime's intermediate attached-state emission. Ownership commits only after resume succeeds, so the prior runtime's `stop`/`error` lifecycle can still settle its own turn.
- The renderer now completes adoption before appending the next user message and opening its optimistic run. A draining terminal event therefore cannot accidentally finish the incoming Runtime Segment.
- An adoption-scoped send guard keeps duplicate submissions blocked during the intentionally idle-looking interval before the new run opens.
- After ownership commits, stale non-Artifact events remain isolated; Artifact lifecycle events continue to finalize through their explicit run/prompt/claim provenance.

Validation: 147 targeted ACP/renderer tests passed; typecheck, lint, formatting, and patch checks passed. The unrelated Notebook timing failures seen only under aggregate local load passed when their affected files were rerun serially (258/258); standard PR CI is running independently.

ewen-poch:
Addressed the snapshot-retention issue in `0937826`.

- The coordinator now remembers event ids accepted while a runtime owns the session. After ownership commits, those already-admitted events remain in aggregated snapshots until the runtime's bounded event window drops them, so an ownership broadcast cannot erase a stop, error, message, or tool event before renderer processing.
- Events first emitted by the draining runtime after adoption are still rejected and never admitted, preserving Runtime Segment isolation.
- Remembered ids are pruned against each runtime's current snapshot window to keep retention bounded.

Regression coverage proves both sides of the boundary: pre-commit owner/tool/stop events survive the ownership flip, while a post-adoption Codex tool event is absent. Validation passed with 147 targeted tests, typecheck, lint, formatting, and patch checks.

ewen-poch:
Addressed the composer-restoration regression in `0efbefd`.

- A duplicate submit suppressed while session adoption is in progress now returns a distinct `null` outcome instead of the `undefined` failure outcome.
- `WorkspacePage` keeps the composer cleared for that intentional suppression; genuine failures still restore the draft and attachments.
- The automatic job-analysis adapter maps the suppression outcome to an unsent result, so completed-job notifications remain retryable rather than being marked consumed.

Regression coverage verifies both the adoption guard outcome and the workspace composer behavior. Validation passed with 174 related tests, typecheck, lint, formatting, and patch checks.

ewen-poch:
Both findings are valid and remain unresolved:

- A detached continuation on the same framework/backend needs a guaranteed fresh agent context for a Branch reset instead of sharing the provider-adoption path.
- The adoption window needs a reactive send/edit lock. Treating every suppressed submit as a duplicate can discard a different draft entered before adoption completes.

No additional code changes are being pushed in this pass. Despite green CI, the PR should not merge until these two handoff cases are addressed.

ewen-poch:
Addressed both handoff findings in `d3580df`.

- **Detached Branch context:** Branch reset now distinguishes a selected-runtime change from simple detachment. An attached selected runtime resets directly. A detached or changed runtime adopts first; when resume reports `contextReset: false`, `resetSessionContext` runs against the newly selected coordinator owner before history replay and prompt dispatch.
- **Different prompt during adoption:** Runtime adoption and Branch reset now publish per-session preparation state. `WorkspacePage` disables draft editing, sending, and message editing for that interval. A synchronous draft/session-key guard closes the React re-render gap, returning before draft clearing or runtime dispatch; therefore a different newer draft remains in the composer instead of being discarded.
- The `null` suppression sentinel and its job-analysis adapter path were removed; `undefined` once again consistently represents an unsent result.

Regression coverage includes detached same-framework reset ordering, cross-framework reset ownership, preparation-state transitions, UI gating, and preservation of a newer draft with only one runtime call. Validation passed: 180 related tests, full suite (8144 passed / 139 skipped), typecheck, lint, formatting, and patch checks.

ewen-poch:
Addressed the adopted-session connection-status finding in `662d4ba`.

- The ownership commit now records the incoming runtime's current snapshot status in `sessionConnectionStatuses`. The attached-state notification remains suppressed during pending adoption, but its status is no longer lost when the resume promise resolves.
- If resume remaps the provider session id, the stale status entry for the requested id is removed together with its old ownership mapping.
- Regression coverage starts with the draining owner recorded as `error`, completes adoption through a `connected` runtime, and verifies that the adopted session is projected as `connected`.

Validation passed: 180 related tests, full suite (8144 passed / 139 skipped), typecheck, lint, formatting, and patch checks.

ewen-poch:
Addressed both findings in `8243762` and `aad74fd`.

- **Late draining terminal events:** The incoming runtime may attach, but `resumeSession` now waits for the prior owner's active prompt to clear before committing ownership or resolving to the renderer. A `stop`/`error` emitted after the incoming provider's resume completes therefore still settles the old Runtime Segment; only after that drain can the renderer append the selected runtime's new turn. Later old-runtime events remain filtered after ownership commits.
- **Prepared-send failure after navigation:** Failure restoration now checks the original `sendRequestKey`. If that conversation is still selected, it restores the live composer; otherwise it writes the captured doc and attachments into `composerDraftsRef[sendRequestKey]`, leaving the newly selected conversation untouched.

Regression coverage proves the incoming provider can finish resume while the coordinator remains pending, the old terminal event still publishes, ownership commits only after prompt-in-flight clears, and later old events are rejected. A workspace test switches from session A to B during preparation failure and verifies A's doc/attachment are restored only when A is reopened. Validation passed: 182 related tests, full suite (8146 passed / 139 skipped), typecheck, lint, formatting, and patch checks.

ewen-poch:
Addressed both findings in `1372e5b` and `fb0de6e`.

- Incoming adoption now revalidates that the runtime is still registered, not retired, and still the pending adopter after the prior owner's active turn drains. If that condition no longer holds, ownership is not committed and the adoption rejects as superseded.
- Interrupted and overflow-recovery retries now snapshot the transcript, conversation graph, and files revision before removing the unanswered turn. If shared send preparation returns unsent, that projection is restored while preserving the live session's newer runtime and failure metadata.
- A restored interrupted retry remains resumable. When the provider is already attached on the next attempt, the flow skips a redundant provider resume and retries the preserved prompt exactly once.

Validation: 185 related tests passed; the full suite passed with 8,149 tests and 139 skipped. Typecheck, lint (0 errors), Prettier, and `git diff --check` also passed.

ewen-poch:
Addressed the finding in `900449f`.

- Composer drafts now carry a per-conversation user-mutation version. Optimistic send clearing and conversation navigation do not advance it; user text edits and attachment intake, cancellation, or removal do.
- A delayed prepared-send failure restores its captured prompt and attachments only when that version is unchanged. If the user has created newer intent on the same conversation, the newer draft remains untouched.
- Staged files belonging only to the superseded failed request are deleted instead of becoming unreachable after the newer draft wins.

Regression coverage reproduced the overwrite with a newer prompt and attachment, then verifies both remain while only the superseded attachment is cleaned up. Validation passed with 186 related tests, typecheck, lint (0 errors), Prettier, and `git diff --check`.

The local repository-wide run reached 8,149 passing tests and one unrelated schema-parity failure because a concurrent worktree regenerated the shared Prisma client from its modified `PermissionGrant` schema. The isolated PR Verify jobs are running against this branch's own schema and generated client.

ewen-poch:
This finding is valid and remains unresolved.

- `publishedRuntimeEventIds` currently makes every previously admitted event sticky across ownership transfer, including `context-overflow` errors and compaction control events.
- Renderer deduplication is hook-local, so a workspace remount or another renderer can observe the retained event without its prior handled-id state and invoke recovery again against the adopted session.
- The appropriate follow-up is to exclude overflow/compaction control-flow events from cross-owner snapshot retention while preserving the sticky window for terminal and display/history events that the renderer still needs to consume.

No additional code changes are being pushed because the requested review limit has been reached. Despite green CI and a clean merge state, the PR should not merge until this stale control-event replay is addressed or explicitly accepted.

ewen-poch:
Rebased the branch onto the latest `origin/main` (`08b6a1a`) with no conflicts. The rewritten head is `3d2fdcb` and was updated using an exact `--force-with-lease` against the prior remote head.

Post-rebase validation:

- 187 related ACP/workspace tests passed.
- Typecheck passed.
- Lint passed with 0 errors (23 existing warnings).
- The local repository-wide run passed 8,169 tests with one schema-parity failure caused by the known concurrent-worktree Prisma generated-client contamination; isolated PR CI is expected to validate this branch's own generated client.

ewen-poch:
Implemented the accepted review fixes in f3fa281:

- Added an explicit renderer event-drain barrier after runtime adoption and before history replay or creation of the next optimistic run.
- Routed edited resends through the common preparation path so selected-runtime adoption happens before reset, the preparation lock covers the full operation, and Branch truncation happens only after preparation succeeds.
- Removed previously admitted overflow and compaction control events from snapshots after ownership transfer while retaining drain-critical chat, tool, and stop events.
- Required run, prompt, and claim provenance for cross-owner Artifact events.
- Replaced Artifact writer substring detection with exact known identities and added a negative delete_artifact_file regression test.

Validation:
- Focused ACP and Artifact tests: 155 passed.
- npm run typecheck: passed.
- npm run lint: 0 errors; repository-existing warnings only.
- Independent Standards review: no findings.
- Independent Spec review: all six accepted findings satisfied.

The full local suite completed with 8,172 passing tests and one unrelated environment failure in prisma-client.test.ts because this worktree resolves a parent-worktree generated Prisma client containing PermissionGrant while the checked-out schema does not. The affected focused suites are green; CI will run with a fresh generated client.

ewen-poch:
Rebased onto the latest main (920b549) and integrated the new specialist session-binding flow with the runtime handoff fixes. The resolved path preserves specialistId during create/resume, keeps the specialist reconfigure barrier ahead of send, and retains adoption preparation locking plus the renderer event drain before opening the new optimistic run. I also updated the rebased WorkspacePage fixture and resume-call assertions for the new main APIs.

Validation after the rebase:
- npm run typecheck: passed
- targeted ACP/workspace/specialist suite: 192 tests passed
- npm run lint: 0 errors (existing warnings only)
- full npm test: 8,474 passed, 139 skipped; the only local failure is the known shared generated Prisma client mismatch where PermissionGrant exists in the parent client but not this checkout runtime DDL. CI performs a fresh dependency/client generation.

ewen-poch:
Addressed the interrupted-resume handoff finding in 0303a0e.

- The interrupted-turn resend now receives the same preparation-state callback and renderer event-drain callback as ordinary sends and edited resends.
- The hook keeps the composer/send/edit gate active through the post-resume preparation window.
- Accepted events from the prior owner are drained before the preserved interrupted turn opens its replacement run.
- Regression coverage verifies the preparation true/false lifecycle and that the drain completes before sendPrompt.

Validation: the targeted ACP/workspace/specialist suite passes with 192 tests, typecheck passes, and lint reports 0 errors.

ewen-poch:
Addressed the valid deleted-session finding in `afa01e0`.

- Existing-session sends now re-read the target after asynchronous adoption/reset and abort if deletion or hydration removed it, so the generic append seam cannot recreate the conversation.
- Interrupted retries explicitly require the previously captured durable session both before and after common-send preparation. Deletion during the outer reconnect therefore cannot trigger a second stale runtime adoption, recreate local state, or dispatch a prompt.
- Regression coverage deletes a session while normal adoption is pending and while interrupted reconnect is pending, then verifies no session resurrection, no orphaned prompt, and no second resume.

I did not apply the rollback suggestion because its premise does not match the current paths: `restoreRemovedTurnProjection` is called only after `removeMessage` in interrupted/overflow retries. `removeMessage` changes `messages`, `conversationGraph`, and `filesRevision`, which are exactly the fields restored. Edited resend does not use this helper; reset failures occur before `truncateSessionFromMessage`, while post-append prompt failures intentionally preserve the failed edited branch.

- 6 affected test files: 194/194 passed.
- `npm run lint`: 0 errors (24 pre-existing warnings).
- Full `npm test`: 560 files and 8,476 tests passed; the sole local failure is the known shared generated-Prisma-client mismatch (`PermissionGrant` columns absent from this worktree runtime DDL), unrelated to these renderer changes and avoided by CI fresh generation.
- Independent Standards and Spec review: no remaining P1/P2 findings.

PR #519, opened 2026-07-30
fix(session-persistence): prevent stale Artifact session saves

ewen-poch:
## Problem

Renderer store autosaves used a local queue, while Artifact finalization called the Session IPC save path directly. An older queued snapshot could therefore arrive after the latest Artifact-bound Session, replace the durable JSON, and make provenance capture fail with Artifact-owning Message is outside its bound Branch.

## Proposed change

- Route store saves, manifest saves, and Artifact finalization saves through one ordered renderer persistence seam.
- Validate every finalized Artifact Message binding before replacing durable Session JSON, including Versions that already have an immutable Message Snapshot.
- Add regression coverage for renderer ordering, the Artifact event path, pre-write rejection, and snapshot-before/snapshot-after binding validation.

## Scope and non-goals

- No schema migration, UI change, or Artifact format change.
- This does not introduce a general Session revision protocol; the fail-closed main-process guard is scoped to finalized Artifact Message bindings.
- This does not rewrite historical Session files that are already corrupt.

- Store autosaves cannot overtake the latest explicit Artifact save -> npm test -- src/renderer/src/lib/session-persistence/session-persistence.test.ts -> passed in the 17-test file.
- Artifact finalization shares the ordered save path -> npm test -- src/renderer/src/lib/acp/workspace-events.test.ts -> passed in the 58-test file.
- A stale finalized binding cannot overwrite durable Session JSON -> npm test -- src/main/session-persistence/coordinator.test.ts -> passed in the 57-test file.
- Stale graphs are rejected before and after Message Snapshot creation -> npm test -- src/main/artifacts/provenance-message-snapshot.test.ts -> passed in the 8-test file.
- Combined targeted suite -> 4 files, 140 tests passed.
- npm run typecheck -> passed.
- npm run lint -> passed with 0 errors; 24 existing warnings are outside the changed files.
- npm test -> 544 files and 8140 tests passed; 11 files and 139 tests skipped by project configuration.

One full-suite run hit unrelated timing failures in ProjectFilesView and notebook force-stop tests. Both files then passed independently, 209/209, and the subsequent full-suite rerun passed.

Uncovered risk: there is no real Electron end-to-end stress test for multi-window writers, and the main guard adds a bounded finalized-Version lookup to Session saves.

## Review focus

- Whether all renderer Session write paths share the intended queue.
- Whether finalized Artifact binding validation is correctly placed before JSON replacement.
- The query scope and fail-closed behavior for Sessions with long Artifact histories.

### DIFF
--- a/src/main/artifacts/provenance-message-snapshot.test.ts
+++ b/src/main/artifacts/provenance-message-snapshot.test.ts
@@ -159,6 +159,17 @@ describe('Provenance Message snapshots', () => {
       getClient: () => Promise.resolve(client)
     })
 
+    const staleSession = structuredClone(session)
+    if (!staleSession.conversationGraph) {
+      throw new Error('Expected the Session conversation graph.')
+    }
+    staleSession.conversationGraph.branches[0].headMessageId = 'prompt-1'
+    staleSession.messages = staleSession.messages.filter((message) => message.id === 'prompt-1')
+    await expect(snapshots.validateFinalizedMessageBindings(staleSession)).rejects.toThrow(
+      'Artifact-owning Message is outside its bound Branch.'
+    )
+    await expect(snapshots.validateFinalizedMessageBindings(session)).resolves.toBeUndefined()
+
     const streamingSession = structuredClone(session)
     const streamingMessage = streamingSession.conversationGraph?.messages.find(
       (message) => message.id === 'message-1'
@@ -189,6 +200,10 @@ describe('Provenance Message snapshots', () => {
       messageCount: 2,
       checksum: expect.stringMatching(/^[a-f0-9]{64}$/u)
     })
+    await expect(snapshots.validateFinalizedMessageBindings(staleSession)).rejects.toThrow(
+      'Artifact-owning Message is outside its bound Branch.'
+    )
+    await expect(snapshots.validateFinalizedMessageBindings(session)).resolves.toBeUndefined()
     const read = await provenance.getVersionProvenance({
       projectId: 'project-1',
       appSessionId: 'session-1',
--- a/src/main/artifacts/provenance-message-snapshot.test.ts
+++ b/src/main/artifacts/provenance-message-snapshot.test.ts
@@ -2,7 +2,7 @@ import { createHash } from 'node:crypto'
 import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { dirname, join } from 'node:path'
-import { afterEach, describe, expect, it } from 'vitest'
+import { afterEach, describe, expect, it, vi } from 'vitest'
 
 import {
   createLinearConversationGraph,
@@ -158,6 +158,7 @@ describe('Provenance Message snapshots', () => {
       storageRoot,
       getClient: () => Promise.resolve(client)
     })
+    const findVersions = vi.spyOn(client.artifactVersion, 'findMany')
 
     const staleSession = structuredClone(session)
     if (!staleSession.conversationGraph) {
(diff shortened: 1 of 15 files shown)


ewen-poch:
Addressed the stale-save recovery concern in 2282cf5.

- The renderer now identifies only the safe non-graph fields actually changed by each autosave (title, pin state, and session preference fields) and includes that explicit intent with the save.
- Finalized Artifact binding mismatches now use a typed conflict. On that conflict, main reloads the latest durable Session, rebases only those explicitly changed fields onto the authoritative graph, revalidates the result, and then saves it. Generic validation failures and conflicts without safe-field intent still fail closed.
- Added renderer, coordinator, and IPC coverage for the rebase path. Type checks and targeted tests pass; the full suite passes 8,481 tests with one unrelated Prisma runtime-DDL baseline failure that reproduces on current main.

ewen-poch:
Addressed the renderer graph refresh concern in fc22b10.

Autosaves that carry safe-field rebase intent now apply the returned durable Session as a full persisted projection only when the submitted source object is still current. This refreshes the renderer to the authoritative graph after a conflict rebase. If a newer renderer mutation has already replaced that source, the existing upload-identity-only merge remains in effect so the older response cannot roll back newer live state. Added regression coverage for both the authoritative replacement and newer-state fallback paths.

ewen-poch:
Addressed both persistence concerns in b87b0bf.

- Pre-save binding validation now distinguishes `FinalizedArtifactBindingConflictError` from transient provenance lookup failures. Confirmed binding conflicts still fail closed/rebase as before; transient database failures are logged and no longer prevent the authoritative Session JSON write, preserving the existing JSON-first durability path.
- Failed session targets now retain the safe-field rebase intent captured by the attempted snapshot. Explicit retries feed that intent back into the forced save instead of recalculating from the already-advanced attempted baseline. The retained intent is accumulated across consecutive failures and cleared only after success or target removal.

Added regression coverage for transient validation availability and forced retry intent.

ewen-poch:
Addressed the binding-validation hot path in 026a0e1.

- The coordinator now fingerprints only the graph topology used to resolve finalized Artifact bindings. Streaming content/status updates and safe metadata changes reuse a confirmed fingerprint, while topology changes still validate before the Session JSON write.
- Artifact/session mutations and catalog reloads invalidate the fingerprint, so newly finalized bindings and externally refreshed authority are revalidated. Transient validation failures are not cached.
- Both validation and pending snapshot capture now select only `rootFrameId`, `agentFrameId`, `messageBranchId`, and `messageId` from Prisma instead of loading large evidence/content fields.

Added regression coverage for streaming saves, topology and Artifact-binding invalidation, transient lookup retries, and the narrow Prisma select.

ewen-poch:
Addressed the specialist binding conflict path in 60b5e66.

- Added a named `saveSessionSpecialistBinding` coordinator entry point that always delegates with `conflictRebaseFields: ["specialistId"]`.
- Routed the main-process specialist persistence caller through that entry point, so a finalized Artifact binding conflict reloads the authoritative Session graph and rebases only the requested specialist UUID, including clearing it with `undefined`.
- Added regression coverage proving the specialist change survives while the authoritative message graph wins.

Focused tests, type checking, formatting, lint, and an independent Standards/Spec review passed. The full suite remains at one previously reproduced PermissionGrant runtime-DDL baseline failure, with 8,485 tests passing.

ewen-poch:
Addressed cross-window safe-field propagation in 32f8807.

`rebaseSafeSessionFields` now assigns `Math.max(authoritative.updatedAt, submitted.updatedAt) + 1`, so every conflict-rebased title, specialist binding, pin, or preference update has a revision strictly newer than both competing snapshots. Other lifecycle clients therefore take the full newer-session path instead of the equal-revision upload-identity-only merge.

The regression test now asserts the rebased revision is greater than both inputs. Coordinator, renderer session-store, and lifecycle-sync tests pass, as do type checking, formatting, lint, and an independent Standards/Spec review. The full suite still has only the previously reproduced PermissionGrant runtime-DDL baseline failure, with 8,485 tests passing.

PR #520, opened 2026-07-30
fix(notebook): activate R inventory and log installers

ewen-poch:
## Problem

On Windows, Conda-backed R environment inventory and fingerprint probes launched `Rscript.exe` with the raw main-process environment instead of the activated Conda PATH used by the kernel. After an agent package operation, the recovery probe could therefore fail to resolve DLLs such as `Library/bin/libgcc_s_seh-1.dll` and block kernel restart even when the environment itself was intact.

Installer and probe failures also lacked bounded, credential-safe child-process details in the main-process `main.log`, making nightly-package failures difficult to diagnose.

## Proposed change

- Give managed and external Conda R inventory/fingerprint probes the same activated PATH layout used by the kernel, including the Windows Conda DLL directories.
- Preserve the existing environment cache/recovery key while carrying the Conda prefix as execution context.
- Add structured installer and probe diagnostics to `main.log`, including exit metadata and bounded stdout/stderr/cmd fields.
- Redact URL credentials, bearer tokens, common secret parameters, and OpenAI-style secret keys from diagnostics.
- Cover the default subprocess boundary and the real rotating `main.log` sink, not only helper functions or injected loggers.

## Scope and non-goals

- Keep strict post-install recovery behavior; this change fixes the environment used by recovery probes rather than weakening the guard.
- Do not reset or reinstall user environments.
- Do not change renderer or spreadsheet behavior.
- Do not change the persisted cache/recovery target identity.

| --- | --- | --- |
| Windows Conda R fingerprint/inventory default spawns receive the activated DLL PATH | `npm test -- src/main/notebook/environment-state-tracker.test.ts src/main/notebook/runtime-diagnostics.test.ts src/main/notebook/runtime-service.test.ts src/main/notebook/runtime-service-logging.integration.test.ts` | Passed: 4 files, 173 tests |
| Installer records reach the real main-process `main.log`; diagnostic fields are bounded and redacted | Same focused test command | Passed, including disk-backed JSONL integration coverage |
| Legacy dirty/recovery state remains addressable after the execution-context addition | Same focused test command | Passed |
| TypeScript validation | `npm run typecheck` | Passed |
| Lint validation | `npm run lint` | Passed with 0 errors and 24 warnings in unrelated existing files |
| Patch whitespace validation | `git diff origin/main...HEAD --check` | Passed |
| Full repository suite | `npm test` | **Not green:** 8,142 passed, 3 failed, 139 skipped. All failures are in the unchanged `src/renderer/src/pages/workspace/previews/spreadsheet-cache-policy.test.ts` (cached row count, retained worksheet count, and pending-request cap). |

Independent standards and spec reviews found no remaining code-hunk or requirement gaps after the default-spawn and real-log-sink tests were added.

## Review focus

- Parity between the R probe PATH and the established kernel Conda activation PATH on Windows.
- Diagnostic size bounds and credential redaction before records reach `main.log`.
- Compatibility of existing dirty/recovery state across the added execution context.

## Uncovered risks

- A packaged Windows nightly smoke test is still needed to prove that the real `Rscript.exe` resolves `libgcc_s_seh-1.dll` and that kernel restart recovers after an agent installs an R package.
- The repository's required full `npm test` gate remains blocked by the three unrelated spreadsheet cache-policy failures above; the PR is Draft until that gate is resolved or accepted by maintainers.

### DIFF
--- a/src/main/notebook/environment-state-tracker.test.ts
+++ b/src/main/notebook/environment-state-tracker.test.ts
@@ -4,7 +4,7 @@ import { join } from 'node:path'
 
 import { afterEach, describe, expect, it, vi } from 'vitest'
 
-import { EnvironmentStateTracker } from './environment-state-tracker'
+import { environmentCaptureProcessEnv, EnvironmentStateTracker } from './environment-state-tracker'
 
 let dataRoot: string | undefined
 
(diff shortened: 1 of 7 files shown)


ewen-poch:
Addressed the Windows `Path` casing issue in 5ae3e26. The R probe environment now resolves the inherited path case-insensitively on Windows, removes all case-variant path keys, and emits one activated `PATH` while preserving inherited system/user entries. The updated test uses a `Path`-cased input and asserts the normalized activated `PATH`; the tracker tests pass 20/20, along with typecheck, ESLint, and Prettier checks.

ewen-poch:
Addressed in d62bd65. The `package installer completed` record is now emitted only after `refreshAfterPackageMutation()` and any final `InstallResult` rewrite, so an inventory verification failure is logged at `warn` with `ok: false` and the same error returned by `managePackages()`. The installer-only duration is still captured immediately after the installer subprocess completes. A regression assertion now covers the refresh-failure path; the focused suite passes 173/173, along with typecheck, ESLint, and Prettier checks.

PR #527, opened 2026-07-30
fix(update): recover Windows in-place upgrades

### LINKED ISSUE #512: [Bug]: In-place upgrade repeatedly fails to remove old application files on Windows
### Preflight checklist

## Bug Description

Open Science cannot be upgraded normally on Windows.

Each upgrade attempt fails with the following error:

> Failed to uninstall old application files. Please try running the installer again.: 2

Restarting the computer and running the installer again does not resolve the issue. However, manually uninstalling the existing version and then reinstalling Open Science works successfully.

## Actual Result

The installer repeatedly fails to remove the old application files, preventing an in-place upgrade.

## Expected Result

The installer should automatically remove or replace files from the previous version and complete the upgrade without requiring users to manually uninstall the application.

## Workaround

Manually uninstall the existing version, then install the new version.

## Additional Information

- Platform: Windows
- Restarting the computer does not resolve the issue.
- Manual uninstall and reinstall works normally.
- Screenshot attached.

Image

Image

[main(8).log](https://github.com/user-attachments/files/30535592/main.8.log)

### Steps to reproduce

/

### Operating system

Windows

### App version

0.8.0

### Provider / model

_No response_

```shell

```

ewen-poch:
## Problem

Windows in-place upgrades can stop at `Failed to uninstall old application files...: 2`. The attached log also records a shutdown-time `EventEmitter.removeListener` receiver error. On custom-drive installs, electron-builder's updated uninstaller can move the old tree across volumes through `%TEMP%`; the reported configuration additionally stores `OpenScience` data inside the installation tree, so that directory must be protected before any old-uninstaller attempt.

Closes #512.

## Proposed change

- Call `BrowserWindow.removeListener` / `off` with the window as their receiver during find-overlay teardown.
- Protect a nested `<install>\OpenScience` data root in a deterministic same-volume sibling before the first old-uninstaller pass, then restore it on success, failure, cancellation, GUI shutdown, and elevated inner-installer flows.
- On a real uninstall failure, stop install-directory processes and retry once with child `TEMP`/`TMP` on the installation volume.
- Treat a non-zero old-uninstaller result as completed only when the old executable is already gone, including after the retry.
- Add EventEmitter regression coverage and source-contract coverage for the NSIS hook order, data preservation, environment restoration, and electron-builder insertion points.

## Scope and non-goals

This changes Windows NSIS upgrade recovery and the find-overlay shutdown path only. It does not alter update rollout, version selection, application data-root configuration, or manual uninstall behavior.

A physical Windows custom-drive upgrade was not available in this environment. The NSIS include was compiled successfully into a minimal x64 Windows installer with electron-builder 26.15.3, but a real `D:`-drive upgrade remains the main manual/CI validation target.

- Destroying the overlay against a real `EventEmitter` no longer throws and removes the resize listener.
- Registered nested data roots are moved outside the install tree before the first old uninstaller and restored on every covered exit path.
- Recovery uses same-volume `TEMP`/`TMP`, restores the parent installer's environment, retries exactly once, and rechecks the executable sentinel.
- `npx vitest run src/main/find-overlay.test.ts build/packaging.test.ts`: 36 passed.
- `npm run lint`: passed with 24 pre-existing warnings and 0 errors.
- `npm test`: 546 files passed, 8,163 tests passed, 139 skipped.
- Minimal `electron-builder --win nsis --x64` compile: passed.

## Review focus

Please focus on the NSIS preserve/restore behavior across assisted elevation and the two registry uninstall passes, plus restoration of `TEMP`/`TMP` after the retry. The remaining validation risk is runtime behavior on a physical Windows machine with the application and data root on a non-system drive.

### DIFF
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -2,14 +2,154 @@
   nsExec::ExecToLog '"$SYSDIR\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -NonInteractive -ExecutionPolicy Bypass -File "$INSTDIR\resources\windows-runtime-cache-uninstall.ps1"'
 !macroend
 
+!ifndef BUILD_UNINSTALLER
+!define MUI_CUSTOMFUNCTION_ABORT restorePreservedOnAbort
+
+Var perMachineInstallDirCache
+Var perUserInstallDirCache
+Var perMachineDataBackup
+Var perUserDataBackup
+Var dataProtectionFailed
+Var dataRestoreFailed
+
+# Move a data root outside the installation before the OLD uninstaller sees it. A deterministic
+# sibling path lets an elevated inner installer or a later retry recover data left by an
+# interrupted outer installer without relying on process-local registers.
+!macro preserveNestedDataRoot DIR BACKUP SLOT
+  StrCpy ${BACKUP} ""
+  ${if} "${DIR}" != ""
+    ClearErrors
+    GetFullPathName $R2 "${DIR}\.."
+    ${if} ${Errors}
+      DetailPrint `Could not safely preserve "${DIR}\OpenScience"; its parent path could not be resolved.`
+      MessageBox MB_OK|MB_ICONSTOP "Open Science could not safely protect its data folder before updating.$\r$\nThe existing data was left untouched."
+      StrCpy $dataProtectionFailed "1"
+    ${else}
+      StrCpy ${BACKUP} "$R2\.open-science-update-data-${SLOT}"
+      ${if} ${FileExists} "${DIR}\OpenScience\*.*"
+        ${if} ${FileExists} "${BACKUP}\*.*"
+          DetailPrint `Could not safely preserve "${DIR}\OpenScience" because the backup path already exists: ${BACKUP}`
+          MessageBox MB_OK|MB_ICONSTOP "Open Science found both the current data folder and an earlier update backup.$\r$\nNo data was changed. Please inspect:$\r$\n${BACKUP}"
+          StrCpy ${BACKUP} ""
+          StrCpy $dataProtectionFailed "1"
+        ${else}
+          ClearErrors
+          Rename "${DIR}\OpenScience" "${BACKUP}"
+          ${if} ${Errors}
+            DetailPrint `Could not safely preserve "${DIR}\OpenScience"; leaving the existing installation untouched.`
+            MessageBox MB_OK|MB_ICONSTOP "Open Science could not safely protect its data folder before updating.$\r$\nThe existing data was left untouched."
+            StrCpy ${BACKUP} ""
+            StrCpy $dataProtectionFailed "1"
+          ${else}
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the AI review finding about matching HKCU/HKLM registrations in commit 2d961e3. The unelevated outer installer now probes sibling-directory write access before protecting a shared path; machine-owned paths are deferred to the elevated inner installer, while user-writable shared paths remain protected immediately. Added a regression contract test for the UAC deferral path. Validation: 33 targeted tests passed, 8,160 full-suite tests passed, typecheck passed, lint has 0 errors, and a minimal Windows x64 NSIS installer compiled successfully.

ewen-poch:
Addressed the P1 UAC handoff finding in commit 0dd3c8d. Normal GUI shutdown no longer restores deterministic data backups, so an unelevated outer installer cannot race the elevated inner installer's later HKCU uninstall pass. Explicit user abort, installation failure, and fatal recovery paths still restore immediately; an unexpected interruption leaves the deterministic backup for the next installer to adopt. Added a contract assertion that .onGUIEnd is not registered. Validation: 33 targeted tests and 8,160 full-suite tests passed, and the minimal Windows x64 NSIS installer compiled successfully.

ewen-poch:
Addressed the P2 shared-path comparison finding in commit 12663ca. Cached HKCU and HKLM InstallLocation values are now normalized by removing harmless trailing separators while preserving drive roots; NSIS string equality is already case-insensitive, so case and separator variants converge before the UAC probe, backup aliasing, and handoff restore checks. Added a regression contract test that pins normalization before the first shared-path branch. Validation: 34 targeted tests and 8,161 full-suite tests passed, typecheck passed, lint has 0 errors, and a minimal Windows x64 NSIS installer compiled successfully.

ewen-poch:
Addressed the P2 final install-mode finding in commit 360c628. A distinct HKLM data root is no longer touched during interactive customInit. Interactive assisted installs now protect it in the instfiles preflight only after the final all-users selection, while preserving electron-builder's existing install-directory sanitizer; silent installs use the mode already resolved by initMultiUser. Current-user installs therefore cannot be blocked by an unrelated machine installation. Added a regression contract test for the interactive and silent paths. Validation: 35 targeted tests and 8,162 full-suite tests passed, typecheck passed, lint has 0 errors, and a minimal Windows x64 NSIS installer compiled successfully.

ewen-poch:
Addressed the P1 restore-before-retry finding in f5028d8. Both uninstall-result hooks now keep the registered data backup outside the installation tree until non-zero-exit recovery completes, so an OpenScience directory recreated by the old process cannot short-circuit process termination and retry. If recovery also preserves data created by that process, it keeps the additional copy at its unique sibling path instead of overwriting the authoritative backup. Added regression coverage for both hooks. Validation: 36 targeted tests and 8,163 full-suite tests passed; typecheck passed; lint has 0 errors; and the Windows x64 NSIS installer compiles successfully.

PR #529, opened 2026-07-30
feat(previews): add TIFF image rendering

ewen-poch:
## Problem

TIFF artifacts are currently treated as generic files. Browser-native image preview cannot decode TIFF, so selecting a `.tif` or `.tiff` artifact shows “Preview unavailable” even though it is an image.

## Proposed change

- Detect TIFF by extension or MIME type and route it to a dedicated image preview.
- Decode TIFF pages in an isolated worker with `tiff@7.1.3`, then render them through the same zoomable canvas experience used by regular images.
- Terminate an in-flight decode worker when page navigation cancels its request, then recreate the session for the new page so synchronous work cannot block later requests.
- Keep multipage navigation available when a later page cannot be decoded, allowing users to return to pages that already rendered successfully.
- Render the first TIFF page in generated-artifact cards, using the same filled, top-aligned thumbnail treatment as regular images while keeping the full PreviewPanel image contained.
- Gate thumbnail work by viewport proximity, cap it at two concurrent decodes, apply thumbnail-specific admission budgets, and downsample in the worker to a 512px maximum dimension before transferring pixels to the renderer.
- Allow renderer fetches from the managed `open-science-preview:` protocol so worker-backed TIFF decoding can read the acquired snapshot.
- Support classic strip-based RGB, grayscale (including 16-bit and floating-point), palette, multipage, uncompressed, LZW, and Deflate TIFFs.
- Bound file size, dimensions, pixel count, decoded memory, metadata traversal, worker lifetime, and compressed-strip expansion before the full decode.
- Bind strict managed-file inspection, acquisition, and protocol streaming to the same `dev`/`ino`/`mtimeNs` identity and hold the admitted file handle while streaming.

## Scope and non-goals

This PR covers classic strip-based TIFF preview inside the existing artifact preview surface. BigTIFF, tiled TIFF, floating-point WhiteIsZero, and compression schemes not handled by the selected decoder remain unsupported and fall back to the existing stable error/download experience.

- TIFF detection, PreviewPanel routing, generated-card thumbnails, thumbnail `cover` versus panel `contain`, CSP access, page controls, image zoom, managed-resource lifecycle, strict acquisition, viewport gating, bounded concurrency, worker downsampling, in-flight page cancellation, bad-page recovery, visibility cleanup, and canvas-failure fallback -> focused Vitest run over the 15 affected test files -> **222 tests passed**.
- Raw/LZW/Deflate decoding, 16-bit/Float32 color and independent alpha normalization, MinIsWhite color and alpha handling, palette and multipage behavior, both Deflate tags, LZW code-width/final-slot boundaries, explicit unsupported compression/tiled/BigTIFF rejection, plus malformed metadata and compression-bomb rejection -> included in the focused run and full suite -> **passed**.
- Same-size/same-timestamp inode replacement between inspection and acquisition -> managed-preview resource regression -> **passed**.
- Node and renderer types -> `npm run typecheck` -> **passed**.
- Repository lint -> `npm run lint` -> **passed with 0 errors** (23 pre-existing warnings, none in changed files).
- Worker-aware production bundle -> `./node_modules/.bin/electron-vite build` -> **passed**; emitted a separate TIFF preview worker bundle.
- Full regression suite -> `npm test -- --maxWorkers=4` -> **551 files passed / 11 skipped; 8185 tests passed / 139 skipped**. (An unconstrained run first hit unrelated worker-pool/time-limit failures; all affected files passed separately, and the bounded-concurrency rerun passed.)
- Final validated commit -> `ed1b38d`.

Uncovered risk: validation uses representative real TIFF fixtures and automated renderer behavior, but does not include a broad manual visual smoke test against third-party TIFF corpora. Unsupported TIFF variants intentionally surface the existing recoverable preview error.

## Review focus

- Worker termination and managed-resource cleanup across success, retry, page changes, timeout, and decode failure.
- Preflight limits for metadata traversal and LZW/Deflate expansion before third-party decode.
- Consistency of TIFF routing with existing image artifact behavior.
- Generated-card thumbnail cropping versus full-image containment in PreviewPanel.
- Thumbnail admission budgets, worker downsampling, concurrency scheduling, and cleanup when cards leave the viewport.
- Strict inspect-to-acquire file identity and same-handle protocol streaming.


ewen-poch:
Addressed the actionable TIFF review points:

- Aborted worker requests now remove themselves from the pending map, clear their timeout, and unregister the abort listener. A regression test confirms that an aborted page request cannot terminate a later decode that reuses the same worker session.
- Float32 grayscale/RGB intensity and alpha now use independent sample ranges. A valid Float32 grayscale-plus-alpha TIFF regression fixture confirms that an opaque alpha value remains fully opaque even when the image intensity range is outside `[0, 1]`.
- I did not add another MinIsWhite inversion in `convertToRgba()`: `tiff@7.1.3` already performs the PhotometricInterpretation=0 inversion in its decoder before returning `ifd.data`, so a second inversion would reintroduce the negative-image bug. I added a MinIsWhite regression test that exercises the dependency path and confirms the final RGBA luminance.

Validation: 202 focused tests passed, the full suite passed with 8,177 tests, and typecheck, lint, and the production build passed.

ewen-poch:
Addressed the valid dictionary-capacity issue:

- Preflight now permits allocation of the legal dictionary entry at index 4095 and rejects only an attempted entry beyond the 4096-entry table.
- I kept the code-width jumps at 511/1023/2047. TIFF LZW uses the early-change convention, and these values match the `tiff@7.1.3` decoder's `bitJumps` table. Moving them to 512/1024/2048 would make preflight parse different stream boundaries from the decoder.
- Added a regression stream that crosses the 9→10→11→12-bit boundaries and allocates slot 4095.

Validation: 203 focused tests passed, the full suite passed with 8178 tests (139 skipped), typecheck passed, lint passed with no errors, and the production Electron build passed.

ewen-poch:
I am not changing the full-table behavior because a regular code after the final dictionary allocation is not a conformant TIFF LZW transition in the decoder contract used here:

- The official libtiff decoder states that once the table reaches this state, the next valid code is `EOI` or `CLEAR`; a regular code is rejected on the next attempted registration ([pinned source](https://github.com/libsdl-org/libtiff/blob/d6217f352cd5eb388a72e44053c985ec15b3375a/libtiff/tif_lzw.c#L518-L528)). Its encoder likewise emits `CLEAR` when the table is full ([pinned source](https://github.com/libsdl-org/libtiff/blob/d6217f352cd5eb388a72e44053c985ec15b3375a/libtiff/tif_lzw.c#L1231-L1241)).
- `tiff@7.1.3`, which performs the actual decode in this PR, has the same behavior: it accepts slot 4095, then rejects an attempted append beyond the 4096-entry table. Allowing that transition only in preflight would therefore move the same failure into the worker decoder rather than make the TIFF previewable.
- The existing boundary regression already verifies that slot 4095 followed by `EOI` is accepted. `CLEAR` is handled before the capacity check and resets the table.

The preflight check remains aligned with both the production decoder and libtiff's conformant transition behavior.

ewen-poch:
Applied the actionable follow-up review findings:

- Fixed WhiteIsZero grayscale+alpha rendering. `tiff@7.1.3` inverts every returned sample for WhiteIsZero, including `ExtraSamples`; the RGBA conversion now restores alpha while preserving the dependency's luminance inversion. A real uncompressed two-component fixture verifies exact grayscale and alpha output.
- Explicitly reject floating-point WhiteIsZero before image-data decode. The pinned decoder applies integer-style inversion to Float32/Float64 data and loses precision irreversibly, so a stable unsupported-preview result is safer than corrupted pixels.
- Added direct regression coverage for raw/uncompressed decode plus unsupported compression, tiled TIFF, and BigTIFF rejection.

I kept the other reviewed trade-offs unchanged: Float32 range scanning runs in the worker under the existing pixel limits; error mapping has a safe fallback; nearest-neighbor downsampling intentionally prioritizes thumbnail cost; and the shared `maxBytes`/strict-acquire branches encode managed-resource policy rather than TIFF-specific behavior.

Final validation after the last edit: 208 focused tests passed; the full suite passed with 551 files / 8183 tests (11 files / 139 tests skipped); typecheck passed; lint passed with 0 errors; and the production Electron build passed. Independent Standards and Spec reviews found no remaining actionable issues.

ewen-poch:
Review assessment:

- The LZW finding is not actionable. The preflight intentionally matches the pinned `tiff@7.1.3` decoder and libtiff behavior: slot 4095 is accepted, while a subsequent regular code would require an out-of-bounds dictionary registration and is rejected. Allowing it only in preflight would merely defer the same failure to the production decoder.
- The managed-preview identity finding is valid. The current resource pins `dev`/`ino` from `acquire()`, but the earlier `inspect()` snapshot contains only size and mtime, so a same-size, same-mtime inode replacement between those steps is not detected. The correction should carry `dev`/`ino` in the inspected snapshot and reject an identity mismatch during acquisition, with a replacement-file regression test.

ewen-poch:
Addressed the valid strict-admission identity finding:

- `inspect()` now returns a main-process-only managed-preview snapshot containing lossless bigint `dev`, `ino`, and `mtimeNs` identity in addition to the renderer-safe size/version metadata.
- `acquire()` rejects any size, timestamp, device, or inode mismatch before minting a capability. Strict protocol streaming continues to `fstat()` the same open handle and now compares the bigint identity through the final verification.
- Added a regression where resolution changes to a different inode with the same size and timestamp; acquisition fails and no capability id is created.
- Kept the identity type in the generic managed-preview module rather than the Office-specific shared contract.

Validation after the final edit: 221 focused tests passed across 15 files; the full suite passed with 551 files / 8184 tests (11 files / 139 tests skipped); typecheck passed; lint passed with 0 errors; and the production Electron build passed. Independent Standards and Spec reviews found no remaining actionable issues.

ewen-poch:
Final review assessment:

- The abort finding is valid. Removing the aborted request from the client pending map prevents its timer from poisoning later requests, but it does not cancel the synchronous decode already running inside the worker. A subsequent page request can remain queued behind that stale decode while its own timeout is active. A complete correction needs session replacement or an explicit worker cancellation protocol coordinated with the renderer cache.
- The page-error finding is also valid. The current decode catch path stores an error at resource scope and disables managed-resource admission, so one unsupported/corrupt page prevents returning to a previously viewable page. The correction should preserve the admitted resource for page-local decode errors and keep error state keyed by page/request, while reserving capability teardown for file-wide read/admission failures.

These renderer lifecycle changes are separate from the completed inspect-to-acquire inode fix and are not included in the current revision.

ewen-poch:
Addressed both P2 findings in `ed1b38d`:

- In-flight page cancellation now fails and disposes the decode session, which synchronously terminates the worker. The renderer removes the disposed session from its cache, so the next page creates a fresh worker instead of queuing behind synchronous work from the old page. A settled session remains reusable for normal page navigation.
- After at least one page has decoded successfully, a later page decode failure is now page-scoped. The managed resource and page controls remain available, so users can navigate back to a valid page. Initial file/read/decode failures remain resource-scoped.

Regression coverage now verifies worker termination on abort and the valid-page → failing-page → valid-page flow. Validation passed: 222 focused tests, typecheck, lint with 0 errors, production build, and the full suite with 8185 tests passed.

ewen-poch:
I’m not changing the preflight for this finding because its downstream-decoder premise does not hold for the pinned `tiff@7.1.3` implementation.

That decoder calls `addStringToTable()` for every non-`CLEAR`/non-`EOI` code. Once `tableLength` is 4096, the next ordinary code writes `stringTable[4096]`, grows the array beyond 4096 entries, and throws its own LZW decoding error. Letting preflight continue at that point would therefore admit a stream that the selected decoder still cannot render and only move the failure to the full decode.

The current boundary regression covers all code-width transitions and the final legal dictionary slot 4095 followed by EOI. Supporting TIFF LZW streams that continue with a saturated dictionary would require changing or replacing the decoder, which is outside this PR’s stated selected-decoder scope.

PR #534, opened 2026-07-31
feat(session-artifacts): add selective artifact downloads

ewen-poch:
## Problem

Session Artifacts can be inspected inside the workspace, but a user cannot export an entire Session collection or choose a subset of its generated files from the Session menu.

## Proposed change

- Add a Download all artifacts action to each Session menu.
- Open a selection dialog that loads every Artifact page, repairs an incomplete Project Files index before presenting the list, selects all items by default, and supports Check all and Uncheck all.
- Add a typed preload and main-process save flow. A single selection uses Save As; multiple selections use one destination folder, avoid overwriting name collisions, and report per-file failures without dropping successful downloads.
- Resolve immutable Artifact Versions within their Project and Source Session, while retaining Session-scoped compatibility for both Project and legacy default-project storage layouts.
- Update the generated web API map and workspace design guidance.

## Scope and non-goals

This PR covers Session Artifact export from the desktop Session menu. It does not export uploads, notebook inputs, Session transcripts, or a packaged Session archive. It does not change Artifact retention or deletion behavior.

- Session menu action, dialog selection, retry, partial-failure state, and Workspace wiring -> npm test -- src/renderer/src/pages/workspace/DownloadSessionArtifactsDialog.interaction.test.tsx src/renderer/src/pages/workspace/WorkspaceSidebar.render.test.tsx src/renderer/src/pages/workspace/WorkspacePage.draft-preservation.test.tsx -> passed.
- Complete pagination and incomplete-index repair -> npm test -- src/renderer/src/pages/workspace/session-artifact-download-data.test.ts -> passed.
- Single and batch saves, collision handling, partial source failures, and legacy namespace fallback -> npm test -- src/main/file-save.test.ts src/main/session-artifact-file-resolver.test.ts -> passed.
- Preload forwarding and generated web API synchronization -> npm test -- src/preload/index.test.ts src/renderer/web/api-map.generated.test.ts -> passed.
- Combined focused validation -> 124 tests passed across 8 affected test files.
- npm run typecheck -> passed.
- npm run lint -> passed with 0 errors; 23 existing warnings are outside this diff.
- npm test -> 8,542 tests passed and 139 skipped across 582 test files.

Independent Standards and Spec reviews found no remaining findings in the final state.

Uncovered risks: native save-dialog behavior is not exercised end-to-end on every supported operating system, and very large real-world Artifact batches were not performance-tested. IPC and filesystem behavior are covered with project-owned unit and interaction tests.

## Review focus

- Project and Source Session scoping for native and legacy Artifact paths.
- Partial batch behavior when a source disappears or a copy fails.
- Index-repair behavior before the dialog treats an empty list as authoritative.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -394,6 +394,7 @@ Workspace-only tokens without a shadcn counterpart, plus shadow tokens. For shar
 - Overlay: `fixed inset-0 bg-black/50`, using Radix state animations for open and close; compact workspace dialogs use `bg-black/25 backdrop-blur-[2px]`.
 - Delete confirmation copy must include the session name and state that session artifacts remain in the project.
 - Rename dialog input uses `h-9 rounded-lg border-border-200 bg-bg-000 text-sm text-text-000 placeholder:text-text-100` and a subtle `ring-border-200/25` focus ring.
+- Session Artifact download dialog: use a scrollable `Dialog` up to `640px` wide and `80svh` high, with a compact header, an artifact checklist, and a persistent footer. Repair an incomplete Project Files index before treating the list as authoritative. Select every Artifact by default; show the selected/total count, file type, and size; disable download when none are selected; and keep failed items selected after a partial batch download.
 
 ### DropdownMenu / Popover / Select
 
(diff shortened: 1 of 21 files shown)


ewen-poch:
Addressed the desktop-only artifact export issue in b6242aa.

- Added `file:save-session-artifacts` to `WEB_UNAVAILABLE_RPC_CHANNELS`, so Web bootstrap omits the capability and direct Web RPC requests are rejected.
- The Session menu now renders the artifact download action only when `saveSessionArtifacts` is available, keeping it hidden in Web mode while preserving the desktop flow.
- Added regression coverage for both the Web RPC boundary and the Sidebar capability behavior.

Validation: `npm run typecheck`, `npm run lint` (0 errors; existing warnings only), focused tests (35 passed), and full `npm test` (8543 passed, 139 skipped).

PR #539, opened 2026-07-31
fix(conversations): harden conversation exports

ewen-poch:
## Problem

Follow-up review of #531 identified three implementation gaps:

- Durable session loading normalizes interrupted `running` and `waiting-permission` states to `error`, so the export guard did not observe the authoritative live state in production.
- An assistant response ending with an unterminated `<think>` block could expose partial private reasoning.
- Prompt-derived document titles kept only the first line/sentence and were truncated at 240 characters, even though truncation belongs at the filename boundary.

## Proposed change

- Require the export service to consult the ACP runtime's live in-flight prompt set, while retaining the durable-status guard as a secondary defense.
- Remove unterminated assistant `<think>` content through the end of the response.
- Derive document titles from the complete whitespace-normalized first user prompt and keep the existing 240-byte limit only in filename sanitization.
- Recognize both renderer and Headless Task API truncated automatic-title formats.
- Add regression coverage for all three review findings.

## Scope and non-goals

This PR only hardens the conversation-export path introduced by #531. It does not change the export UI, Markdown/PDF layouts, or session persistence recovery semantics.

All commands below ran after the last material source edit:

- Live active-session exclusion after durable status normalization -> `npm test -- src/main/session-persistence/conversation-export.test.ts` -> passed (8/8).
- Unterminated reasoning removal and complete prompt-derived titles -> `npm test -- src/shared/conversation-export.test.ts` -> passed (13/13).
- Combined focused behavior -> `npm test -- src/shared/conversation-export.test.ts src/main/session-persistence/conversation-export.test.ts` -> passed (21/21).
- Lint -> `npm run lint` -> passed with 0 errors (23 existing warnings outside the changed files).
- Full regression suite -> `npm test` -> passed (8,605 passed; 139 skipped).
- Formatting -> `npx prettier --check src/shared/conversation-export.ts src/shared/conversation-export.test.ts src/main/session-persistence/conversation-export.ts src/main/session-persistence/conversation-export.test.ts src/main/ipc.ts` -> passed.
- Patch hygiene -> `git diff --check` -> passed.

Independent reviewer confirmation of this final evidence mapping is still pending, so this PR remains a draft.

## Review focus

- The live-state mapping uses `projectName` from the ACP runtime as the project id, matching the runtime/storage boundary's existing convention.
- The reasoning sanitizer intentionally removes from an unmatched assistant `<think>` opener through end-of-response.
- Full prompt text is retained in document metadata/headings; only the filesystem-safe filename is byte-limited.

## Uncovered risks

The Electron Save As dialog and rendered PDF output were not exercised manually; existing automated Markdown/PDF service and rendering tests cover those paths.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -888,7 +888,14 @@ const registerIpcHandlers = async ({
   registerSessionPersistenceIpcHandlers(sessionPersistenceBackend, reviewRepository)
   registerConversationExportIpcHandler(
     createConversationExportService({
-      loadSession: (projectId, sessionId) => sessionRepository.loadSession(projectId, sessionId)
+      loadSession: (projectId, sessionId) => sessionRepository.loadSession(projectId, sessionId),
+      isSessionActive: (projectId, sessionId) =>
+        runtime
+          .getActivePromptSessions()
+          .some(
+            (activeSession) =>
+              activeSession.projectName === projectId && activeSession.sessionId === sessionId
+          )
     })
   )
   registerProjectFilesIpcHandlers(
--- a/src/main/session-persistence/conversation-export.test.ts
+++ b/src/main/session-persistence/conversation-export.test.ts
@@ -45,6 +45,7 @@ const session: PersistedChatSession = {
 
 describe('conversation export service', () => {
   const loadSession = vi.fn()
+  const isSessionActive = vi.fn()
   const showSaveDialog = vi.fn()
   const writeExportFile = vi.fn()
   const createTempDirectory = vi.fn()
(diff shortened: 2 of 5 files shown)


ewen-poch:
Thanks for the review.

- **Headless auto-titles:** Addressed in `12d8e04`. Export title recognition now covers both the renderer's 48-character automatic-title prefix and the Headless Task API's 57-character prefix with its 60-character threshold. A regression test verifies that headless exports recover the complete prompt-derived title.
- **Manual title matching an automatic title exactly:** Not changed in this PR. Reliably distinguishing that explicit rename from an untouched generated title requires persisted title-origin metadata. That data-model change is disproportionate to this ambiguous edge case and is outside this focused export-hardening scope; all non-matching manual titles remain authoritative.

Validation after the change: focused export tests `21/21`, typecheck passed, lint passed with zero errors, and the full suite passed with `8,605` tests (`139` skipped).

PR #540, opened 2026-07-31
feat(workspace): show turn token usage

ewen-poch:
## Problem

Agent responses do not expose the token totals consumed by the complete response turn. This makes it impossible to inspect input, cache, and output usage from the transcript, and Codex's pinned ACP adapter currently reports only its final model request when a turn performs tools.

## Proposed change

- Normalize ACP `PromptResponse.usage` into provider-neutral `inputTokens`, `cacheTokens`, and `outputTokens` totals.
- Combine cache reads and cache writes under the displayed `Cache` total.
- Patch the app-managed Codex ACP adapter to accumulate every model request in the prompt from cumulative counter deltas, without double-counting repeated snapshots or importing resumed-session history.
- Attach the totals only to the final Agent message for the active prompt and persist them in the session JSON/conversation graph.
- Render a compact footer under that final message. When the Agent provides no valid usage, render `Input — / Cache — / Output —` instead of fabricating zeroes.

```mermaid
flowchart LR
  A["Provider token usage"] --> B["ACP prompt response"]
  A --> C["Codex per-request delta accumulator"]
  C --> B
  B --> D["Runtime stop event"]
  D --> E["Final Agent message"]
  E --> F["Session JSON and conversation graph"]
  E --> G["Input / Cache / Output footer"]
```

## Scope and non-goals

- Adds optional, backward-compatible `turnUsage` and `turnUsageUnavailable` message fields.
- No database migration and no new IPC channel.
- Does not estimate missing provider data, calculate monetary cost, or display reasoning/thought tokens.
- Intermediate Agent message IDs from the same prompt do not receive a footer; the final message owns the whole-turn total.

- Whole-turn totals flow from ACP stop events to the final Agent message and durable session graph.
- Codex multi-request prompts sum all requests, ignore duplicate cumulative snapshots, and handle a missing resume baseline.
- Cache read and cache write tokens are combined in the `Cache` value.
- Completed responses show exact formatted totals; missing usage shows em-dash placeholders; streaming and intermediate messages show no footer.
- Old session files remain loadable and malformed usage fields are rejected.

Validation after the final material edit and rebase onto current `main`:

- `npm run lint` — passed with 0 errors and 23 pre-existing warnings in unrelated files.
- `npm test` — 574 files passed, 11 skipped; 8,611 tests passed, 139 skipped.
- Focused persistence/store/UI tests — 109 passed.
- Independent Standards review — no findings.
- Independent Spec review — no findings.

## Review focus

- Codex cumulative-delta behavior in `managed-codex.ts`.
- Final-message ownership of the footer for turns with multiple Agent message IDs.
- Backward compatibility of the optional session JSON fields.

### DIFF
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -7431,6 +7431,13 @@ describe('ACP runtime session management', () => {
     expect(runtime.getSnapshot().contextUsageBySession).toMatchObject({
       s1: { used: 24, size: 200000 }
     })
+    expect(runtime.getSnapshot().events.find((event) => event.kind === 'stop')).toMatchObject({
+      turnUsage: {
+        inputTokens: 31,
+        cacheTokens: 15,
+        outputTokens: 14
+      }
+    })
   })
 
   it('corrects an unpatched external Codex total with its latest request usage at stop', async () => {
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -7486,6 +7486,47 @@ describe('ACP runtime session management', () => {
     })
   })
 
+  it('keeps managed Codex turn totals separate from its latest context snapshot', async () => {
+    const process = new FakeAgentProcess()
+    startFakeAgent(process, ['s1'], {
+      modes: createModes(['read-only', 'agent', 'agent-full-access'], 'agent'),
+      onPrompt: () => ({
+        stopReason: 'end_turn',
+        usage: {
+          totalTokens: 27,
+          inputTokens: 19,
+          cachedReadTokens: 5,
+          outputTokens: 3
+        },
+        _meta: {
+          'open-science/turn-usage': {
+            totalTokens: 60,
+            inputTokens: 31,
+            cachedReadTokens: 8,
+            cachedWriteTokens: 7,
+            outputTokens: 14
+          }
+        }
+      })
+    })
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
+      defaultCwd: '/workspace',
+      spawnAgent: () => asAgentProcess(process),
+      framework: codexFramework
+    })
+
+    await runtime.createSession({ cwd: '/workspace' })
+    await runtime.sendPrompt({ sessionId: 's1', text: 'use a tool and summarize' })
+
+    expect(runtime.getSnapshot().contextUsageBySession).toMatchObject({
+      s1: { used: 24 }
+    })
+    expect(runtime.getSnapshot().events.find((event) => event.kind === 'stop')).toMatchObject({
+      turnUsage: { inputTokens: 31, cacheTokens: 15, outputTokens: 14 }
+    })
+  })
+
   it('uses OpenCode native input and cache-read context usage', async () => {
     const process = new FakeAgentProcess()
     startFakeAgent(process, ['s1'])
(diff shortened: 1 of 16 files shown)


ewen-poch:
Addressed the artifact-only terminal-message lifecycle issue in 5a3603b. Deferred artifacts are now synchronously bound before finishRun selects the token-footer owner, while durable finalization still runs after the run settles. The terminal update also uses a monotonic timestamp so the completed usage payload wins graph synchronization when artifact binding and stop occur in the same millisecond. Added regression coverage for both reported and unavailable usage, including durable persistence. Validation: 8,613 tests passed, 139 skipped; typecheck and lint passed.

ewen-poch:
Addressed the stop-before-artifact ordering issue in 24b9095. The renderer now retains terminal usage scoped to the stopped prompt only when no Agent response exists; a matching late Artifact consumes it while creating the owning Agent message and graph projection. Both reported usage and the unavailable marker are persisted. Added stop-before-artifact regression coverage for both paths. Validation: 8,615 tests passed, 139 skipped; typecheck and lint passed.

ewen-poch:
Addressed both managed-Codex integration issues in cc60ac5. The context patch now uses a distinct local and migrates the previous replacement, so full patch composition produces executable JavaScript. PromptResponse.usage again carries only the latest model-request snapshot for context reconciliation, while accumulated whole-turn usage is transported through the app-owned open-science/turn-usage response metadata and consumed only for the stop-event footer. Added full-composition and runtime-separation regressions. Validation: 8,617 tests passed and 139 skipped; typecheck passed; lint passed with 23 pre-existing warnings.

ewen-poch:
Addressed both whole-turn attribution issues in 0716fee. Codex stop events now accept whole-turn usage only from the app-owned open-science/turn-usage metadata; unpatched or user-managed Codex responses without that metadata are marked unavailable, while providers with turn-accumulated PromptResponse.usage retain the existing fallback. Pending artifact usage is now stored per prompt within each session and consumed only by the matching late artifact, with a bounded per-session cache. Added regressions for external Codex and consecutive artifact-only prompts. Validation: 8,618 tests passed and 139 skipped; typecheck passed; lint passed with 23 pre-existing warnings.

ewen-poch:
Addressed both final-cycle integration issues in 14d07e4. Managed Codex now normalizes omitted cachedInputTokens to zero before cumulative delta calculation and prompt accumulation, including migration of the previously installed turn-usage patch; repeated uncached snapshots no longer invalidate or inflate the total. Headless task completion now projects the terminal stop event into the persisted assistant message, preserving reported usage or the unavailable marker for text and artifact-only completions. Added regressions for uncached repeated updates, patch migration, reported headless usage, and artifact-only unavailable usage. Validation: 8,621 tests passed and 139 skipped; typecheck passed; lint passed with 23 pre-existing warnings.

ewen-poch:
Addressed the cache-field contract gap in 0f1825f. The new regression fixture mirrors the pinned codex-acp 1.1.4 mapping (`cachedInputTokens` -> `cachedReadTokens`), then passes the patched response metadata through `toAcpTurnTokenUsage` and verifies `cacheTokens: 8` together with the input/output totals. This is test-only; production behavior is unchanged.

I kept the 16-entry pending-artifact bound because it is an intentional per-session memory-safety limit for an extremely unlikely 16+ unmatched-turn case. I also kept `turnUsageUnavailable` semantics because a terminal response without matching provider usage should remain explicitly unavailable; a later matching artifact still replaces it with valid usage.

Validation: focused token-usage tests passed (51/51), the four full-suite resource-sensitive failures passed in isolation (246/246), typecheck passed, and lint passed with 0 errors. Two full-suite runs reached 8,618 passing tests but hit the same unrelated timing-sensitive notebook/job-poller failures under full concurrency.

PR #554, opened 2026-07-31
fix(storage): migrate immutable Environment manifests only

ewen-poch:
## Problem

On Windows, a data-root move can fail during provenance validation with:

`Could not verify provenance data: Unfinished Environment operation blocks migration: 6781bee2cc7128dad60abe39756695758edd2dc2f9b42bb53db430253a7d8b43`

The migration validator treated the Environment inventory binding and pending-operation records as durable provenance. That inventory is a mutable cache keyed by the interpreter command, which embeds the managed runtime's absolute data-root path. It cannot be reused after relocation, and stale `.json` / `.tmp` operation files or a dirty binding could therefore permanently block an otherwise recoverable move.

## Proposed change

- Preserve and validate only checksum-addressed immutable Environment manifests during migration.
- Exclude the path-keyed Environment inventory cache from the provenance copy set and migration preflight.
- Fail closed when the authoritative runtime operation journal is corrupt or contains a pending operation.
- Remove only a pre-existing destination Environment inventory cache before staging, while preserving other destination runtime data.
- Keep notebook/artifact manifest-reference validation and the existing Artifact, Upload, session, and SQLite safety checks intact.

## Scope and non-goals

- This does not relax immutable Artifact or Upload provenance validation.
- This does not migrate managed environment prefixes; they are rebuilt under the relocated data root.
- This does not delete unrelated destination runtime files.

| Behavior | Evidence |
| --- | --- |
| The reported Environment key no longer blocks a move because of a dirty/orphaned mutable cache | Storage focused tests cover dirty bindings plus orphan `.json` and `.tmp` operation records |
| Immutable Environment manifests remain available after migration | Migration integration test verifies the manifest is copied |
| Stale destination inventory is not retained while unrelated runtime data survives | Integration test seeds both states and verifies exact cleanup scope |
| Pending or corrupt runtime recovery state still blocks migration | Validator tests plus a migration-level pre-copy regression test |
| Artifact staging remains blocked | Provenance validation regression test |
| Type and repository-wide behavior remain valid | Typecheck, lint, and full test suite |

Checks run after rebasing onto the latest `origin/main`:

- `rtk npm test -- --run src/main/storage/provenance-migration-validation.test.ts src/main/storage/migration-service.test.ts`: 83 passed
- `rtk npm run typecheck`: passed
- `rtk npm run lint`: 0 errors, 23 pre-existing warnings
- `rtk npm test`: 591 files passed, 11 skipped; 8,882 tests passed, 140 skipped; 0 failed
- `rtk git diff --check origin/main...HEAD`: passed

## Independent review

Standards and spec reviews completed with no remaining findings. The first spec review identified that an existing destination inventory could survive a move; the implementation now removes that exact cache before staging, and re-review confirmed the issue is resolved.

## Review focus

Please focus on the mutable-cache versus immutable-manifest boundary, the authoritative runtime-journal safety gate, and the scope of destination cache cleanup.

## Remaining risk

The regression is modeled with cross-platform filesystem integration tests, including the exact reported target key, but has not yet been exercised through a packaged Windows UI migration.

### DIFF
--- a/src/main/storage/migration-service.test.ts
+++ b/src/main/storage/migration-service.test.ts
@@ -19,7 +19,7 @@ import {
   discardStagedCopy,
   MIGRATED_DIRS,
   PROJECT_DATABASE_FILE,
-  RUNTIME_PROVENANCE_DIR,
+  RUNTIME_ENVIRONMENT_MANIFESTS_DIR,
   runDataRootMigration,
   validateNewDataRoot
 } from './migration-service'
@@ -489,7 +489,7 @@ describe('runDataRootMigration (copy phase)', () => {
       expect.objectContaining({
         from: currentDataRoot,
         to: target,
-        dirs: [...MIGRATED_DIRS, RUNTIME_PROVENANCE_DIR]
+        dirs: [...MIGRATED_DIRS, RUNTIME_ENVIRONMENT_MANIFESTS_DIR]
       })
     )
     // runtime/ is excluded from the moved set (non-relocatable; rebuilt on demand).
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the P1 in `3389f11`.

- Migration now treats `runtime/operation-journal.json` as the authoritative crash-recovery safety gate.
- A pending operation blocks the copy before any source data is migrated, and a corrupt journal fails closed instead of being treated as empty.
- Only the path-keyed `environment-inventory` cache is excluded from relocation; immutable Environment manifests are still preserved and validated.
- Added migration-level and validator-level regression coverage for pending and corrupt journals while retaining the original stale-inventory recovery case.

Validation after the change:

- Focused storage tests: 83 passed
- Typecheck: passed
- Lint: 0 errors
- Full suite: 8,882 passed, 140 skipped, 0 failed

PR #559, opened 2026-07-31
feat(conversation): show message timing metadata

ewen-poch:
## Problem

Conversation messages did not expose stable sent/completed timing metadata. User-message controls needed a stable placement without shifting timestamps, and completed Agent Messages had no compact elapsed/Usage summary. Provider usage payloads also differ enough that Codex cache-heavy responses could be rejected as unavailable if cached input was normalized twice.

A conversation that moved from an OpenCode/Codex Bridge model to a native Codex Responses model could retain an app-facing `ses_...` ID. After the Codex process restarted, the runtime sent that non-UUID ID to Codex `session/resume`, which failed with JSON-RPC `-32603` instead of recovering the conversation.

## Proposed change

- Render `Sent MMM D, h:mm AM/PM` below user bubbles, with Branch, copy, and edit controls immediately left of the bubble.
- Render stable `Completed ...` or `Failed ...`, elapsed runtime, and a dashed-underlined `Usage` trigger on one separator-free line below Agent Messages.
- Show a compact Context-window-style Usage popover with a proportional category bar, Input/Cache/Output rows, a divided Total row, and conditional Cache read/write rows only when both fields are actually reported.
- Persist terminal timestamps and turn usage independently from mutable record updates; backfill legacy complete/error messages and interrupted `streaming -> error` restores.
- Normalize turn usage at framework boundaries: preserve the pinned Codex adapter's already-exclusive input/cache categories, aggregate every OpenCode assistant step, and preserve cache read/write details only as a validated pair.
- Reconcile Codex context usage as uncached Input + Cache while excluding Output and reasoning.
- Keep immutable Provenance transcript messages on the same metadata presentation.
- Before a Codex resume, validate the protocol session ID. Preserve a non-UUID app conversation ID while adopting a fresh underlying Codex session and replaying the visible transcript; valid UUID and `urn:uuid:` IDs continue through native resume.

## Scope and non-goals

- Scope includes the conversation renderer, backward-compatible session persistence, ACP turn-usage normalization, the app-managed Codex adapter patch, and Codex session recovery after backend/model transitions.
- This does not add a live in-progress timer, monetary cost estimation, or a new session-file envelope version.
- It does not change Branch semantics or copy/edit behavior.
- If an upstream provider reports no usage counters at all, Usage remains unavailable; the app does not invent estimates.
- Recovery intentionally creates a new agent-side Codex context when the retained ID cannot be parsed by Codex; it does not attempt to preserve unpublished agent-local state outside the persisted transcript.

- Message timestamps, elapsed time, relocated actions, Usage popover lifecycle/layout/accessibility, persistence, and framework normalization -> focused Vitest command over 10 affected files -> 651/651 passed.
- Interrupted `streaming -> error` terminal time -> `npm test -- --run src/shared/session-persistence.test.ts -t 'builds a legacy conversation graph from normalized interrupted messages'` -> 1/1 passed.
- Non-UUID Codex IDs adopt a fresh underlying session while UUID and `urn:uuid:` IDs still resume -> `npm test -- --run src/main/acp/runtime.test.ts -t 'Codex protocol session id|app-facing id is not a valid Codex UUID'` -> 3/3 passed.
- Existing and migrated Codex adapter usage patches remain accepted -> `npm test -- --run src/main/settings/managed-codex.test.ts` -> 53/53 passed.
- Full regression suite after the final material edit and rebase onto latest `origin/main` (`89ae5c0`) -> `npm test` -> 602 files passed, 11 skipped; 9,021 tests passed, 140 skipped.
- Type safety after the final material edit -> `npm run typecheck` -> passed.
- Static checks after the final material edit -> `npm run lint` -> 0 errors and 23 pre-existing warnings.
- Formatting -> changed-file Prettier checks and `git diff --check` -> passed.
- Independent standards/spec review of the final diff -> no remaining actionable findings.

## Review focus

- Codex usage semantics: pinned `codex-acp` already separates cached input, so runtime normalization must not subtract Cache again; context usage must recombine Input + Cache.
- Whole-turn accumulation and the standard PromptResponse fallback when private Codex metadata is absent.
- Conditional Cache read/write preservation for Claude Code/OpenCode without inventing missing or `null` fields.
- Stable terminal timestamps across later artifact reconciliation and interrupted-session restoration.
- Single-line footer, compact popover proportions/Total, and left-of-bubble action alignment.
- Session identity boundary: an app-facing adopted ID remains stable, but only UUID-shaped IDs are sent to Codex resume after an agent-process restart.

Uncovered risk: final layout verification is DOM/SSR based rather than a pixel-level browser comparison. Exact Usage also depends on an upstream provider returning counters; missing counters intentionally render as unavailable. Recovery for an incompatible Codex ID depends on the existing transcript replay path and necessarily resets unpublished agent-local context.


ewen-poch:
Addressed the out-of-range persisted timestamp crash in `df67af5`.

- Message timestamps are converted to `Date` and validated with `getTime()` before `toISOString()` or `Intl.DateTimeFormat` runs.
- Corrupted sent/terminal timestamps are omitted instead of crashing transcript rendering.
- A valid terminal timestamp still renders when only the paired turn start is invalid; elapsed time is omitted.
- Valid Usage metadata remains available when terminal timing metadata is corrupted.

Validation: the focused renderer tests pass (31/31), the full Vitest suite passes (602 files; 9,024 tests), typecheck passes, lint reports 0 errors, and changed files pass Prettier plus `git diff --check`.

ewen-poch:
Addressed the repeated `Date.now()` assertion race in `90ec997`.

The test now captures the emitted `failedAt` once, verifies that it is numeric, and compares the live message, canonical conversation-graph message, and persisted projection against that same value. This preserves the cross-projection consistency assertion without depending on the wall clock remaining in the same millisecond.

Validation: the focused regression passes (1/1), the full Vitest suite passes (602 files; 9,024 tests), typecheck passes, lint reports 0 errors, and Prettier plus `git diff --check` pass.

ewen-poch:
Not changing this calculation. `recordCodexPromptResponseContextUsage` reconciles the latest model-request context snapshot used for the context-window indicator and compaction threshold; it does not consume the cumulative turn usage. The managed Codex adapter publishes that latest request through `PromptResponse.usage`, where uncached input and cached input are mutually exclusive categories, so the context numerator is `inputTokens + cacheTokens`. Whole-turn cumulative usage, including the optional cache read/write breakdown, is carried separately in `open-science/turn-usage` metadata and is used only for the completed message Usage footer. Changing this line would conflate the two accounting paths rather than prevent cumulative usage from affecting context compaction.

ewen-poch:
Added model-turn counts inside the Usage popover header as smaller, right-aligned text. Counts are adapter-specific and only shown when reliable: OpenCode derives them from new assistant model records, Claude Code consumes SDK result num_turns while excluding autonomous result lanes, and managed Codex counts positive per-request usage deltas for Responses and Bridge paths. Unavailable counts remain omitted.\n\nValidation: full npm test passed before the review fix; npm run typecheck passed; lint completed with only existing warnings; the Claude human-origin regression test passes after the review fix. Independent spec and standards reviews found no remaining actionable issues.

ewen-poch:
Addressed both review findings.\n\n- The Codex turn-usage patch now repairs legacy overwritten response metadata before the idempotence guard, including bundles where another response site already has the current shape. Added a mixed-response-site regression test.\n- Artifact provenance no longer maps message createdAt to agent completedAt. Because the immutable provenance snapshot does not persist a reliable terminal timestamp, it now omits the completion label instead of fabricating one; no data-model migration was introduced. Added integration and renderer regression coverage.\n\nValidation: full suite passes (602 files, 9,027 tests), typecheck passes, focused regression files pass, and git diff --check passes.

PR #561, opened 2026-07-31
perf(permissions): avoid full session reconciliation

ewen-poch:
## Problem

Opening **Settings > Permissions** called the Session persistence `loadAll()` path solely to resolve Session-scoped grant labels. That path scans durable Session JSON and may run startup recovery and file-index reconciliation, so the page latency grows with storage work that the permission projection does not need.

## Proposed change

Keep a process-local Session metadata snapshot (`id`, `projectId`, and `title`) inside `SessionPersistenceCoordinator` and have the permission IPC projection read that snapshot instead of invoking `loadAll()`.

The snapshot is refreshed by authoritative loads, updated after durable saves, and pruned after durable Session or Project deletion. Partial scans, snapshot failures, and startup reconciliation failures preserve readable labels while marking Session metadata incomplete.

```mermaid
flowchart LR
    JSON["Session JSON (authority)"] -->|"existing load/save/delete"| Coordinator["SessionPersistenceCoordinator"]
    Coordinator -->|"in-memory id/projectId/title snapshot"| PermissionIPC["permissions:list projection"]
    PermissionIPC --> Panel["Settings > Permissions"]
    PermissionIPC -. "does not call" .-> Reconciliation["Session load/recovery/index reconciliation"]
```

## Scope and non-goals

- No database schema, table, row, migration, or persisted Session JSON changes.
- No changes to permission grant storage or data relationships.
- The snapshot is process-local and is rebuilt through the existing Session hydration path.
- Full Session loading and reconciliation remain authoritative for startup/recovery; this change only removes them from permission label projection.

- Permissions resolves Session-scoped labels from cached metadata without calling full Session storage
  → `npm test -- --run src/main/permission-grants/ipc.test.ts src/main/session-persistence/coordinator.test.ts`
  → 2 files passed, 78 tests passed.
- Snapshot stays fresh across load/save/Session delete/Project delete and preserves incomplete/error degradation
  → same targeted command
  → 78/78 passed, including startup reconciliation failure coverage.
- Node and renderer types remain valid
  → `npm run typecheck`
  → passed.
- Repository lint gate
  → `npm run lint`
  → passed with 0 errors (23 pre-existing warnings outside this change).
- Full regression suite
  → `npm test`
  → 600 files passed, 8,996 tests passed; 11 files / 140 tests skipped by existing suite configuration.
- Independent Standards and Spec reviews of the final commit
  → both completed with no remaining findings.

## Review focus

- Whether snapshot updates occur only after the corresponding durable authority mutation.
- Whether incomplete metadata remains fail-visible without hiding or preventing individual grant revocation.
- Whether any Session mutation path can leave a stale title or deleted Session in the projection.

Uncovered risk: validation structurally proves the expensive `loadAll()` path is not invoked, but does not include an end-to-end wall-clock benchmark with a very large on-disk Session catalog.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -78,6 +78,7 @@ import { registerReviewerIpcHandlers } from './reviewer/ipc'
 import {
   createDefaultReviewRepository,
   createDefaultSessionRepository,
+  loadSessionMetadataAfterProjectRecovery,
   loadSessionsAfterProjectRecovery,
   registerSessionPersistenceIpcHandlers
 } from './session-persistence/ipc'
@@ -941,7 +942,13 @@ const registerIpcHandlers = async ({
         return projectRepository.list()
       }
     },
-    sessions: sessionPersistenceBackend,
+    sessions: {
+      metadataSnapshot: () =>
+        loadSessionMetadataAfterProjectRecovery(
+          projectDeletionCoordinator,
+          sessionPersistenceCoordinator
+        )
+    },
     connectors: {
       get: async () => ({
         ...(await settingsService.getConnectors()),
--- a/src/main/permission-grants/ipc.test.ts
+++ b/src/main/permission-grants/ipc.test.ts
@@ -18,6 +18,38 @@ import { registerPermissionGrantIpcHandlers } from './ipc'
 beforeEach(() => handlers.clear())
 
 describe('permission grant IPC', () => {
+  it('projects Session-scoped grants from cached metadata without loading Session storage', async () => {
+    const loadAll = vi.fn().mockRejectedValue(new Error('full Session load must not run'))
+    const metadataSnapshot = vi.fn().mockResolvedValue({
+      sessions: [{ id: 'session-1', projectId: 'project-1', title: 'Cached session' }],
+      isComplete: true
+    })
+    const sessions = { metadataSnapshot, loadAll }
+    const registry = {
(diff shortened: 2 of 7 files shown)


ewen-poch:
Addressed the P2 race in 6a13989. Session metadata snapshots now enter the existing coordinator queue, so a permissions read waits for any already-queued save or deletion and observes the committed metadata. The read still uses only the in-memory id/projectId/title snapshot and does not trigger Session storage loading or reconciliation. Added a gated in-flight save regression test; typecheck, lint, and the full suite pass (8,997 tests).

ewen-poch:
Addressed the P1 failure-state regression in 0186375. A durable save that later fails file-index synchronization now keeps the updated in-memory Session label but marks the metadata snapshot incomplete, matching the existing degraded projection semantics. The regression test now proves the initial load succeeds and only the subsequent save sync fails. Typecheck, lint, and the full suite pass (8,997 tests).

ewen-poch:
Addressed the P2 recovery-gate regression in bab8d06. Permission Session metadata now passes through the same sticky Project deletion recovery gate as the previous Session loader before reading the in-memory snapshot. Recovery failures reject the metadata read so the permission projection reports Sessions as incomplete instead of publishing stale navigation labels. Added tests for strict recovery-before-snapshot ordering and fail-closed recovery errors. Typecheck, lint, and the full suite pass (8,999 tests).

PR #571, opened 2026-08-01
fix(settings): surface preference write failures

ewen-poch:
## Problem

Several Settings preferences optimistically rolled back after a failed write, but the failure was only logged to the console. Users could see their selection revert without an actionable explanation.

## Proposed change

- Add transient Settings write-error state with safe, preference-specific user-facing copy while preserving raw errors in the console.
- Cover agent framework, reasoning effort, notifications, conversation Skill import, window close behavior, and app icon writes.
- Keep the existing optimistic rollback behavior where applicable.
- Show a fixed inline danger alert below the Settings header. The alert is dismissible, exposes `role="alert"`, and provides a tooltip for its icon-only dismiss action.
- Track write generations per preference: stale work for the same preference cannot overwrite the latest value, while failures from different concurrent preferences remain visible together.
- Serialize persistence for repeated optimistic writes to the same preference while keeping the UI update immediate; failures restore the last confirmed value, including all-failure and earlier-success/later-failure sequences.
- Clear only errors that were already visible when a later Settings write began and then succeeded; a different failure arriving while that write is pending remains visible.
- Include successful active-provider/model writes in the shared lifecycle so they clear an existing stale preference error.
- Record active-provider/model failures with safe copy, preserve rejection semantics for awaited callers, and explicitly consume already-recorded failures at fire-and-forget model selectors.
- Keep framework status refresh failures separate from successful framework writes to avoid false save-failure feedback.
- Preserve framework write rejection semantics. Onboarding invalidates the old readiness result before switching and stops without checking the old framework when persistence fails.

## Scope and non-goals

- No architecture, data model, data relationship, persistence format, or IPC contract changes.
- No success toast or new notification infrastructure.

- Settings write failures display an actionable alert and preserve or restore the expected preference state -> focused store, Settings page, model-selector, composer, and onboarding tests -> 5 files and 184 tests passed.
- The alert is placed above the scrolling content, uses the existing danger treatment, is announced as an alert, exposes a dismiss tooltip, and clears on dismissal -> the focused test set passed.
- Node and renderer TypeScript remain valid -> `npm run typecheck` passed.
- Repository lint rules remain satisfied -> `npm run lint` completed with 0 errors and 23 pre-existing warnings; changed Settings files added no warnings.
- The complete repository suite remains green -> `npm test` -> 609 files passed, 11 skipped; 9,084 tests passed, 140 skipped.
- Patch formatting is clean -> `git diff --check origin/main...HEAD` passed.

Known scope trade-off: callers that invoke these setters while the Settings dialog is closed still do not render this banner. App-level or caller-specific feedback is intentionally deferred so this focused fix does not change onboarding or workspace interaction patterns.

## Review focus

- Whether all in-scope preference writes set safe feedback without exposing raw IPC details.
- Whether latest-write ownership matches the expected behavior for overlapping Settings writes.
- Alert placement, accessibility, dismissal, and consistency with the existing Settings design.

### DIFF
--- a/src/renderer/src/pages/onboarding/AgentStep.render.test.tsx
+++ b/src/renderer/src/pages/onboarding/AgentStep.render.test.tsx
@@ -258,6 +258,42 @@ describe('AgentStep', () => {
     expect(checkEnvironment).toHaveBeenCalledOnce()
   })
 
+  it('does not refresh the old framework environment when a switch fails', async () => {
+    const setAgentFramework = vi.fn().mockRejectedValue(new Error('framework save failed'))
+    const checkEnvironment = vi.fn().mockResolvedValue(undefined)
+    useSettingsStore.setState({
+      agentFrameworkId: 'claude-code',
+      agentFrameworks: threeFrameworks,
+      claude: { resolvedPath: '/bin/claude', version: '2.1.0' },
+      codex: { resolvedPath: '/bin/codex-acp', version: '0.9.0' },
+      preflight: {
+        claudeReady: true,
+        opencodeReady: false,
+        codexReady: true,
+        agentFrameworkId: 'claude-code',
+        agentReady: true,
+        activeProviderReady: false
+      },
+      environmentCheck: environment(true),
+      setAgentFramework,
+      checkEnvironment
+    })
+
+    await renderStep()
+    await act(async () => {
+      container.querySelector<HTMLElement>('[aria-label="Use Codex"]')?.click()
+      await Promise.resolve()
+      await Promise.resolve()
+    })
+
+    expect(setAgentFramework).toHaveBeenCalledWith('codex')
+    expect(checkEnvironment).not.toHaveBeenCalled()
+    const continueButton = [...container.querySelectorAll('button')].find(
+      (button) => button.textContent === 'Continue'
+    )
+    expect(continueButton?.disabled).toBe(true)
+  })
+
   it('enables Continue only when the latest check passes for the selected framework', async () => {
     useSettingsStore.setState({
       agentFrameworks: twoFrameworks,
--- a/src/renderer/src/pages/settings/ActiveModelSelect.tsx
+++ b/src/renderer/src/pages/settings/ActiveModelSelect.tsx
@@ -67,7 +67,7 @@ const ActiveModelSelect = (): React.JSX.Element | null => {
       value={current ? `${current.providerId}${SEP}${current.model}` : undefined}
       onValueChange={(value) => {
         const [providerId, model] = value.split(SEP)
-        void setActiveProvider(providerId, model)
+        void setActiveProvider(providerId, model).catch(() => undefined)
       }}
     >
       <SelectTrigger aria-label="Active model">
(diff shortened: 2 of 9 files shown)


ewen-poch:
Addressed the concurrent-write and stale-error lifecycle findings in `043d567`:

- Write generations are now tracked per preference. A stale result for the same preference cannot apply its snapshot, roll back the latest optimistic value, or publish an error.
- Failures from different concurrent preferences are retained together instead of being suppressed by an unrelated later write.
- A successful active provider/model write participates in the lifecycle and clears errors that were already visible when that write began. It does not clear a different failure that arrives while the write is pending.

I am intentionally not moving the alert to an app-level surface in this PR. The agreed interaction contract is a fixed inline alert in the Settings dialog. Adding app-level or caller-specific feedback would change onboarding and workspace interaction patterns and substantially widen this focused fix; that should be handled as a separate UX-scoped change.

Validation after the last material edit: focused Settings tests passed (130/130), typecheck passed, lint completed with 0 errors and 23 pre-existing warnings, and the full suite passed (9,080 passed, 140 skipped).

ewen-poch:
Addressed the active-provider failure finding in `80a2ee0`:

- `setActiveProvider` now publishes stable, path-safe feedback and logs the raw IPC error.
- It rethrows after recording the failure, preserving the existing control flow for awaited onboarding/provider setup callers.
- The Settings and workspace fire-and-forget model selectors explicitly consume the already-recorded rejection, preventing an unhandled Promise rejection.

The onboarding framework-alert finding remains intentionally deferred for the same scope reason described above. This PR keeps the agreed feedback surface inside the Settings dialog; adding feedback to `AgentStep`/`AgentPanel` would change onboarding interaction and should be reviewed as a separate UX-scoped change.

Validation after the last material edit: focused tests passed (156/156), typecheck passed, lint completed with 0 errors and 23 pre-existing warnings, and the full suite passed (9,081 passed, 140 skipped).

ewen-poch:
Addressed the confirmed-value rollback finding in `1ea5e92`:

- Repeated optimistic writes are serialized per preference, while writes to different preferences still run independently.
- The selected value is still applied to the UI immediately; only the persistence calls are ordered.
- Each in-flight preference sequence retains its last confirmed value. If every write fails, the UI restores the value from before the sequence. If an earlier write succeeds and a later write fails, the UI restores that earlier confirmed value.
- Stale generations still cannot publish an error or overwrite the latest optimistic selection.

Validation after the last material edit: focused tests passed (158/158), typecheck passed, lint completed with 0 errors and 23 pre-existing warnings, and the full suite passed (9,083 passed, 140 skipped).

ewen-poch:
Addressed the framework-switch control-flow finding in `2a974e1`:

- `setAgentFramework` now rethrows after recording safe feedback and raw diagnostics.
- The Settings confirmation handler consumes that recorded rejection and skips the environment check.
- Onboarding invalidates the previous framework readiness result as soon as a switch is requested, so Continue cannot remain enabled from stale readiness.
- A failed onboarding switch is consumed by the switch queue without checking the still-selected old framework; a newer queued user intent can still proceed.

This corrects onboarding control flow without adding a new onboarding or app-level error surface.

Validation after the last material edit: focused tests passed (184/184), typecheck passed, lint completed with 0 errors and 23 pre-existing warnings, and the full suite passed (9,084 passed, 140 skipped).

PR #587, opened 2026-08-01
fix(e2e): fail tests on renderer errors

ewen-poch:
## Problem

Electron E2E fixtures logged renderer `pageerror` events and `console.error` messages but did not fail the test. Renderer regressions could therefore leave a green E2E run.

## Proposed change

- Add a renderer failure gate that records `pageerror` and error-level console events.
- Attach the gate to every launched or restarted renderer page.
- Assert accumulated failures after application teardown so cleanup still runs.
- Keep warnings and lower console levels non-fatal.

## Scope and non-goals

- Changes only E2E failure semantics and fixture plumbing.
- No production behavior, persisted data, user interaction, or Specialist changes.

- Error events fail with aggregated diagnostics while warnings remain allowed: covered by `test/renderer-failure-gate.test.ts`.
- `npm run lint`: passed with 0 errors and the existing 23 warnings.
- `npm test`: 631 files passed, 15 skipped; 9,357 tests passed, 184 skipped.
- `npm run build:e2e`: passed.
- `npm run test:e2e`: 12 passed, 2 Windows-only tests skipped.

## Review focus

- Gate attachment across initial launch and relaunch.
- Assertion ordering relative to Electron teardown.
- Diagnostic quality when multiple renderer failures occur.

### DIFF
--- a/e2e/fixtures/electron-app.ts
+++ b/e2e/fixtures/electron-app.ts
@@ -4,6 +4,7 @@ import { chmod, mkdir, mkdtemp, readdir, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { delimiter, join, resolve } from 'node:path'
 import { _electron as electron, type ElectronApplication, type Page } from 'playwright'
+import { RendererFailureGate } from './renderer-failure-gate'
 
 const APP_ROOT = resolve(process.cwd())
 const FAKE_AGENT_PATH = resolve(APP_ROOT, 'e2e', 'fixtures', 'fake-opencode.mjs')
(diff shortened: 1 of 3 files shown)


ewen-poch:
Addressed in ee1d1d7. The failure gate now attaches live listeners first, then backfills Playwright's consoleMessages() and pageErrors() history before waiting for domcontentloaded, so errors emitted during initial navigation are retained. Added regression coverage for errors emitted before observation begins. Validation passed: typecheck, lint (0 errors), full unit suite (9,358 passed), and Electron E2E (12 passed, 2 Windows-only skipped).

PR #592, opened 2026-08-01
ci(pr-gate): add dependency-aware shadow gate

ewen-poch:
## Problem

Pull request validation is spread across independently triggered workflows, and path-only selection does not model downstream consumers. A Main or shared interface change can affect Renderer behavior even when Renderer files are untouched.

## Proposed change

- Add a manifest-driven change classifier with consumer dependency closure and unknown-to-full fallback.
- Add a parallel shadow `PR Gate` with a stable aggregate conclusion.
- Add base-trusted `CI Integrity` inspection for workflow-sensitive changes.
- Validate editable PR titles in `CI Integrity` and immutable commit history in `PR Gate`, so metadata edits never restart the full matrix.
- Add changed-file formatting, PR policy, stable-status, workflow trust, Windows radius, CLI/SDK, and aggregate-gate regression tests.

## Scope and non-goals

- This PR is shadow-only: existing PR workflows remain active and the new statuses are not required yet.
- AI Review remains independent and never participates in `PR Gate`.
- No product architecture, persisted data, data relationships, or end-user interaction changes.
- The current cross-process lane is explicitly an `Interface contract baseline (shadow)`; complete handler/event/declaration/serialization contracts remain Phase 4.
- actionlint/zizmor security expansion, dependency gates, coverage calibration, package smoke, ruleset activation, and legacy-workflow retirement remain later phases.

- Change classification and dependency radius -> `npm test -- --run scripts/ci` -> 7 files / 87 tests passed.
- Full repository regression -> `npm test` -> 642 files / 9471 tests passed; 15 files / 184 tests skipped.
- Repository lint -> `npm run lint` -> 0 errors; 23 pre-existing warnings remain outside this branch.
- CLI and SDK lane -> `npx vitest run cli packages/open-science` -> 16 files / 153 tests passed.
- Existing web API contract -> `npm run check:web-api-map` -> passed.
- CLI package contents -> `npm run check:cli-package` -> passed.
- Workflow syntax and pins -> `actionlint .github/workflows/pr-gate.yml .github/workflows/ci-integrity.yml` -> passed.
- Changed files -> Prettier and `git diff --check` -> passed.
- Independent standards and specification reviews -> passed after fixes.

## Review focus

- Trust boundary of `pull_request_target` CI Integrity: base-only checkout and no PR-authored code execution.
- Conservative Main/Shared/Preload consumer closure into Renderer and desktop E2E lanes.
- Stable `PR Gate` aggregation semantics for selected, skipped, failed, and canceled jobs.
- Fork contributor behavior and average wall-clock time during shadow observation.

## Uncovered risks

Real GitHub fork approval behavior, merge-group behavior, queue delay, and lane timing can only be measured after this workflow runs on hosted Actions. Do not require the new statuses or retire legacy workflows until shadow observation and the later contract/security phases are complete.

### DIFF
--- a/.github/workflows/ci-integrity.yml
+++ b/.github/workflows/ci-integrity.yml
@@ -3,7 +3,7 @@ name: CI Integrity
 on:
   pull_request_target:
     branches: [main]
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [opened, edited, synchronize, reopened, ready_for_review]
   merge_group:
     types: [checks_requested]
 
@@ -74,6 +74,13 @@ jobs:
 
           echo "base=$base_sha" >> "$GITHUB_OUTPUT"
           echo "head=$head_sha" >> "$GITHUB_OUTPUT"
+
+      - name: Validate pull request metadata
+        env:
+          EVENT_NAME: ${{ github.event_name == 'pull_request_target' && 'pull_request' || github.event_name }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+          POLICY_SCOPE: title
+        run: node scripts/ci/check-pr-policy.mjs
 
       - name: Inspect CI-sensitive changes
         env:
--- a/.github/workflows/ci-integrity.yml
+++ b/.github/workflows/ci-integrity.yml
@@ -1,0 +1,82 @@
+name: CI Integrity
+
+on:
+  pull_request_target:
+    branches: [main]
+    types: [opened, synchronize, reopened, ready_for_review]
+  merge_group:
+    types: [checks_requested]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-integrity-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  integrity:
+    name: CI Integrity
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout trusted base
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha }}
+
+      - name: Setup Node
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install trusted parsing toolchain
+        run: npm ci --ignore-scripts
+
+      - name: Resolve inspected revisions
+        id: revisions
+        shell: bash
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          MERGE_HEAD_REF: ${{ github.event.merge_group.head_ref }}
+          MERGE_HEAD_SHA: ${{ github.event.merge_group.head_sha }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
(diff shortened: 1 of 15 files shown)


ewen-poch:
Addressed the trust-boundary findings in `f50121b`:

- Preflight now materializes the classifier and impact manifest from the exact base revision. The introducing PR has no trusted copy yet, so bootstrap is explicitly fail-closed to the full lane set; later PRs cannot execute their own planning logic.
- The aggregate job checks out the exact base revision before running the evaluator. Its bootstrap fallback is a strict full-result evaluator and is used only while the base does not contain the new evaluator.
- Established `PR Gate` and `CI Integrity` workflow definitions are now protected control-plane files. The trusted integrity check rejects semantic changes and requires an explicit maintainer ruleset bypass, so retaining only the display name cannot produce a self-approved no-op gate.
- CI script fixtures are still listed as sensitive changes, but workflow-content rules now apply only to workflow/composite-action YAML. This removes false positives from intentionally unsafe YAML strings in policy tests; an end-to-end Git replay regression was added, and replaying the complete PR diff now passes CI Integrity.

I did not change the bootstrap observation: a newly introduced `pull_request_target` workflow is not active until it exists on the default branch, so there is no run in this PR that can check out the base and then fail to find the newly introduced scripts. Once active, both scripts are present in the trusted base.

I also kept merge-group commit policy separate from PR commit policy. Merge-group heads are synthetic commits and should not be required to follow Conventional Commits; constituent PR commits are validated before queue entry, while the merge-group run revalidates deterministic product checks against the synthetic merge revision.

ewen-poch:
Implemented the two P1 findings from the latest Codex review in `b3c512a`:

- Trusted control-plane protection now covers the established `PR Gate` and `CI Integrity` workflows plus the base-executed classifier, impact manifest, evaluator, PR policy, and integrity checker. Future semantic edits or path moves intentionally require an explicit maintainer ruleset bypass.
- Action pin validation now parses executable YAML structurally and checks workflow steps, reusable-workflow jobs, and root/nested composite-action steps. Quoted keys and inline mappings are covered without treating ordinary `env`/`with` keys named `uses` as action references.

The same fixed-point review also closed required-check identity spoofing: canonical workflow paths cannot move, `PR Gate`/`CI Integrity` job names and IDs are reserved, and expression-derived names are rejected when their wildcard expansion could equal a reserved context. Safe names with fixed prefixes, such as `Verify (${{ matrix.os }})`, remain allowed.

Validation after the final edit: `scripts/ci` 87/87 tests, full suite 9471 passed with 184 skipped, typecheck passed, focused lint passed, Prettier passed, and both standards/specification reviews reported no remaining findings.

ewen-poch:
No code change is planned for this finding in the current pull request.

The missing base-side `pull_request_target` run is the intentional one-time bootstrap limitation of the approved shadow rollout. Executing the proposed integrity checker from this PR's `pull_request` workflow would still execute PR-controlled policy and would not create an independent trusted path. This repository has no pre-existing maintainer-controlled workflow that can validate a not-yet-present gate.

The bootstrap is constrained as follows: the new statuses are shadow-only and not required, all legacy checks remain active, the bootstrap classifier runs the full lane plan, the workflow and policy changes have project-owned regression tests plus independent review, and `CI Integrity` becomes base-trusted for every subsequent pull request after merge. Ruleset activation and legacy-workflow retirement remain explicitly deferred until shadow observation.

Eliminating the one-time gap requires a rollout change outside this PR: a maintainer must first seed an inert trusted workflow/checker on the default branch (or use an equivalent out-of-band trusted validation), then this branch must be rebased onto that trusted base. That conflicts with the current single-PR bootstrap design, so work is paused after this fifth review for maintainer direction.

PR #608, opened 2026-08-02
fix(codex): bypass proxies for local Responses bridge

ewen-poch:
## Problem

Codex can inherit proxy environment variables while it is configured to call the app-owned Responses compatibility bridge on localhost. A proxy that does not bypass loopback can intercept that request and return a 502 for the local `/v1/responses` URL before the bridge sees it. The supplied screenshot and timing strongly support this path, but the existing logs did not contain enough lifecycle detail to prove where the request stopped.

The same logs also lacked lifecycle diagnostics for `acp:resume-session`, leaving same-session Resume failures or no-response behavior indistinguishable at the IPC boundary.

## Proposed change

- Append loopback hosts and ranges to both `NO_PROXY` and `no_proxy` for the Codex child process only when the local Responses bridge is active.
- Preserve all inherited HTTP/HTTPS proxy variables so the main-process request from the bridge to an overseas model API can continue through the configured proxy.
- Preserve inherited bypass entries from either proxy-bypass variable spelling in both aliases.
- Add correlated, privacy-safe lifecycle diagnostics for local Responses compatibility requests and `acp:resume-session`.
- Whitelist network errors, RPC codes, error kinds, and services; bucket unknown values as `other`.
- Return a fixed client-safe message for unexpected bridge failures instead of exposing raw exception details.

## Scope and non-goals

- This does not change OS-level routing performed by transparent proxies, TUN devices, or VPNs; those still need explicit DIRECT rules for localhost, `127.0.0.0/8`, and `::1`.
- This does not add a speculative behavioral fix for Resume. The new start/completed/failed events provide evidence for the next diagnosis, including a start event with no terminal event.
- New diagnostics do not record request or response bodies, prompts, credentials, model parameters, upstream domains, raw session IDs, workspace paths, backend IDs, tool or Skill names, raw error messages, or unrecognized error metadata.

- [x] Local Responses bridge requests bypass inherited environment proxies.
- [x] External model API traffic retains its existing proxy route.
- [x] Inherited bypass entries from either variable spelling are preserved in both aliases.
- [x] 502 forwarding and transport failures include correlated lifecycle metadata without user content.
- [x] Unexpected bridge failures do not expose raw exception text to the local client.
- [x] Resume success, failure, and missing-terminal-event cases can be distinguished without logging raw identifiers.
- [x] Focused tests: 40 passed.
- [x] Full test suite: 649 files passed, 15 skipped; 9591 tests passed, 184 skipped.
- [x] `npm run typecheck` passed.
- [x] `npm run lint` passed with 0 errors and 23 pre-existing warnings.

## Review focus

Please focus on the boundary between the Codex child and the main-process compatibility bridge, preservation of external proxy routing and inherited bypasses, and whether every new diagnostic or error response remains privacy-safe.

### DIFF
--- a/src/main/acp/ipc.test.ts
+++ b/src/main/acp/ipc.test.ts
@@ -93,14 +93,21 @@ const {
   }
 })
 
-// Spy on the file logger so the create-session failure path can be asserted (routes to main.log, not a
-// bare console.error). errorLogFields stays real so the assertion also covers its output shape.
-const { errorLogSpy } = vi.hoisted(() => ({ errorLogSpy: vi.fn() }))
+// Spy on the file logger so session lifecycle diagnostics can be asserted (routes to main.log, not a
+// bare console). errorLogFields stays real so the create-session assertion also covers its output.
+const { errorLogSpy, infoLogSpy } = vi.hoisted(() => ({
+  errorLogSpy: vi.fn(),
+  infoLogSpy: vi.fn()
+}))
 vi.mock('../logger', async (importOriginal) => {
   const actual = await importOriginal<typeof import('../logger')>()
   return {
     ...actual,
-    createLogger: (scope: string) => ({ ...actual.createLogger(scope), error: errorLogSpy })
+    createLogger: (scope: string) => ({
+      ...actual.createLogger(scope),
+      error: errorLogSpy,
+      info: infoLogSpy
+    })
   }
 })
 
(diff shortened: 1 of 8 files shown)


ewen-poch:
Addressed the actionable proxy-bypass finding and the related security annotation:

- `NO_PROXY` and `no_proxy` now receive the same deduplicated union of inherited bypass entries plus loopback routes. This preserves one-sided configurations for clients that prefer the other alias. Added one-sided regression coverage and updated the Codex bridge integration expectation.
- Unexpected local Responses bridge failures now return a fixed client-safe error message instead of exposing the raw exception text. Privacy-safe phase/error classification remains available in structured logs.

Validation: 40 focused tests passed; the full suite passed with 9,591 tests; typecheck passed; lint completed with 0 errors.

PR #613, opened 2026-08-02
feat(workspace): show completed turn runtime icons

ewen-poch:
## Problem

Completed-message footers do not identify the ACP framework or model that produced a turn, and icons alone are not explicit enough when reviewing historical messages.

## Proposed change

- Resolve each completed message through its existing Conversation Graph runtime segment.
- Reuse the Settings framework and provider icons beside Usage in the token popover.
- Add muted Agent and Model lines at the bottom of the popover, following the Context window diagnostics style.
- Render each label or icon only when its current display value can be resolved; incomplete or deleted historical metadata stays hidden.
- Keep persisted data unchanged: sessions continue to store only framework/backend/model identifiers.

## Scope and non-goals

- No persistence schema, IPC, ACP runtime, dependency, or settings data-model changes.
- No Unknown placeholder or raw internal identifier is exposed for incomplete historical records.

All checks ran after the final material edit.

- Runtime labels and missing historical values -> npm test -- src/renderer/src/pages/workspace/WorkspaceMessageItem.mentions.test.tsx -> 20 tests passed.
- Renderer and shared type contracts -> npm run typecheck -> passed.
- Repository lint -> npm run lint -> passed with 19 pre-existing warnings and 0 errors; changed files report no warnings.
- Regression suite -> npm test -> passed.

Uncovered risk: no new pixel-level visual-regression snapshot was added; the interaction tests verify the popover content, icon source, accessible hover labels, and omission behavior.

## Review focus

Please confirm that resolving runtime identity from message.runtimeSegmentId preserves historical attribution when Settings change, and that the muted Agent and Model lines provide enough clarity without making the completed-message footer noisier.

### DIFF
--- a/src/renderer/src/pages/settings/AgentPanel.tsx
+++ b/src/renderer/src/pages/settings/AgentPanel.tsx
@@ -1,12 +1,5 @@
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { RefreshCw, TriangleAlert } from 'lucide-react'
-
-// Import the bare Mono/Color components straight from their modules: each icon's entry point
-// eagerly attaches its Avatar/Combine companions, which drag in @lobehub/ui (antd-style + an
-// emoji-mart JSON import vitest can't parse). The Mono/Color components are self-contained.
-import ClaudeColor from '@lobehub/icons/es/Claude/components/Color'
-import Codex from '@lobehub/icons/es/Codex/components/Mono'
-import OpenCode from '@lobehub/icons/es/OpenCode/components/Mono'
 import { ExternalTextLink } from '@/components/ExternalTextLink'
 import { Button } from '@/components/ui/button'
 import { selectAnyInstalling, useSettingsStore } from '@/stores/settings-store'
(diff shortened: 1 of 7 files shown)


ewen-poch:
Addressed both findings in `98141ff`:

- Synthesized legacy runtime segments no longer expose the fallback Claude Code identity. Explicitly persisted backend/model values are still retained independently.
- Framework/provider lookup now uses dormant selectors that subscribe while the popover is open, so Settings changes refresh immediately without causing transcript updates while it is closed.

Validation: the focused Workspace tests pass (54/54), lint and typecheck pass. The full suite passed 9,756 tests with one unrelated timeout; its isolated retry passed 22/22.

ewen-poch:
Added compact Settings-reused icons before the `Agent` and `Model` detail labels in `4389280`. Both marks render at approximately 10px inside a 12px row slot; existing omission behavior remains unchanged when metadata or its configured icon cannot be resolved.

Validation: focused test 20/20, full suite 9,757/9,757, typecheck passed, and lint completed with the same 19 pre-existing warnings and no errors.

ewen-poch:
Added the selected Lucide `CircleGauge` SVG before the message-footer `Usage` label in `480e300`. It renders at 12px, inherits the footer text color, remains decorative for accessibility, and keeps the existing trigger interaction unchanged.

Validation: focused test 20/20, full suite 9,757/9,757, typecheck passed, and lint completed with the same 19 pre-existing warnings and no errors.

PR #618, opened 2026-08-02
fix(workspace): use semantic runtime label icons

ewen-poch:
## Problem

The runtime detail rows added in #613 used framework and provider brand icons. Those icons identify the concrete runtime, but the rows need semantic labels matching Settings: Bot for Agent and Brain for Model.

## Proposed change

- Reuse the Lucide Bot and Brain icons already used by Settings.
- Keep the branded framework and provider icons in the Usage header.
- Render the Brain icon whenever a Model row is available, including when its provider no longer resolves.
- Assert the exact semantic SVG icons in the existing renderer test.

## Scope and non-goals

This changes presentation only. It does not change architecture, persisted data, runtime resolution, historical visibility, or interaction behavior.

- Semantic Agent and Model detail icons -> `/Users/eweno/projects/aipoch/open-science/node_modules/.bin/vitest run src/renderer/src/pages/workspace/WorkspaceMessageItem.mentions.test.tsx` -> 20 passed.
- Lint -> `npm run lint` -> 0 errors; 19 existing warnings outside the changed files.
- Regression suite -> `npm test` -> 669 files passed, 15 skipped; 9,764 tests passed, 184 skipped.

Uncovered risk: no screenshot regression was added; the rendered icon components and classes are covered directly by the component test.

## Review focus

Confirm the bottom detail rows use fixed Bot and Brain semantics while the Usage header retains framework and provider branding.

### DIFF
--- a/src/renderer/src/pages/workspace/WorkspaceMessageItem.mentions.test.tsx
+++ b/src/renderer/src/pages/workspace/WorkspaceMessageItem.mentions.test.tsx
@@ -452,8 +452,12 @@ describe('WorkspaceMessageItem turn token usage', () => {
     const details = document.body.querySelector('[data-slot="turn-runtime-details"]')
     expect(details?.textContent).toContain('Agent: Codex')
     expect(details?.textContent).toContain('Model: gpt-test')
-    expect(details?.querySelector('[data-slot="turn-runtime-agent-detail-icon"]')).not.toBeNull()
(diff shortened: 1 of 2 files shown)


ewen-poch:
The detail-row icons now follow the semantic Settings navigation: Bot for Agent and Brain for Model. Framework and provider branding remains limited to the Usage header, so the two locations communicate different information without changing runtime metadata or historical visibility.

PR #623, opened 2026-08-02
perf(acp): reduce agent context overhead

ewen-poch:
## Problem

Simple ACP tasks can accumulate disproportionate input/cache traffic because stable app guidance is copied into ordinary prompt history, Notebook observations can carry duplicated or unbounded long-tail output, and agents may guess connector methods before loading the matching progressive Skill. A planned provider/model reconnect can also briefly look like an unexpected disconnect and expose a false Resume action.

## Proposed change

- Install stable Open Science guidance once in each backend's native system/developer-instruction layer for OpenCode and Codex, while preserving Claude Code session metadata delivery and fallback behavior for injected/legacy backends.
- Keep connector discovery progressive, but make the compact baseline require loading the matching `mcp-*` Skill before the first `host.mcp` call and forbid guessed server/method names across Claude Code, OpenCode, Codex Responses, and Codex bridge routes.
- Compress duplicated Notebook guidance and apply whole-result budgets to execution observations and `notebook_state`.
- Preserve ten recent Notebook run summaries, retain output preview only for the newest run, remove duplicated stream payloads, and keep full persisted output available in the Notebook UI.
- Return compact Artifact and package-management receipts when verbose provenance is not needed by the next inference.
- Treat deliberate provider/model reconnect teardown as idle rather than closed so the renderer does not present Resume for an expected process restart.
- Integrate dynamic Skill-import availability with the session capability owner introduced on current `main`.

## Scope and non-goals

- No persistent data-model or data-relationship migration.
- No removal of full Notebook run data or renderer output; only model-facing MCP projections are shortened.
- No unified capability gateway or extra describe inference.
- No live in-process model switching. The current reconnect/resume path remains the fallback; a possible adapter-owned live-switch seam is recorded only in an ignored internal design document and is not part of this PR.
- No new user workflow. The visible behavior change is limited to avoiding a false Resume action during planned reconnects.

- Stable backend guidance is absent from repeated user prompt text and remains included in context accounting.
- Connector baselines contain skill-first routing instructions but no connector operation schema or guessed method examples.
- Notebook execution replies stay within the global result budget, `notebook_state` returns up to ten compact runs, and full data remains durable.
- Planned reconnects end in idle without emitting closed; unexpected disconnect behavior remains unchanged.
- Rebased onto `origin/main` at `c6d6ee4`.
- `npm run lint` — passed with 0 errors and 19 pre-existing warnings outside this diff.
- `npm test` — passed with 670 files / 9,854 tests passing and 15 files / 184 tests skipped. The suite was run with localhost loopback enabled because HTTP/RPC tests bind `127.0.0.1`.
- `git diff origin/main...HEAD --check` — passed.

External-provider E2E token/cache measurements remain the main follow-up risk: unit coverage verifies prompt lifecycle and result bounds, but exact savings still depend on each provider's cache accounting and agent inference count.

## Review focus

- Whether native OpenCode/Codex instructions remain stable across reconnect and model configuration paths without leaking into ordinary user history.
- Whether Notebook projections preserve the execution facts needed for the immediate next inference while keeping durable/UI output intact.
- Whether planned reconnect state remains distinguishable from a genuine process or transport loss.
- Whether the session capability owner remains the sole boundary for dynamic app-tool availability after the rebase.

### DIFF
--- a/src/main/acp/context-usage-static-context.test.ts
+++ b/src/main/acp/context-usage-static-context.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from 'vitest'
+import { Tiktoken } from 'js-tiktoken/lite'
+import cl100kBase from 'js-tiktoken/ranks/cl100k_base'
 
+import { NOTEBOOK_SYSTEM_PROMPT_APPEND } from '../notebook/mcp-server'
 import { contextUsageMcpSections } from './context-usage-static-context'
 
 describe('contextUsageMcpSections', () => {
@@ -32,6 +35,20 @@ describe('contextUsageMcpSections', () => {
     const text = sections.map((section) => section.text).join('\n')
     expect(text).toContain('mcp.open-science-notebook.notebook_execute')
     expect(text).not.toContain('mcp__open_science_notebook__notebook_execute')
+  })
+
+  it('keeps the notebook schema plus scoped guidance within the static context budget', () => {
+    const [{ text: schema }] = contextUsageMcpSections('codex', {
+      artifacts: false,
+      notebook: true,
+      skillImport: false
+    })
+    const tokenizer = new Tiktoken(cl100kBase)
+
+    // Baseline before deduplication was about 5.2k cl100k tokens (3.6k schema + 1.6k prompt).
+    expect(
+      tokenizer.encode(`${NOTEBOOK_SYSTEM_PROMPT_APPEND}\n${schema}`).length
+    ).toBeLessThanOrEqual(3_200)
   })
 
   it('uses bridge aliases for Codex MCP tools delivered through a compatibility proxy', () => {
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -12670,6 +12670,50 @@ describe('ACP runtime session management', () => {
     expect(states).not.toContain('closed')
   })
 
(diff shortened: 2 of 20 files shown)


ewen-poch:
Addressed the P1 in b6ad4f7. `disconnectForPlannedReconnect` now captures the teardown's connection generation and publishes `idle` only while that generation is still current, so an older teardown cannot overwrite a newer connected state. Added regression coverage that overlaps an old teardown with a successful new connection. Validation: focused regression test, typecheck, lint (0 errors), and the full suite (9,855 tests passed).

ewen-poch:
Addressed the P1 in 98b85bc. The compact notebook state now falls back to the latest run's bounded text/plain display result when no diagnostic stream is present, preserving successful expression and connector results for recovery without returning images or older run payloads. Added regression coverage for latest-only recovery and image exclusion. Validation: focused regression test, typecheck, lint (0 errors), and the full suite (9,856 tests passed).

ewen-poch:
Addressed the P1 in f72e10c. The notebook guidance now uses the framework-neutral `write_artifact_file` from `open-science-artifacts` form, allowing the existing framework renderer to produce valid Claude and OpenCode callable names while remaining valid for Codex. Added assertions against the malformed qualified form. Validation: 49 focused tests, typecheck, lint (0 errors), isolated flaky-test retries, and the full suite with bounded concurrency (9,856 tests passed).

PR #634, opened 2026-08-02
perf(agent-context): reduce skill and notebook payloads

ewen-poch:
## Problem

ACP agents repeatedly receive duplicated connector guidance and verbose Notebook results. Codex compatibility paths also spend a separate model request selecting a Skill when the user already names a connector such as PubMed.

## Proposed change

- Compact generated connector Skills while retaining exact tool schemas, return shapes, examples, reuse guidance, and the custom MCP security boundary.
- Project every Notebook MCP result to the fields needed for the next step. Bound execution results to 12,000 characters, state to 6,000 characters with 10 recent runs, and control results to 8,000 characters.
- Keep runtime and environment discovery complete through bounded offset/limit pagination with nextOffset.
- Route explicitly named mcp-* connector Skills locally before building the bounded inference catalog. Preserve semantic model selection for ordinary Skills and unnamed connectors.
- Correct existing connector examples that used Python boolean literals in JavaScript REPL calls.

## Scope and non-goals

- No renderer, database, persisted Notebook data, or data-relationship changes.
- No unified connector gateway or extra describe inference.
- Full execution history and artifacts remain persisted; only agent-facing tool receipts are shortened.
- Existing sessions may need a new agent session to rematerialize updated Skill documents.

All checks below ran after the final rebase onto main:

- Connector Skill compaction, calling guidance, and custom MCP security boundary -> targeted Vitest files in the command below -> passed.
- Notebook execution/state/control projection and discovery pagination -> src/main/notebook/mcp-server.test.ts -> passed.
- Explicit connector routing, bounded semantic fallback, and bridge parity -> both selector/compatibility test files -> passed.
- `npm test -- --run src/main/connectors/skill-doc.test.ts src/main/connectors/custom-skill-doc.test.ts src/main/notebook/mcp-server.test.ts src/main/settings/responses-bridge.test.ts src/main/settings/native-responses-compatibility.test.ts src/main/settings/service.test.ts` -> 371 passed.
- `npm run typecheck` -> passed.
- `npm run lint` -> 0 errors; 19 pre-existing warnings in unmodified files.
- `npm test` -> 9,944 passed, 184 skipped.

Independent standards and specification reviews reported no remaining findings.

## Review focus

- Whether projected Notebook fields preserve every next-step decision while removing duplicated payloads.
- Whether pagination preserves runtime/environment discoverability under the response budgets.
- Whether local connector matching remains narrow enough to avoid bypassing semantic selection for ordinary Skills.

## Uncovered risk

Actual token savings vary by provider tokenizer, cache accounting, loaded connector, and tool payload shape. Unit tests enforce payload budgets and routing behavior, but do not reproduce provider-side billing counters.

### DIFF
--- a/src/main/connectors/custom-skill-doc.test.ts
+++ b/src/main/connectors/custom-skill-doc.test.ts
@@ -30,10 +30,10 @@ describe('renderCustomSkillDoc', () => {
     expect(md).not.toContain('name: mcp-myserver')
     expect(md).toContain('source: connector')
     expect(md).toMatch(/description: ".*Use when.*"/)
-    expect(md).toContain('## When to Use')
+    expect(md.match(/Use when/g)).toHaveLength(1)
     expect(md).toContain('search')
     expect(md).toContain('fetch')
-    expect(md).toContain('"type": "object"')
+    expect(md).toContain('"type":"object"')
     // Runtime routing still uses the display name (the key McpClientManager registers under).
     // No-arg tools render without a third argument (a literal ... would reach the bridge as Ellipsis).
     expect(md).toContain('host.mcp("myserver", "search")')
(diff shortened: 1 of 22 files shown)


ewen-poch:
Addressed. Connector-local routing now requires an authoritative `source: "connector"` marker supplied by `SettingsService.codexSkillCatalog`; a user-authored or imported Skill named `mcp-*` remains on the semantic selector path. Added regression coverage for both provenance propagation and the non-connector `mcp-*` case. Verified with the focused settings tests (302 passed), typecheck, lint (0 errors), and the full suite (9,945 passed).

PR #646, opened 2026-08-03
fix(acp): persist usage during graceful shutdown

ewen-poch:
## Problem

Graceful app quit reaped ACP runtimes before terminal stop responses and renderer Session writes were guaranteed to settle. This could lose the latest context usage or completed-turn usage even though the agent returned it during cancellation.

## Proposed change

- Cancel active ACP prompts and wait for interaction ownership release before backend teardown, bounded to four seconds.
- Ask the live Electron renderer to fetch the final ACP snapshot, process terminal events, synchronize context usage, and flush the ordered Session persistence queue.
- Bound the renderer acknowledgement to 1.5 seconds and continue teardown if the renderer disappears or does not respond.
- Keep Claude Code, OpenCode, and managed Codex usage calculation in their existing adapters; coordinate them at their shared terminal-response boundary.

Quit flow: confirmation → ACP cancel and drain → renderer final snapshot drain → Session queue flush → backend teardown.

## Scope and non-goals

- No Session schema, data relationship, or UI changes.
- Hard process termination cannot produce exact final usage; existing last-known context usage remains authoritative.
- This change does not replace framework-specific usage reconciliation.

All listed checks ran after the final material edit and after rebasing onto the latest main branch.

- Cancelled ACP response retains terminal usage → npm test -- --run src/main/acp/runtime.test.ts -t "keeps a cancelling prompt in flight until the agent returns its stop response" → passed.
- Quit preparation cancels and waits, with timeout fallback → npm test -- --run src/main/acp/runtime-coordinator.test.ts → passed.
- Renderer writes flush in order → npm test -- --run src/renderer/src/lib/session-persistence/session-persistence.test.ts → passed.
- Renderer acknowledgement ordering and main timeout cleanup → npm test -- --run src/renderer/src/hooks/useQuitPersistenceFlush.test.ts src/main/session-persistence/renderer-flush.test.ts → passed.
- App quit ordering → npm test -- --run src/main/app-lifecycle.test.ts → passed.
- Full regression suite → npm test → passed.
- Type safety → npm run typecheck → passed.
- Static analysis → npm run lint → passed.
- Web API surface remains stable → npm run check:web-api-map → passed.

## Review focus

Independent review found one blocker to resolve before marking this ready:

- When no Electron renderer exists, including headless mode or macOS after the final window closes, cancel and drain can receive terminal usage but no renderer projects it into Session JSON. A main-owned persistence path or retained renderer is still required.

Additional uncovered risks:

- No adapter-specific Stop → cancelled response → stop event → Session JSON integration test yet for Claude Code, OpenCode, and managed Codex.
- Connection loss and cancel timeout persistence are not covered end to end.
- The Electron-specific IPC wrapper and final runtime-snapshot drain are covered through lower-level fakes rather than a production-wrapper integration test.

### DIFF
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -660,6 +660,60 @@ describe('AcpRuntimeCoordinator', () => {
     expect(released).toBe(true)
   })
 
+  it('cancels active prompts and waits for their terminal responses before quit teardown', async () => {
+    const prompt = createDeferred<unknown>()
+    let created!: ReturnType<typeof createFakeRuntime>
+    const coordinator = new AcpRuntimeCoordinator((callbacks) => {
+      created = createFakeRuntime({
+        frameworkId: 'claude-code',
+        sessionIds: ['session-1'],
+        callbacks,
+        prompt: () => prompt.promise
+      })
+      return created.runtime
+    })
+    const session = await coordinator.createSession({ cwd: '/workspace' })
+    const running = coordinator.sendPrompt({ sessionId: session.sessionId, text: 'keep usage' })
+    await Promise.resolve()
+
+    let prepared = false
+    const preparing = coordinator.prepareForQuit(1_000).then(() => {
+      prepared = true
+    })
+    await Promise.resolve()
+
+    expect(created.cancelPrompt).toHaveBeenCalledWith({ sessionId: 'session-1' })
+    expect(prepared).toBe(false)
+
+    prompt.resolve({ stopReason: 'cancelled' })
+    await running
+    await preparing
+    expect(prepared).toBe(true)
+  })
+
+  it('bounds quit preparation when an agent never returns a terminal response', async () => {
+    const prompt = createDeferred<unknown>()
+    let created!: ReturnType<typeof createFakeRuntime>
+    const coordinator = new AcpRuntimeCoordinator((callbacks) => {
+      created = createFakeRuntime({
+        frameworkId: 'codex',
+        sessionIds: ['session-1'],
+        callbacks,
+        prompt: () => prompt.promise
+      })
+      return created.runtime
+    })
+    const session = await coordinator.createSession({ cwd: '/workspace' })
+    const running = coordinator.sendPrompt({ sessionId: session.sessionId, text: 'never stops' })
+    await Promise.resolve()
+
+    await expect(coordinator.prepareForQuit(0)).resolves.toBeUndefined()
+    expect(created.cancelPrompt).toHaveBeenCalledWith({ sessionId: 'session-1' })
+
+    prompt.resolve({ stopReason: 'cancelled' })
+    await running
+  })
+
   it('blocks user prompts on startup admission while allowing recovery continuations through', async () => {
     const admission = createDeferred<void>()
     let createdRuntime!: ReturnType<typeof createFakeRuntime>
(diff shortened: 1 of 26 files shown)


ewen-poch:
Addressed both P1 findings in e5f923d:

- No-tray titlebar closes now use a dedicated quit classification that prevents BrowserWindow destruction and requests an unconfirmed app quit. This keeps the renderer alive through cancel-and-drain and the bounded Session flush, while preserving the active-work confirmation.
- ACP cancellation ownership now includes pending prompt reservations. A cancellation accepted during preflight marks that reservation cancelled, so the existing checkpoint stops it before provider dispatch.
- The renderer quit hook also tolerates an absent sessions bridge, fixing the partial-bridge unit and coverage failures.

Validation: npm test, npm run typecheck, npx eslint --no-cache ., and npm run check:web-api-map all pass.

The broader truly rendererless case (headless or macOS after the last window has already closed) remains the documented draft risk; solving it requires a main-owned persistence path or retained-renderer architecture.

ewen-poch:
Fixed the static-check failure in 9363e6b. The persistence drain is a module-level function with stable identity, so the hook now references it directly instead of passing a named function to useCallback, which violates the React Hooks memo lint rule.

Validation: npm run lint, npm test (10,028 passed), and npm run typecheck all pass.

ewen-poch:
Addressed the renderer lifetime race in 528ba89. Once quit cleanup starts, every subsequent close is classified as a held quit, including on macOS and tray hosts, so the BrowserWindow cannot be destroyed while ACP preparation or Session persistence flushing is pending. Normal platform close behavior remains unchanged before shutdown starts.

Added a regression test that reproduces a second close during pending quit preparation. Validation: 84 lifecycle/window tests pass, npm test passes (10,029 passed), npm run lint passes, and npm run typecheck passes.

ewen-poch:
Resolved the main-branch conflict in e17d7f2 by preserving #649’s latest-wins Session write coalescing and FIFO barriers while retaining this PR’s bounded queue flush. The existing flush regression now verifies that both explicit and coalesced writes settle before quit persistence completes.

Validation: persistence/quit tests pass (29 tests), npm run typecheck passes, npm run lint passes, formatting passes, and npm run check:web-api-map passes. Unrelated full-suite timing failures reproduced only under the 700-file concurrent local run; all affected files pass in isolation (129 tests), so CI will provide the final isolated-run signal.

ewen-poch:
Addressed the quit-time auto-review race in bd2e07f. The renderer now cancels all scheduled auto-review timers before draining terminal cancellation events, rejects any later stop-event scheduling, and stops retry loops before another reviewer start. The suppression is scoped to the renderer’s committed quit handshake, so normal auto-review behavior is unchanged while the app remains active.

Added regression coverage for cancelling an already scheduled review, blocking a review scheduled by the drained stop event, and enforcing the order: suppress reviews, drain events, flush persistence, acknowledge. Validation: npm test passes (10,034 tests), npm run typecheck passes, npm run lint passes, formatting passes, and npm run check:web-api-map passes.

ewen-poch:
Addressed the quit-admission race in cc13e5f. prepareForQuit now closes a main-process admission latch before taking the session snapshot. User prompts, app continuations, compaction, reviewer activities, and reviewer builds are rejected after the latch; prompts waiting on the startup admission guard are checked again at dispatch. Reviewer correction prompts now use the existing pending/active ownership tracking, so work admitted before the latch is included in cancel-and-drain, and reviewer sessions that finish building after quit begins are immediately disposed. Added concurrency coverage for each path. Full npm test passes: 685 files passed, 15 skipped; 10,038 tests passed, 184 skipped.

PR #661, opened 2026-08-03
fix(updater): show Windows update preparation

ewen-poch:
## Problem
Windows updates become invisible after Open Science quits, so users may reopen the desktop shortcut while NSIS is replacing files and mistake the blank shortcut or delayed relaunch for a failed update.

## Proposed change
- publish an `applying` state before backend teardown and lock duplicate update actions
- explain the wait before Open Science closes
- show the native assisted NSIS UI during Windows file replacement, then force-relaunch
- return to an actionable error if backend teardown throws or is degraded

## Scope and non-goals
No architecture, persistence, data-model, or data-relation changes. No custom updater helper or new dependency. The native assisted installer may require the user to finish its standard flow.

- apply feedback and duplicate guard: targeted updater and dialog tests, 50 passed
- `npm run typecheck`: passed
- `npm run lint`: passed with 18 pre-existing warnings outside this diff
- `npm test -- --maxWorkers=4`: 691 files and 10,084 tests passed; 15 files and 184 tests skipped
- independent standards/spec review: 0 findings on both axes

## Review focus
Please verify the assisted NSIS experience on Windows, especially custom install paths and automatic relaunch.

## Uncovered risk
No Windows packaged update was executed locally; unit tests verify the electron-updater contract, but the native NSIS progress/relaunch flow still needs CI or release-candidate validation.


ewen-poch:
Addressed both findings in f2f9166:

- The apply lifecycle now uses an independent in-flight flag. Background checks return early, updater event transitions cannot replace the applying state, and concurrent apply calls remain locked out.
- Install-gate failure paths now preserve the existing update metadata, including `latest`, so the actionable error dialog and manual fallback remain visible.

Added regression coverage for both cases. The targeted updater suite passes 27/27, with typecheck and ESLint also passing.

ewen-poch:
Addressed the installer-error finding in 772462d:

- Updater errors now clear the apply lock and preserve update metadata before publishing the actionable error state.
- An updater error received while the install gate is pending aborts the remaining install handoff.
- Synchronous `quitAndInstall` failures are caught and surfaced through the same retry-capable error UI.

Added regression coverage for both error paths. The targeted updater suite passes 29/29, with typecheck and ESLint also passing.

ewen-poch:
Addressed the lifecycle-race finding in c1b8332:

- Applying is now exclusive across check, download, cancel, and repeated apply calls.
- Updater errors received while the install gate is still preparing are treated as stale events from prior operations and cannot release the apply latch.
- Once `quitAndInstall` handoff begins, updater errors are attributed to the installer and restore the actionable error state.

Regression coverage now exercises stale pre-handoff errors, blocked concurrent operations, post-handoff errors, and synchronous installer failures. The targeted updater suite passes 30/30, with typecheck and ESLint also passing.

ewen-poch:
Leaving this finding unchanged for now. The update install gate has already stopped active agent and notebook work before `quitAndInstall`, so the normal lifecycle confirmation fast-paths without a user prompt. The remaining cancellation case is limited to independent guards such as an active data migration.

Handling that edge case would require a new cross-module cancellation protocol or changing update-driven quit confirmation behavior. That trade-off is not justified without evidence of the scenario occurring in production; we will revisit it if telemetry or user reports show it is material.

PR #713, opened 2026-08-04
fix(workspace): filter images when replaying text-only sessions

### LINKED ISSUE #621: [Bug]:
### Preflight checklist

### What happened?

Error invoking remote method 'acp:send-prompt': RequestError: Internal error: API Error: 400 Model Incompatible with image: step-router-v1

### Steps to reproduce

生成的方案没有技术路线图。更换阶跃星辰的这个step-router-v1模型，让open science补充路线图，然后就报错了，可能是模型提供方不能在生图和非生图模型间自由路由

### Operating system

Windows

### App version

0.9.1

### Provider / model

Step Plan · step-router-v1

```shell
Detected OS: Windows
Agent framework: Claude Code
Runtime: Electron 39.8.10, Chrome 142.0.7444.265, Node 22.22.1

Runtime log not attached automatically (it can contain local paths and prompts).
Reveal it from Settings → General → Diagnostics and attach after reviewing.
```

ewen-poch:
## Problem

Switching a conversation with image history to a text-only model successfully resumes the native ACP session, but the resumed Claude Agent SDK context still contains the old image blocks. The next text-only request is therefore rejected with `Model Incompatible with image`.

## Proposed change

- Keep the native `session/resume` attempt.
- When the destination model is text-only and the local history contains image media, reset the resumed Agent context through the existing context-reset path.
- Replay the text transcript while omitting historical image blocks and image uploads.
- Apply the same media filtering to the existing fallback replay path.

## Scope and non-goals

- No provider- or model-specific special case; `step-router-v1` remains configured as text-only.
- No architecture, persistence schema, or data-relationship changes.
- Local chat history and images remain visible; only the provider-bound replay omits unsupported media.
- No new prompt, banner, or notification is added.

- Native resume remains first, then incompatible hidden image context is reset: `npm test -- src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts` -> 101 passed.
- Text-only replay preserves the text preamble and omits both historical image uploads and inline images: covered by the same 101-test module suite.
- Renderer types remain valid: `npm run typecheck:web` -> passed.
- Repository lint: `npm run lint` -> 0 errors; 17 pre-existing warnings outside the changed files.
- Portable regression suite after the final rebase: `npm test` -> 727 files passed, 15 skipped; 10,791 tests passed, 184 skipped.
- A local characterization probe using the installed Claude Agent SDK and a loopback Anthropic-compatible endpoint reproduced that native resume re-sends the old image and triggers the same 400; the probe was temporary and is not part of this PR.

Uncovered risk: no live request was sent to the Step Plan production gateway; the provider behavior is covered by the reported error and the local protocol-level reproduction.

## Review focus

- Confirm that capability downgrade resets only sessions whose replayable history contains image media.
- Confirm that text-only replay never forwards `historyAttachments` or `historyImages`.

Closes #621

### DIFF
--- a/src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts
+++ b/src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts
@@ -2809,7 +2809,7 @@ describe('recovering from a request-size overflow', () => {
     vi.unstubAllGlobals()
   })
 
-  const seedOverflowedConversation = (): void => {
+  const seedOverflowedConversation = (includeHistoryImage = false): void => {
     // A completed prior turn (replayed as text) followed by the unanswered turn that overflowed.
     useSessionStore.getState().appendUserMessage({
       sessionId: 'session-1',
@@ -2824,6 +2824,14 @@ describe('recovering from a request-size overflow', () => {
       eventId: 'event-1',
       content: 'Here is what it shows'
     })
+    if (includeHistoryImage) {
+      useSessionStore.getState().appendAgentMessageChunk({
+        sessionId: 'session-1',
+        streamId: 'assistant-image-1',
+        eventId: 'image-event-1',
+        image: { mimeType: 'image/png', data: 'aGVsbG8=', byteLength: 5 }
+      })
+    }
     useSessionStore.getState().finishRun('session-1')
     useSessionStore.getState().appendUserMessage({
       sessionId: 'session-1',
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the overflow replay gap.

- Forwarded the active provider's image-input capability through overflow recovery.
- Text-only overflow replay now keeps the text preamble while omitting historical image attachments and image blocks.
- Extended the overflow recovery test to cover image-bearing history with a text-only model.

Validation: the focused runtime suite passes 101/101 tests, the full suite passes 10,791 tests with 184 skipped, Web typecheck passes, and lint reports 0 errors.

PR #715, opened 2026-08-04
feat(workspace): add branch in new session

ewen-poch:
## Problem

Users cannot continue a conversational session from its current active path in a separate session without manually reconstructing context or changing the source conversation.

## Proposed change

- Add an integrated split Send control for all conversational sessions with a **Branch in new session** action.
- Immediately create and select a local pending session, copy only the source session's active message path, and use the submitted prompt as the new title.
- Reuse the existing bounded ACP replay path for text, images, and upload references; creation failures remain retryable in the new session and never mutate or roll back to the source session.
- Keep artifacts, project files, and notebook state owned by the source session. Generated message entries retain their version references, with missing preview metadata resolved through one batched artifact descriptor command rather than by copying artifact records.
- Register the descriptor resolver through Electron IPC, preload/Web contracts, and the application-command owner introduced on current `main`.

```mermaid
flowchart LR
  A["Source session: active path"] -->|"snapshot messages and refs"| B["New local pending session"]
  B --> C["Persist and select immediately"]
  B --> D["Create ACP session"]
  D --> E["Replay bounded context"]
  E --> F["Send the new prompt"]
  A -. "unchanged" .-> A
```

## Scope and non-goals

- No persisted session-schema fields are added; branch/retry markers remain transient and are stripped from persistence.
- Existing replay limits remain in force (including the approximately 12,000-character text budget and existing image/upload limits).
- Existing inactive message branches are not copied.
- Artifact files, ArtifactVersion rows, project Files entries, and notebook state are not duplicated.
- This PR does not add replay summarization or general-purpose session move support.
- The planning notes under `docs/internal` remain local and are intentionally excluded from this PR.

- Selecting **Branch in new session** creates and navigates to the new session immediately.
- The source session remains unchanged; only its active branch is copied.
- Plain prompts, attachment-only prompts, interrupted histories, retries, replay races, and generated artifact previews are covered by targeted tests.
- The split Send control renders as one continuous primary surface, retains keyboard focus/pressed/disabled states, and expands both touch hit targets without changing its 32 px visual height.

Final evidence after the last production edit and final rebase onto `b5bf717`:

- `npm test` — 728 files passed, 15 skipped; 10,821 tests passed, 184 skipped.
- `npm run typecheck` — Node and Web passed after the final test-fixture update.
- `npm run lint` — 0 errors; 23 pre-existing warnings.
- `npm run check:web-api-map` — passed.
- Targeted command-owner, artifact IPC, and composer interaction tests — passed.

Uncovered risk: no manual Electron visual/E2E pass was run; the interaction shape and state classes are covered by renderer tests.

## Review focus

- Active-path snapshot semantics and transient retry markers in the session store.
- Bounded replay/race handling in the ACP runtime.
- Artifact-version descriptor batching and the no-copy ownership boundary.
- Application-command inventory changes (`217` internal / `215` local Web / `159` remote Web).


ewen-poch:
Addressed the replay retry regression. The transient replay marker now survives ACP session binding, bound-session retries replace the original prompt and replay the copied history, and the marker is cleared only after the matching sendPrompt call succeeds. Added a regression test for ACP creation succeeding followed by the initial prompt failing; the retry reuses the same ACP session, preserves exactly one current prompt, and includes the copied history. Validation: 184 focused tests passed, typecheck passed, and the full suite passed (10,822 tests; 184 skipped).

ewen-poch:
We are intentionally not persisting the branch-specific replay marker. If the app closes while replay is in progress, the persisted running turn is normalized to an interrupted session on restore; the existing Resume path finds and removes the unanswered user turn, adopts the runtime, and replays the prior transcript before resending it. We accept the narrower edge case where sendPrompt has already failed and the app is then restarted, rather than widening PersistedChatSession for transient branch retry state. No code change is planned for this finding.

ewen-poch:
Fixed the cross-platform workspace E2E regression exposed by the split send control. When Send transitioned from the 64px split control to the 32px running Cancel button, adjacent controls shifted under the stationary pointer and opened the Context window hover popover, which intercepted the permission Deny button. The running state now preserves the split-control width and right-aligns Cancel, eliminating the layout shift. Added a layout regression test. Validation: affected permission E2E passed, build/typecheck passed, focused component tests passed, and the full suite passed (10,823 tests; 184 skipped).

ewen-poch:
Addressed the copied activity collision. Historical tool activity and group IDs are now namespaced with the child session snapshot ID, and activityGroupId/activityIds are remapped consistently. Added a regression test that simulates the fresh runtime reusing the original toolCallId/groupId and verifies the new live records do not overwrite the copied history. Validation: session store tests 80/80, typecheck and lint passed; the full suite passed 10,823 tests with one unrelated timing-sensitive shell test flake, which passed 10/10 on immediate isolated rerun.

ewen-poch:
Addressed both findings. Immutable history Upload Versions now retain their source ownership and are resolved directly within the current Project; only unversioned staged history and current-turn uploads enter the existing finalization path. Artifact Version descriptor resolution now requires the visible Session scope, validates that Session against main-process persistence authority, and constrains the batch query to the validated Project. Added regressions for source-owned upload replay, staged-history compatibility, same-Project historical descriptor access, and rejection of an invalid Project/Session scope. No Session or database schema changes were introduced. Validation: 74 focused tests passed, lint/format/typecheck passed, and the full suite passed (10,824 tests; 184 skipped).

ewen-poch:
Addressed the two Branch-specific upload retry regressions without changing persistence or the Upload model. Branch creation now passes the original staged attachments into the existing pending-session projection, which retains the main-issued path only in renderer memory until finalization; Session JSON remains path-free. A retry with no replacement attachments now inherits the marked Branch prompt's uploads, covering both pre-bind staged paths and post-bind immutable Version references. This intentionally matches the existing new-session behavior: closing the app before any pending upload acquires a Version remains a platform-wide limitation, not a Branch-specific durability change. The branch was also rebased onto current main. Validation: focused runtime/store tests 185/185, typecheck and lint passed, and the full suite passed (10,850 tests; 184 skipped).

ewen-poch:
Addressed the transient Artifact Version resolver failure. In-flight ids are still claimed before IPC to prevent duplicate lookups, but a rejected batch now releases only those claims; a later Session update retries naturally. Missing Versions remain cached only after a successful lookup returns no descriptor. Added a regression covering an ownership/persistence-race rejection followed by successful resolution. No persistence fields, schema, ownership checks, or UX behavior changed. Validation: focused interaction tests 12/12, web typecheck and lint passed, and the full suite passed (10,851 tests; 184 skipped).

ewen-poch:
Addressed the failure-state regression without reintroducing legacy absolute paths into branched Session data. Branch history media preparation now catches an unreplayable path-only legacy upload, settles the already-selected child through the existing error/retry state, and does not create an ACP Session or dispatch the prompt. Legacy path publication remains owned by the existing startup/live-save migration; Branch continues to copy only immutable Upload Version references and never duplicates bytes. Added a regression for a path-only legacy image. No schema, persistence field, ownership relationship, or new UX state was introduced. Validation: runtime tests 106/106, web typecheck and lint passed, and the full suite passed (10,852 tests; 184 skipped).

ewen-poch:
Addressed the mounted-Session retry gap. A rejected descriptor batch now releases its in-flight cache claims and schedules one bounded 250 ms retry while the same Session/cache remains mounted; each Version id can schedule that delayed retry only once, so persistent failures cannot create an IPC loop. Successful empty lookups remain cached as confirmed missing. The regression test keeps the exact same activeSession object and verifies the copied GENERATED card resolves on the delayed retry. No persistence, schema, ownership, or UX model changes were introduced. Validation: focused interaction tests 12/12, web typecheck passed, and lint completed with no errors. The local full suite was stopped after an abnormal no-output wait; the fresh PR Gate run is the authoritative complete-suite check.

ewen-poch:
Addressed the copied timeline ordering finding. Branched messages now retain the source timeline's effective sort index (including the renderer's fallback for hydrated messages), while copied activities keep their existing indexes, so exact-timestamp tool rows remain between the same surrounding messages. Added regression assertions for the tool row's position. No schema, data relationship, architecture, or new interaction state was introduced. Validation: session-store tests 80/80, formatting passed, and lint completed with no errors (17 unrelated existing warnings). The local web typecheck was stopped after an abnormal no-output wait; the fresh PR Gate is the authoritative typecheck/full-suite check.\n\nThe legacy path-only upload finding is intentionally unchanged. Branch snapshots do not persist or silently replay arbitrary absolute paths, and they must not omit an unrecoverable historical image and continue with incomplete context. The child therefore remains in the existing error/retry state without dispatching the prompt; the existing startup/live-save migration can publish legacy paths to immutable Upload Versions before a later retry. This preserves the previously agreed Branch behavior and avoids a Branch-specific persistence or Upload-model change.

ewen-poch:
Addressed both recovery findings without adding fields, schema, IPC, or a new ownership relationship. Before ACP Session creation, Branch now deduplicates staged uploads from the copied history, finalizes them under the source Session, and replaces the existing upload references in both source and child messages with the same source-owned immutable Version refs. The main prompt owner also rejects any unreconciled staged history defensively, so a child can never consume the source bytes. A reconciliation failure leaves the immediately selected child in the existing error/retry state without dispatching the prompt. Pending-session retry now catches replay-media reconstruction failures and settles the same error state instead of rejecting the send promise. Validation: focused renderer/main tests 112/112, node and web typecheck passed, targeted lint and formatting passed. The local full suite was stopped after the recurring abnormal no-output wait; the fresh PR Gate is the authoritative full-suite check.

ewen-poch:
Addressed the legacy-history distinction. The prompt owner now classifies only unversioned .pending uploads as staged and rejects those until source reconciliation. A source-owned legacy upload remains unchanged and is resolved through the existing Project-scoped repository ownership checks; it is never re-finalized for the target Session. Branch snapshots still strip source-owned legacy absolute paths, so this does not broaden Branch persistence or change the agreed legacy retry behavior. Added regressions proving staged history is preserved/rejected and source-owned legacy history is readable without a finalization call. No fields, schema, IPC, ownership relationship, or UX behavior changed. Validation: focused tests 113/113, node/web typecheck passed, targeted lint and formatting passed.

ewen-poch:
No code change is planned for the legacy session.artifacts finding. Branch intentionally does not copy session-level Artifact ownership or Files entries; historical GENERATED cards are retained through immutable native Version references resolved under the original Project/Session ownership. Metadata-only compatibility artifacts without a native Version are not durable cross-Session references, and copying them into the child would conflict with the confirmed no-Artifact-copy model. Supporting those entries should be a separate legacy resolver or migration rather than widening Branch persistence.

ewen-poch:
Corrected the staged-history guard after the full Coverage lane exercised the generic ACP runtime path. Branch still reconciles copied staged uploads under the source Session before dispatch. Outside Branch, a genuine .pending history capability may still be finalized for the target Session as before; source-owned legacy history is excluded from finalization and remains Project-scoped. This restores the two existing runtime scenarios without changing fields, schema, ownership, or UX. Validation: prompt-owner and renderer regressions passed; the full runtime file confirmed the two upload failures are fixed, with only three unrelated local sandbox listen EPERM failures; node/web typecheck, targeted lint, and formatting passed.

ewen-poch:
Review disposition:

- The stale pending replay marker is a valid edge case. It can be fixed by clearing the existing transient marker when its prompt is no longer present. This requires no persisted field, schema, or data-model change.
- The artifact descriptor retry can race child-session persistence. A robust fix should trigger resolution from persistence acknowledgement or the post-save lifecycle instead of adding arbitrary retries. This also requires no persisted field, but it changes save/lookup lifecycle timing, so no change has been applied yet.

The final PR Gate run passed. These two review items remain open.

ewen-poch:
Addressed both open review findings:

- Pending replay state is now cleared when removeMessage or truncateSessionFromMessage removes its prompt, and appendUserMessage drops an already-stale marker instead of preserving it. Regression coverage exercises removal, edited resend, and an already-missing prompt.
- Copied Artifact descriptor resolution now waits for the existing Session persistence queue to settle before its one retry. The fixed 250 ms timer was removed, and the test holds persistence open to verify that resolution does not retry early.

No persisted Session field, schema, data-model, or interaction change was introduced. The branch was rebased onto current main and remains one commit (ffa4d7d). Focused affected-file tests pass: 452/452; lint, formatting, and Node/Web typechecks pass.

ewen-poch:
Legacy compatibility disposition:

- Fixed path-only legacy Upload replay. The branch snapshot now retains the source-owned path only while the child is pending, allowing the existing source-Session finalization path to publish an immutable Upload Version. Source and child messages are then replaced with the same Version reference before the child can persist. No new field, schema, or data-model change was introduced.
- Legacy Artifact ids remain intentionally outside this PR. Supporting them would require either copying session.artifacts into the child, which conflicts with the confirmed no-Artifact/no-Files-copy boundary, or adding a second project-session-JSON lookup mode to artifactResolver. Native immutable Artifact Version ids remain supported. The historical message text is preserved; only legacy generated-file cards are unavailable in a branched Session.

Regression coverage now verifies path-only Upload finalization and replay. After rebasing current main, 197 focused tests and Web typecheck pass. Head: e0d878c.

ewen-poch:
Addressed the branch-startup cancellation race:

- Cancel now settles an unbound pending Session locally because it has no ACP identity to cancel yet.
- Pending startup revalidates ownership of the active prompt before creation and after asynchronous creation/finalization steps, so a cancelled branch cannot bind or dispatch its prompt after history reconciliation finishes.
- Added a regression test that holds history finalization open, cancels the visible pending branch, then verifies that no ACP Session is created and no prompt is sent.

The repeated legacy Artifact recommendation remains intentionally deferred per the documented compatibility boundary in the previous comment; no Artifact metadata or Files are copied into the child.

No field, schema, data-model, or interaction change was introduced. Affected tests pass: 261/261; lint, formatting, and Node/Web typechecks pass. Head: 698f4b4.

Issue #722, opened 2026-08-04
[Bug]: Codex subscription artifact finalization fails because the terminal message uses a different runtime segment


bobo-xxx:
## Preflight checklist

- [x] I searched existing issues and this has not been reported yet.
- [x] I'm running a recent version of Open Science.

## What happened?

At the end of an analysis in the session **“Reproduce the bioinformatics analysis”**, Open Science generated two artifact files but failed to finalize them:

```text
Generated file finalization failed: Error invoking remote method 'artifacts:finalize-run': ArtifactFinalizationProofError: Artifact finalization message does not match the declared durable ownership.
```

The generated files were:

```text
pcec_download_manifest.tsv
pcec_seurat_image.tsv
```

Both artifact versions remained in the `pending` state and had no finalized `messageId`.

I expected the generated artifacts to be finalized and attached to the completed agent response.

Inspection of `main.log`, the persisted conversation graph, artifact provenance, and the artifact database showed that the artifact claim and terminal message used different runtime segments.

The artifact run was declared under:

```text
runtimeSegmentId: runtime-segment-1785827961915-65
backendId: codex:builtin-codex-subscription
promptMessageId: message-1785827961915-102
```

The terminal agent message was stored under:

```text
messageId: message-1785828579323-107
runtimeSegmentId: runtime-segment-1785828582914-66
backendId: codex:builtin-codex-isolated
responseToMessageId: message-1785827961915-102
```

The root frame, agent frame, message branch, prompt message, and artifact IDs matched. Only the runtime segment and backend identity changed.

The session graph repeatedly alternated between:

```text
codex:builtin-codex-subscription
codex:builtin-codex-isolated
```

Based on the Open Science 0.10.0 code, the configured Codex subscription provider is initially represented as `codex:builtin-codex-subscription`, while backend resolution maps it to `codex:builtin-codex-isolated`. This appears to create a new runtime segment when the terminal response is persisted.

Artifact finalization requires the terminal message's `runtimeSegmentId` to match the artifact claim's declared `runtimeSegmentId`. The otherwise matching proof is therefore rejected as `invalid-proof`.

The identifiers above were verified against the exact session named **“Reproduce the bioinformatics analysis”**, not the older sessions with similar truncated titles.

## Steps to reproduce

1. Run Open Science 0.10.0 on Windows 11.
2. Configure the **Codex subscription** provider.
3. Select the Codex framework and `gpt-5.6-sol`.
4. Start a new analysis session.
[...]

PR #725, opened 2026-08-04
fix(reviewer): preserve correction artifact provenance

ewen-poch:
## Problem

Reviewer-owned correction prompts omitted the durable conversation-graph provenance. After a pending Session was bound to its durable runtime id, ArtifactTurnOwner could synthesize a root Frame from the runtime id that did not match the durable graph, causing Artifact finalization to fail closed.

## Proposed change

- Materialize the already-loaded durable Session graph before the correction turn.
- Reuse the existing active-conversation helper to pass the full Frame, Branch, Runtime Segment, ancestry, and prompt binding through sendPrompt.
- Cover a durable graph whose root still uses a pending Session id while the correction targets a different runtime Session id.

## Scope and non-goals

- No schema, migration, durable graph rewrite, dependency, or user-interface change.
- The separate Runtime Segment mismatch reported in #722 was fixed by 92c2072 and has been closed.

- Reviewer correction durable ownership -> npm test -- --run src/main/reviewer/correction.test.ts -> 7 passed.
- Main-process type safety -> npm run typecheck:node -> passed.
- Source lint -> npm run lint -> passed with 0 errors and 17 existing warnings outside this diff.
- Portable suite -> npm test -> 11,024 passed and 2 unrelated 15-second timeouts in unmodified Notebook and ProjectFilesView tests. The affected files passed in isolation: ProjectFilesView 69/69, Notebook runtime 180/180, completion-gate 22/22, and the Notebook state-budget case 1/1.
- Independent review -> Standards 0 findings; Spec 0 findings.

The branch was rebased over non-overlapping upstream commits after validation; the focused Reviewer test was rerun post-rebase and passed 7/7.

## Review focus

Confirm that correction prompts use the durable active graph identity rather than any Session-id-derived fallback.

Related to #722.

### DIFF
--- a/src/main/reviewer/correction.test.ts
+++ b/src/main/reviewer/correction.test.ts
@@ -575,6 +575,61 @@ describe('single-round auditor correction', () => {
     expect(msg).toContain('[fail]')
   })
 
+  it('marks checks unaddressed when correction provenance is incomplete', async () => {
+    const process = new FakeAgentProcess()
+    startFakeAgent(process, {
+      reviewerSessionId: 'reviewer-session-1',
+      mainSessionId: 'main-session-1',
+      simulateFindingsViaHttp: true,
+      checksToSubmit: [
+        {
+          status: 'fail',
+          claim: 'Test claim',
+          evidence: 'Test evidence',
+          locator: { blockRef: { blockIndex: 0 }, contentHash: 'hash1' }
+        }
+      ]
+    })
+
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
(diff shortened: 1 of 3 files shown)


ewen-poch:
Addressed the P1 in 3aa0281. Correction provenance derivation now runs inside the existing fail-closed fix-loop handling. A legacy graph without a Runtime Segment does not receive a fabricated ownership context; its findings are marked unaddressed and no correction prompt is sent. Added regression coverage for this case. Reviewer correction tests pass 8/8, and affected-file ESLint passes.

PR #738, opened 2026-08-04
fix(local-rpc): route Windows app tools through named pipes

ewen-poch:
## Problem

On Windows, the app-owned stdio MCP processes start successfully but their inner loopback HTTP calls back to the Electron main process can fail with the opaque `fetch failed` error. This affects Notebook state/execution (including `bash_execute`), Artifact provenance, and Skill import because they share the same local RPC boundary.

Reviewer uses a separate app-local Streamable HTTP MCP endpoint on the same loopback boundary. Its session therefore fails before `submit_findings`, and recent readiness handling surfaces only the generic `Internal error` from ACP.

## Proposed change

- Use authenticated Windows named pipes for app-local Notebook, Artifact, Skill, and control-REPL RPC while preserving the existing HTTP/JSON protocol.
- Expose the existing scope-enforcing Reviewer MCP server through a thin stdio proxy backed by the same named-pipe transport on Windows.
- Keep loopback HTTP unchanged on macOS and Linux.
- Preserve the underlying socket error in local RPC diagnostics and structured ACP error details in Reviewer logs.
- Add no dependency; the shared transport uses Node's built-in HTTP and stream APIs.

```mermaid
flowchart LR
  A["Agent / Claude MCP child"] -->|stdio| B["App MCP process or Reviewer proxy"]
  B -->|"Windows named pipe, authenticated HTTP"| C["Electron main process"]
```

## Scope and non-goals

- No database schema, persisted data, data relationship, or user-interface change.
- Reviewer tool definitions, scope validation, token authentication, and findings submission remain in the main process.
- The previously reported white-screen/context slowdown is not changed here; the supplied log showed no renderer crash after v0.9.0, and that crash path was addressed separately by #610.

Validation below reflects the state after the last material edit.

| --- | --- | --- |
| Issued capability → execution request → REPL environment → real socket call | `npm test -- --run src/main/notebook/runtime-service.test.ts src/main/notebook/local-rpc-server.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/repl-loop.integration.test.ts` | Pass (253 tests, 30 gated skips) |
| Real local-socket HTTP plus Reviewer MCP list/call/submission path | `npm test -- --run src/main/local-rpc-transport.test.ts src/main/notebook/mcp-server.test.ts src/main/reviewer/mcp-server.test.ts` | Pass (77 tests) |
| Full main/renderer regression suite, including capability ownership and lifecycle | `npm test` | Pass (11,142 tests; 184 gated skips) |
| Node and renderer types | `npm run typecheck` | Pass |
| Repository lint | `npm run lint` | Pass with 0 errors (17 pre-existing warnings in unrelated files) |
| Bundled Electron entry, including the lazy Reviewer proxy entry | `npm run build:e2e` | Pass |
| Patch hygiene | `git diff --check` | Pass |

One intermediate full-suite run hit the unchanged `completion-gate.execute-control` 15-second timeout under concurrent load. Its file passed 22/22 in isolation, and the final full suite then passed cleanly.

Uncovered risk: this macOS workspace cannot execute an actual packaged Windows build. The transport is covered with real Unix-domain sockets using Node's cross-platform `socketPath` API and the Electron bundle succeeds, but Windows CI or a packaged Windows smoke test should confirm host endpoint-security behavior.

## Review focus

- Named-pipe fetch streaming, abort, error, and server lifecycle behavior, including propagation through every issued RPC capability.
- Reviewer stdio proxy authentication and enforcement that Reviewer receives exactly one app-owned MCP capability.
- Windows-only transport selection and unchanged macOS/Linux loopback behavior.


ewen-poch:
Addressed the first Codex review in `eb6270b`:

- P1: `socketPath` now propagates from `ensureStarted()` through `issueSessionConnection`, `issueSkillImportConnection`, and `issueControlConnection`.
- P2: all production consumers are socket-aware. The control REPL receives `OPEN_SCIENCE_MCP_RPC_SOCKET_PATH` and uses Node HTTP over `socketPath`; Notebook, Artifact, Skill, and Reviewer already use the shared socket-aware client/proxy. A source scan finds no remaining direct production fetch of an issued endpoint.
- Added a real child-process regression that invokes `host.mcp` over a local socket and verifies the Bearer token, plus seam tests from capability issuance through executor environment.

Validation after the fix: targeted impact set 253 passed; final full suite 11,142 passed; typecheck, lint (0 errors), and `build:e2e` passed.

PR #751, opened 2026-08-05
feat(agent-runtime): hot-switch ACP models and providers

ewen-poch:
## Problem

Changing an ACP model or API provider currently reconnects the agent process even when the active
adapter can update the existing session. That loses the lightweight transition, may trigger native
resume or transcript replay, and makes a model-only selection look like a process failure in logs.

The four supported routes have different constraints: Claude Code, OpenCode, Codex Responses, and
Codex bridge cannot safely share one unconditional switching rule. Provider base URLs, credentials,
image capabilities, adapter catalogs, and transport type all affect whether a live switch is valid.

## Proposed change

- Apply registered, compatible model/provider targets to the existing ACP generation and sessions.
- Keep the current in-flight message on its original model. User selections remain enabled; a
  latest-wins pending target is applied after generation activity drains, while new prompts wait at
  an admission barrier.
- Route compatible Claude, OpenCode, and Codex API-key providers through generation-owned loopback
  transports so endpoint, credential, and model changes can occur without exposing real keys to the
  ACP child process.
- Preserve existing fail-closed reconnect behavior for framework, auth-lane, or wire-route changes;
  unregistered targets; unsafe image downgrades; and Codex same-slug capability conflicts.
- Emit `session transcript replay dispatched` only when a history preamble is actually sent to ACP.
- Isolate the pending/barrier/drain/fallback state machine in `AcpModelChangeWorkflow`. Connection
  bootstrap remains in `AcpConnectionLifecycleWorkflow`; the mainline `AcpConnectionCloseWorkflow`
  coordinates model-change cancellation/drain during close, reconnect, and retirement; and the
  connection adapter transfers transport leases to the connection resource owner.

```mermaid
flowchart LR
  Select["Persist provider/model selection"] --> Eligible{"Registered compatible target?"}
  Eligible -->|No| Reconnect["Existing deferred reconnect"]
  Eligible -->|Yes| Busy{"Generation busy?"}
  Busy -->|Yes| Pending["Replace pending target; latest wins"]
  Pending --> Drain["Wait for current activity to finish"]
  Busy -->|No| Apply["Retarget transport and session config"]
  Drain --> Apply
  Apply --> Same["Same process and session; no replay"]
  Apply -->|Rejected| Reconnect
```

## Scope and non-goals

- No Prisma migration, persisted settings schema change, renderer IPC contract change, or new UI.
- The current message is never moved between models mid-generation.
- Framework changes, subscription/shared/isolated auth changes, and
  `codex-response`/`codex-bridge` route changes still reconnect.
- Custom providers still expose the existing configured model shape; this does not add a persistent
  multi-model provider catalog.
- The local design record under `docs/internal/` is intentionally Git-ignored and is not part of this
  pull request.

All listed checks ran after the last material edit and after rebasing onto `origin/main@ab70b78`.

| Behavior / risk | Final check | Result |
| --- | --- | --- |
| Busy latest-wins switching, prompt admission barrier, close/reconnect ordering, fallback recovery | `npm test -- src/main/settings/backend-resolver.test.ts src/main/acp/model-change-workflow.test.ts src/main/acp/connection-close-workflow.test.ts src/main/acp/agent-connection-adapter.test.ts src/main/acp/connection-lifecycle-workflow.test.ts src/main/acp/connection-resource-owner.test.ts src/main/acp/runtime.test.ts` | 7 files and 494 tests passed |
| Connection adapter and transport lease ownership | Same affected-ACP test command | Covered by the 494 passing tests |
| Node and renderer type consumers | `npm run typecheck` | Passed |
| Source lint | `npm run lint` | Passed with 0 errors; 17 pre-existing warnings are in unchanged files |
| Cross-module and loopback transport regression coverage | `npm test` | 762 files passed, 14 skipped; 11,273 tests passed, 184 skipped |

The complete suite was run outside the restricted filesystem sandbox because project-owned bridge
and RPC tests must listen on `127.0.0.1`; the sandbox run failed with `listen EPERM` before the same
suite passed with loopback access.

Independent review identified four compatibility gaps: Codex cross-provider same-slug capability
metadata, clearing a previous reasoning override when the new target resolves to `default`, avoiding
inactive Claude bridge targets, and preserving a newer effort selection behind a queued model
switch. All now fail closed, reset explicitly, or serialize through the existing barrier and have
regression coverage.

Uncovered risk: real third-party provider credentials were not used in automated tests. Transport
behavior was validated with loopback upstreams and fake keys; existing reconnect remains the fallback
when a live adapter rejects a target.

## Review focus

- The boundary between `AcpModelChangeWorkflow`, runtime target policy, connection lifecycle, and
  the newly extracted connection-close workflow from #750.
- Atomic ordering of transport retargeting, ACP config-option updates, generation state commit, and
  reconnect recovery.
- Image downgrade and Codex same-slug fail-closed conditions.
- Lease transfer/release behavior across failed connection candidates, retirement, and shutdown.

### DIFF
--- a/src/main/acp/agent-connection-adapter.test.ts
+++ b/src/main/acp/agent-connection-adapter.test.ts
@@ -153,9 +153,11 @@ describe('AcpAgentConnectionAdapter', () => {
     expect(connectionHooks.markProcessExitExpected).toHaveBeenCalledWith(process, 1)
   })
 
-  it('reaps an untransferred process tree and releases its bridge lease exactly once', async () => {
+  it('reaps an untransferred process tree and releases every transport lease exactly once', async () => {
     const process = new FakeAgentProcess()
-    const release = vi.fn(async () => undefined)
+    const releaseBridge = vi.fn(async () => undefined)
+    const releaseAnthropic = vi.fn(async () => undefined)
+    const releaseProviderTransport = vi.fn(async () => undefined)
     const backend: ResolvedAgentBackend = {
       framework: { ...claudeCodeFramework, spawn: () => asAgentProcess(process) },
       executablePath: '/bin/agent',
@@ -164,7 +166,15 @@ describe('AcpAgentConnectionAdapter', () => {
         selectSkills: vi.fn(async () => []),
         registerReviewerSession: vi.fn(),
         unregisterReviewerSession: vi.fn(() => false),
-        release
+        release: releaseBridge
+      },
+      anthropicBridgeLease: {
+        setTarget: vi.fn(() => true),
+        release: releaseAnthropic
+      },
+      providerTransportLease: {
+        setTarget: vi.fn(() => true),
+        release: releaseProviderTransport
       }
     }
     const candidate = await openCandidate(process, backend)
@@ -174,7 +184,9 @@ describe('AcpAgentConnectionAdapter', () => {
 
     expect(terminateProcessTree).toHaveBeenCalledOnce()
     expect(terminateProcessTree.mock.calls[0]?.[0]).toBe(process)
-    expect(release).toHaveBeenCalledOnce()
+    expect(releaseBridge).toHaveBeenCalledOnce()
+    expect(releaseAnthropic).toHaveBeenCalledOnce()
+    expect(releaseProviderTransport).toHaveBeenCalledOnce()
   })
 
   it('rejects a transfer to a different owner epoch without consuming the candidate', async () => {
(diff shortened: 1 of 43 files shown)


ewen-poch:
Addressed in `d107f8e`.

`resolveActiveModelChangeTarget()` now reuses the existing Claude bridge-catalog decision and only emits `anthropicBridgeTargetId` when the active generation would actually own an Anthropic bridge lease. A single-provider Claude generation now keeps the lightweight ACP model-option path instead of forcing a reconnect.

Added regression coverage for the no-bridge case. `backend-resolver.test.ts` passes 34/34, targeted lint passes, and `typecheck:node` passes.

ewen-poch:
Addressed in `337e172`.

`applyReasoningEffortChange()` now waits on the existing model-change admission barrier before updating the live session. A queued model target can still apply its captured effort as part of the switch, but the latest persisted effort is then applied to the new model/session, so the queued target cannot overwrite the newer selection. This reuses the existing barrier and adds no queue or state.

Added a regression test for a model switch queued behind an active prompt followed by a newer effort selection. The full suite passes (762 files, 11,273 tests), full typecheck passes, and lint completes with 0 errors (17 pre-existing warnings).

PR #754, opened 2026-08-05
feat(acp): make context replay agent aware

ewen-poch:
## Problem

When an ACP session cannot be resumed after switching agents or starting a replacement session, the app currently replays a character-truncated tail of the transcript. That can lose the original task, split user/assistant turns, replay unrelated media, and spend the same context budget across Claude Code, OpenCode, direct Codex Responses, and the Codex bridge even though their context paths differ.

Notebook continuity is also only available through a later tool call, so a replacement agent can begin its first turn without the small amount of live execution state needed to continue safely.

## Proposed change

- classify replay targets as `claude-code`, `opencode`, `codex-response`, or `codex-bridge` and derive a bounded token budget from the target and advertised context window
- build replay from user-led turns, retaining the original task plus a contiguous recent suffix instead of slicing the final characters
- preserve role labels, avoid orphan assistant replies, account conservatively for CJK text, and truncate oversized messages within their role
- replay attachments and images only when their source message survives text selection
- share one cumulative text budget across injected file contents
- peek at already-loaded Notebook execution state and inject a compact, one-shot continuity block on the first replayed prompt
- keep replay and Notebook continuity private to the agent prompt; the user-facing message remains the text the user entered

```mermaid
flowchart LR
  A["ACP context reset"] --> B["Resolve target agent"]
  B --> C["Select original task + recent turns"]
  C --> D["Select media from admitted messages"]
  B --> E["Peek loaded Notebook state"]
  D --> F["Compose private replay prompt"]
  E --> F
  F --> G["Replacement agent first turn"]
  H["Current user message"] --> F
  H --> I["User-facing conversation bubble"]
```

## Scope and non-goals

- This changes context assembly and first-turn injection behavior.
- It does not change database fields, schemas, relationships, migrations, or persisted session data.
- Notebook state is read only from an already-loaded in-memory session; the peek does not create or reload a session.
- It does not change Notebook/MCP/function-call return projections, permission-message presentation, connector Skill materialization, or connector result schemas.

The checks below ran after the final rebase/conflict resolution.

- target-specific replay budgets, original-task anchoring, contiguous recent turns, CJK accounting, and media correlation -> `npm test -- --exclude docs/internal/2026-08-05-context-lifecycle-replay-spike.test.ts` -> 11,283 passed, 184 skipped
- cumulative file-content admission and one-shot Notebook continuity -> same complete tracked Vitest suite -> passed
- context-reset runtime composition and private prompt/user-message separation -> `npm test -- --run src/main/acp/runtime.test.ts` -> 415 passed
- Node and renderer contracts -> `npm run typecheck` -> passed
- linted source/configuration -> `npm run lint` -> 0 errors; 17 existing warnings outside this diff

The excluded `docs/internal` spike is gitignored, absent from this PR and clean CI checkouts, and intentionally captures pre-implementation behavior. No platform-specific persistence or migration lane is required because the change adds neither.

## Review focus

- whether the four target policies leave sufficient headroom for the active turn and agent-owned system context
- whether turn grouping and oversized-message fallback always preserve a usable user instruction
- whether media selection stays aligned with admitted transcript messages
- whether Notebook continuity remains one-shot, read-only, compact, and invisible in the user-facing message

Uncovered risk: provider behavior is verified at local interfaces and runtime fakes, not with live end-to-end sessions against every external ACP agent.

### DIFF
--- a/src/main/acp/attachment-content.ts
+++ b/src/main/acp/attachment-content.ts
@@ -165,33 +165,7 @@ export const isDatasetAttachment = (name: string, mimeType?: string): boolean =>
   )
 }
 
-// Raster image extensions that vision-capable agents can read as pixels, mapped to the MIME type to
-// inline them as. SVG is deliberately absent: it is vector/text, not a raster type the providers accept
-// as image content. Matches the accepted set in ACP_MESSAGE_IMAGE_MIME_TYPES.
-const IMAGE_EXTENSION_MIME = new Map<string, string>([
-  ['png', 'image/png'],
-  ['jpg', 'image/jpeg'],
-  ['jpeg', 'image/jpeg'],
-  ['gif', 'image/gif'],
-  ['webp', 'image/webp'],
-  ['avif', 'image/avif']
-])
-
-// The image MIME type to inline a file as, or undefined when it is not a supported raster image. A
-// concrete image/* MIME wins; a missing or generic MIME (browsers omit it for some drag/drop and paste
-// sources) falls back to the file extension — so a `.png` upload is still sent as pixels rather than a
-// bare file link the model can only reach by writing code. A concrete non-image MIME is authoritative.
-export const imageAttachmentMimeType = (name: string, mimeType?: string): string | undefined => {
-  const essence = mimeEssence(mimeType)
-
-  if (essence && essence.startsWith('image/')) {
-    return essence === 'image/svg+xml' ? undefined : essence
-  }
-
-  if (essence && !GENERIC_MIME_TYPES.has(essence)) return undefined
-
-  return IMAGE_EXTENSION_MIME.get(fileExtension(name))
-}
+export { imageAttachmentMimeType } from '../../shared/uploads'
 
 // Human-readable byte size for the notice (binary units, one decimal past KB).
 export const formatBytes = (bytes: number): string => {
--- a/src/main/acp/prompt-content-owner.test.ts
+++ b/src/main/acp/prompt-content-owner.test.ts
@@ -230,7 +230,7 @@ describe('AcpPromptContentOwner', () => {
       references: [],
       codexSkillInputs: [],
       skillImportEnabled: false,
-      fileTextBudget: 500
+      fileTextBudget: 2_000
     })
 
     const fileText = contentBlocks(result.content)
@@ -244,7 +244,7 @@ describe('AcpPromptContentOwner', () => {
     expect(fileText.join('\n')).toContain('END-TWO')
     expect(
       fileText.reduce((total, text) => total + estimateHistoryTokens(text), 0)
-    ).toBeLessThanOrEqual(500)
+    ).toBeLessThanOrEqual(2_000)
     expect(fileText.join('\n')).not.toContain('a'.repeat(1_000))
     expect(fileText.join('\n')).not.toContain('b'.repeat(1_000))
   })
--- a/src/main/acp/prompt-content-owner.test.ts
+++ b/src/main/acp/prompt-content-owner.test.ts
@@ -5,6 +5,7 @@ import { join } from 'node:path'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 
 import type { UploadedAttachment } from '../../shared/uploads'
+import { estimateHistoryTokens } from '../../shared/history-preamble'
 import { UploadRepository } from '../uploads/repository'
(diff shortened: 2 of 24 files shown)


ewen-poch:
Addressed both review findings in `f717641`:

- Historical uploads now preserve every selected file. Image capability checks only treat `image/*` attachments as image input, so text/PDF/archive resources continue through the existing resource-link path on text-only models.
- Budget projection now retains a media-only Assistant message with an explicit placeholder and keeps its message index, so its selected image is replayed.

Validation: 118 focused replay/runtime tests passed, full typecheck passed, affected-file lint passed, and the complete tracked Vitest suite passed (11,284 passed; 184 skipped).

ewen-poch:
Addressed both findings in c410bb4:

- Legacy path-only historical files remain available to provider replay as resource links, but are excluded from Notebook turn-input registration unless they carry an immutable Version identity. Current uploads keep their existing registration behavior.
- The existing 10-upload replay cap now prioritizes selected image uploads, then fills remaining slots with the newest non-image files while preserving chronological output order.

Validation after rebasing onto the latest main: 126 focused tests passed, full typecheck passed, and affected lint passed. The complete tracked Vitest suite also passed before the conflict-free rebase: 11,285 passed and 184 skipped.

ewen-poch:
Addressed all three findings in 7a49dc8:

- Replay media is now selected with the active model capability before dispatch. Branch creation, pending-session retry, edited resend, and normal reset replay omit unsupported historical images while retaining the text preamble and replayable files.
- The 10-upload cap is capability-aware: vision models keep image-first selection, while text-only models remove image candidates before filling the cap with the newest documents.
- Renderer replay selection and main-process prompt materialization now use the same extension-aware raster classifier, including generic or missing MIME metadata such as application/octet-stream for a .png file.

Validation: 150 focused tests passed, full typecheck and affected lint passed, and the complete tracked Vitest suite passed with 11,294 tests passed and 184 skipped.

ewen-poch:
Addressed both replay-budget findings in 4e137a6:

- Full-history admission now accumulates formatted turn costs and stops as soon as the budget is exceeded. Only admitted turns are materialized; oversized single-message prefix/suffix selection uses linear scans without code-point arrays or repeated slice/join work.
- Replay and attachment-preview admission now use UTF-8 byte length as a dependency-free tokenizer upper bound. The final packet is checked against the same bound before dispatch.

A local spike with 20,000 messages and about 20.1M source characters produced a 6,408-byte replay under an 8,000 budget in about 23ms without materializing the full transcript.

Validation: 150 focused tests and full typecheck passed after rebasing onto the latest main. Before the conflict-free rebase, affected lint passed and the complete tracked Vitest suite passed with 11,294 tests passed and 184 skipped.

ewen-poch:
Addressed the image-history compatibility finding in 5f3a5f9.

The text-only runtime adoption check now scans all persisted messages for inline images and image uploads, independently of the bounded replay selection. The replay payload remains budgeted and still omits unsupported image media after the reset.

Added a regression test with an image turn deliberately omitted by a 600-byte replay budget. Validation passed: 121 focused tests, full typecheck, affected lint, and the complete tracked Vitest suite with 11,302 passed and 184 skipped.

ewen-poch:
Addressed both replay-policy findings in 3de497f.

- Interrupted resume and overflow recovery now derive the replay target and context window from the persisted Session framework, backend, and model. The provider is resolved from its stored backend identity rather than the currently selected Settings entry.
- If a persisted Codex provider is no longer available, the resolver uses the conservative Codex bridge policy. Descriptor-free non-workspace callers, including reviewer context recovery, also default to the 8,000-byte bridge cap instead of the 16,000-byte Claude cap.

No new persisted fields or runtime metadata channel were added. Validation passed: 137 focused tests, full typecheck, affected lint, and the complete tracked Vitest suite with 11,305 passed and 184 skipped.

ewen-poch:
Addressed the Notebook continuity finding in e28cc42. Prompt requests now carry an explicit transient contextReset signal whenever the provider context was replaced, independently of whether the bounded transcript contains replayable text. Notebook handoff reads that signal for interrupted first turns, branch/edit resets, runtime switches, overflow replacement, forced-skill fallback replacement, reviewer activity, and Task runs; ordinary continuations remain unchanged. The signal is request-scoped only: no Session persistence, data-model, relationship, or UI changes. Validation passed: 615 focused tests, full typecheck, affected lint, and the complete tracked Vitest suite with 11,307 passed and 184 skipped.

PR #763, opened 2026-08-05
feat(settings): add settings keyboard shortcut

ewen-poch:
## Problem

Settings can only be opened through visible UI controls, without the standard platform keyboard shortcut. Cmd/Ctrl+W also falls through to preview or window handling while transient modals are active.

## Proposed change

- Open the existing Settings panel with Cmd+, on macOS and Ctrl+, on Windows and Linux.
- Ignore the open shortcut until startup is interactive or while another top-level modal is active.
- Route Cmd/Ctrl+W through dismissible surfaces before preview panes and the window.
- Return a Settings breadcrumb page to its current panel root before closing Settings on the next chord.

## Scope and non-goals

- Renderer-only interaction change.
- No IPC, data model, persistence, settings content, dependency, or application-menu framework changes.
- Blocking approval, recovery, and quit-confirm dialogs keep their explicit decision semantics and are not silently dismissed.

All listed local checks ran after the last material edit.

- Settings open shortcut, startup/modal gates, close precedence, and breadcrumb navigation -> npm test -- src/renderer/src/App.test.tsx src/renderer/src/hooks/useCloseActivePaneShortcut.test.ts src/renderer/src/pages/settings/SettingsPage.render.test.tsx -> passed, 96 tests.
- Source lint -> npm run lint -> passed with 0 errors and 17 pre-existing warnings outside the changed files.

The final PR Gate run provides the complete portable-suite and platform evidence. Native OS accelerator delivery remains the only uncovered manual risk.

## Review focus

Please review the top-level modal guard, close-shortcut precedence, and Settings breadcrumb-to-root behavior.

### DIFF
--- a/src/renderer/src/App.test.tsx
+++ b/src/renderer/src/App.test.tsx
@@ -447,6 +447,22 @@ describe('App startup routing', () => {
     dialog.remove()
   })
 
+  it('ignores a closed dialog portal when opening Settings', async () => {
+    mocks.settings.isLoaded = true
+    await render()
+    const dialog = document.createElement('div')
+    dialog.setAttribute('role', 'dialog')
+    dialog.dataset.state = 'closed'
+    document.body.appendChild(dialog)
+
+    window.dispatchEvent(
+      new KeyboardEvent('keydown', { key: ',', metaKey: true, cancelable: true })
+    )
+
+    dialog.remove()
+    expect(mocks.settings.openSettings).toHaveBeenCalledOnce()
+  })
+
   it('closes the update dialog before underlying surfaces', async () => {
     mocks.settings.isLoaded = true
     mocks.settings.isSettingsOpen = true
--- a/src/renderer/src/App.test.tsx
+++ b/src/renderer/src/App.test.tsx
@@ -35,7 +35,10 @@ const mocks = vi.hoisted(() => {
       init: vi.fn().mockResolvedValue(undefined),
       retry: vi.fn().mockResolvedValue(undefined)
     },
-    preview: { fileDialogItem: undefined as unknown | undefined },
+    preview: {
+      fileDialogItem: undefined as unknown | undefined,
+      expandedToolItemId: null as string | null
+    },
     loadProjects: vi.fn().mockResolvedValue(undefined),
     deepLinkNavigation: vi.fn(),
     lifecycleSync: vi.fn(() => ({
@@ -291,6 +294,7 @@ describe('App startup routing', () => {
     mocks.compute.pendingApprovals = []
     mocks.skillImport.pending = []
     mocks.preview.fileDialogItem = undefined
+    mocks.preview.expandedToolItemId = null
     mocks.deepLinkNavigation.mockClear()
     mocks.lifecycleSync.mockClear()
     mocks.syncWindowFindAppearance.mockClear()
@@ -450,6 +454,18 @@ describe('App startup routing', () => {
     })
 
     expect(mocks.globalSearch.props?.open).toBe(false)
+  })
+
+  it('does not open Settings under an expanded preview modal', async () => {
+    mocks.settings.isLoaded = true
+    mocks.preview.expandedToolItemId = 'project-files'
+    await render()
+
+    window.dispatchEvent(
+      new KeyboardEvent('keydown', { key: ',', metaKey: true, cancelable: true })
+    )
+
+    expect(mocks.settings.openSettings).not.toHaveBeenCalled()
   })
 
   it('shows startup progress until settings have loaded', async () => {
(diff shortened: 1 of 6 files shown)


ewen-poch:
Fixed in bf3ffe4 by requiring isSessionPersistenceHydrated before handling Cmd/Ctrl+,. This matches the existing App recovery semantics: initial loading and hard failures remain unhydrated, while partial-load warnings are hydrated and interactive. The App test now covers settings startup, session loading, hard failure, and recovery. The targeted test, renderer typecheck, and lint pass.

ewen-poch:
Addressed the modal-integration finding in 1b07a79. Opening Settings is now blocked by the existing top-level modal gates, and Cmd/Ctrl+W follows the close ladder for dismissible surfaces: global search, Settings breadcrumb/root, transient file/tool previews, preview tab/pane, then the window. A Settings breadcrumb page returns to its current panel root before the next chord closes Settings. Blocking approval, recovery, and quit-confirm dialogs retain their explicit decision semantics and are not silently dismissed. The 96 targeted tests, renderer typecheck, and lint pass.

ewen-poch:
Fixed in 522192d. The close handler now checks UpdateDialog before global search and Settings. A dismissible update dialog closes through its existing closeDialog action; the applying state consumes the shortcut without closing the underlying Settings surface. The regression test covers UpdateDialog open over Settings, and the 97 targeted tests, renderer typecheck, and lint pass.

ewen-poch:
Fixed in a34f7f9. The Settings-open guard now tracks expandedToolItemId in addition to fileDialogItem, so Cmd/Ctrl+, cannot open Settings beneath an expanded Files/tool preview. The regression test covers the expanded-preview state. The 98 targeted tests, renderer typecheck, and lint pass.

ewen-poch:
Addressed. Preview modal state now participates in Cmd/Ctrl+W and Cmd/Ctrl+, guards only while the navigation view is Workspace. Home therefore falls through to window close or Settings even if preview state persists. Regression tests cover both paths while preserving the existing per-project preview state.

ewen-poch:
Addressed both findings. Cmd/Ctrl+W now detects a nested Settings dialog or alert dialog and delegates one Escape event to Radix before breadcrumb or parent handling, so dismissible children close first while existing non-dismissable protections remain effective. The direct parent-close path also resets the mobile navigation drawer. The existing shortcut ladder and mobile navigation tests cover both behaviors.

ewen-poch:
Addressed both findings. Cmd/Ctrl+, now refuses to open Settings while any active dialog or alert dialog is present, covering workspace-local workflows without enumerating each modal. The Settings close ladder is now nested dialog, mobile navigation drawer, breadcrumb root, then parent Settings. Regression tests cover the workspace-dialog gate and drawer-before-breadcrumb behavior.

ewen-poch:
Addressed. Shortcut-driven breadcrumb closure now moves the existing history index back when a parent entry is available. Direct deep links with no parent replace the current entry with the breadcrumb root instead of pushing a duplicate. Regression coverage verifies that Back stays disabled after both normal drill-down closure and direct-link closure.

ewen-poch:
Addressed. Nested Settings dialog detection now targets only open Radix dialog and alert-dialog layers via data-state="open". The inline Specialist ZIP preview therefore no longer consumes Cmd/Ctrl+W and proceeds through the normal breadcrumb history path. Regression coverage keeps the real nested Radix dialog case and adds the inline role="dialog" case.

ewen-poch:
Addressed the active-dialog Escape finding in 44b9b4a. Cmd/Ctrl+W now dispatches Escape to the last active nested dialog element instead of document, so element-bound handlers such as Link Safety receive it. The regression test renders the real LinkSafetyModal and verifies its close callback is invoked.

ewen-poch:
Addressed the inactive-expansion finding in 77acf73. Expanded preview state now counts as a visible modal only when the preview panel is open and the expanded tool is the active item. Cmd/Ctrl+W applies the same visibility check before clearing expansion state, so switching to another preview closes that active tab instead of consuming the shortcut. Regression tests cover both Settings-open and close-pane paths.

ewen-poch:
Addressed the fullscreen-viewer finding in 09d9103. The Settings shortcut guard now reuses STREAMDOWN_FULLSCREEN_SELECTOR alongside dialog and alert-dialog detection, covering both table and Mermaid fullscreen layers. The App regression test verifies Cmd/Ctrl+, is ignored while a Streamdown table fullscreen layer is present.

ewen-poch:
Addressed the closed-portal finding in dfabc97. The Settings shortcut selector now excludes role=dialog and role=alertdialog elements with data-state=closed while continuing to block open Radix surfaces and custom dialogs without data-state. A regression test verifies a closing portal no longer suppresses Cmd/Ctrl+,.

PR #774, opened 2026-08-05
fix(acp): correct Claude refusal attribution

ewen-poch:
## Problem

A live Claude Code session using a custom Anthropic-compatible provider completed two native `WebSearch` calls and produced a useful partial answer, then ended with `stopReason: "refusal"`. Claude Code appended fixed Anthropic AUP wording, misleadingly attributing the selected provider refusal to Anthropic.

## Proposed change

Replace only the fixed Claude Code AUP prefix with provider-neutral wording while normalizing visible ACP assistant events. Scope the rewrite to Claude Code sessions and preserve the remaining message, tool results, partial answer, and actual stop reason.

## Scope and non-goals

- No provider response or stop-reason rewriting
- No automatic retry or safety bypass
- No architecture, persistence, schema, data-model, or data-relationship changes
- User messages and non-Claude frameworks remain unchanged

All checks below ran after the last material source edit.

- Targeted behavior: `npm test -- src/main/acp/runtime-events.test.ts src/main/acp/session-update-projector.test.ts` — 25 passed
- Node types: `npm run typecheck:node` — passed
- Lint: `npm run lint` — passed with 0 errors and 17 pre-existing warnings outside changed files
- Formatting: Prettier check on all changed files — passed
- Full portable suite: `npm test` — 767 files passed, 14 skipped; 11,359 tests passed, 184 skipped

Renderer consumer coverage is excluded because the event shape is unchanged; only an exact assistant-text prefix is normalized. Platform/E2E lanes are excluded because the change does not touch platform-specific behavior, persistence, or build configuration.

Known residual risk: if Claude Code changes its refusal wording, exact-prefix normalization intentionally stops matching until the prefix is updated.

## Review focus

Please verify exact-prefix matching, Claude-only routing, and preservation of user/non-Claude messages.

### DIFF
--- a/src/main/acp/runtime-events.test.ts
+++ b/src/main/acp/runtime-events.test.ts
@@ -29,6 +29,37 @@ describe('ACP runtime event normalization', () => {
     })
   })
 
+  it('rewrites Claude Code policy attribution only for assistant messages', () => {
+    const text =
+      'API Error: Claude Code is unable to respond to this request, which appears to violate our Usage Policy (https://www.anthropic.com/legal/aup). Try rephrasing the request in a new session or change your model.'
+    const notification: SessionNotification = {
+      sessionId: 'session-1',
+      update: {
+        sessionUpdate: 'agent_message_chunk',
+        content: { type: 'text', text }
+      }
+    }
+
+    expect(toAcpRuntimeEvent(notification, 'event-refusal', 1710000000000, true).text).toBe(
+      'The selected model declined to complete this response under its safety policy. Try rephrasing the request in a new session or change your model.'
+    )
+    expect(toAcpRuntimeEvent(notification, 'event-other-agent', 1710000000000).text).toBe(text)
+    expect(
+      toAcpRuntimeEvent(
+        {
+          sessionId: 'session-1',
+          update: {
+            sessionUpdate: 'user_message_chunk',
+            content: { type: 'text', text }
+          }
+        },
+        'event-user-refusal',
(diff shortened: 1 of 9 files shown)


ewen-poch:
Thanks. I traced the refusal producer in the pinned `@agentclientprotocol/claude-agent-acp@0.60.0`. When `stop_reason === "refusal"`, the adapter sends the complete `lastRefusalExplanation` in one `sendUpdate` call as a single `agent_message_chunk`; that value is captured whole from `stop_details.explanation`. Regular assistant deltas can be split, but this adapter-authored policy message does not use that path.

Adding per-message buffering would introduce delayed output plus flush and lifecycle state for an input the current producer cannot emit, so the exact per-chunk normalization is intentionally kept stateless.

ewen-poch:
Rebased onto the latest `main` without conflicts. The source diff is unchanged. Post-rebase validation passes: 26 targeted ACP tests and `npm run typecheck:node`.

ewen-poch:
Addressed in `960a274`. The exact-prefix normalizer is now shared and re-applied after complete stream assembly in both the renderer session store and TaskRunner output. A regression test splits the prefix across two chunks and verifies both the visible message and persisted projection use the provider-neutral wording.

This keeps the producer fast path while avoiding per-message buffering or lifecycle state. Validation: 124 targeted tests, Node/Web typechecks, lint with 0 errors, formatting, and the full portable suite with 11,385 tests passed.

PR #816, opened 2026-08-06
fix(acp): suppress synthetic Codex compaction warning

ewen-poch:
## Problem

Native Codex emits a warning after context compaction. `@agentclientprotocol/codex-acp` 1.1.4 flattens that warning into an unscoped `agent_message_chunk`, so Open Science persisted and rendered adapter-authored text as part of the assistant response.

## Proposed change

Suppress the exact known Codex compaction warning at the session update projection boundary only when:

- the active framework is Codex;
- the chunk has no `messageId`; and
- its normalized text exactly matches the upstream warning.

The separate context-compacted notice and all model-scoped/provider messages remain visible.

## Scope and non-goals

- No architecture, data-model, persistence-schema, or ACP protocol changes.
- No renderer or shared-interface changes.
- No change to compaction triggers, token accounting, or prompt continuation.
- Other provider warnings remain untouched; wording changes fail open.

- Synthetic Codex warning is suppressed while model-scoped and non-Codex copies remain visible → `npm test -- src/main/acp/session-update-projector.test.ts` → 10/10 passed.
- Electron main-process types remain sound → `npm run typecheck:node` → passed.
- Linted source remains valid → `npm run lint` → 0 errors; 17 pre-existing warnings outside this diff.
- Full portable suite remains green → `npm test` → 802 files passed, 11,799 tests passed; 14 files / 184 tests skipped.

Renderer, preload, shared-contract, and platform lanes are excluded because the final diff changes only main-process projection behavior without changing an interface or adapter schema.

Independent review is pending the PR Reviewer.

## Review focus

Please verify the exact-text and missing-`messageId` guards are narrow enough to avoid suppressing genuine assistant output.

### DIFF
--- a/src/main/acp/session-update-projector.test.ts
+++ b/src/main/acp/session-update-projector.test.ts
@@ -210,6 +210,50 @@ describe('AcpSessionUpdateProjector', () => {
     expect(effects.every(Object.isFrozen)).toBe(true)
   })
 
+  it('suppresses only the unscoped Codex compaction warning', () => {
+    const projector = createProjector()
+    const warning =
+      'Warning: Heads up: Long threads and multiple compactions can cause the model to be less accurate. Start a new thread when possible to keep threads small and targeted.\n\n'
+    const notification: SessionNotification = {
+      sessionId: 'session-1',
+      update: {
+        sessionUpdate: 'agent_message_chunk',
+        content: { type: 'text', text: warning }
+      }
+    }
+    const routing = {
+      framework: 'codex' as const,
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed in commit 17ceafa: the suppression now requires the raw update to be agent_message_chunk, so identical user_message_chunk content remains visible. Added a regression assertion for the unscoped user-chunk case. Validation: targeted projector tests, Node typecheck, lint with 0 errors, and the full test suite all pass.

PR #818, opened 2026-08-06
fix(runtimes): keep setup actions pending until refresh

ewen-poch:
## Problem

During app-managed Python or R setup, the renderer can receive the terminal `done` progress event before the provisioning IPC call returns. After the call returns, runtime discovery can also take a few seconds to refresh. Both gaps allowed the shared runtime panel to fall back to an enabled **Download and set up** action even though the original setup was still settling.

The panel is reused by both Settings → Runtimes and the onboarding Notebook runtime step, so both surfaces were affected.

## Proposed change

- Keep explicit per-language provision/repair runs in the preparing state until their IPC call returns, without changing automatic provision completion semantics.
- Track the panel's post-provision discovery refresh per language.
- Keep **Cancel** available while provisioning is still in flight.
- Show a disabled **Finishing setup…** action while the environment list refreshes, preventing a duplicate install request.
- Add regression coverage for the early `done` event and the delayed discovery refresh.

## Scope and non-goals

- No architecture, persistence, data-model, or data-relationship changes.
- No changes to the Windows cache ACL implementation or provisioning backend.
- The interaction change is limited to accurately preserving the setup lifecycle in the existing shared UI.

Checks below ran after the last material edit.

- Explicit setup remains pending after an early `done` event → targeted store regression → passed.
- Setup is not re-enabled during environment refresh → targeted shared-panel regression → passed.
- Settings and onboarding consumers remain compatible → four-file impact set → **57/57 passed**.
- Renderer types → `tsc --noEmit -p tsconfig.web.json --composite false` → passed.
- Changed source/tests lint → ESLint on the four changed files → passed.
- Full portable Vitest suite → **11,701 passed, 260 skipped, 38 failed** out of 11,999 tests. The remaining failures are outside the changed renderer files and are tied to the local Windows/CI environment: symlink creation permissions, `SetAccessControl` denial, Linux/CI script assumptions, local RPC proxy behavior, and unrelated 15-second timeouts.

Uncovered risk: a packaged Windows build was not manually driven through a real runtime download in this environment. The deterministic renderer/store regressions cover the two state gaps directly.

## Review focus

- The boundary between terminal progress events and explicit IPC completion in `notebook-env-store`.
- The disabled finishing state during the shared panel's environment refresh.

### DIFF
--- a/src/main/notebook/application-commands.test.ts
+++ b/src/main/notebook/application-commands.test.ts
@@ -268,15 +268,18 @@ describe('Notebook application commands', () => {
     await expect(
       router.dispatcher.invoke(notebookEnvironmentStatusCommand, invocation([] as const))
     ).resolves.toMatchObject({ pythonReady: true })
-    await router.dispatcher.invoke(notebookEnvironmentProvisionCommand, invocation(['r'] as const))
+    await router.dispatcher.invoke(
+      notebookEnvironmentProvisionCommand,
+      invocation(['r', 'provision-operation'] as const)
+    )
     await router.dispatcher.invoke(
       notebookEnvironmentRepairCommand,
-      invocation(['python'] as const)
+      invocation(['python', 'repair-operation'] as const)
     )
     await router.dispatcher.invoke(notebookEnvironmentCancelCommand, invocation([] as const))
 
-    expect(provision).toHaveBeenCalledWith('r')
-    expect(repair).toHaveBeenCalledWith('python')
+    expect(provision).toHaveBeenCalledWith('r', 'provision-operation')
+    expect(repair).toHaveBeenCalledWith('python', 'repair-operation')
     expect(cancel).toHaveBeenCalledWith(undefined)
     expect(lifecycle.startup).not.toHaveBeenCalled()
     expect(router.dispatcher.commandNames()).toEqual(
--- a/src/main/notebook/env-ipc.test.ts
+++ b/src/main/notebook/env-ipc.test.ts
@@ -89,17 +89,29 @@ describe('registerNotebookEnvIpcHandlers', () => {
     })
     registerNotebookEnvIpcHandlers(createLifecycle(provisioner))
 
-    await registered.get('notebook-env:provision')?.({}, 'r')
-    await registered.get('notebook-env:repair')?.({}, 'python')
+    await registered.get('notebook-env:provision')?.({}, 'r', 'provision-operation')
+    await registered.get('notebook-env:repair')?.({}, 'python', 'repair-operation')
 
     expect(sent).toEqual([
       {
         channel: 'notebook-env:progress',
-        progress: { phase: 'fetch-r', message: 'Downloading R', progress: 0.4, scope: 'r' }
+        progress: {
+          phase: 'fetch-r',
+          message: 'Downloading R',
+          progress: 0.4,
+          scope: 'r',
+          operationId: 'provision-operation'
+        }
       },
       {
         channel: 'notebook-env:progress',
-        progress: { phase: 'repair', message: 'Repairing Python', progress: 0.2, scope: 'python' }
+        progress: {
+          phase: 'repair',
+          message: 'Repairing Python',
+          progress: 0.2,
+          scope: 'python',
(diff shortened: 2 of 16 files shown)


ewen-poch:
Addressed in \378afa5e\. The existing per-language explicit-run counter now determines whether a completed provision/repair call is the final outstanding call; only that final call settles the shared \preparing\ and error state and reapplies recovery status. Added a regression test covering two overlapping R provision calls and verified the shared Settings/Onboarding impact set (58 tests), renderer typecheck, and changed-file lint.

ewen-poch:
Addressed in `d7093800`. Tagged `error` progress now follows the same explicit-run guard as `done`: it updates the event payload but does not settle `preparing` while a Settings-triggered provision or repair IPC call remains active. The explicit call's finalizer publishes the failure after the last same-language call returns. Added a regression test for a tagged R error arriving before the provision promise rejects; the shared Settings/Onboarding impact set now passes 59 tests, plus renderer typecheck and changed-file lint.

ewen-poch:
Addressed in `be92c1e3`. Explicit runs now retain per-language progress ownership rather than only a count. Terminal progress closes that run's ownership window; if a newer automatic run publishes progress before the older IPC reply arrives, the older finalizer no longer settles or clears the newer state. Added a regression covering explicit `done` → automatic start → delayed explicit IPC completion. The shared Settings/Onboarding impact set passes 60 tests, plus renderer typecheck and changed-file lint.

ewen-poch:
Addressed in `6e83c087`. Renderer-generated operation IDs now flow through preload/IPC into the notebook environment lifecycle and return on progress for explicit provision and repair calls. Automatic maintenance progress omits the ID, so its terminal event cannot claim a queued explicit setup; local operation ownership also preserves newer automatic or other-window progress.

Added the requested regression where an automatic R run finishes before the queued explicit run emits progress. The affected 5 test files pass all 140 tests, Node and renderer typechecks pass, and changed-file lint passes. The full suite was also run; its remaining failures are the existing Windows sandbox ACL/symlink/cache and shell-environment failures.

ewen-poch:
Addressed in `5432b111` without reintroducing the newer-progress overwrite from the previous review. A logged provision/repair operation that fails before emitting progress now projects one redacted, language-tagged `error` event; the lifecycle adds the renderer operation ID, so the explicit run owns its terminal state and settles normally. If a newer automatic run has already published progress, its untagged ownership still remains intact.

Added regressions at both boundaries: lifecycle correlation for a pre-progress failure, and renderer settlement after an older automatic run. The affected 6 test files pass all 145 tests, Node and renderer typechecks pass, and changed-file lint passes. The full suite was rerun; remaining failures are the existing Windows sandbox ACL/symlink/cache and shell-environment failures.

ewen-poch:
Addressed in `40aac480`. The per-language progress cursor now records whether the latest event is terminal. When the last local provision/reset call finishes after newer automatic or another-window terminal progress, it clears only the local optimistic `preparing` state while preserving that newer operation's progress and error. Newer non-terminal progress still keeps the card preparing, so the prior ownership guarantee remains intact.

Added a regression for explicit completion after newer automatic `done` progress. Also updated the three RuntimesPanel bridge assertions that caused the Unit tests and Coverage macOS failures to accept the new operation ID argument. The affected 7 test files pass all 169 tests, Node and renderer typechecks pass, and changed-file lint passes. The full local suite was rerun; remaining failures are the existing Windows sandbox ACL/symlink/cache and shell-environment failures.

ewen-poch:
Addressed in `4d18adce`. Terminal fallback now only applies when terminal progress is newer than the explicit run's starting revision. When a newer terminal event belongs to another operation, the finalizer preserves that operation's error if present; otherwise it falls back to the explicit IPC failure while retaining the newer progress.

This avoids both swallowing the explicit failure and overwriting a newer automatic failure without adding state fields or abstractions. Added regressions for newer automatic success and newer automatic failure.

Validation:
- TypeScript checks and lint pass.
- Relevant suite: 7 files, 171 tests passed.
- The full local suite was rerun; its remaining failures are the existing Windows sandbox/ACL, symlink, cache-path, and timeout failures, outside the changed files.

PR #824, opened 2026-08-06
feat(artifact-provenance): add on-demand code reconstruction

ewen-poch:
## Problem

Artifact Provenance can retain the immutable producer block and Execution Log, but it cannot always present a standalone script that reproduces the final Artifact. Generating that script during save or preview-open would add model cost and latency even when the user does not need it, and running it in the active chat would pollute conversational context.

## Proposed change

- Add an explicit **Generate script** action to the Provenance Code tab. Opening the tab performs only a cache/state lookup; the model is called only after the user clicks.
- Build a bounded, version-scoped context from the immutable Execution Log, producer run, inputs, and relevant environment metadata. Evidence content is treated as untrusted data.
- Resolve the provider/model selected at click time through the existing backend resolver and run a tool-less, non-persistent, one-shot ACP session in an isolated temporary cwd/profile.
- Reuse the existing framework and transport paths for Claude Code, OpenCode, Codex native/Responses compatibility, and Codex bridge-backed configurations rather than introducing another model SDK.
- Cache successful reconstruction by Artifact Version plus execution/context/prompt checksums, and expose lookup/generation through typed main/preload/renderer contracts.
- Label reconstruction as LLM-generated, link back to the raw Execution Log, surface bounded-evidence warnings and input files, and download Python/R/Bash/text scripts with the correct extension.
- Show an accessible centered loading state in the code content area while reconstruction is running, instead of leaving the captured producer block visible.
- Refresh an already-open Provenance panel when a newly finalized Version is selected, and keep every reconstruction toolbar state on one line with right-side ellipsis.

```mermaid
flowchart LR
  A["User clicks Generate script"] --> B["Renderer IPC"]
  B --> C["Version-scoped reconstruction service"]
  D["Immutable Execution Log + provenance"] --> C
  C --> E["Existing backend resolver"]
  E --> F["Isolated, tool-less ACP runtime"]
  F --> G["Validated script response"]
  G --> H["Checksum-bound provenance cache"]
  H --> I["Code preview + script download"]
```

## Scope and non-goals

- No model call on Artifact save, application startup, preview open, or cache lookup.
- No mutation of the immutable Execution Log; it remains the raw source of truth.
- No additional model SDK, database migration, or persistent chat session.
- No attempt to execute or validate the reconstructed script automatically.
- Uploaded inputs remain optional; when captured, their immutable Versions are linked from the Code tab.

All commands below ran after the final material state was rebased onto the current `origin/main` at push time.

- On-demand generation, bounded context, cache validation, and unavailable/error states → `npm test` → passed, including artifact reconstruction service/repository/IPC tests.
- Tool-less isolated ACP execution and backend/provider-model capture → `npm test` → passed, including reconstruction runner and backend resolver/transport compatibility tests.
- Renderer generation flow, content-area loading state, Execution Log navigation, correct script downloads, newly finalized Version refresh, and single-line status layout → `npm test` → passed, including `ArtifactProvenancePanel.render.test.tsx` and file-save tests.
- Main/preload/renderer/shared contract compatibility → `npm run typecheck` → passed.
- Generated renderer API inventory → `npm run check:web-api-map` → passed.
- Source lint → `npm run lint` → passed with 0 errors and 17 pre-existing warnings outside this change.

Independent review of this final evidence map is still pending; this PR remains Draft and is not marked fully verified until that review confirms coverage.

## Consumer and platform coverage

Main, preload, renderer, shared contracts, provenance persistence, settings/backend resolution, and framework adapters are included because the feature crosses those consumers. The complete portable Vitest suite was used instead of a narrower directory-only set.

Packaged Electron E2E and OS-specific provider lanes were not run locally: deterministic reconstruction requires real configured external providers/CLIs, and this change does not add OS-specific UI behavior. PR Gate remains authoritative for any selected consumer or platform lanes.

## Uncovered risks

- Real-provider behavior, latency, and output quality across every Claude Code/OpenCode/Codex transport combination.
- OS-specific temporary-profile cleanup and credential discovery under abrupt process termination; stale profiles are swept after 24 hours.
- Reconstructed scripts are LLM output and may require user review, especially when the retained Execution Log was bounded.

## Review focus

- Whether tool denial, non-persistence, profile isolation, and prompt-injection boundaries are sufficient across all backend paths.
- Whether cache identity invalidates correctly for Artifact Version, Execution Log, context, prompt, framework, and model changes.
- Whether Codex native, Responses compatibility, and bridge transports remain semantically aligned.
- Whether the Provenance UI clearly distinguishes reconstruction from immutable evidence and communicates in-flight generation.


ewen-poch:
Addressed the review feedback in `872e367`:

- Provider-default reconstructions now prefer runtime-reported model metadata and fall back to the stable `provider-default` marker, so successful Claude runs are returned and cached instead of being rejected. Added a regression test for the provider-default path.
- Updated the renderer contract inventory assertions for the two new reconstruction request methods, fixing the deterministic Static checks and Unit tests failures.
- Unified the Code and Execution Log action bars through one shared layout class, and retained animated loading feedback in both the Generate button and content area.

Validation: focused tests pass (109/109), typecheck passes, lint reports 0 errors, the web API map check passes, and the unrelated `provisioner-runtime` timeout tests pass in isolation (19/19).

ewen-poch:
Addressed the latest review feedback:\n\n- Prompt boundary hardening: artifact evidence now escapes angle brackets as JSON Unicode escapes before insertion into the evidence envelope, so scripts, outputs, filenames, and metadata cannot terminate the delimiter while remaining semantically valid JSON.\n- Prompt cache invalidation: bumped the reconstruction prompt version so any earlier cache entry created with the weaker boundary is regenerated on demand.\n- Claude prompt ownership: application guidance is no longer copied into the resolved Claude backend. Session presentation remains the single owner, while connector instructions are preserved and the isolated reconstruction backend still installs its one-shot reconstruction prompt.\n- Regression coverage: added a malicious closing-tag evidence case and a Claude application-guidance ownership test.\n\nValidation:\n- Focused tests: 46 passed.\n- Typecheck and generated Web API map check passed.\n- Full suite on the rebased branch: 11,983 passed and 184 skipped; one unchanged test from current main fails locally with React act timing, so the standard PR runner is the deciding signal.

ewen-poch:
Addressed the latest review and CI findings:

- Data-root migration lifecycle: reconstruction generation now holds the existing `withDataRootWrite` lease across evidence reads, the isolated model run, and the cache commit. Both Electron IPC and application-command dispatch use this shared handler owner, so migration drains active generation and rejects new generation while a move is pending.
- Regression coverage: added a test proving migration drain remains blocked until an in-flight reconstruction finishes.
- Current-main CI contract: updated the provenance Messages integration test to locate the message and tool-group rows independently. This preserves the horizontal `px-0 md:px-0` contract while accepting the intentional compact vertical spacing now used by `WorkspaceActivityGroup` on main.

Validation:

- Affected tests: 33 passed.
- Full suite: 11,985 passed and 184 skipped across 814 passing files.
- Typecheck and generated Web API map check passed.
- Lint completed with 0 errors and the same 17 pre-existing warnings.

ewen-poch:
- The macOS coverage job was cancelled twice at the exact 8-minute job timeout; neither run reported a test failure or coverage-threshold failure.
- A prior successful run on this branch completed the same job in about 5 minutes, showing the suite is sensitive to runner variance.
- Increased only the `coverage_macos` timeout from 8 to 12 minutes, matching the unit-test job budget; test scope and coverage requirements are unchanged.
- Validated the workflow contract with `npm test -- scripts/ci/pr-gate-workflow.test.ts` (14 tests passed).

ewen-poch:
- Reverted the `coverage_macos` timeout change because `.github/workflows/pr-gate.yml` is a protected control-plane file and CI Integrity requires an explicit maintainer ruleset bypass for any modification.
- The two macOS coverage cancellations remain infrastructure timeouts: both reached the exact 8-minute job limit without a test or coverage-threshold failure. The product diff does not attempt to bypass that policy.
- The fifth and final Codex review identified a separate P1 isolation gap: a normal ACP session can still expose auto-approved application MCP tooling during reconstruction.
- No ACP architecture change is being made without design confirmation. The proposed direction is an explicit one-shot no-capability/no-MCP session mode while retaining the selected provider/model and cancellation checks.

PR #835, opened 2026-08-06
feat(permissions): allow live permission profile changes

### LINKED ISSUE #828: [Feature]: Allow permission settings to be changed while a task is running
### Preflight checklist

Image

Users may need to adjust the agent’s permissions after a task has already started.

Currently, permission settings may be unavailable or ineffective during task execution. Users may need to stop the task and start it again simply to grant additional permissions or reduce the agent’s access.

This interrupts long-running workflows and may cause completed progress to be lost.

### Proposed solution

Allow users to view and modify the permission mode while a task is running.

Users should be able to:

- Switch between available permission modes
- Grant or revoke specific permissions for the current session
- Review permissions already granted during the session
- Clear previously saved session permissions

Permission changes should apply to subsequent tool calls and actions without restarting the task.

If the agent is currently waiting for approval, changing the permission mode should immediately update or resolve the pending request where appropriate. The interface should clearly indicate that permission changes do not affect actions that have already been executed.

### Alternatives considered

_No response_

### Additional context

_No response_

ewen-poch:
## Problem

Permission modes were read-only while an Agent turn was running or waiting for approval. Users had to stop the turn before changing the approval posture, even though ACP sessions already support live mode updates.

Closes #828.

## Proposed change

- keep the permission mode control editable while a Session is running or waiting for permission
- apply the Agent's live ACP mode before committing the Session profile
- re-evaluate pending provider tool requests in `AcpPermissionBroker` with the existing automatic-permission policy
  - Auto releases only requests accepted by the existing conservative low-risk policy
  - Full releases eligible provider requests through the provider's one-shot allow option
  - Ask, high-risk requests, and app-owned approvals continue to require explicit user action
- clarify that permission changes affect future actions and do not change completed actions

## Scope and non-goals

- no architecture layer, database schema, data-model, or relationship changes
- no new proactive rule editor; existing session/project/global grant flows remain unchanged
- app-owned privileged confirmations, including Specialist actions, are never auto-approved

Checks ran after the final rebase onto the latest `origin/main`.

- pending provider request re-evaluation and app-approval isolation -> `vitest run src/main/acp/permission-broker.test.ts -t "re-evaluates pending tool requests"` -> passed
- live `session/set_mode` while a prompt waits for permission -> `vitest run src/main/acp/runtime.test.ts -t "changes permission profile while a prompt is waiting"` -> passed
- running/waiting UI editability and controls interactions -> focused Workspace/Composer tests -> 41 passed
- main-process contracts -> `npm run typecheck:node` -> passed
- renderer contracts -> `npm run typecheck:web` -> passed
- lint -> `npm run lint` -> 0 errors; 17 existing warnings outside this diff
- complete portable suite -> `npm test` (`vitest run`) -> 809 files passed, 14 skipped; 11,901 tests passed, 184 skipped

Consumers and platform lanes: main ACP and renderer consumers are covered by focused tests plus both process typechecks. No persistence, migration, packaging, or platform-specific code changed, so no additional platform lane was included.

Uncovered risk: the runtime integration test uses the repository's in-memory ACP Agent. Live Claude, Codex, and OpenCode adapter processes are not launched locally in this suite; review should confirm their documented live mode behavior remains compatible.

## Review focus

- broker re-evaluation must remain session-scoped and reuse only the projected provider one-shot option
- app-owned approvals must remain human-only
- the runtime must commit the new profile before releasing a pending request so subsequent tool calls observe it

### DIFF
--- a/src/main/acp/permission-broker-registry.test.ts
+++ b/src/main/acp/permission-broker-registry.test.ts
@@ -229,6 +229,38 @@ describe('ACP permission broker with durable grants', () => {
     await expect(response).resolves.toEqual({ outcome: { outcome: 'cancelled' } })
   })
 
+  it('re-evaluates a delayed grant lookup against a provider mode downgrade', async () => {
+    const { registry, finishResolve } = controlledEmptyGrantRegistry()
+    const emitted: Parameters<ConstructorParameters<typeof AcpPermissionBroker>[0]>[0][] = []
+    const broker = new AcpPermissionBroker((request) => emitted.push(request), undefined, registry)
+    const response = broker.requestPermission(shellRequest('session-1'), {
+      profile: 'ask',
+      projectId: 'project-1'
+    })
+
+    broker.setLivePermissionProfile('session-1', {
+      selectedProfile: 'full',
+      effectiveProfile: 'full',
+      availableModeIds: ['default', 'bypassPermissions'],
+      currentModeId: 'bypassPermissions',
+      fullAccessAvailable: true
+    })
+    broker.setLivePermissionProfile('session-1', {
+      selectedProfile: 'ask',
+      effectiveProfile: 'ask',
+      availableModeIds: ['default', 'bypassPermissions'],
+      currentModeId: 'default',
+      fullAccessAvailable: true
+    })
+    finishResolve()
+    await new Promise<void>((resolve) => setImmediate(resolve))
+
+    expect(emitted).toHaveLength(1)
+    expect(broker.getPendingRequests()).toHaveLength(1)
+    broker.cancelAllPending()
+    await expect(response).resolves.toEqual({ outcome: { outcome: 'cancelled' } })
+  })
+
   it('fails closed and reports when a remembered approval cannot be persisted', async () => {
     const registry = {
       resolve: vi.fn().mockResolvedValue(undefined),
--- a/src/main/acp/permission-broker-registry.test.ts
+++ b/src/main/acp/permission-broker-registry.test.ts
@@ -108,6 +108,38 @@ const mcpRequest = (
   ]
 })
 
+const controlledEmptyGrantRegistry = (): {
+  registry: PermissionGrantRegistry
+  finishResolve: () => void
+} => {
+  let finish: (() => void) | undefined
+  const registry = {
+    resolve: vi.fn(
+      () =>
+        new Promise<undefined>((resolve) => {
+          finish = () => resolve(undefined)
+        })
+    ),
+    remember: vi.fn(),
+    list: vi.fn().mockResolvedValue([]),
+    listCached: vi.fn().mockReturnValue([]),
+    revoke: vi.fn(),
+    extendUndo: vi.fn(),
+    restore: vi.fn(),
+    prune: vi.fn(),
+    finalizeOwnerDeletion: vi.fn(),
+    subscribe: vi.fn().mockReturnValue(() => undefined)
+  } satisfies PermissionGrantRegistry
+
+  return {
+    registry,
+    finishResolve: () => {
+      if (!finish) throw new Error('Grant lookup has not started.')
+      finish()
+    }
+  }
+}
+
 describe('ACP permission broker with durable grants', () => {
   it('cancels a request while its durable grant lookup is still pending', async () => {
     let finishResolve: (() => void) | undefined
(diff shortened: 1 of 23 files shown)


ewen-poch:
Addressed the P1 concurrency finding in 2aafe18.

- Permission-profile changes are serialized per session, with queued requests coalesced using latest-wins revisions.
- A superseded request re-checks its revision after the provider mode call and cannot commit the stale profile or release pending permissions.
- Live changes force the selected provider mode so a stale in-flight mode application is explicitly reverted by the latest request.
- Added a regression test covering an overlapping Full → Ask change while a permission request is pending.

Validation: 456 ACP runtime/configurator tests passed; Node typecheck and affected-file lint passed. The full suite passed 11,901 tests with one unrelated completion-gate timeout, and that timed-out test passed on immediate isolated rerun.

ewen-poch:
Addressed the pending-request race in 499f703.

- The runtime now passes a per-session revision guard through the permission context into the broker.
- The broker checks that guard before evaluating and immediately before releasing each pending provider request, and stops the re-evaluation pass as soon as the profile change is superseded.
- The runtime commits the profile only after broker re-evaluation completes and the same revision is still current.
- Added a broker-layer regression test that supersedes the profile at an async response boundary and verifies the remaining request stays pending.

Validation: 809 test files passed (14 skipped), with 11,903 tests passed (184 skipped); Node typecheck and affected-file lint passed.

ewen-poch:
Addressed the durable-grant lookup race in 1714f62.

- The broker now tracks the active live profile together with its revision guard.
- When an asynchronous grant lookup returns without a match, the request is re-evaluated against the still-current live profile before it can be enqueued.
- A current Full/Auto transition can therefore release the crossing request, while a superseded transition fails closed and leaves it pending.
- Live profile state is cleared with broker/session teardown.
- Added durable-registry tests for both the current-profile and superseded-profile lookup boundaries.

Validation: 809 test files passed (14 skipped), with 11,905 tests passed (184 skipped); Node typecheck and affected-file lint passed.

ewen-poch:
Addressed both regressions in ce41ffc.

- Split the composer menu gates: permission mode remains editable during a running/waiting turn, while auto-review, specialist, and compute controls retain the original running/interaction read-only gate.
- Added UI coverage proving that only permission mode is editable under that state.
- The runtime now rejects live permission-profile changes while the Session owns a native compaction interaction, while normal prompt interactions remain eligible.
- Added runtime coverage for the compaction guard.

Validation: 809 test files passed (14 skipped), with 11,906 tests passed (184 skipped); Node and Web typechecks plus affected-file lint passed.

ewen-poch:
Addressed the capability-less live-profile race in b2e487c. All provider permission requests now pass through a shared final live-profile decision before enqueueing, covering durable-grant misses, capability-less requests, and legacy grant paths while keeping app-owned approvals manual. Added regression coverage for provider requests arriving after the pending-request snapshot. Validation: 11,907 tests passed; node/web typechecks and affected ESLint checks passed.

ewen-poch:
Addressed the stale early auto-approval path in 91b8ee4. The broker now resolves every automatic decision through the current live profile and revision: a current live profile overrides the request policy, while a superseded revision fails closed. This covers both the immediate automatic branch and the final pre-enqueue recheck. Added a Full-to-Ask regression test with a delayed request carrying stale Full policy. Validation: 11,908 tests passed; node/web typechecks and affected ESLint checks passed.

ewen-poch:
Addressed superseded profile persistence in 0dae768. Attached-session changes now persist the profile selected in the runtime snapshot rather than the originally requested value; a snapshot without a committed profile fails without writing stale state. Detached sessions still persist the requested preference for the next resume. Added regression coverage proving a superseded Full request persists the committed Ask profile. Validation: 11,909 tests passed; node/web typechecks and affected ESLint checks passed.

ewen-poch:
Addressed provider mode synchronization in 0db26d5. A current_mode_update now commits the derived profile to both the session aggregate and the broker live-profile state, so automatic decisions and visible/effective state cannot diverge. Delayed durable-grant lookups re-evaluate against the synchronized profile before enqueueing or approving. Added downgrade coverage for Full-to-Ask mode updates and delayed grant resolution. Validation: 11,910 tests passed; node/web typechecks and affected ESLint checks passed.

ewen-poch:
Addressed in 624a1bd. The runtime now registers the requested live permission profile before awaiting the provider mode change, with revision/session/connection validity checks so requests during a Full-to-Ask transition cannot inherit stale Full access. If provider configuration fails, the broker restores the committed profile. Permission controls are also disabled while send preparation/session replacement is in flight, while remaining editable during normal running and permission-wait states. Added regression coverage for in-flight downgrade enforcement, failed-mode rollback, and the renderer preparation gate. Validation: targeted tests (460 passed), Node/Web type checks, affected-file ESLint, and the full suite (11,912 passed, 184 skipped).

ewen-poch:
Addressed in 2b21504. For the current-mode race, the broker now treats a user-requested permission transition as authoritative until the runtime commits or rolls it back; provider mode projections use a dedicated entry point and cannot replace that guarded transition, while ordinary provider-driven mode changes still synchronize the committed live profile. For provider-session replacement, the replacement workflow explicitly clears the old live-profile binding before detaching the provider session, so the replacement uses its newly committed aggregate profile instead of an isCurrent callback tied to the disposed session. Added regressions for a stale bypassPermissions notification during an in-flight Full-to-Ask downgrade and for Full behavior after same-ID provider-session replacement. Validation: targeted ACP tests (540 passed), Node/Web type checks, affected-file ESLint, and the full suite (11,913 passed, 184 skipped).

ewen-poch:
Addressed in 2a0f77e. Provider current-mode projection is now transition-aware end to end: the projector first asks the broker whether the provider profile can commit, and only then updates the Session aggregate and publishes state. While a user profile transition owns the permission posture, the broker rejects that provider projection, so stale mode notifications cannot mutate either broker decisions or the runtime snapshot. Successful transitions remain committed once by the runtime; failed transitions leave the previously committed aggregate untouched. Added projector regression coverage for rejected provider-mode projection. Validation: targeted ACP tests (541 passed), Node/Web type checks, affected-file ESLint, and the full suite (11,914 passed, 184 skipped).

ewen-poch:
Addressed in 478f749. The attached-session resume path now invalidates the previous broker live-profile override immediately after committing the resumed permission profile to the Session aggregate. Subsequent permission requests therefore resolve from the resumed aggregate instead of a stale pre-resume Full override. Added a Full-to-Ask attached-resume regression that verifies the next command prompts rather than auto-approves, plus a resumer ownership assertion. Validation: targeted runtime/resumer tests (461 passed), Node/Web type checks, affected-file ESLint, and the full suite (11,915 passed, 184 skipped).

Issue #853, opened 2026-08-06
[Bug]: Web UI permission profile switch causes page crash


justemu:
## Bug Description

When accessing Open Science via the web gateway (LAN/phone browser), changing the Permission Profile in the composer UI causes the web page to crash/disconnect. The session becomes unresponsive and must be refreshed.

This bug does **not** occur in the native Electron desktop application — only when accessing via web browser.

## Steps to Reproduce

1. Start Open Science with `open-science start --no-open`
2. Access the web UI via `open-science url` (or LAN URL)
3. Create or open a session
4. In the composer, click the Permission Profile selector
5. Switch from `ask` to `auto` or `full`
6. **Result:** Web page crashes / becomes unresponsive

## Expected Behavior

Permission profile should be changeable via the web UI without crashing, same as in the native desktop app.

## Impact

This bug makes the web interface unusable for users who need to change the approval mode. Since the default permission profile is hardcoded as `ask` (see #852), and there is no settings.json option to change the default, web-only users are forced to:

1. Use the native desktop app (not always possible on remote servers)
2. Accept the `ask` default and manually approve every tool invocation
3. Use the CLI with `--approval-profile full` (requires technical knowledge)

## Environment

- Open Science version: 0.11.0
- Access method: Web gateway (token-authenticated)
- Browser: Chrome/Firefox
- Backend: OpenCode agent framework

## Suggested Fix

1. Fix the web UI permission profile switching (likely a WebSocket/state sync issue)
2. As a workaround, add the `defaultPermissionProfile` setting (see #852) so web users can set their preferred profile without needing to switch it in the UI

justemu:
## Console Error Logs

Reproduced on Open Science 0.11.0 via web gateway access (Edge browser). When switching Permission Profile in the composer UI:

```
mermaid-GHXKKRXX-Bm_n_y0X.js:995 Uncaught TypeError: Cannot read properties of undefined (reading 'onCatalogChanged')
    at mermaid-GHXKKRXX-Bm_n_y0X.js:995:62545
    at bb (mermaid-GHXKKRXX-Bm_n_y0X.js:9:94050)
    at yz (mermaid-GHXKKRXX-Bm_n_y0X.js:9:109468)
    at yc (mermaid-GHXKKRXX-Bm_n_y0X.js:9:109352)
    ... [repeated ~150 stack frames]

[...]

PR #856, opened 2026-08-06
feat(settings): add configurable default permission profile

### LINKED ISSUE #852: [Feature]: Allow configurable default permission profile
## Problem

The default permission profile is hardcoded as `ask` in `src/shared/permission-profiles.ts`:

```typescript
export const DEFAULT_PERMISSION_PROFILE: PermissionProfileId = 'ask'
```

This means every new session starts with the most restrictive approval mode. For team environments where multiple users share an Open Science instance, this creates significant friction:

- New users get stuck on permission dialogs
- Research tasks require constant manual intervention
- The approval profile must be manually changed for every new session

## Proposed solution

Add a configurable default permission profile in `settings.json`:

```json
{
  "defaultPermissionProfile": "auto"
}
```

Valid values: `ask`, `auto`, `full`.

## Implementation

Add `defaultPermissionProfile` to `SettingsSnapshot` and a helper function in `permission-profiles.ts`.

## Use case

Team lead deploying Open Science for a research team. Wants new users to start with `auto` approval so low-risk operations proceed without interruption.

justemu:
## Summary

Add a `defaultPermissionProfile` field to `settings.json` that allows users to configure the initial permission profile for new sessions. Defaults to `ask` (current behavior) when not set.

Closes #852

### `src/shared/permission-profiles.ts`
- Add `getDefaultPermissionProfile(settings)` helper function
- Falls back to `DEFAULT_PERMISSION_PROFILE` (`"ask"`) when not configured or invalid

### `src/shared/settings.ts`
- Add `defaultPermissionProfile?: PermissionProfileId` to `SettingsSnapshot`
- Import `PermissionProfileId` from `./permission-profiles`

## Usage

Set in `settings.json`:

```json
{
  "defaultPermissionProfile": "auto"
}
```

Valid values: `ask`, `auto`, `full`.

- `ask` (default): Requires explicit user confirmation for all tool invocations
- `auto`: Auto-approves low-risk operations (file reads, web searches)
- `full`: Bypasses all permission checks

## Next steps

This PR adds the configuration foundation. A follow-up PR will:
1. Update session creation code to read from settings
2. Add UI in Settings > General to configure the default
3. Wire up the settings store to persist the preference

## Testing

- [ ] New helper function correctly reads from settings
- [ ] Falls back to `"ask"` when settings is undefined
- [ ] Falls back to `"ask"` when value is invalid
- [ ] Type-safe: only accepts valid `PermissionProfileId` values


ewen-poch:
Thanks for working on this. A configurable default permission profile is valuable, especially for the web-gateway use case described in #853. Keeping ask as the safe fallback and validating against ask/auto/full are both good choices.

Could you please complete the end-to-end wiring in this PR? In its current form, the setting does not take effect:

- SettingsSnapshot is the renderer-facing view, but the persisted StoredSettings type does not contain defaultPermissionProfile.
- sanitizeSettings rebuilds settings.json from an allowlist and does not copy this field, so a manually configured value is discarded during load.
- getSettingsView does not emit the field, and the renderer settings store does not retain it.
- WorkspacePage still initializes and resets newConversationPermissionProfile to DEFAULT_PERMISSION_PROFILE, so every new conversation continues to start with ask.
- getDefaultPermissionProfile currently has no callers. Its validation also overlaps with the existing normalizePermissionProfile helper.

A minimal complete change would be:

1. Add the optional field to StoredSettings and validate it in sanitizeSettings with isPermissionProfileId.
2. Project a normalized PermissionProfileId through the settings snapshot/store.
3. Use that value when initializing and resetting genuinely new conversations, while preserving the profile already stored on existing, resumed, or branched sessions.
4. Add tests showing that auto/full survive settings loading and are used for a new conversation, invalid or non-string values fall back to ask, and existing session profiles are unchanged.

Please also include the final behavior-to-command-to-result validation mapping required by CONTRIBUTING.md. Once those pieces are wired together, this will solve the issue rather than only introducing an unused configuration shape.

ewen-poch:
I completed the end-to-end wiring in da57fe43.

What changed:
- StoredSettings now preserves defaultPermissionProfile, and sanitizeSettings accepts only ask, auto, or full from hand-edited settings.json.
- The main-process preferences projection and renderer settings store expose a normalized value with ask as the safe fallback.
- WorkspacePage applies the configured value when initializing or resetting a genuinely new conversation. Existing and resumed sessions continue using their persisted profile; branch behavior was left unchanged.
- The new helper now reuses normalizePermissionProfile instead of duplicating its validation.
- Focused coverage was added for valid/invalid persisted values, main-to-renderer projection, settings-store caching, and new-versus-existing conversation behavior.

- settings persistence/projection and renderer behavior -> focused Vitest files -> passed.
- default permission service projection -> 2 focused tests -> passed.
- Node typecheck -> passed.
- Web typecheck -> passed.
- npm run lint -> passed with 17 pre-existing warnings and no errors.
- Full npm test -> 12,090 passed; one unrelated completion-gate timeout failed under the full parallel load. That exact test passed when rerun alone (1/1).

No Settings UI or write endpoint was added; settings.json remains the configuration surface requested by #852.

ewen-poch:
Implemented the Settings UI follow-up in 5ad070ec.

## Proposed change

- Add a New conversations section to Settings → Permissions with the three existing permission modes.
- Keep the current-session Composer control separate: this preference applies only to newly created conversations.
- Require an explicit amber confirmation before Full access becomes the persistent default, and keep a visible warning while it is selected.
- Add the complete repository → settings service → application command / IPC → preload → renderer-store write path with optimistic rollback.
- Preserve the existing Remembered permissions list, filters, revoke flow, and responsive Settings layout.

## Validation

- settings persistence, transport, IPC, preload, renderer store, Permissions UI, and renderer-contract projection → targeted Vitest set → 326 passed
- default-permission service projection and persistence → targeted SettingsService test → 3 passed
- Settings page consumer, capability composition, and write coordinator → targeted Vitest set → 63 passed
- node and web contracts → npm run typecheck → passed
- source quality → npm run lint → passed
- generated Web contract map → npm run check:web-api-map → passed

## Review focus

Please check the distinction between the default for new conversations and the per-conversation Composer override, plus the Full access confirmation copy. No global styles, navigation, existing-session behavior, or permission-grant semantics were changed.

Visual-regression snapshots do not currently cover the Permissions panel; the change reuses the existing SettingsSection, SettingsRow, Select, dialog chrome, semantic palette, and responsive breakpoints, with render tests covering the new hierarchy and confirmation flow.

ewen-poch:
Rebased this PR onto the latest main and resolved the renderer contract inventory conflict. I also fixed the startup failure caused by the new settings:set-default-permission-profile command not being included in every certified command and renderer-surface inventory.

Verified locally:
- focused command composition, renderer catalog, preload, local Web, and permission-setting tests
- Node and Web type checks
- lint with zero errors
- E2E production build

The PR is mergeable again, and CI has restarted on the rebased head.

PR #857, opened 2026-08-06
fix(web): prevent crash when accessing Electron-only APIs via web gateway

### LINKED ISSUE #853: [Bug]: Web UI permission profile switch causes page crash

justemu:
## Summary

Fix a crash that occurs when accessing Open Science via the web gateway and switching the Permission Profile or navigating to Settings > Specialists.

Closes #853

## Root cause

The web gateway build does not install Electron-only event contracts. When the renderer code calls `window.api.specialist.onCatalogChanged(...)`, the `window.api.specialist` object is undefined, causing an unhandled `TypeError: Cannot read properties of undefined (reading 'onCatalogChanged')`.

The infinite recursion in the stack trace (`yz → yc → yz → yc ...`) triggers a React re-render loop that crashes the page.

### `src/renderer/web/api-installer.ts`
- Add a new branch to handle `unavailable` event contracts
- Provides a no-op stub that returns a no-op unsubscribe function `() => {}`
- Callers can safely call `subscribe(listener)` without crashing

### Defense-in-depth: optional chaining
- `src/renderer/src/pages/settings/SpecialistsPanel.tsx`
- `src/renderer/src/pages/workspace/SpecialistSubmenu.tsx`
- `src/renderer/src/pages/workspace/WorkspacePage.tsx`

All `window.api.specialist.*` calls now use optional chaining (`?.`) as defense-in-depth.

## Testing

- [ ] Web gateway: Settings > Specialists tab loads without crash
- [ ] Web gateway: Permission Profile switch does not crash
- [ ] Native desktop: All specialist operations still work
- [ ] No-op stub returns a callable unsubscribe function

### DIFF
--- a/src/renderer/src/pages/settings/SpecialistsPanel.render.test.tsx
+++ b/src/renderer/src/pages/settings/SpecialistsPanel.render.test.tsx
@@ -121,6 +121,23 @@ afterEach(() => {
 })
 
 describe('SpecialistsPanel', () => {
+  it('renders safely when the web surface omits the specialist API', async () => {
+    const specialistApi = window.api.specialist
+    delete (window.api as { specialist?: Window['api']['specialist'] }).specialist
+    useSpecialistStore.setState({ items: [], isLoaded: false })
+
+    try {
+      await act(async () => {
+        root.render(<SpecialistsPanel view={{ kind: 'list' }} onNavigate={vi.fn()} />)
+      })
+
+      expect(document.body.textContent).toContain('No specialists yet')
+      expect(useSpecialistStore.getState().isLoaded).toBe(true)
+    } finally {
+      window.api.specialist = specialistApi
+    }
+  })
+
   // Shared preview fixture: builtin + owned selected by default, referenced skill unchecked.
   const exportPreviewFixture = (
     overrides: Partial<{
--- a/src/renderer/src/pages/settings/SpecialistsPanel.tsx
+++ b/src/renderer/src/pages/settings/SpecialistsPanel.tsx
@@ -156,7 +156,8 @@ const SpecialistsPanel = ({ view, onNavigate }: SpecialistsPanelProps): React.JS
     void load()
 
     // Subscribe to catalog-changed push events so the list stays in sync.
-    const unsub = window.api.specialist?.onCatalogChanged(() => void load())
+    if (typeof window.api?.specialist?.onCatalogChanged !== 'function') return () => {}
+    const unsub = window.api.specialist.onCatalogChanged(() => void load())
     return unsub
   }, [load])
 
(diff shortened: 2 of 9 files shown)


ewen-poch:
Thank you for tracking this down and putting together a focused fix. Identifying the missing Electron-only event contract is useful, and the no-op unsubscribe idea is a reasonable direction for event consumers.

Could you please adjust the implementation before we merge it? There are a few related paths that still leave the web gateway vulnerable:

- The issue reports both `onCatalogChanged` and `list` as missing. This PR stubs the event, but `specialist.list` remains unavailable on the web surface.
- Installing the event stub creates `window.api.specialist`. That means the namespace guard in `WorkspacePage` now passes, after which `loadSpecialists()` calls the missing `specialist.list()`. This can make the failure happen earlier rather than prevent it.
- `SpecialistSubmenu` and `SpecialistsPanel` both call `load()` unconditionally, so opening the composer menu or Settings > Specialists can still reach the same missing `list()` call.
- In `SpecialistsPanel`, optional chaining `exportContributionTemplate()` can produce `undefined`, but the next line still reads `result.saved`.

I think the key is to decide what the supported web behavior should be:

1. If Specialists are intentionally unavailable in the web gateway, please guard on the exact methods that are required, such as `typeof window.api?.specialist?.list === "function"`, and hide or disable the Specialist UI instead of using namespace existence as the capability check.
2. If Specialists are expected to work on the web, the required methods should be exposed through the web RPC surface rather than only stubbing the events.

Whichever direction you choose, could you also add regression coverage for:

- unavailable Electron events returning a callable no-op unsubscribe;
- a partial or absent `specialist` API in the Permission Profile menu;
- Settings > Specialists under the web API surface;
- ensuring no call to a missing `specialist.list()` is made.

Thanks again for the useful investigation. With these paths covered, this should become a solid fix for #853.

PR #885, opened 2026-08-07
fix(conversation): keep streamed messages in normal paint flow

### LINKED ISSUE #882: [Bug]: User reply disappears and assistant text fragments render out of order
### Summary

A user reply can disappear from the conversation timeline, while fragments of the following assistant message are rendered in the wrong positions. This makes the conversation appear corrupted and makes it difficult to tell which response belongs to which turn.

### Observed behavior

In the attached reproduction:

1. The user sends `找一篇疾病的生信文章`.
2. The assistant asks the user to choose a disease and an analysis type.
3. The user replies with the requested choices, but that user message bubble is no longer visible in the timeline.
4. The next assistant response is visually corrupted:
   - The intended phrase appears to be `明白：聚焦**癌症**，...`.
   - `明` / `癌症**，...` appears near the left side of the response.
   - `白：聚焦**` is detached and rendered much farther to the right and lower in the conversation.
   - The remaining paragraphs and completion metadata are therefore difficult to associate with the correct turn.

This looks like a conversation renderer/layout ordering issue rather than normal model output, but the root cause is not yet confirmed.

### Steps to reproduce

The issue may be intermittent, but the observed sequence was:

1. Open an existing project and conversation.
2. Send a new prompt.
3. Let the assistant complete a response that asks for clarification.
4. Send the requested clarification.
5. Wait for the next assistant response.
6. Observe the conversation timeline.

### Expected behavior

- Every submitted user message remains visible in chronological order.
- Assistant Markdown is rendered as one continuous message in source order.
- No text fragment is positioned outside its message container.
- Completion time, elapsed time, and usage metadata stay attached to the correct turn.
- Reloading or reopening the session preserves the same correct message order.

### Actual behavior

- The user's clarification message is missing from the visible timeline.
- A fragment of the following assistant Markdown is rendered in a detached position.
- The response reads out of order and appears mixed with neighboring turn content.

### Environment

- OS: macOS
- Model shown in the composer: `gpt-5.6-sol`
- Date visible in the session: Aug 7, 2026
- App version: not visible in the screenshot
- Conversation contains an earlier generated artifact and several completed turns

### Acceptance criteria

- [ ] A newly submitted user message always has a visible, stable message row.
- [ ] Streamed assistant text preserves token/block order after completion.
- [ ] Markdown fragments cannot escape or be repositioned outside the owning message container.
- [ ] Message virtualization or incremental rendering does not reuse stale positions from another turn.
- [ ] Reopening the conversation renders the same correct message sequence.
- [ ] Add a regression test covering a user clarification followed by a streamed Markdown response containing mixed Chinese text and bold markup.
- [ ] Verify the fix with long conversations and conversations containing generated artifacts.

### Additional diagnostic suggestions

If available, compare the persisted message records with the rendered DOM to determine whether:

- the user message is missing from storage, or only hidden by the renderer;
- the assistant text is persisted in the correct order but laid out incorrectly;
- virtualization, keyed list reuse, streaming Markdown blocks, or height measurement becomes stale after a new turn.

ewen-poch:
## Problem

A newly submitted user clarification could disappear from the visible conversation while fragments of the following streamed Markdown response were painted out of order. The persisted messages, active conversation branch projection, and transcript keys remain stable; the corruption occurs at the renderer paint boundary.

## Proposed change

- Keep mutable transcript rows in normal layout and paint flow by removing `content-visibility: auto` and its estimated intrinsic size from the shared `MessageScrollerItem`.
- Add a component regression test that prevents visibility/size containment from returning to live message rows.
- Add a transcript regression covering a user clarification followed by streamed Chinese Markdown containing bold markup.

## Scope and non-goals

- No visual design, spacing, typography, message ordering, persistence, or conversation-graph changes.
- No ACP protocol or adapter changes; `claude-code`, `opencode`, `codex-response`, and `codex-bridge` all converge on the same renderer component.
- This intentionally trades the previous off-screen row rendering optimization for stable rendering. It does not add a replacement virtualization system.

- Mutable message rows stay in normal paint flow -> `npm test -- src/renderer/src/components/ui/message-scroller.test.tsx src/renderer/src/pages/workspace/WorkspaceMessageScroller.render.test.tsx` -> **passed (40/40)**
- User clarification remains before streamed mixed Chinese/bold Markdown -> same targeted Vitest command -> **passed**
- Renderer contracts and consumers remain type-safe -> `npm run typecheck:web` -> **passed**
- Source lint -> `npm run lint` -> **passed with 0 errors** (repository-existing warnings remain)
- Final patch whitespace -> `git diff --check` -> **passed**
- Real Electron rendering of the issue sequence -> `npm run build:e2e` plus a temporary deterministic Electron scenario -> **passed**; the user row, continuous Markdown, and attached completion metadata rendered correctly

Node, preload, persistence, ACP adapter, and platform-specific lanes are excluded from the local impact set because the production change is limited to a renderer-only CSS class list and does not change an interface. The uncovered risk is performance in exceptionally long transcripts now that Chromium cannot skip off-screen message subtrees; no transcript-scale benchmark was run.

## Review focus

- Confirm correctness should take precedence over the removed `content-visibility` optimization for mutable streamed rows.
- Confirm the regression coverage captures both the shared paint boundary and the reported Chinese Markdown turn order.

Fixes #882

### DIFF
--- a/src/renderer/src/components/ui/message-scroller.test.tsx
+++ b/src/renderer/src/components/ui/message-scroller.test.tsx
@@ -1,0 +1,49 @@
+// @vitest-environment jsdom
+import { act } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, describe, expect, it } from 'vitest'
+
+import {
+  MessageScroller,
+  MessageScrollerContent,
+  MessageScrollerItem,
+  MessageScrollerProvider,
+  MessageScrollerViewport
+} from './message-scroller'
+
+let container: HTMLDivElement | undefined
+let root: Root | undefined
+
+afterEach(() => {
+  if (root) act(() => root?.unmount())
+  container?.remove()
+  root = undefined
+  container = undefined
+})
+
+describe('MessageScrollerItem', () => {
+  it('keeps mutable transcript rows in normal layout and paint flow', async () => {
+    container = document.createElement('div')
+    document.body.appendChild(container)
+    root = createRoot(container)
+
+    await act(async () => {
+      root?.render(
+        <MessageScrollerProvider>
+          <MessageScroller>
+            <MessageScrollerViewport>
+              <MessageScrollerContent>
+                <MessageScrollerItem messageId="message-1">Streaming message</MessageScrollerItem>
+              </MessageScrollerContent>
+            </MessageScrollerViewport>
+          </MessageScroller>
+        </MessageScrollerProvider>
+      )
+    })
+
+    const item = container.querySelector<HTMLElement>("[data-message-id='message-1']")
+    expect(item).not.toBeNull()
+    expect(item?.className).not.toContain('content-visibility')
+    expect(item?.className).not.toContain('contain-intrinsic-size')
+  })
+})
--- a/src/renderer/src/components/ui/message-scroller.test.tsx
+++ b/src/renderer/src/components/ui/message-scroller.test.tsx
@@ -22,7 +22,7 @@ afterEach(() => {
 })
 
 describe('MessageScrollerItem', () => {
-  it('keeps mutable transcript rows in normal layout and paint flow', async () => {
+  it('contains stable rows while keeping mutable rows in normal paint flow', async () => {
     container = document.createElement('div')
     document.body.appendChild(container)
     root = createRoot(container)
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the containment performance concern in fa9281b.

- Stable and historical rows retain `content-visibility: auto` and the intrinsic-size placeholder.
- Only messages with `status === "streaming"` opt out of containment, keeping mutable Markdown in normal paint flow.
- Regression coverage now verifies both the contained stable path and the streaming opt-out path.

Validation: targeted renderer tests (40/40), full test suite (12,168 passed; 184 skipped), web typecheck, Prettier, and lint (0 errors).

ewen-poch:
The Windows core rerun passed without code changes. The initial failure was an unrelated `micromamba-cache-acl.integration.test.ts` timeout at the 30-second boundary; the rerun completed the full Windows core lane in 2m48s, and the final PR Gate passed.

PR #886, opened 2026-08-07
fix(agent-runtime): harden Plan approval and Connector discovery

ewen-poch:
## Problem

The incident artifacts showed that all PubMed Connector calls succeeded while probes of guessed server names were rejected locally. The shared Connector guidance says not to guess, but did not expose the exact enabled `mcp-*` Skill names.

A pending Session Plan could also be approved or rejected through the Agent RPC without evidence that the current Plan Artifact had first received human feedback. Existing logs did not record the response source, which made the missing approval-card report impossible to attribute. The observed `0/13` remained explained by the absence of `update_step_status` calls.

## Proposed change

- List the exact enabled Connector Skill names in the compact baseline instructions delivered to Claude Code, OpenCode, Codex Responses, and Codex bridge sessions.
- Record a live, one-shot Agent decision authorization for the exact Plan Artifact Version and prompt interaction after routed human feedback.
- Reject Agent decisions for a pending Plan when that authorization is absent; keep direct human Approve/Dismiss commands authoritative.
- Consume or invalidate authorization on decision, Plan replacement, direct human decision, interaction release, or Session cleanup.
- Add structured, content-free audit logs for Plan generation, response source, decision, revision, and step-status updates.

## Scope and non-goals

- No persistence schema change; decision authorization remains owned by the existing in-process Session Plan interaction owner.
- No renderer change for `0/N` progress and no inferred or automatic step completion.
- Does not duplicate the Session Plan completion and persistent call-record fixes already present in v0.11.1.

- Exact Connector discovery and delivery across all four backend routes -> focused Vitest set -> passed.
- Human-feedback causality, direct human authority, replacement/release cleanup, concurrency, audit redaction, and step updates -> focused Vitest set -> passed.
- Combined focused command covering 5 files -> 94 tests passed.
- Node and web TypeScript projects -> `tsc --noEmit -p tsconfig.node.json --composite false` and `tsc --noEmit -p tsconfig.web.json --composite false` -> passed.
- Changed-file ESLint -> passed; full ESLint completed with 0 errors and 17 unrelated existing warnings.
- Full Vitest fallback was attempted on Windows: 12,063 tests passed; 34 tests in 9 unrelated files failed because this host cannot create required symlinks/ACLs, lacks Bash for workflow fixtures, or timed out in unrelated long-running concurrency cases. No Connector or Session Plan test failed.

## Review focus

- Confirm that Agent decision authority cannot outlive its exact Artifact Version or interaction sequence.
- Confirm direct human decisions remain authoritative and clear any stale Agent authorization.
- Confirm audit payloads contain identifiers and fixed-vocabulary state only, never user feedback, Plan text, step titles, or notes.
- Confirm exact Connector Skill names reach all four supported execution paths.

Independent review is still required before marking the change verified under the repository verification policy.

### DIFF
--- a/src/main/acp/runtime-plan-composition.test.ts
+++ b/src/main/acp/runtime-plan-composition.test.ts
@@ -253,6 +253,23 @@ describe('ACP Session Plan approval causality', () => {
 })
 
 describe('ACP Runtime Session Plan composition', () => {
+  it('clears exact decision authorization when prompt supersession releases first', () => {
+    const harness = createHarness()
+    if (harness.interaction.kind !== 'prompt') throw new Error('Expected a prompt interaction.')
+    const authorization = {
+      sessionId: 'session-1',
+      artifactVersionId: 'version-1',
+      interactionSequence: harness.interaction.sequence
+    }
+    harness.interactions.authorizeAgentDecision(authorization)
+    const cancel = harness.workflow.capturePromptCancellation('session-1')
+
+    harness.sessionInteractions.release(harness.interaction)
+    cancel()
+
+    expect(harness.interactions.isAgentDecisionAuthorized(authorization)).toBe(false)
+  })
+
   it('authorizes one Agent decision from restored pending Plan feedback without execution authority', async () => {
     const harness = createHarness()
     const request: AcpPromptRequest = {
--- a/src/main/acp/runtime-plan-composition.test.ts
+++ b/src/main/acp/runtime-plan-composition.test.ts
@@ -17,6 +17,7 @@ vi.mock('../logger', async (importOriginal) => {
   }
 })
 
+import type { AcpPromptRequest } from '../../shared/acp'
 import type { ActivePlanProjection, PlanResponseCommand } from '../../shared/session-plan/contract'
 import { SessionPlanInteractionOwner } from '../session-plan/session-plan-interaction-owner'
 import { composeAcpRuntimeBaseOwners } from './runtime-base-composition'
@@ -252,6 +253,38 @@ describe('ACP Session Plan approval causality', () => {
 })
 
 describe('ACP Runtime Session Plan composition', () => {
+  it('authorizes one Agent decision from restored pending Plan feedback without execution authority', async () => {
+    const harness = createHarness()
+    const request: AcpPromptRequest = {
+      sessionId: 'session-1',
+      text: 'The restored Plan looks good.',
+      planContinuation: {
+        projectId: 'project-1',
+        artifactVersionId: 'version-1',
+        expectedRevision: 1,
+        pendingAction: 'review'
+      }
+    }
+
+    const protectedPlan = await harness.workflow.prompt.preflight(request)
+    if (harness.interaction.kind !== 'prompt') throw new Error('Expected a prompt interaction.')
+    const interaction = harness.interaction
+    const admitted = await harness.workflow.prompt.admit(request, interaction, protectedPlan)
+
+    expect(admitted.protectedPending).toMatchObject({ approval: 'pending' })
+    expect(
+      harness.interactions.isAgentDecisionAuthorized({
+        sessionId: 'session-1',
+        artifactVersionId: 'version-1',
+        interactionSequence: interaction.sequence
+      })
+    ).toBe(true)
+    expect(harness.interactions.executionBindingFor('session-1')).toBeUndefined()
+
+    harness.workflow.prompt.beforeRelease('session-1', interaction)
+    harness.sessionInteractions.release(interaction)
+  })
+
   it('builds a fresh frozen workflow without publishing or requiring Plan capability', async () => {
     const options = { appVersion: 'test', defaultCwd: '/workspace' }
     const create = (): ReturnType<typeof composeAcpRuntimePlanWorkflow> => {
--- a/src/main/acp/runtime-plan-composition.test.ts
+++ b/src/main/acp/runtime-plan-composition.test.ts
@@ -1,13 +1,255 @@
 import { readFileSync } from 'node:fs'
 import { resolve } from 'node:path'
 
-import { describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
+const loggerSpies = vi.hoisted(() => ({ info: vi.fn(), error: vi.fn() }))
+vi.mock('../logger', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../logger')>()
+  return {
+    ...actual,
+    createLogger: () => ({
+      debug: vi.fn(),
+      info: loggerSpies.info,
+      warn: vi.fn(),
+      error: loggerSpies.error
+    })
+  }
+})
+
+import type { ActivePlanProjection, PlanResponseCommand } from '../../shared/session-plan/contract'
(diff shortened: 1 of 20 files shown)


ewen-poch:
Addressed the authorization-lifecycle race in 523ce2b8: after service.respond resolves, the runtime now requires successful consumption of the exact one-shot authorization and revalidates the current prompt interaction sequence before binding execution or resolving approval. Revoked or superseded decisions are rejected with interaction-mismatch, and a regression test covers cancellation/supersession while the response is pending.

Validated with the focused Session Plan suite (36 tests) and the Node typecheck.

ewen-poch:
Addressed the durable-state race in c447a4f. PlanService now evaluates the one-shot authorization as a synchronous commit precondition after its asynchronous reads and before invoking the runtime-context patch. A revoked or superseded authorization therefore returns interaction-mismatch without issuing a persistence write or changing the Plan approval state. If the decision wins the race, its execution binding remains pinned to the exact authorizing interaction rather than a successor.

Regression coverage verifies both the no-write revocation path and the post-commit binding boundary. Validated with 68 focused tests, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Addressed the failed-commit authorization loss in 9030395. The runtime now tracks whether the commit precondition consumed the one-shot authorization and restores it if persistence rejects while the exact authorizing prompt interaction is still current. It deliberately does not restore authorization after cancellation or supersession.

A regression test covers a failed first persistence attempt followed by a successful retry in the same interaction. Validated with 69 focused tests, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Addressed the feedback-ownership race in c77a4b9. The workflow now captures the parked approval owner and exact prompt interaction before awaiting feedback persistence, then revalidates the same approval route, interaction sequence, and prompt message afterward. A superseded or missing owner cannot authorize a successor; the stale parked approval is explicitly rejected and cleaned up.

A regression test covers prompt replacement while feedback persistence is pending and verifies that no successor authorization remains. Validated with 70 focused tests, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Addressed the idempotent approval binding race in 80b8a6a. The Agent path now binds its authorizing interaction only when it actually commits the Plan decision (changed: true). If a concurrent human decision already committed the same approval, the runtime preserves the current/successor execution binding and clears the stale one-shot Agent authorization.

A regression test covers the human-wins/idempotent-Agent path and verifies that the successor binding is not overwritten. Validated with 71 focused tests, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Addressed stale feedback publication in b33e515. Feedback ownership is now revalidated before emitting the renderer user-message event, so canceled, superseded, or mismatched Plan feedback is rejected and cleaned up without creating a ghost message. The valid feedback path is unchanged.

The supersession regression test now verifies that no stale message event is published. Validated with 71 focused tests, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Addressed both integration issues in 21964e3.

Restored pending-Plan review turns now receive a one-shot decision authorization scoped to the exact protected Plan version and newly activated human-feedback interaction. This permits the Agent to interpret an explicit restored-card response and decide the Plan, while still granting no execution authority until approval; ordinary unrelated messages remain unauthorized.

Connector guidance now labels the advertised catalog as globally enabled and explicitly makes Allowed Specialist Skills for this session authoritative. Restricted specialists are instructed not to load or call any mcp-* skill absent from their effective whitelist, matching the existing framework whitelist and per-call capability gate without widening the backend/session contract.

Validated with 131 focused tests, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Addressed stale durable feedback in 60dc203. Feedback ownership is now enforced as a synchronous commit precondition inside the serialized session-persistence queue, after the latest session load and before saveSession. If cancellation or supersession wins, the append aborts without creating a durable message; if feedback wins, the message is committed and may be published. This avoids message-graph rollback and does not change the persisted schema.

Regression coverage verifies that a failed ownership precondition never calls saveSession, leaves durable messages empty, emits no renderer message event, and settles the stale approval. Validated with 167 focused tests, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Addressed the supersession cleanup in c7c108ef. capturePromptCancellation now snapshots the exact prompt interaction sequence before release and clears only that one-shot Agent decision authorization when the cancellation/supersession callback runs. This covers the lifecycle path where the interaction owner releases before normal beforeRelease finalization. A regression test releases the prompt first, invokes the captured cancellation, and verifies that the exact authorization is gone. Validated with 31 focused tests, Node TypeScript type-checking, ESLint, Prettier, and git diff --check.

ewen-poch:
Addressed the decision persistence race in b7fef514. The Agent decision authorization guard is now forwarded through PlanService and evaluated inside SessionPersistenceCoordinator's serialized queue, after loading the latest durable Session and validating its runtime-context revision, immediately before constructing and saving the patch. Cancellation or supersession while the mutation is queued therefore aborts without calling saveSession or changing Plan state. Successful commits retain the existing exact-interaction execution binding, and failed saves retain the existing rollback-safe authorization restoration path. Regression coverage verifies both service forwarding and coordinator no-write behavior. Validated with 158 focused tests, Node TypeScript type-checking, ESLint, Prettier, and git diff --check.

ewen-poch:
Addressed custom MCP discovery in 0f726b29. Backend connector guidance now consumes the same canonical provisioned mcp-* Skill names used by runtime materialization, so enabled runnable custom MCP servers appear alongside bundled connectors in the global catalog. The renderer still deduplicates and validates canonical mcp-* names, and the existing per-specialist whitelist remains authoritative. Regression coverage verifies custom Skill inclusion in both the instruction renderer and backend resolution paths. Validated with 100 focused tests, Node TypeScript type-checking, ESLint, Prettier, and git diff --check.

ewen-poch:
Addressed backend provisioning scope in 8c8a9a78. Claude now receives the canonical bundled-plus-custom connector Skill catalog because its app-owned Skill root contains both document types. Codex and OpenCode receive only enabled bundled connector Skill names, matching their current materializeAgentSkills behavior and avoiding advertisement of unavailable custom Skills. The specialist whitelist remains authoritative in every framework. Regression coverage verifies the framework-specific catalogs across Claude, Codex, and OpenCode targets. Validated with 113 focused tests, Node TypeScript type-checking, ESLint, Prettier, and git diff --check.

ewen-poch:
Addressed both findings in 6fedb1a4. Agent approval now installs its captured execution binding only when no later interaction sequence is already bound, preserving a successor binding while still allowing the approving interaction to win when it reaches the commit boundary first. A regression test covers a changed approval completing after sequence 8 is bound. Custom connector advertisement now comes from an in-memory successful-materialization projection owned by ConnectorRuntimeSettingsProjection: the projection clears before refresh and publishes canonical custom mcp-* names only after custom Skill synchronization succeeds. Failed or incomplete discovery therefore advertises no custom Skills, while bundled catalogs remain framework-specific. Validated with 119 focused tests, Node TypeScript type-checking, ESLint, Prettier, and git diff --check. An additional 347 tests passed; 7 unrelated Windows symlink tests could not create links under the current account (EPERM during setup).

PR #899, opened 2026-08-07
fix(windows-update): stabilize updater certification handoff

ewen-poch:
## Problem

The v0.11.1 Windows updater certification completed a real differential download and requested `quitAndInstall`, but then waited until timeout. The direct silent installer drill on the same runner passed.

The certification had two unstable process-boundary assumptions: Playwright attached a debugger to the packaged app while it was shutting down, and the test treated a PowerShell `VersionInfo.ProductVersion` result as the post-install authority even after NSIS had detached.

## Proposed change

- Drive the published previous release through its existing authenticated loopback Web RPC for `update:check`, `update:download`, and `update:apply`, without attaching a debugger.
- Attach a native Windows process observer to the exact detached pending NSIS executable before applying the update, then retain its real exit code.
- Preserve the real published-previous -> PR-built-current differential path and persist validated download evidence before the NSIS apply handoff.
- Wait for the updated executable to be committed, then use the launched application's authenticated bootstrap as the authoritative version check.
- Keep raw authenticated app output in memory only and redact token query values from every diagnostic path.

The temporary PR-only Windows dry-run workflow and gated NSIS trace used during diagnosis were removed after the live drill passed.

## Scope and non-goals

This changes certification orchestration only. It does not change production updater behavior, portable ZIP behavior, release feeds, publishing, installer UI, or macOS/Linux behavior.

- Real v0.11.0 -> v0.11.1 differential updater dry-run: [passed](https://github.com/aipoch/open-science/actions/runs/31189995737).
- Final PR Gate after watcher cleanup: [passed](https://github.com/aipoch/open-science/actions/runs/31194853906).
- Focused packaging/config/certification suite: 38 passed.
- ESLint, Prettier, Node TypeScript, and `git diff --check`: passed locally or in PR Gate.
- Complete local `npm test`: 12,165 passed; the remaining 30 failures are Windows host capability/timing cases (symlink/ACL privileges, Bash workflow fixtures, and notebook/settings/office timeouts), while the GitHub complete module suite passed.

## Review focus

Please verify that the headless RPC invokes the same production update commands, the process observer targets only the exact pending installer, authenticated output is redacted before reaching CI, and the final application bootstrap remains the authoritative version assertion.


ewen-poch:
Addressed both blocking review findings in `393fb5cb`:

1. **Checkout credentials** — added `persist-credentials: false` to the temporary Windows dry-run checkout step. The release-download step continues to receive only the explicitly scoped `GH_TOKEN`.
2. **Portable ZIP behavior** — `NsisUpdater.installDirectory` is now set only when the running executable's directory contains the expected `Uninstall Open Science.exe` marker. Portable ZIP builds therefore do not get pinned to their extraction directory. Added a regression test for the portable case alongside the NSIS case.

I also made the certification artifact persist its differential-download evidence before installation and forward the packaged Electron process output, so any remaining online failure includes the actual updater/installer handoff diagnostics.

Validation:
- targeted Vitest: 41 passed
- Node typecheck: passed
- ESLint: passed
- Prettier: passed
- full Vitest attempted: 12,164 passed / 29 host-capability failures (Windows symlink/ACL, Bash-backed workflow tests, and host timing); the PR Gate remains authoritative.

ewen-poch:
Follow-up from the second online dry-run: the first NSIS marker check used the display name, but electron-builder derives `UNINSTALL_FILENAME` from `PRODUCT_FILENAME`. With `win.executableName: open-science`, the actual marker is `Uninstall open-science.exe`, so the updater correctly omitted `/D` under the earlier check.

Commit `8c6aafb7` now derives the exact marker from the running executable filename. The tests assert both the installed-NSIS and portable-ZIP paths against `Uninstall open-science.exe` (41 targeted tests passing). The next dry-run should show `/D=<isolated install directory>` in the captured installer arguments.

ewen-poch:
Follow-up on the NSIS-only `installDirectory` change:

- The third live Windows dry-run showed that the product-specific uninstaller filename assumption was still wrong: `electron-updater` launched the installer with `--updated,/S,--force-run` and no `/D=...` argument.
- Commit `7a5c99e2` now confirms an installed NSIS application using the same stable convention as the existing Windows certification harness: exactly one `Uninstall *.exe` entry beside the running executable. It no longer guesses the product-specific filename.
- The portable-build test supplies a normal extracted directory with no NSIS uninstaller and verifies that `installDirectory` remains unset.
- Local validation passes: 41 targeted tests, Node type-check, ESLint, Prettier, and `git diff --check`.

The pushed commit has triggered another real Windows updater dry-run; I will verify that the live installer command contains `/D=<current installation directory>` and that the end-to-end upgrade completes before removing the temporary workflow.

ewen-poch:
Agreed: the previous-release executable owns the updater process, so an updater-side change in this PR cannot affect the `0.11.0 -> 0.11.1` transition.

Commit `ace5c769` addresses that review finding directly:

- Removed the `ElectronUpdaterStrategy.installDirectory` change and its tests. This also removes the portable-ZIP and host-dependent path-join surface entirely.
- Moved custom-directory recovery into the **target NSIS installer**, which is the new code actually executed during the previous-to-current upgrade.
- In `customInit`, the installer caches and normalizes both registered installation paths before the old uninstaller can delete them, then restores `$INSTDIR` only for `--updated` and only from the path matching the resolved install mode. Ordinary silent installs with an explicit `/D` remain authoritative.
- Kept the dry-run as the real published-previous-release to PR-built-current upgrade, so it now exercises the changed target installer rather than substituting a synthetic updater fixture.

Validation passes locally: 63 focused tests, Node type-check, ESLint, Prettier, and `git diff --check`. A new live Windows dry-run has been triggered for the corrected implementation.

ewen-poch:
Final diagnosis update after the live `ace5c769` run:

- The target-installer `$INSTDIR` override still failed identically, proving that directory rediscovery was not the cause. That ineffective NSIS change has been removed.
- Every Playwright-driven run ended the old process with `Waiting for the debugger to disconnect...`; the direct silent-installer drill on the same runner continued to pass.
- Commit `34a6e5e4` keeps the real published `0.11.0` updater and the real differential download, but invokes its production `update:check`, `update:download`, and `update:apply` commands through the app's existing authenticated loopback Web RPC. No debugger is attached to the process that must exit before NSIS can replace it.
- The harness now validates the RPC response at each phase, waits for the real process exit, forwards packaged-process output, and retains the differential-download evidence on installer failure.

Local validation passes: 63 focused tests, Node type-check, ESLint, Prettier, and `git diff --check`. The latest live Windows dry-run will be the final end-to-end verdict.

ewen-poch:
Follow-up on the live Windows certification failure:\n\n- The debugger-free RPC path confirmed that update check, differential download, quitAndInstall, and orderly application shutdown all complete successfully.\n- The remaining failure was observed only after electron-updater detached the NSIS installer; the harness previously inferred completion from the app exit and then polled the target file, which discarded the installer's actual result.\n- Commit 97d9ddf5 starts a read-only process observer before update:apply, matches the exact pending installer path, waits on the native process handle, and preserves the real NSIS exit code. This removes the detached-handoff race and will make any remaining installer failure immediately actionable.\n- The observer was exercised against a real external Windows process locally; 26 focused tests, Node type-checking, ESLint, Prettier, and git diff --check pass.\n\nThe PR-only Windows dry-run is running again on this head. I will use its native installer result as the verdict before removing the temporary workflow.

ewen-poch:
The native watcher made the remaining failure precise in run 31184771720: the detached NSIS process completed in about 21 seconds with exit code 0, but the expected custom install directory was empty. This rules out debugger attachment, process termination, installer timeout, and a non-zero NSIS failure; the installer is successfully selecting a different $INSTDIR.\n\nI also verified that v0.11.0 and v0.11.1 retain the same ppId, product identity, electron-builder 26.15.3, and deterministic NSIS GUID (65c5229-0b29-5716-a0fe-d8755e62f3ca), so this is not an installer-identity migration. Commit fe7e9407 adds a diagnostic trace gated by the PR-only OPEN_SCIENCE_INSTALLER_TRACE environment variable. It records the target installer's initial registered HKCU/HKLM locations, selected mode, $INSTDIR, post-uninstall state, and final install directory. The trace is uploaded only by the temporary dry-run and will be removed with that workflow after the live certification passes.

ewen-poch:
The latest dry-run stopped during packaging because the temporary gated NSIS trace macro passed a delimited string containing $isUpdated as one macro argument; makensis consequently parsed the delimiter as part of the variable name. Commit 1840da27 writes each trace field separately. I also compiled the hook locally with the same cached NSIS 3.0.4.1 toolchain and -WX -INPUTCHARSET UTF8, so warnings remain fatal just as they are in electron-builder. The 35 focused packaging/config/certification tests and formatting checks pass. I am following the replacement Windows dry-run now.

ewen-poch:
Run 31187813462 still stopped during packaging, before certification. The remaining issue was that electron-builder exposes `${isUpdated}` as a generated LogicLib condition, not an NSIS `$isUpdated` variable. Commit a2238bfe evaluates that condition into a runtime 0/1 register before writing the gated trace. I updated the local NSIS compile harness to reproduce electron-builder's generated `_isUpdated` condition macro; NSIS 3.0.4.1 passes with `-WX`, and all 35 focused tests pass.

ewen-poch:
Run 31188544004 reached the real differential handoff. The target NSIS process exited 0, and the gated trace recorded `updated=1`, `mode=CurrentUser`, and both final `$INSTDIR` and HKCU `InstallLocation` at the requested isolated `...\app` directory. This rules out the earlier install-directory drift hypothesis. The remaining timeout was the certification's redundant PowerShell `VersionInfo.ProductVersion` poll, which collapsed an empty probe result into the misleading `<missing>` label. Commit ea632c37 now waits only for a non-empty executable to be committed after the observed installer exits, then uses the existing authenticated application bootstrap launch as the authoritative version check. That validates the executable that actually runs instead of trusting mutable PE metadata. The 35 focused tests, ESLint, Prettier, and diff checks pass; I am following the replacement live dry-run.

ewen-poch:
The live v0.11.0 -> v0.11.1 updater certification is green in run 31189995737. It completed the differential download, observed the exact detached NSIS process exit 0, launched the updated application, verified the current version through the authenticated bootstrap, and preserved the upgrade profile sentinel. I have now removed the temporary PR-only dry-run workflow and gated NSIS trace in commit 41e3bbe9, as planned. The final PR diff contains only the certification implementation and its tests; the replacement PR Gate is running.

ewen-poch:
The two remaining actionable diagnostic findings are addressed in ad7b92d:

1. **Authenticated output:** packaged-app stdout/stderr is retained only in memory for endpoint parsing. It is no longer forwarded live, and every error/diagnostic path redacts `token` query values before the text can reach Actions logs or artifacts. A focused regression test covers multiple token parameters.
2. **Observation timing:** the validated differential observation is now written immediately after `update:download` reports ready and before the process observer and `update:apply` handoff. A nonzero or missing NSIS process therefore no longer removes the download evidence from the failure artifact.

The focused suite now passes 36 tests, ESLint/Prettier/diff checks pass, and Node typecheck passes. Per the agreed review-round cap, I will not start another review iteration; I am only following the final deterministic PR Gate for this commit.

ewen-poch:
Final validation is complete. PR Gate run 31192566833 passed on attempt 2. Attempt 1 had an unrelated Windows settings-persistence E2E failure (the dark class was not observed after relaunch); the same product code had passed immediately before this diagnostics-only commit, and the failed job passed unchanged on rerun. All deterministic lanes are now green: static checks, Windows core/E2E, macOS build/E2E, both full module shards, and the aggregate gate. No temporary workflow or NSIS trace remains in the PR.

ewen-poch:
Regarding the post-cap watcher-cleanup suggestion: I am not opening another implementation/review round for it in this PR. The observer is certification-only, not shipped product behavior; it has a one-minute process-appearance deadline, a five-minute native wait, and a 370-second parent `runProcess` hard timeout, after which the existing process-tree termination path runs. Therefore it is bounded and cannot persist beyond Actions job cleanup. Adding abort semantics to the shared smoke-process runner would widen the final change after the agreed 3-5 review-round cap. The successful path exercised by the live dry-run settles the observer normally (PID observed, exit 0). I would treat explicit early cancellation as a follow-up diagnostics optimization rather than a merge blocker.

ewen-poch:
The installer-watcher cleanup finding is addressed in 4213e0d. This supersedes my earlier post-cap deferral because the review budget was explicitly reopened.

1. `runProcess` now accepts an optional `AbortSignal`. Cancellation uses the existing process-tree terminator and does not reject until that cleanup has settled.
2. The updater certification retains an `AbortController` for the detached NSIS watcher, attaches rejection handling immediately, and cancels and awaits the watcher together with the packaged app in every failure path.
3. The normal success path is unchanged: the observer result is still awaited and its real installer exit code is preserved.
4. Regression coverage verifies signal forwarding and proves an aborted process remains pending until termination completes.

Validation:
- Focused Vitest: 28 passed
- ESLint: passed
- Prettier check: passed
- Node typecheck: passed
- `git diff --check`: passed
- Full local Vitest: 12,165 passed / 30 host-dependent failures (the existing Windows symlink/ACL, Bash-fixture, and timing-sensitive suites); the focused changed suites are green, and the new PR Gate will provide the authoritative cross-platform result.

ewen-poch:
Final validation for 4213e0d is complete:

- AI PR Review 31194852209: **mergeable**, no actionable findings
- PR Gate 31194853906: **passed** in one attempt, including static checks, Windows core/E2E, macOS build/E2E, both full module shards, and the aggregate gate
- CI Integrity 31195607705: **passed** after the final PR-description update
- Review threads: none open
- Focused packaging/config/certification suite: 38 passed

The temporary updater dry-run workflow and gated NSIS trace remain removed. The previously successful live differential updater certification remains the product-path evidence: https://github.com/aipoch/open-science/actions/runs/31189995737

PR #915, opened 2026-08-07
feat(notebook): preview kernel image outputs

ewen-poch:
## Problem

R notebook cells can produce figures through base graphics, ggplot2, grid, or lattice, but saved-only plots were only reported as files and were not shown as cell output. Conversation tool rows also received compact notebook results, so they could not present locally captured figures. Python already rendered captured figures in Notebook preview, but its text/image interaction did not match the requested output layout.

## Proposed change

- Hydrate full notebook runs locally in the renderer by the compact `runId`, without copying figure payloads into session messages.
- Share one output presentation across Notebook preview, the Session notebook dialog, and expanded conversation tool rows.
- Keep text output independently collapsible while rendering every figure in its own always-visible frame when the parent cell or row is expanded.
- Present every captured figure occurrence and every unique saved image path from the run.
- Preview saved PNG/JPEG/WebP/GIF/AVIF/SVG images, decode TIFF files, and render PDF output without cropping its first-page preview.
- Add real-R regression coverage for base graphics, `ggsave`, grid, and lattice, including a single run that saves four images through mixed graphics paths.

## Scope and non-goals

- The kernel and MCP response contracts are unchanged.
- Figure bytes remain in local notebook run state and managed preview resources. They are not added to tool content, agent context, or replay preambles.
- Captured figures and saved files are both preserved. The current run format has no reliable captured-figure-to-saved-path provenance, so the UI does not guess that two outputs are equivalent and risk hiding a valid image.
- This does not add package-specific production handling for ggplot2, grid, or lattice; saved-file observation remains package-independent.

All checks below ran after the final material edit.

- R inline and saved-output behavior -> `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npm test -- --run src/main/notebook/r-loop.integration.test.ts src/main/notebook/kernel-executor.test.ts` -> 91 passed.
- Node contracts and real-R test types -> `npm run typecheck:node` -> passed.
- Renderer hydration and output presentation types -> `npm run typecheck:web` -> passed.
- Repository lint -> `npm run lint` -> passed with 17 pre-existing warnings and no errors.
- Full portable fallback -> `npm test -- --run --maxWorkers=8` -> 841 files passed, 12,311 tests passed, 186 skipped.
- Independent Standards and Spec reviews -> no remaining findings.

Uncovered risk: `gridSVG::grid.export()` was not exercised because `gridSVG` is not installed locally. Standard grid device output is covered, and SVG files use the existing saved-image preview route. No UI E2E or screenshot-diff lane was run; the shared renderer surfaces are covered by component tests.

## Review focus

- Confirm the `runId`-based hydration remains renderer-local and cannot enter context or replay.
- Check mixed captured/saved output ordering and the decision to preserve all outputs when provenance is unavailable.
- Check TIFF/PDF managed resource cleanup and sizing inside narrow conversation rows and wider Notebook preview surfaces.


ewen-poch:
Adjusted the saved-PDF figure preview sizing after visual verification:

- Notebook PDF first pages now render at 768px instead of reusing the 220px artifact-tile resolution.
- The render width is part of the thumbnail job/cache key, so previews at different resolutions cannot reuse the wrong cached image.
- Loaded intrinsic PDF frames now wrap the displayed page rather than retaining an empty fixed-height canvas, removing the large gaps between figures.
- The displayed page remains aspect-ratio constrained and capped at 16rem high.

Validation: 34 targeted renderer tests passed, Web typecheck passed, lint completed with only the 17 pre-existing warnings, and the full suite passed (841 files / 12,313 tests; 186 skipped).

PR #936, opened 2026-08-07
fix(session-resume): continue interrupted turns without duplicate messages

ewen-poch:
## Problem

`Resume` restored the ACP provider session but did not continue the interrupted turn. Earlier retry behavior also removed and re-appended the unanswered user prompt, which could render the same user message two or more times and could repeat already-completed work.

The expected interaction is one durable user turn: keep its original message, mark it as interrupted, reconnect or replace provider context, then continue that turn without requiring the user to send “continue”.

## Proposed change

- Keep exactly one original user message and render the yellow italic `This turn was interrupted.` marker on it.
- Show a Resume loading row immediately after the click. Keep it visible through provider restoration and hidden continuation startup; after the first provider update, use the normal working indicator.
- Reconstruct an app-owned hidden continuation in Main from durable Session state. The continuation uses `suppressUserMessage`, so it cannot publish another visible user event.
- For native Resume, include the durable original request and prompt media in that hidden continuation so a disconnect before provider acceptance cannot lose the task.
- If the provider session resumes natively, send the hidden continuation into that restored context.
- If native resume is unavailable and the session fresh-adopts, create a new Runtime Segment and replay a bounded active-Branch transcript, including useful partial assistant output, before the hidden continuation.
- Keep recovery and replay authority durable until the provider produces its first update. A failed resume or continuation remains retryable without losing the prior transcript.
- If the same originating prompt is already in flight, treat Resume as attachment to that run and do not dispatch another continuation.
- Validate the recovery prompt, active Conversation Branch, framework, and Runtime Segment in Main before continuing.

## Framework spike and strategy

ACP `session/resume` restores provider context but does not start another turn, so all four paths use app-owned continuation after restoration.

| ACP path | Resume strategy | Fresh-adoption strategy |
| --- | --- | --- |
| Claude Code | Native `session/resume`, then hidden continuation | Replay bounded active transcript, then hidden continuation |
| OpenCode | Native resume restores the Session/config/MCP context, then hidden continuation | Replay bounded active transcript, then hidden continuation |
| Codex Responses | Resume the persisted provider thread, then hidden continuation | Replay bounded active transcript, then hidden continuation |
| Codex Bridge | Resume only while the Bridge continuity token is valid, then hidden continuation | Replace Bridge context, replay bounded active transcript, then hidden continuation |

The protocol boundary follows [ACP v1 session resume](https://github.com/agentclientprotocol/agent-client-protocol/blob/main/docs/protocol/v1/session-setup.mdx). Provider behavior was also checked against the installed/upstream Claude Code, OpenCode, Codex Responses, and Codex Bridge adapters.

## Scope and non-goals

- This changes the existing manual Resume path and its recovery provenance. Accepting a newer authored turn atomically abandons the older turn's Resume authority while retaining its interruption marker and conversation history.
- The only shared persistence addition is the closed optional `plan-first` turn intent needed to reconstruct an interrupted first turn after restart. Unknown values are discarded.
- History replay construction is shared between Renderer and Main so both apply the same text/media limits.
- Provider-native SubAgent or child-agent executions are not resumed. Resume continues the parent ACP turn only; child-run recovery is deferred until child session/thread identity and checkpoint state are durable.
- Resume cannot guarantee exactly-once behavior for external side effects that completed at the provider but were never recorded before interruption. The continuation explicitly avoids repeating completed work and completed tool calls when the retained transcript shows them.
- No provider hidden-reasoning content is persisted.

## User-visible behavior

1. The interrupted turn keeps its single original user bubble and interruption marker.
2. Clicking Resume shows the Resume indicator immediately.
3. After provider restoration, the app starts a hidden continuation; no duplicate user bubble is added.
4. The first provider update replaces the Resume state with the normal agent activity indicator.
5. If restoration or continuation startup fails, the interrupted state remains and Resume can be tried again.
6. Sending a newer user message instead abandons the older turn's Resume action without removing its interruption marker or history.

## Validation

- Added coverage for attached and detached sessions, stale event drain, provider-first-update gating, native resume failure, continuation failure/retry, fresh-adoption replay, removed prompts, compaction recovery, session-scoped indicators, and no duplicate user event.
- Added a four-framework continuation matrix for Claude Code, OpenCode, Codex Responses, and Codex Bridge.
- Added Conversation Graph coverage proving old activity remains on the old Runtime Segment while resumed output uses the fresh segment.
- Added coverage proving a newer authored turn clears stale recovery authority without removing the interrupted message or context replay.
- Added Workspace and Task coverage proving fresh-context replay authority survives pre-acceptance prompt rejection and is consumed only after acceptance/success.
- Main/Preload/Renderer application-command and contract inventories pass.
- `rtk npm run typecheck:node` — passed.
- `rtk npm run typecheck:web` — passed.
- `rtk npm run lint` — passed with 0 errors and 17 pre-existing warnings outside this diff.
- Focused Resume, persistence, graph, runtime, UI, and contract suites — passed.
- A full local run passed 839 test files before reaching three environment/aggregate-only failures: two timeout cases pass in isolation; the remaining Prisma test uses the root worktree's generated client, which contains `NotificationInboxItem` while this branch's schema does not. Clean PR Gate runs remain authoritative for the aggregate.

## Review focus

- Main-owned continuation reconstruction and renderer trust boundary.
- Recovery durability until the provider's first update.
- Native resume versus fresh-adoption behavior across all four ACP paths.
- Conversation Graph Runtime Segment ownership for resumed output.
- The absence of any Resume path that appends or republishes the original user message.


ewen-poch:
Addressed the review points:

- Manual Resume now waits for the renderer runtime-event drain after a successful provider resume and before clearing recovery state or unlocking the composer. The same barrier is applied when the runtime is already attached. A regression test holds the drain open and verifies that the session remains interrupted until queued events are settled; Resume still never calls `sendPrompt`.
- Removed the overwritten initial value of `resumedProviderSessionId`; it is now captured once from the attached provider session.
- Clarified the scope boundary: interrupted provider-native SubAgent/child-agent executions are not resumed. Resume restores only the parent ACP session. First-class child-run recovery is deferred until child session/thread identity and checkpointed run state can be persisted.

Validation:

- Resume impact suite: 9 files, 850 tests passed.
- Typecheck and API-map checks passed.
- Lint passed with 0 errors and 17 pre-existing warnings.
- Full portable suite: 12,356 tests passed; the only failure is the unrelated existing `NotificationInboxItem` Prisma/runtime-DDL mismatch, reproduced in isolation.

ewen-poch:
Addressed the current review findings:

- Task/headless now derives replay history from the recovery cutoff, never includes the interrupted prompt, and consumes both `resumeRecovery` and pending replay state when starting a new authored turn. Added coverage for prompt-cutoff and full-history recovery on attached task sessions.
- Replaced the scalar replay cutoff with a persisted discriminated scope: `all` for interrupted control operations such as compaction, and `before-message` for interrupted prompts. Workspace consumes the captured marker only after the next prompt is accepted; feature-preview scalar state migrates on load.
- Changed the user-facing marker to `This turn was interrupted.` to comply with the repository language standard.

Validation:

- Resume/Task impact suite: 10 files, 878 tests passed.
- Typecheck and API-map checks passed.
- Lint passed with 0 errors and 17 pre-existing warnings.
- Full portable suite: 12,356 tests passed. Aggregate-only logger and completion-gate timing failures passed in isolation; the unrelated existing `NotificationInboxItem` Prisma/runtime-DDL mismatch remains reproducible in isolation.

ewen-poch:
Addressed the overflow-recovery persistence finding:

- The replacement returned by `resetSessionContext` is now captured and its framework, backend, provider session ID, and continuity token are stored immediately under the stable application session ID.
- Identity is updated before cancellation and retry guards run, so a cancellation or restart cannot revive the discarded provider owner.
- Added a regression assertion that starts with stale provider metadata, performs overflow replacement, and verifies the new identity in the durable session projection.

Validation:

- Resume/Task impact suite: 10 files, 878 tests passed.
- Focused Workspace runtime suite: 119 tests passed.
- Typecheck and API-map checks passed.
- Lint passed with 0 errors and 17 pre-existing warnings.
- Full portable suite: 12,359 tests passed. The aggregate-only completion-gate timeout passed 22/22 in isolation; the unrelated existing `NotificationInboxItem` Prisma/runtime-DDL mismatch remains reproducible in isolation.

ewen-poch:
Addressed the stale recovery-cutoff finding:

- A missing `before-message` cutoff now falls back to replaying the currently selected Branch instead of aborting the send.
- The original marker remains durable until prompt acceptance. This keeps failed sends retryable and preserves the narrower cutoff if the user switches back before a prompt is accepted; successful sends consume it normally.
- Added a regression test that fresh-adopts an interrupted session, creates a Branch without the interrupted prompt, and verifies that the next send proceeds with only the selected Branch history.

Validation:

- Resume/Task impact suite: 10 files, 879 tests passed.
- Focused Workspace runtime suite: 120 tests passed.
- Typecheck and API-map checks passed.
- Lint passed with 0 errors and 17 pre-existing warnings.
- Full portable suite: 12,361 tests passed; the only failure is the unrelated existing `NotificationInboxItem` Prisma/runtime-DDL mismatch, which remains reproducible in isolation.

ewen-poch:
Addressed the original-prompt retention finding in `116586af`:

- Native Resume now includes the durable original request text in the app-owned hidden continuation and restores prompt-bound uploads, inline images, referenced artifacts, selected Skills, and the closed `plan-first` intent.
- The continuation remains `suppressUserMessage`, so this context cannot create another visible user message.
- Fresh adoption continues to carry the original request through the bounded active-Branch replay; its continuation prompt stays generic so the request is not injected twice.
- Added a regression test for a disconnect before provider prompt retention, including immutable upload reconstruction, plus an assertion that fresh-adoption replay does not duplicate the task in the continuation text.

Validation: focused continuation tests pass (8/8), Node and Web typechecks pass, and lint passes with 0 errors (17 pre-existing warnings).

ewen-poch:
Addressed in `c4a84c66`.

Accepting a new authored user turn now atomically clears the prior turn's `resumeRecovery`, transient interrupted-session state, and resumed-output Runtime Segment ownership. The original prompt remains in history with its interruption marker, while the new turn can still use the existing pending history replay to preserve context.

Added a regression test covering in-memory and persisted recovery state. The focused session-store suite passes (109 tests); the broader Resume suites pass (297 tests), and both TypeScript targets plus lint pass.

ewen-poch:
Addressed in `6dc65500`.

- Workspace fresh-adoption now persists its replay marker through continuation startup. A pre-acceptance failure remains retryable, and an attached retry derives `contextReset` from that marker while reusing the prepared Runtime Segment. The marker is consumed only after Main observes the provider's first update.
- Task/headless runs now clear the old `resumeRecovery` authority when accepting a newer authored turn, but retain `pendingHistoryReplay` in the prepared Session. A prompt rejection preserves it; successful prompt completion consumes it before the final Session projection is saved.

Added regressions for both rejection paths and for durable Workspace persistence. The focused Resume/Task/UI suites pass (325 tests), both TypeScript targets pass, and lint passes with 0 errors (17 pre-existing warnings).

ewen-poch:
Addressed the resumed-turn notification finding in `2e479076`.

- Interrupted-turn continuation now registers the durable original user prompt with `TaskNotificationService` before provider startup, restoring completion notifications and unread-task tracking after restart.
- A rejected continuation startup removes exactly its tracking token; attaching to an already-running recovered prompt does not create another track.
- Resume lifecycle actions were also moved into the existing run-projection owner to satisfy the current `main` Session Store architecture gate. The public store API and user-visible behavior are unchanged.

Validation: 320 focused Resume/notification/architecture tests passed; Node and Web typechecks passed; lint passed with 0 errors. The full suite passed 12,463 tests; four aggregate timeout cases passed individually, and the remaining local failure is the known shared generated-Prisma-client mismatch.

PR #944, opened 2026-08-07
ci(pipelines): optimize nightly and release workflows

ewen-poch:
## Problem

Nightly rebuilt and republished on every `main` push, so frequent pushes cancelled expensive native
builds and wasted hosted runners. The reusable release build also serialized verification before the
platform matrix, Windows full coverage had two long serial shards, and the advisory historical
Windows upgrade drill kept stable Release runs open while waiting for a Windows runner. Transient CDN
403 responses could also fail an otherwise healthy package build immediately.

## Proposed change

- Batch Nightly at minute 17 of every hour and skip a batch when the rolling `nightly` tag already
  points at the current `main` commit.
- Prepare Nightly certification metadata and checksums in the read-only build workflow. Move real
  publication into a serialized `workflow_run` workflow with `cancel-in-progress: false`; the
  privileged publisher never checks out or executes triggering-run code, while manual Nightly runs
  stop after producing the no-side-effect `nightly-ready` artifact.
- Run reusable-workflow verification beside the platform matrix while retaining fail-closed publish
  dependencies at every caller.
- Split Windows Full Test from two to three shards while retaining `maxWorkers=1`.
- Retry transient runtime-bundle CDN failures three times with bounded 1s/2s backoff.
- Dispatch the historical Windows upgrade drill after stable assets publish, and run it as an
  independent advisory workflow against those published assets.

## Scope and non-goals

- Stable Release asset certification, macOS notarization, and publish dependencies remain blocking.
- The advisory Windows upgrade drill does not become a stable-release gate.
- Runner images, application code, signing policy, package contents, and release artifact retention
  are unchanged.
- Manual Nightly runs validate publication preparation but cannot overwrite the rolling release.

Final Test Impact Set mapping (all listed focused checks ran after the final material edit):

- Workflow topology, trigger guards, concurrency, artifact handoff, action pinning, and advisory
  isolation -> `vitest run scripts/ci scripts/verify-runtime-bundle.test.ts scripts/windows-release-workflows.test.ts`
  -> 14 files, 176 tests passed after rebasing onto current `main`.
- TypeScript consumers -> `npm run typecheck` -> passed.
- Linted workflow tests/runtime script -> `npm run lint` -> passed with 17 pre-existing warnings and
  no errors.
- Formatting -> Prettier check for all changed files -> passed.
- Full Nightly build plus read-only publish preparation -> Actions run 31223910062 -> passed; all
  four platform builds, Verify, evidence aggregation, checksums, and the `nightly-ready` artifact
  succeeded without changing a Release.
- Privileged workflow security boundary -> CodeQL check 93014242278 -> passed with no new alerts and
  zero annotations after removing the `workflow_run` source checkout.
- Complete Windows portable suite -> Actions run 31221519591 -> all three serial shards passed in
  10.1-11.4 minutes.
- Local `npm test` exercised 12,571 tests: 12,266 passed before the three intentionally stale workflow
  assertions were updated. The remaining local failures require unavailable Windows symlink/ACL or
  subprocess capabilities, or were full-suite timing failures. Both target-runner suites above passed.

Uncovered pre-merge risks: GitHub only enables the new `workflow_run` and `repository_dispatch`
entrypoints from the default branch, so the real rolling Release mutation and post-release advisory
dispatch cannot be exercised before merge without side effects. Their event and side-effect guards
are covered by contract tests; the manual Nightly path exercises the complete read-only preparation
and the privileged publisher consumes artifacts without executing them.

## Review focus

- Confirm the `Nightly` completion guard admits only successful scheduled `main` runs and that the
  tag comparison safely suppresses empty hourly batches and duplicate publication.
- Confirm every real publish operation is excluded from the manual dry-run path.
- Confirm stable Release dispatch happens only after asset publication and remains advisory.
- Confirm the reusable build still reports verification failure to callers even though native jobs
  no longer depend directly on `verify`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,9 +31,9 @@ on:
         type: boolean
         default: false
 jobs:
-  # Gate every packaged build on the quality suite so nightly (push to main) and release (tag) can
-  # never publish installers that were never tested. Runs once on Apple Silicon to match the
-  # release/verification environment (see pr-check.yml for why macOS). The build matrix `needs` this.
+  # Run the quality suite once on Apple Silicon beside the native package matrix. The reusable
+  # workflow still fails when this job fails, so every caller's publish boundary remains fail closed
+  # without putting the full verification duration on the matrix's critical path.
   verify:
     name: Verify (lint + typecheck + test + package)
     # Skipped by the notarization dry-run (skip_verify), which only needs a signed mac build.
@@ -100,10 +100,10 @@ jobs:
 
   build:
     name: Build ${{ matrix.name }}
-    needs: [verify, setup]
-    # Run once the matrix is resolved. Tolerate a skipped verify (skip_verify dry-run) but still
-    # block on a real verify failure — matching the fail-closed behavior when verify runs.
-    if: ${{ always() && needs.setup.result == 'success' && (needs.verify.result == 'success' || needs.verify.result == 'skipped') }}
+    needs: setup
+    # Start as soon as the matrix is resolved. The sibling verify job remains part of this reusable
+    # workflow's result, so release/nightly callers cannot reach publish if verification fails.
+    if: ${{ needs.setup.result == 'success' }}
     runs-on: ${{ matrix.os }}
     strategy:
       # Don't let one platform's failure cancel the others.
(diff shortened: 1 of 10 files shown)


ewen-poch:
Addressed the stale-publication ordering issue. After acquiring the workflow-level publication lock, the publisher now compares the candidate SHA with the currently published nightly SHA. Only an ahead candidate may publish; identical or behind candidates are skipped, while divergent history and unexpected API failures fail closed. Contract coverage was added, and the complete CI script test set passes (176/176). Implemented in ab5fd54.

PR #952, opened 2026-08-08
fix(notebook): separate captured output from saved artifacts

ewen-poch:
## Problem

Notebook cell output mixed two different concepts:

- immutable figures captured by the Python/R kernel and persisted in `run.outputs`;
- mutable files recorded in `workingFiles` or later attached as Artifacts.

When a run both captured a figure and saved TIFF/PDF/image output, the Notebook and message tool block could render both. Reusing the saved path in a later run also made a historical cell show the latest file bytes after an app restart. Jupyter export had the same semantic overlap because it appended Artifact files to the captured cell outputs.

This was a read/export-time presentation error. Historical run bytes and Artifact Versions were not rewritten.

## Change

- Treat kernel-captured `run.outputs` as the sole authority for Notebook and message cell figures.
- Stop resolving `workingFiles` as Notebook output, including saved-only PNG/TIFF/PDF files.
- Export `.ipynb` cell outputs only from `run.outputs`; Artifacts remain independent files and are no longer auto-inlined.
- Keep every captured figure occurrence, including multiple valid pages from one run.
- Capture R plots drawn to built-in file devices by replaying R's native display list into the kernel PNG device when the file device closes.
- Cover base graphics, ggplot2, grid, and lattice file-device output while keeping the saved file independent from the captured PNG.
- Remove the now-unused Artifact path-resolution and MIME-inlining export path.

## Behavior

- `plot()` / an open Matplotlib figure: captured as PNG, persisted as base64 with the run, and shown inline.
- `savefig()` / `tiff()` / `pdf()`: produces a working file for `write_artifact_file`; the saved bytes are not inferred as cell output.
- A figure closed before Python end-of-cell capture may have no inline output.
- An R plot drawn to a built-in file device is also replayed to the kernel PNG capture device, so a single plot produces one immutable inline image plus its independent working file.
- Artifact cards and general Files/Artifact TIFF/PDF previews are unchanged.

## Scope

- No persistence schema, field, data-model, or migration changes.
- Existing base64 capture persistence in `run.json` remains unchanged.
- Existing histories are corrected at render/export time; no data repair is required.

## Validation

- Targeted Renderer, export, and application tests: 8 files / 68 tests passed.
- R kernel integration file: 39/39 passed with real R, including PDF/TIFF, an unused TIFF device, base graphics, ggplot2, grid, and lattice.
- Complete `kernel-executor` file: 54/54 passed, including saved-file observation plus canonical PNG capture.
- `npm run typecheck`: passed for Node and Web.
- `npm run lint`: 0 errors; 17 unrelated pre-existing warnings.
- Full non-sandbox suite: 847 files / 12,389 tests passed, 14 files / 187 tests skipped. The only failure is the unrelated existing `NotificationInboxItem` Prisma runtime DDL parity failure in `src/main/projects/prisma-client.test.ts`.

## Review focus

- Confirm the capture-only cell-output boundary is applied consistently across Notebook, message details, and `.ipynb` export.
- Confirm saved files remain available exclusively through the Artifact/Files workflows.
- Confirm R file-device display-list replay produces one captured PNG without reading or inlining the saved file.


ewen-poch:
Addressed the review concern by defining the output boundary explicitly instead of conditionally suppressing saved files. Notebook/message figures and Jupyter cell outputs now come exclusively from immutable kernel-captured run.outputs. workingFiles and Artifacts remain independent saved files for the Artifact/Files workflows and are never inferred as cell display output. In the reviewed save-fig-1/close-it/leave-fig-2-open case, fig-2 is the inline captured output while fig-1 remains available as an Artifact; no cross-format identity heuristic or data-model change is required. Added coverage for saved-only files, captured-plus-saved runs, export behavior, and an R TIFF file-device plot.

ewen-poch:
Updated the R behavior to match the clarified requirement:

- `tiff()` / `pdf()` / other built-in file-device plots now keep the saved working file and also produce one immutable `image/png` kernel capture in `run.outputs`.
- The kernel enables R native display-list recording on the file device, records before `dev.off()`, and replays into the kernel-owned PNG device. It never reads or converts the saved TIFF/PDF/image file.
- Base graphics, ggplot2, grid, and lattice are covered. A single saved plot produces exactly one capture; an unused TIFF device produces none.
- Notebook, message details, and `.ipynb` export still consume only `run.outputs`; working files remain in the independent Artifact/Files workflow.
- No persistence fields, schema, data model, architecture, or interaction changes were introduced.

Validation: R loop 39/39, complete kernel-executor file 54/54, Node typecheck passed, and changed-file lint passed.

ewen-poch:
Addressed the P2 cross-request replay finding.

- Reproduced it with two real R loop requests: request A opened a TIFF device and plotted; request B plotted on its own capture device and then closed A's device. Before the fix, request B returned two PNG figures.
- File devices are now registered to the request that opened them. `dev.off()` records/replays a file device only when it belongs to the current request; closing an older device still closes/saves the file but does not inject stale history into the later run.
- Same-block `tiff(); plot(); dev.off()` / PDF behavior is unchanged and still produces exactly one canonical PNG capture.
- Added the cross-request regression test and tightened the Python `savefig()` integration test to require exactly one capture.

Validation: R loop 40/40, Python loop 5/5, kernel executor 54/54, renderer/export regressions 26/26, Node typecheck, changed-file lint, and `git diff --check` all pass. No persistence fields, schema, data model, architecture, or UI interaction changes were introduced.

ewen-poch:
Added same-run multi-format capture deduplication for R file devices.

- Reproduced the issue with one R request saving the same plot as `same-plot.pdf` and `same-plot.tiff`: both files were written, but two PNG captures were emitted.
- File devices now derive a request-local capture key from the normalized target path without its extension. A later device with the same key still writes/closes its saved file but does not replay a duplicate Notebook capture.
- Different paths/plots continue to produce separate captures. No file-content conversion or cross-format image comparison is involved.
- Parent-directory normalization avoids treating macOS `/var` and `/private/var` aliases as different keys; the same logic supports relative and Windows paths.

Validation: focused PDF/TIFF + unused-device + cross-request tests 3/3, complete R loop 41/41, the mixed-graphics executor case 1/1, Node typecheck, changed-file lint, and `git diff --check` pass. The full executor run reached 53/54 with only the mixed-graphics case exceeding its 15-second timeout under load; that exact case then passed alone in 8.5 seconds. No persistence, schema, data-model, architecture, or UI changes were introduced.

ewen-poch:
This behavior is intentional under the clarified product requirement. Within one R request, file-device targets that share the same normalized path stem are treated as one logical plot exported to multiple formats. All physical files are still written and retained, while Notebook, message output, and notebook export receive one canonical capture. Distinct logical plots should use distinct path stems. Using the full path would defeat the required PDF/TIFF multi-format deduplication, while content-based comparison is not reliable here because equivalent device replays can produce different PNG bytes and would add cross-format complexity. No code change is planned for this finding.

PR #963, opened 2026-08-08
feat(skills): add authenticated GitHub imports

### LINKED ISSUE #959: [Feature]: Support GitHub token for authenticated API requests during skill import
## What problem does this solve?

When importing skills from a large GitHub repository via **UI → Skills → Add Skill → Import from GitHub**, the feature hits GitHub's unauthenticated API rate limit (60 requests/hour) and fails with a `403` error.

**Reproduction steps:**
1. Open the Skills panel → Add Skill → Import from GitHub
2. Enter a URL to a repository with many skills, e.g. `https://github.com/aipoch/medical-research-skills`
3. The scan/review operation makes dozens of unauthenticated GitHub API calls (repo metadata, commit SHA resolution, recursive tree listing, per-skill directory walking + file downloads)
4. After approximately 60 API calls, subsequent requests receive HTTP 403 with rate-limit headers

**Root cause:** All GitHub API requests in `src/main/skills/github-import.ts` use unauthenticated headers (`User-Agent: open-science`, `Accept: application/vnd.github+json`) with no `Authorization` header. GitHub's unauthenticated rate limit is **60 requests/hour**, while an authenticated request (with a personal access token) raises the limit to **5,000 requests/hour** -- an 83x increase.

The rate-limit error is already caught gracefully in `searchGitHubSkillRepositories` (returning a user-facing message), but for the repo scan + import path (`scanRepoForSkills`, `fetchSkillPreview`, `fetchSkillFiles`), the error surfaces as a generic `GitHub API request failed (403)` without mentioning that authentication would solve it.

## Proposed solution

Add a **GitHub personal access token** configuration option so that skill import requests are authenticated:

1. **Settings UI**: Add a "GitHub token" field under Settings → Skills (or Settings → Connectors) where users can paste a [fine-grained or classic personal access token](https://github.com/settings/tokens). The token only needs the `public_repo` scope for public repositories, or `repo` scope for private repos.

2. **Backend plumbing**: When a token is configured, inject `Authorization: Bearer <token>` into the `GITHUB_HEADERS` object in `github-import.ts` before each API call. This applies to all GitHub API interactions:
   - `searchGitHubSkillRepositories` (keyword search)
   - `scanRepoForSkills` (repo tree scan)
   - `fetchSkillPreview` (single-skill preview)
   - `fetchSkillFiles` (full import)

3. **Token storage**: Store the token using the same `safeStorage` encryption pattern already used for API keys in `StoredProvider` (see `keyRef`/`keyMask` in `src/main/settings/types.ts`), so the plaintext token is never written to disk.

4. **Graceful degradation**: When no token is configured, behavior remains unchanged (unauthenticated requests with the 60 req/hour limit). The error message on rate-limit failure should be updated to hint that configuring a token would resolve the issue.

## Alternatives considered

- **Use GitHub CLI token detection (`gh auth token`)**: Could work as a fallback, but adds a CLI dependency and may surprise users. A settings-based approach is more explicit and portable.
- **Use `GITHUB_TOKEN` environment variable**: Reasonable for power users, but not discoverable from the UI. Could be added as a secondary source.
- **Batch import via zip/upload instead of API**: Already supported as an alternative path, but the GitHub import UX is more convenient for discovering and browsing skills.

## Additional context

- GitHub rate limit documentation: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api
- The affected code is in `src/main/skills/github-import.ts` -- the `GITHUB_HEADERS` constant and all functions that use it
- The `netFetch` wrapper (Electron net) already handles proxy configuration; adding auth headers follows the same pattern
- The `medical-research-skills` repo has dozens of skills across multiple directories, making it a realistic stress test for this feature

ewen-poch:
Closes #959

## Problem

GitHub Skill discovery and import use anonymous REST requests, so larger scans can exhaust GitHub's low unauthenticated rate limit. The Settings import page had no way to configure a credential, and search, scan, preview, Settings import, and conversation import did not share an authenticated request seam.

## Proposed change

- Add a compact GitHub token control to the Import from GitHub header with verify/save, replace, masked status, failure, and remove states.
- Verify candidate tokens with GitHub before replacing an existing credential, then persist only a `safeStorage`-encrypted reference and display mask.
- Apply `Authorization: Bearer` only to exact HTTPS `api.github.com` and `raw.githubusercontent.com` hosts.
- Reuse the authenticated fetch seam for repository search, direct scan, lazy preview, Settings import, and conversation-requested import.
- Make GitHub rate-limit failures point users back to the token control.
- Extend and certify the Electron/Web application-command inventories for the three local-only credential commands.

```mermaid
flowchart LR
  UI["Import from GitHub"] --> Verify["Main verifies candidate"]
  Verify --> Store["safeStorage encrypted reference"]
  Store --> Catalog["Skill catalog fetch seam"]
  Catalog --> Search["Search / scan"]
  Catalog --> Preview["Preview / import"]
  Catalog --> Conversation["Conversation import"]
```

## Scope and non-goals

- Keeps anonymous GitHub behavior when no decryptable token is configured.
- Does not add OAuth, synchronize credentials, or expose plaintext to renderer snapshots.
- Does not send the token to `github.com` lookalikes, non-HTTPS URLs, or arbitrary download hosts.
- Does not change Skill import size, depth, file-count, or request-count safety limits.

- A token can be verified, encrypted, saved, replaced, and removed without exposing plaintext.
- A failed replacement leaves the prior credential intact.
- Search, scan, preview, Settings import, and conversation import all use the configured credential.
- Missing or undecryptable credentials continue anonymously.
- The application starts with the updated certified command inventory.

Final evidence mapping (all commands ran after the last material source edit and the rebase onto `0a3b7270`):

| Behavior / risk | Check | Result |
| --- | --- | --- |
| Credential lifecycle, encryption, replacement safety, trusted-host header restriction, authenticated import paths, UI states, renderer/preload/application-command contracts, and startup inventory | 12-file impacted Vitest set covering composition, Settings commands/repository/catalog/IPC, GitHub import, preload, renderer contracts/surfaces, token UI, and SkillsPanel | 12 files, 316 tests passed |
| SettingsService GitHub façade and proxy-aware request wiring | `npm test -- src/main/settings/service.test.ts -t GitHub` | 6 passed, 232 skipped |
| Main/preload/renderer TypeScript contracts | `npm run typecheck` | passed |
| Changed source/test lint | ESLint over the 26 changed TypeScript/TSX files | passed with 0 warnings and 0 errors |
| Generated Web API compatibility map | `npm run check:web-api-map` | passed |
| Electron main/preload/renderer bundle, including the certified 230-command startup inventory | `npm run build:e2e` | passed |
| Full fallback selected because `docs/design.md` has no module owner | `npm test` | 12,350 passed; 31 environment-related failures in 8 unrelated files |

Uncovered local risks: the full Windows run cannot execute Bash workflow tests or create symlinks/ACLs in this environment, and several unrelated notebook/office tests time out under full-suite concurrency. None of the 31 remaining failures touches this diff; the focused impacted set and build are green. PR Gate remains authoritative for portable and platform lanes. Independent CI/reviewer confirmation of this evidence mapping is still pending, so the change is not marked independently verified here.

## Review focus

- Confirm the trusted-host allowlist is narrow enough and all GitHub Skill request paths use the wrapper.
- Confirm candidate verification precedes encryption/persistence and failure cannot overwrite the existing token.
- Confirm the local-only renderer contracts and 230/228/166/62 application-command inventory counts stay synchronized.
- Check the compact credential disclosure in light/dark Settings themes and narrow widths.


ewen-poch:
Addressed both review findings in 2065c5f0:

- Token management is now capability-gated: the control stays hidden when the status call is a remote-Web local-only rejecting stub, while desktop and local-Web behavior remain unchanged.
- HTTP 403 is now classified as rate limiting only when GitHub response metadata indicates an exhausted limit or retry window; bare 403 responses report forbidden access and permission guidance. Token verification uses the same distinction.

Validation: 49 focused tests passed, full Node/Web typecheck passed, and lint completed with 0 errors.

PR #998, opened 2026-08-09
feat(shortcuts): improve keyboard search navigation

ewen-poch:
## Problem

Composer `/` skill and `@` artifact result popups only accepted the highlighted item with Enter. Settings panels also exposed several search fields without a consistent keyboard shortcut, forcing keyboard users to tab through the panel before filtering.

## Proposed change

- Let plain Tab confirm the highlighted `/` skill or `@` artifact result while preserving Shift+Tab and modified Tab navigation.
- Add a shared Settings search input that focuses with Cmd+K on macOS and Ctrl+K on Windows/Linux, displays restrained platform keycaps, and exposes `aria-keyshortcuts`.
- Apply the shortcut to Skills, Connectors, Specialists, runtime package filtering, and Specialist capability pickers.
- Give the topmost nested Settings dialog priority so a background search cannot steal focus.
- Document the interaction contract in `docs/design.md`.

## Scope and non-goals

This change is limited to composer mention selection and existing Settings search/filter fields. It does not change the app-level command palette, add a shortcut to the GitHub repository query form, or alter search matching behavior.

- Composer Tab selection, modifier preservation, and empty/loading behavior -> targeted renderer tests -> 216 tests passed across 9 files.
- Settings Cmd/Ctrl+K focus, platform hints, nested-dialog ownership, and accessibility metadata -> targeted renderer tests -> included in the same 216 passing tests.
- Affected Node and renderer types -> `npm run typecheck` -> passed.
- Source and repository lint rules -> `npm run lint` -> passed with 0 errors (17 pre-existing warnings outside this diff).
- Final impact classification -> `node scripts/ci/module-test-impact.mjs affected --base main --head HEAD --explain` -> full fallback because `docs/design.md` is not assigned to a module owner.
- Complete portable suite -> `npm test -- --maxWorkers=4 --testTimeout=60000` -> 871 files passed, 14 skipped; 12,669 tests passed, 190 skipped.

The default-timeout full run first exposed four unrelated 15-second timeout flakes under concurrent load. All three affected files passed in isolation (83 tests), and the unfiltered suite passed after retaining four workers and raising the per-test timeout to 60 seconds.

Uncovered risk: no manual cross-platform Electron UI pass was run locally. macOS and the shared Windows/Linux Ctrl path are covered at component level; exact-head CI remains authoritative for platform lanes and independent verification.

## Review focus

- Confirm that Tab is prevented only when a selectable composer result exists and that Shift+Tab remains untouched.
- Confirm that the topmost Settings dialog owns Cmd/Ctrl+K when searches are nested.
- Check the in-field keycap spacing against the existing Settings toolbar at narrow widths.


ewen-poch:
Addressed the P2 about closing and hidden Settings searches. The shortcut hook now ignores inputs beneath closed, hidden, or aria-hidden ancestors, so the app-level Cmd/Ctrl+K handler remains available during close transitions. Added regression coverage for both closing and hidden dialog states. Validation passed: targeted test 5/5, full typecheck, lint, and the complete Vitest suite with 12,671 tests passing.

PR #1021, opened 2026-08-09
feat(projects): add project pinning

ewen-poch:
## Problem

Frequently used Projects cannot stay at the top of the Home Project list because every Project follows the same recent-activity ordering. When the entire Recent sessions list is empty, the Project rows also provide no artifact visibility.

## Proposed change

- Persist a per-Project pinned state and expose Pin/Unpin as the first Project menu action with the existing star icon language.
- Group pinned Projects ahead of unpinned Projects while preserving the existing recent-activity order inside both groups.
- Keep pin-only updates from changing `updatedAt`, so pinning does not masquerade as research activity.
- Update only the `pinned` column for pin-only patches, preventing a stale timestamp from overwriting concurrent Project activity.
- Show an authoritative artifact count on each Project row only when the complete Recent sessions catalog is empty and that Project's Files index is complete.
- Document the Home surface behavior in `docs/design.md`.

## Scope and non-goals

- Reuses the existing Project update command and Project Files overview contract; no new dependency or IPC surface.
- Does not change the activity ordering inside pinned or unpinned groups.
- Does not display partial artifact-index counts.
- Subscribes to the existing Project Files change event only while artifact counts are visible, and refetches only the affected Project.

- Pin state persists through the Project schema and legacy migration: targeted repository, Prisma integration, and shared-contract tests pass.
- Concurrent pin/activity regression: the repository test fails against the original read/write timestamp implementation and passes with the pin-column-only update; the real SQLite round trip also preserves `updatedAt`.
- Pin/Unpin menu behavior, pinned grouping, within-group activity ordering, and empty-Recent artifact visibility: Home render/persistence tests pass.
- Targeted suite: 5 files, 49 tests passed.
- `npm run lint`: passed with 17 pre-existing warnings and no errors.
- Prisma schema validation: passed.
- Generated web API contract check: passed.
- Web production build: passed.
- Desktop, 390 px mobile, and isolated empty-Recent visual QA: passed with no horizontal overflow.
Checks above were rerun after rebasing onto `688ec334`. Cross-platform behavior remains delegated to CI.

## Review focus

- Pin-only `updatedAt` preservation in `ProjectRepository.update`.
- Pinned-first grouping without disturbing recent-activity ordering inside either group.
- Artifact visibility gating on both an empty complete Session catalog and a complete Project Files index.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -580,7 +580,8 @@ Workspace-only tokens without a shadcn counterpart, plus shadow tokens. For shar
 - Recent sessions: the secondary line is always the owning Project name, never a prompt preview or a
   repeat of the Session title. Once the Session catalog is complete, if the entire Recent sessions
   list is empty, each Project row also shows its artifact count from a complete Project Files index;
-  partial index counts are omitted.
+  partial index counts are omitted. While those counts are visible, Project Files change events
+  refresh the affected Project so index repair and file changes do not leave stale totals.
 - List row: `h-10 rounded-lg px-3 hover:bg-accent hover:text-accent-foreground`.
 - Inline more actions: default `opacity-0`, then `opacity-100` on hover or focus-visible.
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -548,8 +548,10 @@ Workspace-only tokens without a shadcn counterpart, plus shadow tokens. For shar
 - Projects title: use `GalleryVerticalEnd`; activity counters sit beside each Project name and split
   `running` from `waiting on you` rather than presenting one ambiguous total.
 - Project actions: use the shared Session dropdown styling and size the surface to its content so
-  inline padding stays balanced. The first action is `Settings`, opening `Project Settings` for Name
-  and Description; its edit action is labelled `Save`.
+  inline padding stays balanced. The first action is `Pin project` / `Unpin project`, using an outline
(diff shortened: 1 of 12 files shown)


ewen-poch:
Resolved the concurrent activity timestamp finding in 5981f262. Pin-only patches now issue a parameterized UPDATE that writes only the pinned column, so they neither trigger Prisma updatedAt automation nor restore a stale timestamp. A repository regression reproduces the old rollback under concurrent activity, and the real SQLite round trip verifies pinned persistence with an unchanged updatedAt. The branch is also rebased onto 9574fe88.

ewen-poch:
Resolved the artifact refresh finding in 097c33e8. Home now subscribes to the existing Project Files change event only while counts are visible, refetches only the affected active Project, and removes the listener when counts are hidden or Home unmounts. Per-Project request versions prevent an older initial overview response from overwriting a newer event refresh. The Home regression verifies 114 to 115 artifacts after an upsert event and listener cleanup. Targeted Project/Home suite: 5 files, 49 tests passed; Web typecheck passed; lint passed with 0 errors and 17 existing warnings. The full local suite passed 12,832 tests with three unrelated 15-second contention timeouts, and both timeout files then passed 32 of 32 in isolation. The branch remains rebased onto 688ec334.

PR #1043, opened 2026-08-10
fix(compute): coordinate jobs with owner deletion

ewen-poch:
## Problem

Project and Session deletion bypassed the independent Compute Job lifecycle. Deleting an owner could therefore leave promoted or running remote work, remote Job directories, persisted Job rows, and pending Job notifications behind. The previous cleanup order also removed Job history before durable owner deletion had committed.

## Proposed change

- Keep `ComputeJobLifecycle` as the sole owner of Job status-transition intent and add owner-deletion begin/commit/abort intents.
- Serialize Job creation, conditional status transitions, owner barriers, and owner row deletion through the repository mutation queue.
- Acquire a reference-counted dispatch handoff lease before publishing a direct submitted row, then transfer it synchronously to `dispatchJob`.
- Arm the owner deletion barrier before Session/Project reads, upload migration, or provenance preparation, so no Job can enter during fallible preparation.
- Pause owner-scoped queued promotion in `ConcurrencyManager`, drain active promotion/dispatch handoff, then pause and drain poller/background work.
- During prepare, resolve and validate a complete in-memory remote cleanup plan while retaining Job rows and notification history; do not contact the remote host.
- Commit Session JSON deletion or the durable Project Session tombstone before executing any irreversible remote cleanup.
- Execute the idempotent remote cleanup plan after owner authority commits, then delete Job/notification rows only after every cleanup succeeds.
- If remote cleanup or Job-row deletion fails after authority commits, retain Job rows, notifications, the admission barrier, and runtime pause for background retry; never restore Session/Project authority.
- If preparation or owner-authority deletion fails before authority commits, compensate file/provenance state and abort the Compute barrier.
- At startup, restore local Project-intent and orphan-Session Job barriers before the Compute poller starts, without performing SSH work.
- Start the application normally, then run remote deletion recovery and orphan reconciliation in a lifecycle-owned background loop that logs and retries failures every 30 seconds.
- Broadcast already-persisted Job updates when an existence recheck fails transiently; only confirmed absence suppresses a delayed update.
- Resume owner-scoped queue draining after an aborted deletion so eligible queued Jobs do not wait for an unrelated terminal transition.
- Exclude the Electron IPC composition root from changed-file coverage, matching the existing non-logic composition-root policy.

```mermaid
flowchart LR
  A[Project or Session deletion] --> B[Begin lifecycle barrier]
  B --> C[Pause owner promotion and drain runtime]
  C --> D[Build and validate cleanup plan]
  D --> E[Commit Session JSON deletion or Project Session tombstone]
  E --> F[Stop remote process and remove Job directory]
  F --> G[Delete Job and notification rows]
  F -. cleanup fails .-> H[Retain Job rows, notifications, runtime pause, and barrier]
  H --> I[Retry full idempotent plan in background]
  I --> F
  S[Application startup] --> T[Restore local barriers only]
  T --> U[Start Compute runtime and application]
  U --> I
```

## Scope and non-goals

Included: main-process Compute lifecycle/repository/runtime, Session persistence deletion, durable Project deletion recovery, startup orphan barrier restoration, background recovery retry, module-impact certification, and coverage configuration.

Not included: schema changes, a user-facing manual cancel command, retained Job audit tombstones, renderer interaction changes, or live-cluster SSH integration.

Related issue: none provided.

- Changed-file PR coverage command: `npx vitest run --coverage --changed origin/main` -> 41 files passed, 1 skipped; 1,144 tests passed, 15 skipped; statements 88.38%, branches 82.72%, functions 80.15%, lines 91.45%.
- Compute, Project deletion, and Session persistence certification set -> 36 files passed, 1 skipped; 727 tests passed, 15 skipped.
- Focused startup barrier/retry tests -> 47 tests passed.
- Complete Compute IPC test file -> 54 tests passed.
- Latest affected Session/Project/Compute deletion set -> 8 files and 231 tests passed.
- `npm run typecheck` -> passed; latest follow-up also passed `npm run typecheck:node`.
- `npm run lint` -> passed with existing repository warnings and no errors; changed files have no warnings.
- `git diff --check` -> passed.
- Full fallback `npm test` -> 908 files passed, 13 skipped; 13,061 tests passed, 275 skipped. Ten failures require Windows symlink privilege. The remaining nine load-sensitive failures all passed when their unchanged cases were rerun in isolation (9 passed, 358 skipped).

Platform coverage: local validation ran on Windows; macOS CI is authoritative for the selected PR unit/coverage lane. Remote cleanup is intentionally POSIX-only because it executes through an SSH shell; generated POSIX paths are covered and Windows-style remote paths are rejected.

## Review focus

- Authority-before-remote-cleanup ordering and explicit Job-row deletion semantics.
- Owner-scoped admission/promotion barrier and dispatch quiescence.
- Idempotent full-plan retry while durable intent, Job rows, notifications, and barriers remain fail-closed.
- Local-only startup barrier restoration before poller activation.
- Remote path validation and process-group termination.

Uncovered risk: behavior against a real scheduler/SSH host is covered through runner contract tests rather than a live cluster. An unresolved provider alias or unreachable host no longer prevents Session/Project authority deletion once preparation succeeds, but it retains the owner barrier and Job history until background recovery can complete.

### DIFF
--- a/scripts/ci/module-impact.json
+++ b/scripts/ci/module-impact.json
@@ -177,6 +177,7 @@
       "ownerPaths": [
         "src/main/compute/compute-host-profile-owner.ts",
         "src/main/compute/compute-job-lifecycle.ts",
+        "src/main/compute/job-deletion-owner.ts",
         "src/main/compute/compute-job-workflow-owner.ts",
         "src/main/compute/compute-remote-operation-owner.ts",
         "src/main/compute/concurrency-manager.ts",
@@ -191,6 +192,7 @@
       "testFiles": {
         "owner": [
           "src/main/compute/compute-job-lifecycle.test.ts",
+          "src/main/compute/job-deletion-owner.test.ts",
           "src/main/compute/compute-service.architecture.test.ts",
           "src/main/compute/compute-host-profile-owner.test.ts",
           "src/main/compute/compute-job-workflow-owner.test.ts",
--- a/scripts/ci/module-test-impact.test.ts
+++ b/scripts/ci/module-test-impact.test.ts
@@ -41,6 +41,7 @@ describe('module test impact commands', () => {
     expect(plan.testFiles).toEqual(
       expect.arrayContaining([
         'src/main/compute/compute-job-lifecycle.test.ts',
+        'src/main/compute/job-deletion-owner.test.ts',
         'src/main/compute/compute-service.test.ts',
         'src/main/compute/ssh-runner.test.ts',
         'src/main/compute/ipc.test.ts',
@@ -65,6 +66,7 @@ describe('module test impact commands', () => {
 
   it.each([
     'src/main/compute/compute-job-lifecycle.ts',
+    'src/main/compute/job-deletion-owner.ts',
     'src/main/compute/job-repository.ts',
     'src/main/compute/concurrency-manager.ts',
     'src/main/compute/job-dispatcher.ts',
@@ -81,6 +83,7 @@ describe('module test impact commands', () => {
       expect.arrayContaining([
         'src/main/compute/compute-service.architecture.test.ts',
         'src/main/compute/compute-job-lifecycle.test.ts',
+        'src/main/compute/job-deletion-owner.test.ts',
         'src/main/compute/concurrency-manager.test.ts',
         'src/main/compute/job-dispatcher.test.ts',
         'src/main/compute/job-poller.test.ts',
--- a/src/main/compute/compute-job-lifecycle.test.ts
+++ b/src/main/compute/compute-job-lifecycle.test.ts
@@ -90,6 +90,31 @@ describe('ComputeJobLifecycle', () => {
     expect(publish).toHaveBeenCalledOnce()
   })
 
+  it('blocks later transitions and retains rows until owner deletion commits', async () => {
+    const owner = { projectId: 'project-1', sessionId: 'session-1' }
+
+    await lifecycle.beginOwnerDeletion(owner)
+
+    await expect(lifecycle.promoteQueued('queued-job')).resolves.toEqual({ kind: 'ignored' })
+    await expect(repository.get('queued-job')).resolves.toMatchObject({ status: 'queued' })
+    await expect(repository.get('submitted-job')).resolves.not.toBeNull()
+
+    await lifecycle.deleteOwnerRows(owner)
+
+    await expect(repository.get('queued-job')).resolves.toBeNull()
+    await expect(repository.get('submitted-job')).resolves.toBeNull()
+    await expect(repository.get('running-job')).resolves.toBeNull()
+  })
+
+  it('reopens transitions when prepared owner deletion aborts', async () => {
+    const owner = { projectId: 'project-1', sessionId: 'session-1' }
+
+    await lifecycle.beginOwnerDeletion(owner)
+    await lifecycle.abortOwnerDeletion(owner)
+
+    await expect(lifecycle.promoteQueued('queued-job')).resolves.toMatchObject({ kind: 'applied' })
+  })
+
   it('keeps an applied promotion successful when its observer throws', async () => {
     publish.mockImplementation(() => {
       throw new Error('observer failed')
--- a/src/main/compute/compute-job-lifecycle.ts
+++ b/src/main/compute/compute-job-lifecycle.ts
@@ -1,5 +1,5 @@
 import type { ComputeJob, ComputeJobStatus } from '../../shared/compute'
-import type { ComputeJobRepository, UpdateJobRequest } from './job-repository'
+import type { ComputeJobOwner, ComputeJobRepository, UpdateJobRequest } from './job-repository'
 
 export type ComputeJobTransitionResult = { kind: 'applied'; job: ComputeJob } | { kind: 'ignored' }
 
@@ -16,6 +16,18 @@ export class ComputeJobLifecycle {
     private readonly repository: ComputeJobRepository,
     private readonly onApplied: (job: ComputeJob) => void = () => undefined
   ) {}
+
+  beginOwnerDeletion(owner: ComputeJobOwner): Promise<void> {
+    return this.repository.beginOwnerDeletion(owner)
+  }
+
+  deleteOwnerRows(owner: ComputeJobOwner): Promise<void> {
+    return this.repository.deleteByOwner(owner)
+  }
+
+  abortOwnerDeletion(owner: ComputeJobOwner): Promise<void> {
+    return this.repository.abortOwnerDeletion(owner)
+  }
 
   async promoteQueued(jobId: string): Promise<ComputeJobTransitionResult> {
     return this.apply(jobId, ['queued'], {
--- a/src/main/compute/compute-job-workflow-owner.test.ts
+++ b/src/main/compute/compute-job-workflow-owner.test.ts
@@ -11,6 +11,7 @@ import type { ComputeHostRepository } from './repository'
 import type { ResolvedSshTarget, SshRunner } from './ssh-runner'
 import type { ScpRunner } from './scp-runner'
 import type { ConcurrencyManager } from './concurrency-manager'
+import { sharedDispatchTracker } from './dispatch-tracker'
 
 // ---------------------------------------------------------------------------
 // Helpers
@@ -208,6 +209,66 @@ const makeJobRepo = (
 }
 
 describe('ComputeJobWorkflowOwner.submitJob', () => {
+  it('tracks a committed submitted row through dispatcher registration', async () => {
+    const runner = makeFakeRunner({
+      exitCode: 1,
+      stdout: '',
+      stderr: 'dispatch stopped',
+      truncated: false,
+      timedOut: false
+    })
+    const { repo: jobRepo, createCalls } = makeJobRepo()
+    const { repo } = makeRepo()
+    const broker = {
+      request: vi.fn(),
+      requestWithContext: vi.fn(() => Promise.resolve('once' as const)),
+      respond: vi.fn()
+    } as unknown as ComputeApprovalBroker
+    let handoffWait: Promise<void> | undefined
+    const concurrencyManager = {
+      enqueue: vi.fn(async () => 'can_dispatch' as const),
+      admit: vi.fn(
+        async (
+          _params: { sessionId: string; providerId: string },
+          commit: (status: 'submitted' | 'queued') => Promise<void>
+        ): Promise<'submitted'> => {
+          await commit('submitted')
+          const request = createCalls.mock.calls[0]?.[0] as
+            import('./job-repository').CreateJobRequest | undefined
+          expect(request).toBeDefined()
+          expect(sharedDispatchTracker.has(request!.id)).toBe(true)
+          const settled = vi.fn()
+          handoffWait = sharedDispatchTracker.waitFor([request!.id]).then(settled)
+          await Promise.resolve()
+          expect(settled).not.toHaveBeenCalled()
+          return 'submitted'
+        }
+      )
+    } as unknown as ConcurrencyManager
+    const service = makeOwner(
+      runner,
+      repo,
+      broker,
+      jobRepo,
+      undefined,
+      undefined,
+      undefined,
+      concurrencyManager
+    )
+
+    const result = await service.submitJob(
+      'ssh:biowulf',
+      'test dispatch handoff',
+      'echo hi',
+      {},
+      { sessionId: 's1', projectId: 'p1' }
+    )
+
+    expect(handoffWait).toBeDefined()
+    await handoffWait
+    expect(sharedDispatchTracker.has(result.job_id)).toBe(false)
+  })
+
   it('returns job_id + remote_workdir immediately (before dispatch)', async () => {
     // Runner should never be called for submit_job itself (dispatch is background).
     const runner = makeFakeRunner({
--- a/src/main/compute/compute-job-workflow-owner.ts
+++ b/src/main/compute/compute-job-workflow-owner.ts
@@ -12,6 +12,7 @@ import type {
 import { getNotebookSessionRoot } from '../notebook/repository'
 import type { ComputeApprovalBroker } from './compute-approval-broker'
 import type { ConcurrencyManager, SessionStatus } from './concurrency-manager'
+import { sharedDispatchTracker } from './dispatch-tracker'
 import { computeRemoteWorkdir, dispatchJob, hashCommand } from './job-dispatcher'
 import type { StagedInputEntry } from './job-dispatcher'
 import type { ComputeJobRepository } from './job-repository'
@@ -269,7 +270,12 @@ export class ComputeJobWorkflowOwner {
     const commandHash = hashCommand(command)
     const inputManifest = stagedEntries.length > 0 ? JSON.stringify(stagedEntries) : undefined
     const jobRepository = this.jobRepository
+    let dispatchHandoffHeld = false
     const createRow = async (initialStatus: 'submitted' | 'queued'): Promise<void> => {
+      if (initialStatus === 'submitted') {
+        sharedDispatchTracker.begin(jobId)
+        dispatchHandoffHeld = true
+      }
       await jobRepository.create({
         id: jobId,
         providerId: host.providerId,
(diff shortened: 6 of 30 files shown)


ewen-poch:
Resolved the merge conflict that caused the AI Review checkout failure.

- Preserved both the Compute Job deletion participant wiring and the latest Side Chat composition in `src/main/ipc.ts`.
- Kept the existing 660-line Compute IPC architecture gate by simplifying the equivalent delayed-broadcast lookup chain; no architecture, data-model, or user-interaction behavior changed.
- Confirmed the PR diff remains limited to the original 17 files.

Validation after the resolution:
- Complete PR impact set: 10 files, 257 tests passed.
- `npm run lint`: passed with the existing repository warnings.
- `git diff --check origin/main...HEAD`: passed.

Two latest-main Side Chat tests still fail only on this Windows host because their assertions expect POSIX `/cwd` separators; those files are unchanged by this PR and Linux CI remains the authoritative lane.

ewen-poch:
Addressed both P1 lifecycle findings after rebasing onto the latest Compute ownership refactor:

- Job rows and notification history now remain intact during remote cleanup. Session JSON deletion or the durable Project Session tombstone commits first; only then does `ComputeJobLifecycle` commit owner-scoped Job row deletion. A failed Session row commit restores the Session JSON and aborts the barrier, while committed Project tombstones retain their recovery intent for retry.
- `ConcurrencyManager` now provides an owner-scoped promotion barrier that drains any active promotion/dispatch handoff. Repository creation and conditional status transitions are serialized behind the same owner barrier, and the poller/background runtime is drained before the final snapshot and remote cleanup.
- Startup now recovers pending Project deletions and reconciles orphan Job owners before starting the Compute poller.
- The previous changed-coverage failure is fixed by explicitly excluding the non-logic Electron IPC composition root. The CI-equivalent command passed with 1,108 tests and coverage of 87.26% statements / 81.56% branches / 77.91% functions / 90.32% lines.

Additional validation: full typecheck passed, lint passed with existing warnings only, the Compute certification set passed 531 tests, and the architecture gate remains at the existing 660-line limit.

ewen-poch:
Addressed the deleted-host and nullable-path review findings without skipping remote cleanup:

- A terminal Job no longer requires its Compute Host database row. When the row has been deleted, cleanup recovers the SSH alias from the persisted, canonical `ssh:<alias>` provider ID and resolves the target through the normal SSH configuration path.
- A legacy Job with no persisted `remote_workdir` now derives the exact dispatcher fallback via `computeRemoteWorkdir(host.scratchRoot, jobId)` when the Host row is available.
- Unsafe or genuinely unrecoverable cleanup still fails closed; no Job rows are committed away until the remote cleanup succeeds.

Regression tests cover both cases. The focused lifecycle set passed 158 tests (5 skipped), Node typecheck and lint passed, and the CI-equivalent changed-coverage run passed 1,110 tests (15 skipped) at 87.30% statements / 81.69% branches / 77.97% functions / 90.35% lines.

ewen-poch:
Closed the dispatch-start race with an explicit lease handoff:

- Direct submit now acquires a `DispatchTracker` lease before publishing a `submitted` Job row.
- `dispatchJob` synchronously acquires its own overlapping lease before its first await; only then does submit release the handoff lease.
- `DispatchTracker` is reference-counted, so owner deletion and the poller remain blocked until the final handoff/dispatch lease ends. Row-creation failures release the handoff lease in `finally`.
- Queued promotion continues to use the existing owner-scoped `ConcurrencyManager` drain.

Regression tests cover the committed-row handoff window and overlapping leases. The dispatch/deletion certification set passed 154 tests (5 skipped), Node typecheck and lint passed, and the CI-equivalent changed-coverage run passed 1,112 tests (15 skipped) at 88.00% statements / 82.89% branches / 78.91% functions / 90.79% lines.

ewen-poch:
Addressed the startup-blocking recovery finding while preserving fail-closed deletion semantics:

- Startup now restores Project-level barriers for durable deletion intents and legacy tombstones, plus Session-level barriers for orphan Job owners, using local authority only. This phase performs no SSH work and completes before the Compute poller starts.
- Remote Project recovery and orphan reconciliation now run after runtime startup in a lifecycle-owned background loop. A transient failure is logged and retried every 30 seconds, so an unreachable host no longer prevents the application from opening.
- Barriers restored from durable authority are retained when remote cleanup fails; queued promotion and new Job creation remain blocked while the deletion intent, Job rows, and notification history stay intact.
- Project commit releases covered local queue barriers only after Job-row deletion succeeds. Online pre-commit aborts still release their barriers normally.

Regression coverage verifies local-only barrier restoration, multiple orphan owners, Project-level failure retention, safe retry, and startup ordering. The focused suite passed 47 tests; the broader Compute/Project/Session certification passed 727 tests (15 skipped). The CI-equivalent changed-coverage run passed 1,138 tests (15 skipped) at 88.28% statements / 82.84% branches / 80.05% functions / 91.37% lines. Full typecheck and changed-file lint passed.

ewen-poch:
Addressed the transient repository lookup case in `86e2230a`.

- A successful `null` lookup still suppresses a delayed update after owner deletion.
- A rejected lookup is now treated as unknown rather than deletion evidence, so the already-persisted transition is broadcast and the renderer does not silently miss it.
- Added regression coverage for both the transient-failure path and the confirmed-deletion path.

Validation: complete Compute IPC tests (54 passed), architecture guard (9 passed, `ipc.ts` remains at 660 lines), Node typecheck, changed-file lint, and changed coverage (1,139 passed / 15 skipped).

ewen-poch:
Reviewed both findings in `c71fc32d`.

**Owner barrier after row deletion:** no lifecycle change was needed because the reported admission gap does not occur. `beginOwnerDeletion()` records the owner in `ComputeJobRepository`; `commitOwnerDeletion()` delegates only to `deleteByOwner()` and never clears that record. Every Job create is serialized through the same repository mutation queue and calls `assertOwnerMutable()`. `ConcurrencyManager.resumeOwner()` only re-enables queue scans—it cannot clear the repository barrier or admit a new Job. Added an integration assertion that a create after `deleteByOwner()` still rejects, and that only explicit `abortOwnerDeletion()` restores admission.

**Queue drain after abort:** addressed. `resumeOwner()` now triggers the existing `onJobCompleted()` drain after removing the owner pause, so eligible queued Jobs resume without waiting for an unrelated terminal transition. The regression test now observes automatic promotion rather than manually invoking a drain.

Validation: affected Compute lifecycle tests (69 passed), Node typecheck, changed-file lint, and changed coverage (1,139 passed / 15 skipped; 88.30% statements, 82.84% branches, 80.10% functions, 91.38% lines). The repository-wide run passed 13,056 tests; its Windows privilege/load-sensitive failures all passed or were isolated as documented in the PR body.

ewen-poch:
Addressed in `6586eef5`.

- Session authority restoration remains the release condition: if restoring the Session JSON fails or is unknown, the Compute barrier stays fail-closed.
- Once Session JSON restoration succeeds, provenance/file-index compensation now runs inside a `try` whose `finally` independently calls `abortSessionJobDeletion()`. A later compensation failure therefore still surfaces to the caller, but can no longer strand the repository barrier, owner queue pause, or poller pause.
- Added a regression where Job-row commit fails, Session JSON restoration succeeds, and file-index restoration fails; the test verifies the original compensation error is preserved and Compute admission is released.

Validation: affected Session/Compute deletion tests (136 passed), Node typecheck, changed-file lint, and changed coverage (1,140 passed / 15 skipped; 88.31% statements, 82.84% branches, 80.10% functions, 91.38% lines). The repository-wide run passed 13,056 tests; all Windows privilege/load-sensitive failures were isolated as documented in the PR body.

ewen-poch:
Addressed both findings in `cfc660ba`.

- **Barrier ordering:** Session and Project deletion now prepare the existing Compute Job owner barrier before reading deletion inputs, migrating uploads, or preparing provenance. The Compute participant already self-aborts if its own preparation fails; later Session failures use the local compensation path, while Project failures use the coordinator's durable-state-aware abort path.
- **Missing Session authority:** the Session path now records a confirmed pre-existing absence. If Job-row commit then fails, it restores the soft-deleted file-index token and aborts Compute Job deletion without creating Session JSON authority. Failures after a successful Job commit retain the committed deletion behavior.

Added three regressions covering Session ordering, Project ordering, and the missing-Session commit failure. Validation passed: 218 affected Session/Project/Compute deletion and architecture tests, Node typecheck, changed-file lint, `git diff --check`, and changed coverage (1,143 passed / 15 skipped; 88.31% statements, 82.92% branches, 80.10% functions, 91.39% lines).

ewen-poch:
Addressed in `ace382b0`.

- `prepare*` now arms admission/promotion/runtime barriers, drains dispatch handoff, and builds a fully validated in-memory cleanup plan. It does not contact the remote host.
- Session JSON deletion or the durable Project Session tombstone commits before `commit*`.
- `commit*` executes the idempotent remote plan, then calls the explicitly named `deleteOwnerRows()` only after every cleanup succeeds.
- A remote or row-deletion failure after authority commit leaves Job rows, the owner barrier, and runtime pause intact. The existing startup barrier restoration/background orphan reconciliation retries the full idempotent plan; Session authority is no longer restored.
- Pre-authority failures still compensate file/provenance state and abort the barrier.

Added regressions for no pre-authority remote mutation, partial multi-Job cleanup retry, startup recovery, Session found/missing authority behavior, and Project tombstone retention.

Validation: 231 affected tests, Node typecheck, changed-file lint, `git diff --check`, and changed coverage (1,144 passed / 15 skipped; 88.38% statements, 82.72% branches, 80.15% functions, 91.45% lines). Full fallback passed 13,061 tests; the 10 Windows symlink failures require privilege, and all 9 load-sensitive failures passed in isolation.

ewen-poch:
Addressed in `f1a57f75`.

- Every persisted-handle or `job.pid` PID now passes a remote cwd ownership guard before any TERM/KILL signal.
- The command resolves the canonical Job work directory, checks `/proc/<pid>/cwd` on Linux, and falls back to `lsof` cwd output for supported macOS direct-SSH hosts.
- If the directory was already removed by an earlier successful cleanup, or process ownership cannot be proven, cleanup skips process signaling and keeps the directory removal idempotent.
- A later Job failure can therefore retry the full plan without signaling a PID that has since been reused by an unrelated process.

Regression coverage asserts that persisted PIDs are routed only through the ownership-guarded helper, including both cwd lookup paths.

Validation: 231 affected Session/Project/Compute tests, Node typecheck, zero-warning ESLint on the changed files, and `git diff --check`.

ewen-poch:
Reviewed both findings in `3f71ad46`.

**Owner barrier after row deletion:** no lifecycle change was needed. `releaseCommittedOwnerBarriers()` releases only the local queue/runtime pause and coordinator bookkeeping; it deliberately does not call `ComputeJobLifecycle.abortOwnerDeletion()`. The Repository's `deletingProjects` / `deletingSessions` admission gate therefore remains active after `deleteOwnerRows()`, and both Job creation and conditional transitions stay serialized behind that gate. The existing integration regression verifies that a create after owner-row deletion still rejects until an explicit Repository barrier abort.

**Startup poller recovery:** addressed. `findNonTerminal()`, `findTerminalUnharvested()`, and `findErrorUnnotified()` now uniformly exclude owners with an active deletion barrier. Because startup awaits Project/orphan barrier restoration before registering the Compute runtime, the first poller tick can no longer poll, harvest, or emit error notifications for those owners. `findByOwner()` remains unfiltered so deletion recovery can still inspect and remove the retained rows.

Validation: 264 affected Session/Project/Compute tests including the complete poller test file, Node typecheck, zero-warning ESLint on the changed files, and `git diff --check`.

ewen-poch:
Reviewed both findings in `44d9557e`.

**Session tombstone after post-authority failure:** addressed. If the deletion operation rejects, the outer coordinator now re-reads Session authority and clears the in-memory tombstone only when the JSON is explicitly `found`. A missing, unreadable, or unverifiable authority state retains the tombstone fail-closed. A Compute cleanup failure after JSON deletion can therefore no longer reopen `saveSession` or other Session mutations and resurrect the owner. Pre-authority failures whose JSON is confirmed live still clear the tombstone as before.

**Project barrier through the deletion tail:** no Compute lifecycle change was needed. `commitProjectJobDeletion()` deletes the owner rows and releases only local queue/runtime pauses; it does not call `abortOwnerDeletion()`. The Repository's project admission gate remains armed through the remaining Project deletion tail and continues rejecting creates and conditional transitions. The owner-deletion integration test verifies that a create remains rejected after row deletion until an explicit Repository barrier abort.

Regression coverage now verifies both the post-authority rejected save and the pre-authority confirmed-live recovery behavior.

Validation: 264 affected Session/Project/Compute tests including the complete poller test file, Node typecheck, zero-warning ESLint on the changed files, and `git diff --check`.

ewen-poch:
Addressed in `591231e7`.

- The cleanup command now rejects a Job work-directory entry that is itself a symlink.
- It resolves the accepted directory once with `cd -- <validated-path> && pwd -P`, so mutable symlinked parents are removed from the path used for destructive operations.
- PID cwd ownership checks, recursive removal, and the final absence check now all use that canonical runtime value.
- The original validated path remains only for the initial symlink check and locating `job.pid`; it is no longer passed to `rm -rf`. Any PID read through that path must still match the canonical Job cwd before a signal is sent.
- A previously completed cleanup leaves the canonical value empty and remains an idempotent success.

Regression coverage asserts the symlink guard, canonical recursive removal/absence check, and absence of `rm -rf` against the original path.

Validation: 264 affected Session/Project/Compute tests including the complete poller test file, Node typecheck, zero-warning ESLint on the changed files, and `git diff --check`.

ewen-poch:
Addressed the retained child Session / parent Project recovery gap in `1ce22024`.

- Added Project-scoped orphan reconciliation that revalidates Project and Session authority before retrying cleanup, so a live or unreadable Session is never treated as an orphan.
- Project deletion preflight now finishes retained orphan Session cleanup for that Project before preparing the Project-wide Compute deletion.
- Background recovery now reconciles orphan Session jobs before replaying durable Project deletion intents.
- A parent Project abort no longer throws or clears a different retained child Session plan; the child plan and any restored durable Project barrier remain fail-closed for retry.
- Regression coverage verifies retained-child recovery, parent preparation after recovery, Project scoping, and the background ordering contract.

The late-approval finding did not require a change: approval completion still has to pass through `jobRepository.create()`, and the Repository owner-admission gate remains armed after committed deletion; pending approvals are process-local and cannot survive restart.

Validation after the final edit: 7 affected files / 265 tests passed; full typecheck passed; full lint completed with 0 errors (17 existing warnings, none in changed files); complete portable suite reached 13,064 passed / 275 skipped. The 10 symlink failures require Windows symlink privilege, and all 8 load-sensitive failures passed when their three files were rerun in isolation (128 tests passed).

ewen-poch:
Addressed the canonical ancestor-symlink cleanup risk in b960150f. The cleanup command now canonicalizes the configured scratch root and Job workdir independently, derives the only allowed canonical path as <scratch-root>/.openscience/jobs/<job-id>, and exits before PID access, signaling, or rm unless they match exactly. This preserves the direct-workdir symlink guard and canonical removal. Validation: focused Compute tests 26/26, affected tests 265/265, typecheck, zero-warning ESLint on the changed files, and the unrelated load-sensitive failures passed when rerun individually; the known Windows symlink-privilege failures remain environment-specific.

ewen-poch:
Reviewed this reported lock cycle and did not change production code because withDataRootWrite is not a mutually exclusive writer lock. It synchronously acquires an activity-count lease; concurrent writers are allowed, and only data-root migration waits for the writer count to drain. Therefore a poller harvest from another async context does not wait behind the deletion lease, so 
untime.pause() cannot form the proposed cycle. I exercised the exact shape with a pre-existing harvest acquiring withDataRootWrite while deletion held its own lease and awaited it through 
untime.pause(); it completed in 3 ms instead of timing out. When migration is already pending, a new writer is rejected rather than queued, which also cannot create this cycle. The intentional remaining wait is the harvest operation itself, which must drain before remote Job cleanup. The temporary repro was removed; migration-state.test.ts plus job-deletion-owner.test.ts pass 28/28 and the worktree is clean.

ewen-poch:
Addressed the concurrent child/parent deletion finding in 8b8dc13c. A foreign owner prepare now snapshots the active plan outcome inside the serialized operation queue, then waits outside that queue so the owning Session commit or abort can still run. A successful child commit releases the waiter and the Project retries preparation; a failed remote cleanup propagates the same retained-plan error to the waiting Project instead of hanging. Same-owner prepare remains idempotent. The regression was red before the fix with the reported Another Compute Job owner deletion is already prepared error and is now green. Validation: 19/19 owner tests, 7 affected files / 267 tests, Node typecheck, zero-warning ESLint on both changed files, and git diff --check.

ewen-poch:
Addressed both recovery lifecycle findings in bf400bbd. Compute owner liveness is now tri-state: an unreadable Session returns unknown, startup restores a fail-closed admission barrier, reconciliation defers all cleanup while authority remains unknown, and a later live read releases only that temporary barrier. This lets application startup continue without converting unreadable authority into deletion authority. ProjectDeletionRecoveryLoop.stop() now disables retries, clears the timer, and awaits the tracked active recovery Promise; application-module disposal continues to enforce the existing bounded timeout. Both regressions were red before the fix and are now green. Validation: 7 affected files / 270 tests, Node typecheck, zero-warning ESLint across all 6 changed files, and git diff --check.

ewen-poch:
Addressed the final stale-broadcast finding in fdb8839d. The publisher still checks existence before the potentially expensive summary scan, then revalidates the Job row after the last await and immediately before the synchronous renderer broadcast. A confirmed null row now drops the callback; transient lookup failures retain the existing fail-open publication behavior. The regression returned a Job for the first lookup and null for the post-summary lookup: it failed before the fix and now passes. Validation: Compute IPC 54/54, complete affected set 7 files / 270 tests, Node typecheck, zero-warning ESLint, git diff --check, and the existing 660-line Compute IPC architecture gate.

PR #1046, opened 2026-08-10
fix(composer): move attachment limits into tooltip

ewen-poch:
## Problem

The persistent attachment-limit copy makes the composer action menu feel visually abrupt and gives Attach files disproportionate vertical weight.

## Proposed change

- Place a small CircleHelp affordance beside Attach files.
- Show the existing file type, 10 GB limit, and large-file behavior in a tooltip on hover or keyboard focus.
- Remove the persistent helper block below the menu item.

## Scope and non-goals

This changes only composer menu presentation and its focused interaction test. Upload selection, validation, size limits, and attachment handling are unchanged. The WorkspacePage consumer interface is unchanged, so no additional consumer test is required. The markup has no platform-specific behavior, so platform lanes are excluded from the local Test Impact Set.

- Attachment guidance presentation -> npm test -- src/renderer/src/pages/workspace/ConversationPanel.interaction.test.tsx -> passed, 77 tests.
- Source lint -> npm run lint -> passed with 0 errors; 17 pre-existing warnings remain in unrelated files.

The full portable suite was excluded because CodeGraph identified ConversationPanel as the sole owner, WorkspacePage as its production consumer, the public props interface did not change, and the project-owned interaction test directly exercises the changed behavior.

## Review focus

Please verify that the tooltip composition preserves DropdownMenuItem selection behavior and that the compact help affordance reads clearly in both themes. Local visual E2E was not run; tooltip collision placement remains covered by the shared Radix primitive rather than this focused test.

### DIFF
--- a/src/renderer/src/pages/workspace/ConversationPanel.interaction.test.tsx
+++ b/src/renderer/src/pages/workspace/ConversationPanel.interaction.test.tsx
@@ -655,19 +655,27 @@ describe('ConversationPanel composer intake', () => {
     })
   })
 
-  it('moves file type and per-file size guidance into the Attach files tooltip', () => {
+  it('keeps attachment limits discoverable in the tooltip and touch fallback', () => {
     renderPanel()
 
+    const guidance = 'Any file type · 10 GB per file. Large files are linked, not embedded.'
     expect(container.querySelector('[data-testid="menu-attach-files"]')?.textContent).toBe(
       'Attach files'
     )
     expect(
       [...container.querySelectorAll('[data-testid="tooltip-content"]')].some(
-        (node) =>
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the touch discoverability concern in 84204c27. Coarse-pointer layouts now render compact visible attachment-limit guidance, and the same fallback is shown whenever Attach files is disabled; enabled pointer and keyboard layouts retain the existing tooltip-only presentation. The tooltip and fallback share one source string. Updated interaction coverage verifies the tooltip content, coarse-pointer display class, and disabled-state visibility. Validation passed: 77 targeted tests, web typecheck, and lint with 0 errors.

PR #1047, opened 2026-08-10
fix(acp): preserve restored continuation provenance

ewen-poch:
## Problem

After an app restart, persisted permission and elicitation continuations could resume with only the originating prompt id. The resumed ACP runtime then synthesized Frame, Branch, and Runtime Segment identities from the provider session instead of the durable Session graph. Artifact writes could therefore complete under a current-run root that the strict finalizer correctly rejected as unrelated to the durable Session graph.

An out-of-order Main permission lifecycle event could also project an older pending runtime-context revision over a newer settled revision. The permission was accepted, but the old card remained visible; clicking it again then correctly reached Main as a stale restored request. Independently, a stale renderer callback could resubmit an already accepted restored permission after the first IPC call settled.

## Proposed change

- Restore permission and elicitation continuations through a Main-owned context owner that loads the durable Session and derives the complete provenance context and history replay.
- Validate pending and revised elicitation authority against the active Frame, visible Branch ancestry, prompt, Runtime Segment, durable request identity, and flat projection.
- Persist and publish Main-owned revision prompts before resuming the provider so later renderer artifact events extend the same Session graph.
- Keep strict Artifact finalization unchanged and validate restored ownership against the captured current-run context.
- Coalesce concurrent and stale permission responses after success, while releasing failed attempts and explicit Main re-arms of the same request id.
- Guard the renderer fallback projection by runtime-context revision so an IPC result cannot overwrite a newer Main-owned rearm.
- Reject Main permission-authority projections whose runtime-context revision is older than the renderer's current revision, so a late pending snapshot cannot resurrect a settled card.
- Preserve the quit boundary: only durable waits detached by actual shutdown are converted to cancellation.

```mermaid
sequenceDiagram
  participant R as Renderer
  participant M as Main ACP runtime
  participant S as Durable Session
  participant P as Provider
  R->>M: restored permission or elicitation response
  M->>S: load and validate graph authority
  M->>S: persist continuing or revision prompt
  M-->>R: publish authoritative Session revision
  M->>P: continue with durable provenance
  P-->>M: complete or fail
  M->>S: settle or rearm pending authority
  M-->>R: publish newer durable revision
  Note over R,M: older permission revisions are ignored
```

## Scope and non-goals

- The Artifact finalizer remains fail-closed; this does not weaken graph ownership checks.
- Existing artifact runs already persisted with mismatched provenance are not silently rewritten.
- This does not reinterpret or accept invalid `write_artifact_file` arguments. Provider tool calls that fail the `source` schema remain separate input-validation failures.
- No provider protocol or permission-grant semantics are changed.

Checks against the final HEAD:

- Restored permission lifecycle and response-race coverage, including a settled revision followed by a stale pending replay -> 175 targeted tests passed.
- Full Vitest run -> 13,093 tests passed; 8 unrelated tests timed out under full parallel load. All 5 affected files were rerun with one worker and passed 594/594.
- Main, preload, shared, and renderer type contracts -> `npm run typecheck` -> passed.
- Repository source and test standards -> `npm run lint` -> passed with 0 errors (17 pre-existing warnings).
- Formatting and patch integrity -> Prettier on changed files and `git diff --check` -> passed.
- Independent Spec and Standards reviews of the final HEAD -> clean, with no actionable findings.

The Test Impact classifier selected full fallback because an ACP test path has no known module owner. Packaged/manual restart behavior remains an external verification risk; deterministic restart, stale-projection, and response-race paths are covered at the runtime and renderer seams.

## Review focus

- Durable Session graph authority versus provider-session identity during restored continuations.
- Ordering of Main-owned revision publication before hidden provider continuation.
- Permission response attempt lifetime: stale-success suppression, direct-failure retry, explicit rearm retry, and revision protection against a fast Main rearm.
- Monotonic permission-authority projection: an older pending revision must not replace a newer settled revision.

### DIFF
--- a/src/main/acp/application-commands.test.ts
+++ b/src/main/acp/application-commands.test.ts
@@ -50,7 +50,7 @@ const createDependencies = (): AcpApplicationCommandDependencies => ({
     cancelPrompt: vi.fn(async () => snapshot),
     deleteSession: vi.fn(async () => snapshot),
     respondToPermission: vi.fn(async () => snapshot),
-    respondToElicitation: vi.fn(() => snapshot),
+    respondToElicitation: vi.fn(async () => snapshot),
     getSessionPlanProjection: vi.fn(async () => null),
     respondSessionPlan: vi.fn(async () => ({ projection: {} as never, changed: true })),
     setPermissionProfile: vi.fn(async () => snapshot),
--- a/src/main/acp/client-interaction-owner.test.ts
+++ b/src/main/acp/client-interaction-owner.test.ts
@@ -1,0 +1,184 @@
+import type {
+  CreateElicitationRequest,
+  CreateElicitationResponse,
+  RequestPermissionRequest,
+  RequestPermissionResponse
+} from '@agentclientprotocol/sdk'
+import { describe, expect, it, vi } from 'vitest'
+
+import type { AgentFrameworkId } from '../../shared/settings'
+import { AcpClientInteractionOwner } from './client-interaction-owner'
+
+const elicitationRequest = (): CreateElicitationRequest => ({
+  mode: 'form',
+  sessionId: 'provider-session',
+  toolCallId: 'ask-1',
+  message: 'Choose an approach',
+  requestedSchema: {
+    type: 'object',
+    properties: {
+      question_0: { type: 'string', enum: ['minimal', 'expanded'] }
+    }
+  }
+})
+
+const permissionRequest = (): RequestPermissionRequest => ({
+  sessionId: 'provider-session',
+  toolCall: {
+    toolCallId: 'tool-1',
+    title: 'Run command',
+    kind: 'execute',
+    status: 'pending'
+  },
+  options: [{ optionId: 'allow-once', name: 'Allow once', kind: 'allow_once' }]
+})
+
+const codexApprovalRequest = (): CreateElicitationRequest => ({
+  ...elicitationRequest(),
+  toolCallId: 'mcp-tool-1',
+  message: 'Allow this MCP tool?',
+  _meta: { codex_approval_kind: 'mcp_tool_call' }
+})
+
+const createOwner = (
+  options: {
+    frameworkId?: AgentFrameworkId
+    hasTrustedCodexMcpToolCall?: boolean
+    consumeTrustedCodexMcpToolCall?: boolean
+  } = {}
+): {
+  owner: AcpClientInteractionOwner
+  requestElicitation: ReturnType<typeof vi.fn>
+  requestPermission: ReturnType<typeof vi.fn>
+  consumeTrustedCodexMcpToolCall: ReturnType<typeof vi.fn>
+  hasTrustedCodexMcpToolCall: ReturnType<typeof vi.fn>
+} => {
+  const requestElicitation = vi
+    .fn<(request: CreateElicitationRequest) => Promise<CreateElicitationResponse>>()
+    .mockResolvedValue({ action: 'accept', content: { question_0: 'minimal' } })
+  const requestPermission = vi
+    .fn<(request: RequestPermissionRequest) => Promise<RequestPermissionResponse>>()
+    .mockImplementation(async (request) => ({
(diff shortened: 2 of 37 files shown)


ewen-poch:
Addressed the permission-card dismissal timing in 8234ff07. The workspace runtime now removes a request from the visible permission queue as soon as its response starts, while retaining the existing response coalescing and stale-response protection. If the IPC response fails, the request becomes visible again for retry; successful responses remain hidden while the authoritative snapshot settles. This does not change the permission protocol, durable authority model, or UI component contract.

Validation: 176 targeted permission/lifecycle tests passed, the full Vitest suite passed (13,102 tests), typecheck passed, and lint completed with 0 errors.

ewen-poch:
Introduced AcpClientInteractionOwner as the single semantic ingress for ACP client interactions. Ordinary elicitation from Claude Code, OpenCode, and Codex remains structured user input, while provider permission requests and trusted Codex MCP approval elicitations route to the existing permission owner. Trusted ask_user_question approvals are still accepted silently; untrusted or cross-framework Codex approval metadata now fails closed. Durable storage remains unchanged: Permission stays at runtimeContext.permission and Ask User stays at activity.elicitation.\n\nValidation: 549 ACP interaction/runtime tests passed; ownership tests passed; full typecheck passed; lint completed with 0 errors and 17 pre-existing warnings. The full suite passed 13,109 tests, with one unrelated concurrent architecture-scan timeout that passed independently (2/2).

ewen-poch:
Implemented the restored-permission replay hardening:

- Main now publishes the exact `permissionRequestId` on restored-permission lifecycle events; Renderer no longer infers correlation from mutable Session state.
- Accepted restored responses remain coalesced and hidden through delayed pending projections. Only a new matching `REARMED` lifecycle event can release the one-shot request.
- Main publishes a transient coordinator-wide snapshot revision. A single Renderer revision owner rejects older push, action, initial-pull, and quit-persistence snapshots before stale lifecycle events can mutate or persist Session state.
- The persistent Session schema is unchanged.

Validation: `npm run typecheck`; `npm run lint` (0 errors, 17 pre-existing warnings); targeted permission/runtime tests (259 passed); full `npm test -- --run` suite passed. Final Standards and Spec reviews are clean.

ewen-poch:
Addressed the elicitation revision persistence race. After the Renderer creates the revised Session branch, it now awaits the existing ordered Session persistence queue before invoking respondToElicitation, so Main validates and appends against the durable fork. The existing rollback path remains active if persistence or IPC fails.\n\nAdded a deterministic deferred-persistence test proving that continuation IPC is not sent before the fork is durable and verifying reset -> persistence -> continuation ordering.\n\nThe branch is rebased onto the latest origin/main. To satisfy the new workspace-runtime completion gate introduced on main, the permission response attempt/tombstone state machine was moved verbatim into a private owner module; exact permissionRequestId correlation, replay suppression, terminal retention, cleanup, persistent formats, and user interaction remain unchanged.\n\nValidation: the affected ACP suite passed 823/823 before the behavior-preserving extraction; the post-extraction workspace runtime tests passed 154/154 and the architecture suite passed 11/11 with a single worker. Typecheck and targeted lint pass. A full local run reached 13,170 passed and 190 skipped; six tests hit timeout/scheduling limits while several other worktrees were running Vitest/TypeScript concurrently, including unrelated notebook/R tests. The deterministic architecture gate exposed by the rebase was fixed and then passed independently.

ewen-poch:
Fixed the macOS Module tests failure. The new permission response attempt owner imports the public ChatSession type, so it is a direct production consumer of the session-store facade and must be listed in the explicit consumer contract. Added the owner path to that contract; no production behavior changed.\n\nValidation: src/renderer/src/stores/session-store.test.ts passed 121/121.

ewen-poch:
Rebased the branch again after origin/main advanced by two workspace composer commits. The rebase completed without conflicts. The permission and elicitation behavior remains unchanged on the new base.\n\nPost-rebase validation: 10 affected ACP, workspace-runtime architecture, and session-store contract test files passed 955/955; full typecheck passed; CodeGraph is synchronized and git diff --check is clean.

PR #1048, opened 2026-08-10
fix(uploads): restore missing immutable upload bytes

ewen-poch:
## Problem

Startup reconciliation can repeatedly fail in `reconcile-derived-state` when SQLite has a ready Upload Version and the deterministic legacy upload still exists, but the Version's immutable `content` file is missing. The raw `ENOENT` keeps storage recovery incomplete, so Retry cannot clear the saved-conversations banner.

## Proposed change

- Treat only a missing ready-Version content file as recoverable; existing corrupt content still fails.
- Verify the deterministic legacy source against Session/Project/Upload/Version authority, regular-file identity, size, SHA-256, and a stable snapshot.
- Publish verified bytes atomically with a hard link before the existing private-claim cleanup.
- Keep checksum mismatches and source races fail-closed, and remove a recovered hard link if its shared inode is mutated during cleanup.
- Add a production-path regression covering mismatch retention, successful recovery, idempotence, and in-place mutation.

## Scope and non-goals

- No schema, renderer, or user-facing copy changes.
- Does not guess from non-deterministic paths or delete unverifiable legacy bytes.
- Does not change handling for a present but corrupt immutable Version file.

- [x] The reported path-free ready Upload state no longer throws `ENOENT` when a checksum-valid legacy copy exists.
- [x] A checksum mismatch does not publish immutable bytes or delete the legacy file.
- [x] A cleanup-time in-place mutation does not leave corrupted immutable bytes.
- [x] Retry is idempotent after successful recovery.
- [x] Node and web TypeScript checks pass.
- [x] ESLint passes for the changed files.
- [x] Declared Test Impact Set passes: 14 files, 304 tests passed, 1 existing Windows symlink-permission test excluded.
- [x] Full Vitest suite executed after the final edit: 12,962 passed, 275 skipped, 34 failed in unrelated/environment-sensitive tests. The only Upload failure is the same existing symlink test (`EPERM` creating a symlink); the new regression passes in the full run.

## Review focus

- The authority and source-verification gates before hard-link publication.
- The inode-specific rollback when cleanup observes a mutated hard-linked source.
- Fail-closed behavior for checksum mismatches and races.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -45,6 +45,7 @@ files:
   - '!node_modules/pdfjs-dist/**/*.map'
 asarUnpack:
   - resources/**
+  - node_modules/@aipoch/safe-file-publisher-native/build/Release/*.node
   # The ACP agent runs as a spawned child process, so its entry must live outside the asar.
   - node_modules/@agentclientprotocol/claude-agent-acp/**
 # The Prisma generated client (`node_modules/.prisma`) is produced by `prisma generate` and is NOT a
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
       "dependencies": {
         "@agentclientprotocol/claude-agent-acp": "^0.60.0",
         "@agentclientprotocol/sdk": "^1.2.1",
+        "@aipoch/safe-file-publisher-native": "file:packages/safe-file-publisher-native",
         "@anthropic-ai/tokenizer": "^0.0.4",
         "@bokuweb/zstd-wasm": "^0.0.27",
         "@electron-toolkit/preload": "^3.0.2",
@@ -140,6 +141,10 @@
           "optional": true
         }
       }
+    },
+    "node_modules/@aipoch/safe-file-publisher-native": {
+      "resolved": "packages/safe-file-publisher-native",
+      "link": true
     },
     "node_modules/@ant-design/colors": {
       "version": "8.0.1",
@@ -22303,6 +22308,11 @@
         "type": "github",
         "url": "https://github.com/sponsors/wooorm"
       }
+    },
+    "packages/safe-file-publisher-native": {
+      "name": "@aipoch/safe-file-publisher-native",
+      "version": "0.0.0",
+      "hasInstallScript": true
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   "dependencies": {
     "@agentclientprotocol/claude-agent-acp": "^0.60.0",
     "@agentclientprotocol/sdk": "^1.2.1",
+    "@aipoch/safe-file-publisher-native": "file:packages/safe-file-publisher-native",
     "@anthropic-ai/tokenizer": "^0.0.4",
     "@bokuweb/zstd-wasm": "^0.0.27",
     "@electron-toolkit/preload": "^3.0.2",
--- a/packages/safe-file-publisher-native/binding.gyp
+++ b/packages/safe-file-publisher-native/binding.gyp
@@ -1,0 +1,21 @@
+{
+  "targets": [
+    {
+      "target_name": "safe_file_publisher_native",
+      "sources": ["src/safe_file_publisher_native.cc"],
+      "defines": ["NAPI_VERSION=8"],
+      "conditions": [
+        ["OS=='win'", {
+          "msvs_settings": {
+            "VCCLCompilerTool": {
+              "AdditionalOptions": ["/std:c++17"]
+            }
+          }
+        }],
+        ["OS!='win'", {
+          "cflags_cc": ["-std=c++17"]
+        }]
+      ]
+    }
+  ]
+}
--- a/packages/safe-file-publisher-native/index.cjs
+++ b/packages/safe-file-publisher-native/index.cjs
@@ -1,0 +1,4 @@
+'use strict'
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports -- Native Node-API bindings load through CommonJS.
+module.exports = require('./build/Release/safe_file_publisher_native.node')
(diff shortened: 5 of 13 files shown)


ewen-poch:
Addressed both findings in 7371da1e:

- Hard-link compatibility: unsupported link errors now fall back to an exclusive sibling temp copy. The copy and source snapshot are verified before atomic rename, and a valid crash leftover can be resumed safely.
- Parent containment: recovery now walks the final parent chain with `lstat`, rejects symlink and non-directory components, creates missing directories one level at a time, and verifies canonical containment before publication.

Added regression coverage for forced `EPERM` fallback and a symlinked/junction final parent. The declared impact set passes (304 passed, 1 known local symlink-permission test skipped), with Node typecheck and ESLint passing.

ewen-poch:
Addressed the stale-temp liveness issue in 014936de:

- Each copy attempt now uses a UUID-scoped sibling temp path with exclusive creation, so a truncated temp left by a crashed attempt cannot block a later recovery.
- Normal copy/validation failures clean up only the temp inode owned by that attempt; checksum and source-stability verification still run before atomic publication.
- Added regression coverage showing that a pre-existing truncated recovery temp does not prevent the ready Version from being restored.

The focused regression, Node typecheck, ESLint, and the declared impact set all pass (304 passed, 1 known local symlink-permission test skipped).

ewen-poch:
Addressed the raced-final/parent-swap P1 in `a178245d`:

- Added a narrow native Node-API publication adapter that atomically refuses an existing destination.
- Linux/macOS anchor the storage root and walk each parent component by descriptor with symlink following disabled before using `renameat2(RENAME_NOREPLACE)` / `renameatx_np(RENAME_EXCL)`.
- Windows holds non-reparse root and parent handles, verifies the parent handle remains inside the root, and publishes with `SetFileInformationByHandle` using `ReplaceIfExists = FALSE`.
- A raced destination now remains untouched and the legacy source remains retryable; a swapped parent fails closed without publishing through it.
- Added deterministic regression coverage for both a concurrent final file and a parent-to-junction/symlink swap. The affected suite passes (304 passed, 3 skipped), along with node typecheck, ESLint, and the electron-builder config tests.

The native binding could not be compiled locally because this Windows host does not have the MSVC C++ workload. The native tests are included and will run after CI installs/builds the binding on each platform.

ewen-poch:
Addressed the non-adversarial round-4 feedback in e3902773:\n\n- Recovery retries reclaim only UUID-named regular recovery temp files older than 24 hours, and recheck the file snapshot before removal.\n- Fresh recovery attempts are preserved; the regression test covers stale removal and fresh preservation.\n- The native CommonJS loader now uses a .cjs extension with a narrow lint exception, fixing the Static checks failure.\n\nThe two P1 findings assume that another same-user/local process intentionally replaces already-validated pathnames or inodes during the syscall window. That adversarial filesystem threat model is out of scope for this recovery fix, so this PR does not expand the native transaction/read architecture further.\n\nValidation: ESLint, Prettier, Node typecheck, the focused recovery regression, electron-builder config tests, and the 304-test impact set pass. The full suite completed with 12,976 passing and 20 known local Windows failures caused by symlink/ACL permissions, Windows path-separator expectations, and suite-load timeouts.

ewen-poch:
Final-round disposition:\n\n- The hard-link parent-swap finding remains out of scope because it requires an adversarial same-user/local process to replace the validated directory during the syscall window.\n- The renameat2 capability finding is a legitimate portability trade-off for older Linux kernels and filesystems that do not support no-replace rename. A safe fallback needs additional native publication design and validation. Per the agreed five-round review cap, I am pausing without a sixth implementation round rather than adding an unreviewed fallback.

ewen-poch:
Final CI diagnosis: the only macOS Module-test failure is `atomic-no-replace-publisher.test.ts` expecting `ELOOP` for a symlinked publication parent while the macOS native adapter returns `EIO` with `Could not anchor the publication parent.` The operation still fails closed and does not publish through the symlink. Static checks, Windows core/E2E, macOS build/E2E, CodeQL, and the other gates pass. Per the five-round cap, I recorded this platform-specific assertion mismatch without another code push.

ewen-poch:
Addressed the actionable round-5 portability and CI findings in `900883ae`:

- The symlink-parent regression now asserts the cross-platform contract (publication throws and no destination is created) instead of requiring macOS to normalize its native failure to `ELOOP`.
- Linux now falls back from unavailable `renameat2(RENAME_NOREPLACE)` support (`ENOSYS`, `EOPNOTSUPP`, or capability `EINVAL`) to same-directory `linkat` publication followed by best-effort source unlink. `linkat` atomically refuses an existing destination.
- If neither native no-replace mechanism is supported, the binding reports `ENOTSUP`; recovery returns an unsafe residual, preserves the verified legacy source, cleans its owned temporary, and does not abort the overall reconciliation workflow.
- Added regression coverage for the unsupported-publication outcome.

The hard-link parent-swap finding remains intentionally out of scope because it requires an adversarial same-user/local process to replace a validated directory during the syscall window.

Validation: ESLint, Prettier, Node typecheck, the focused regressions, electron-builder config tests (5 passed), and the declared impact set (304 passed, 3 skipped) pass. The full suite ran with 12,951 passing and 45 environment/load-sensitive Windows failures; the one additional Upload timeout passes in isolation.

ewen-poch:
Round disposition for 900883ae:

- Both remaining P1 findings require an adversarial same-user or local process to replace already validated pathnames or directories during syscall windows.
- Per the agreed threat model, those pathname-swap attacks are out of scope, so no further native architecture expansion is planned in this PR.
- The actionable portability and macOS CI findings are addressed, and all required CI checks, including PR Gate, now pass.

PR #1077, opened 2026-08-10
fix(database): add a versioned migration startup gate

ewen-poch:
## Problem

The packaged Electron application maintained its SQLite schema through runtime DDL and column guards without a durable migration history. That made schema ownership diffuse, complicated legacy adoption, and allowed database compatibility failures to surface only after database-dependent runtime composition had begun.

## Changes

- add an application-owned, forward-only migration ledger with an immutable `0001` baseline, checksum validation, classified failures, and legacy database adoption;
- generate canonical SQLite target SQL from Prisma plus the explicit CHECK-constraint contract, and keep unsafe SQL behind the private migration executor;
- gate desktop and headless startup on database verification, keeping the loading shell available for checking, migration, retryable failures, incompatible history, and quit;
- certify fresh install, legacy adoption, reopen, downgrade blocking, path handling, native-engine presence, and the effective `sqlite_version()` across packaged release lanes;
- document the schema-change workflow in `CONTRIBUTING.md`.

This PR does not add Project/Review/ComputeJob foreign keys or change their deletion semantics. Those remain separate migrations after their ownership policies are approved.

## User impact

Compatible databases are checked or upgraded while the loading surface is shown. A newer database, invalid migration history, modified checksum, missing native engine, or failed upgrade remains blocked on that surface; database-dependent IPC and the workspace do not start. No new setting or normal workspace interaction is introduced.

Downgrades are forward-only from the first ledger-aware release: an older ledger-aware build detects unknown later migrations and blocks instead of opening the database.

## Verification

Final compressed head:

- `npm run db:schema:check`
- `npm run typecheck`
- focused ledger/startup/release suite: 9 files, 128 tests passed
- final tree equivalence check against the fully tested pre-compression head

Identical final tree before history compression:

- `npm run lint`: 0 errors, 17 pre-existing warnings
- `npm test -- --maxWorkers=4`: 923 files passed, 15 skipped; 13,265 tests passed, 192 skipped

## Risks and follow-ups

- Native macOS ARM64/x64, Windows x64, and Linux x64 release runners remain authoritative for installed-package certification JSON.
- Installed-process read-only, locked, and disk-full fault injection can be added if release policy requires proof beyond classifier and startup-gate tests.
- Project-owned Review/preview relations and ComputeJob ownership remain separately reviewed follow-up migrations.


ewen-poch:
Addressed. Quit requests received during an active database attempt are now latched. A blocked startup can still exit directly, while a verified startup releases a pending quit only after the application lifecycle is installed, so backend cleanup runs through the normal shutdown owner. Added coverage for both verified and blocked paths.\n\nAlso fixed the PR Gate findings: the schema generator now resolves Prettier configuration from the generated target path for platform-stable drift output, and the Electron fixture polls the database-ready IPC from the test process before invoking runtime handlers.

PR #1078, opened 2026-08-10
ci(pr-gate): reduce cross-platform runner contention

ewen-poch:
## Problem

CI wall time is dominated by scarce hosted-runner capacity rather than a lack of macOS shards. Recent full PR samples showed macOS coverage merge work waiting behind the five-runner macOS cap, while the advisory Windows full-suite workflow ran after every main push and frequently overlapped newer work.

## Proposed change

- add a focused, no-side-effect `unit-coverage` dispatch mode to PR Gate for validating the real macOS shard and coverage topology
- keep the two full macOS test shards, but move full report download, merge, coverage upload, and threshold enforcement to Ubuntu
- make the cross-platform Vitest blob merge tolerate the absence of locally discoverable Ubuntu test paths while preserving coverage threshold failures
- batch Windows full-suite coverage hourly at minute 47
- skip an hourly Windows run when the latest successful workflow already covered the current main SHA; manual dispatches always run
- retain the existing three Windows shards after a four-shard experiment showed no material wall-clock improvement

## Scope and non-goals

- Changes are limited to two GitHub Actions workflows and their topology tests.
- No product source code or runtime behavior changes.
- No third macOS shard is added because observed macOS capacity serialized existing shards.
- No Windows test failures are suppressed; the workflow remains advisory and fail-closed when it runs.
- The known atomic publisher `EIO` failure is not fixed here.

- Focused PR Gate dry-run: [run 31383567512](https://github.com/aipoch/open-science/actions/runs/31383567512) passed on commit `597b6137`.
  - macOS shards passed in 5m34s and 3m43s.
  - Ubuntu `Module tests and coverage` passed in 1m24s, including artifact download, cross-platform merge, coverage thresholds, upload, and final deterministic gate.
  - The run observed the two macOS shards execute serially because of runner availability, confirming that another macOS shard would add queue pressure.
- Windows four-shard experiment: [run 31384719168](https://github.com/aipoch/open-science/actions/runs/31384719168).
  - job times were 15.47m, 10.03m, 10.60m, and 9.60m; test-step times were 11.62m, 6.42m, 6.85m, and 6.18m.
  - the 3/4 shard hit the pre-existing atomic publisher `EIO`.
  - the 15.47m long tail did not improve on the historical three-shard baseline, so the final workflow retains three shards.
- `npm run lint`: passed with 0 errors and 17 warnings in unchanged files.
- Affected workflow tests: 5 files and 77 tests passed.
- Prettier and `git diff --check`: passed.
- Full local `npm test` was attempted: 903 files passed and 15 failed (41 failed tests), with Windows-local symlink/ACL/file-lock failures and load-related timeouts in untouched suites. Exact-head GitHub Actions remains the authoritative full validation.

## Review focus

- Verify that full plans use Ubuntu only for report merge and threshold enforcement, while selective affected-module tests remain on macOS.
- Verify that the focused dispatch cannot select release or platform E2E bundles.
- Verify that the hourly Windows planner fails open on API lookup errors, skips only an already-successful main SHA, and never skips manual dispatches.
- Verify that the final matrix and command remain three-way sharded.
## Merge prerequisite

- CI Integrity intentionally rejects changes to the protected PR Gate control plane. An explicit maintainer ruleset bypass is required before merge; this PR does not weaken or bypass that guard.

### DIFF
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -504,7 +504,9 @@ jobs:
         # the unit bundle without unit_macos. Merge once so thresholds see the complete suite.
         if: ${{ needs.unit_shard.result != 'skipped' }}
         continue-on-error: true
-        run: npx vitest run --merge-reports=vitest-reports --coverage
+        # macOS blob reports retain their source test paths. Ubuntu can merge their results and
+        # coverage, but local discovery finds no tests; keep coverage thresholds authoritative.
+        run: npx vitest run --merge-reports=vitest-reports --coverage --passWithNoTests
       - name: Upload Module coverage report
         if: ${{ always() && (steps.unit_macos_related.outcome != 'skipped' || steps.unit_macos_full.outcome != 'skipped') }}
         continue-on-error: true
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -7,6 +7,15 @@ on:
   merge_group:
     types: [checks_requested]
   workflow_dispatch:
+    inputs:
+      dry_run:
+        description: Focused no-side-effect validation plan
+        required: false
+        default: classified
+        type: choice
+        options:
+          - classified
+          - unit-coverage
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


ewen-poch:
Regarding the protected control-plane finding: this PR intentionally changes `.github/workflows/pr-gate.yml` to move the coverage merge from macOS to Ubuntu and to add the focused no-side-effect dry-run. Moving the implementation to a non-protected path would not update the trusted gate and would therefore conflict with the intended change. The CI Integrity failure is expected and will not be weakened or bypassed in code. The PR documents the required explicit maintainer ruleset bypass under **Merge prerequisite**. No code change is appropriate for this finding.

PR #1084, opened 2026-08-10
fix(workspace): persist choices only after next

ewen-poch:
## Problem

Selecting an option in a multi-question Ask User card persisted that answer immediately. Draft restoration treats persisted answers as completed steps, so a projection refresh could advance to the next question before the user clicked Next.

## Proposed change

Keep the current question answer local while the user selects or edits it. Persist completed choice answers only through the existing Next submit path.

## Scope and non-goals

This changes only the Ask User choice confirmation boundary and its interaction regression test. It does not change card styling, final submission, or restored draft behavior.

- WorkspaceElicitationCard interaction tests: 14 passed.
- ConversationPanel interaction tests: 77 passed.
- Web and Node typechecks: passed.
- Lint: passed with 0 errors and 17 pre-existing warnings.
- Full portable Vitest: 13,282 passed and 193 skipped. Seven resource-contention timeouts passed when rerun at low concurrency. One unrelated existing failure remains in project prisma client: PermissionGrantSeed runtime DDL is missing id and appliedAt.

Uncovered risk: the complete local suite is not fully green because of that existing Prisma schema baseline failure; exact-head CI remains authoritative.

## Review focus

Confirm that option clicks no longer invoke the draft completion callback and that Next remains the only non-final action that persists and advances a question.

### DIFF
--- a/src/renderer/src/pages/workspace/WorkspaceElicitationCard.interaction.test.tsx
+++ b/src/renderer/src/pages/workspace/WorkspaceElicitationCard.interaction.test.tsx
@@ -236,9 +236,7 @@ describe('WorkspaceElicitationCard choice question', () => {
     })
 
     expect(onRespond).not.toHaveBeenCalled()
-    expect(onDraftChange).toHaveBeenLastCalledWith([
-      { fieldId: 'question_0', value: 'multi-omics' }
-    ])
+    expect(onDraftChange).not.toHaveBeenCalled()
     expect(container.querySelector('h3')?.textContent).toBe(
       'What should this skill primarily cover?'
     )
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the P1 in cc8e63b8. WorkspaceElicitationCard now keeps a confirmed-values snapshot and updates only the active question when Next is submitted, so a later selection made before Back is not persisted. The regression coverage exercises Q2 select → Back → Q1 Next and asserts that the draft still contains only Q1. Verified with 14/14 card interaction tests, 77/77 ConversationPanel interaction tests, Web typecheck, and Prettier.

PR #1085, opened 2026-08-10
feat(connectors): surface runtime status and recovery

ewen-poch:
## Problem

Enabled custom MCP connectors could fail during startup or skill discovery while Settings still presented them as connected. Expired OAuth credentials and unavailable stdio commands also lacked a clear, recoverable interaction path.

## Proposed change

- Project runtime discovery status into Connector settings without persisting transient failures.
- Distinguish Connected, Disabled, Unavailable, Sign-in required, and Checking states.
- Offer Retry for unavailable connectors, Sign in again for OAuth failures, and Configure when authentication is not configured.
- Make Retry authoritative by closing the cached MCP client, clearing the failure gate, invalidating permissions, rediscovering Connector skills, and returning the refreshed settings snapshot.
- Add the retry command across the Electron application-command, IPC, preload, renderer, and generated contract surfaces as a local-only operation.
- Keep startup non-blocking and avoid exposing raw connector errors to the renderer.

## Scope and non-goals

- Durable enabled state remains separate from transient runtime availability.
- Disabling a connector suppresses transient runtime failures immediately; structural configuration failures remain visible.
- This does not add remote-Web retry support, global startup modals, or startup toasts.
- This does not persist raw runtime errors or expose paths, credentials, or tokens in the UI.

The final Test Impact Set ran after the last material edit and after rebasing onto the current main branch.

- Runtime status classification and settings projection -> targeted Connector and Settings tests -> passed.
- OAuth reauthentication, unavailable retry, configure fallback, and accessible waiting state -> ConnectorsPanel render and workflow tests -> passed.
- Cached-client reset and authoritative rediscovery -> Connector service and MCP client manager tests -> passed.
- Electron, IPC, preload, renderer, local-Web, and generated contract wiring -> command composition and renderer contract tests -> passed.
- Combined targeted set -> npm test with 16 explicit test files -> 16 files passed, 338 tests passed.
- Runtime process contracts -> npm run typecheck -> passed for Node and Web.
- Source validation -> npm run lint -> passed with 0 errors and 17 pre-existing warnings in untouched files.
- Generated Web API map -> npm run check:web-api-map -> passed.
- Full fallback -> npm test -> 926 files passed, 15 skipped, 1 failed; 13,348 tests passed, 193 skipped. The sole failure was a 30-second timeout in the unchanged ACP oversized skill-edit permission test.
- Timeout follow-up -> isolated ACP test rerun -> passed, with 1 passed and 505 skipped.
- Independent Standards and Spec reviews -> passed with no remaining findings.

Uncovered risk: no packaged Electron UI E2E was run, so native focus behavior is covered by renderer tests and accessibility semantics rather than a packaged-app journey. CI remains authoritative for platform lanes and the complete suite under CI resources.

## Review focus

- Failure classification boundaries between unauthenticated and unavailable states.
- Separation of durable connector configuration from transient runtime availability.
- Retry ordering and whether stale MCP clients can survive recovery.
- Local-only authority and command inventory consistency across process boundaries.
- Accessibility behavior of the OAuth waiting toggle and recovery controls.


ewen-poch:
Addressed the late runtime failure propagation issue. ConnectorSettingsModule.listConnectors now waits for the runtime projection refresh that is already in flight before building the renderer snapshot, so the first mounted Settings view receives Unavailable or Sign-in required instead of a transient false Connected state. The fix reuses the existing snapshot path and does not add a new event channel or persisted state. Added deterministic regression coverage for the late-settle race and for waiting on the projection refresh queue. Validation: 4 affected test files passed with 76 tests, Node and Web typechecks passed, and lint completed with 0 errors.

ewen-poch:
Addressed the post-mutation runtime status gap while preserving the existing fire-and-forget mutation contract. Connector refresh completion now publishes a signal-only `settings:connector-runtime-changed` application event. Each renderer Connector store subscribes once and re-reads `listConnectors()`, so enable/update/authenticate failures reconcile to Unavailable or Sign-in required without blocking the mutation or duplicating snapshot payloads across the event interface. Retry also notifies other open windows. Added deterministic store and workflow ordering coverage and updated Electron/Web contract inventories. Validation: 19 targeted test files passed with 350 tests; Node and Web typechecks passed; lint completed with 0 errors; generated Web API map check passed.

ewen-poch:
Addressed both runtime-status findings. `listConnectors()` no longer awaits the MCP refresh queue; the projection exposes a transient checking state and publishes the existing runtime-changed signal when refresh begins and settles, so Settings opens immediately, shows Checking…, and re-reads the final authoritative status. `ConnectorService` now reports classified custom-server dispatch failures and recoveries through a narrow callback into the same projection. Dispatch and discovery failures are stored separately before being combined, preventing one source from incorrectly clearing the other during concurrent work. No runtime status is persisted and no transport details reach the renderer. Also completed the event-interface test mocks that caused the previous Module test job to fail. Validation: full Vitest suite passed with 927 files / 13,354 tests (15 files / 193 tests skipped); Node and Web typechecks passed; lint completed with 0 errors; formatting and generated Web API map checks passed.

ewen-poch:
Updated the host.agents connector catalog to consume the same runtime availability resolver as Settings. list_connectors now reports known runtime unavailable/unauthenticated states, so the existing create/update/attach gate rejects new attachments before ProfileService is called; detach remains unchanged. Stored configuration still owns mainEnabled, and no runtime state is persisted. Added service-level regression coverage for both the public read model and the attachment gate. Full suite: 927 files passed (15 skipped), 13,356 tests passed (193 skipped).

ewen-poch:
Addressed the connector reconciliation race. All snapshot-producing loads and mutations now share a monotonic generation at the existing renderer reconciliation seam; a response commits only when no newer reconciliation has started. This prevents an older runtime-event listConnectors response from overwriting a newer authoritative mutation result without serializing or delaying mutations. Added a deterministic out-of-order completion regression test. Full suite: 927 files passed (15 skipped), 13,357 tests passed (193 skipped); Node/Web typechecks passed and lint completed with 0 errors.

ewen-poch:
Rebased the branch onto current origin/main (c1f94e2c). Git applied all eight existing commits without content conflicts. Post-rebase validation identified upstream renderer-contract inventory drift, so the two existing inventory tests now pin 341 Electron paths and 245 invoke channels. No production behavior, architecture, data model, or interaction flow changed in that follow-up. Validation: full Vitest suite passed with 932 files / 13,471 tests (15 files / 193 tests skipped); Node and Web typechecks passed; lint completed with 0 errors; generated Web API map check passed.

ewen-poch:
Addressed both P2 findings:\n\n- Runtime refreshes are now coalesced while any Connector snapshot mutation is in flight. The mutation response keeps the existing generation-based authority, then one deferred listConnectors refresh reconciles runtime state after all active mutations settle. A deterministic regression test covers an event arriving before the mutation response.\n- Retry is now shown only when an unavailable custom server is still enabled. Configuration-invalid records already project as unavailable and disabled, so they retain Edit but no longer offer a recovery action that cannot succeed; runtime failures remain retryable.\n\nNo persisted state, shared contract, or architecture changed. Validation: 2 targeted files passed with 36 tests; full Vitest passed with 932 files / 13,473 tests (15 files / 193 tests skipped); Node/Web typechecks passed; lint completed with 0 errors; target formatting and generated Web API map checks passed.

ewen-poch:
Added the explicit Promise<void> return type required by the repository's no-implicit-function-return-type static policy. The earlier full-repository lint used its cache and did not surface the new-function error; a no-cache ESLint run now passes for the affected file. The connector slice regression test remains 18/18 passing, and formatting passes.

ewen-poch:
The failed macOS job was rerun once and reproduced the same unrelated visual-regression failure: e2e/visual-regression.spec.ts:318 never finds the completed-session Mark as read button, including the test's internal retry. Electron/Web builds and functional, workspace, and accessibility journeys pass; Windows E2E, Windows core, static checks, module tests, CodeQL, and the AI review all pass. The failing visual scenario and completed-session UI are outside this Connector PR's file scope, so no out-of-scope test or product change was made.

PR #1095, opened 2026-08-11
ci(database): enforce versioned schema migrations

ewen-poch:
## Problem

The application now owns a forward-only database migration ledger, but PR Gate does not reject a schema contract change that forgets its migration version. A released migration can also be edited or replaced without an immediate policy failure.

## Proposed change

Extend the existing protected PR policy check so pull requests and merge groups:

- require changes to `prisma/schema.prisma` or `prisma/sqlite-check-constraints.json` to add a migration;
- keep migration files present on the current target revision immutable;
- require new migrations to use the next contiguous `NNNN-description.ts` version and a matching `NNNN_description` id;
- require the exported migration to be imported and appended to `MIGRATION_MANIFEST` in version order;
- keep imports and complete manifest entries for released migrations immutable, including checksum and backup policy;
- ignore declarations and registrations that only appear in comments or string content.

The policy uses the merge base to identify the PR's changes, then checks released migrations and the next version against the current target revision. This rejects stale branches that try to reuse a migration version already added to the target branch.

## Scope and non-goals

This changes CI policy only. It does not change the database schema, migration runtime, generated schema, release packaging, or application behavior. Existing schema generation and migration tests remain responsible for validating DDL and upgrade behavior.

- version policy behavior and Git revision CLI -> `npx --no-install vitest run scripts/ci/check-pr-policy.test.ts` -> pass (16 tests);
- existing `0001` migration history -> `node scripts/ci/check-pr-policy.mjs` against `f44ce72b^..f44ce72b` -> pass; the historical `0002` introduction intentionally rewrote the `0001` manifest entry before this future-facing policy existed and is now correctly rejected by the stricter invariant;
- exact branch commit and migration policy after the final commit -> `node scripts/ci/check-pr-policy.mjs` with `origin/main...HEAD` revisions -> pass;
- Node and Web types -> `npm run typecheck` -> pass;
- repository lint -> `npm run lint` -> pass (0 errors; 17 pre-existing warnings);
- complete portable Vitest suite outside the loopback-restricted sandbox -> `npm test -- --maxWorkers=2` -> pass (932 files / 13,464 tests; 15 files / 193 tests skipped by suite configuration);
- formatting and whitespace -> Prettier check and `git diff --check` -> pass.

No runtime consumer or platform-specific lane was added locally because the diff only changes the global PR policy input and its tests. The full portable suite was included because global gate inputs require the full fallback. Exact-head required CI remains authoritative.

## Review focus

- Confirm that the two schema contract paths are the correct trigger boundary.
- Confirm that migration immutability and contiguous numbering are evaluated relative to the current target revision while PR changes come from the merge base.
- Confirm that comment stripping and complete manifest/import prefix comparison match the established migration convention without permitting released checksum or backup-policy changes.

Uncovered risk: `scripts/ci/check-pr-policy.mjs` is an established protected gate control-plane file, so this introducing PR is expected to require the repository's explicit maintainer ruleset bypass. After merge, the base-checked protection prevents ordinary PRs from silently weakening this policy.

### DIFF
--- a/scripts/ci/check-pr-policy.mjs
+++ b/scripts/ci/check-pr-policy.mjs
@@ -5,6 +5,8 @@ import { appendFileSync } from 'node:fs'
 import { resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 
+import { parseNameStatus } from './classify-pr-changes.mjs'
+
 const allowedTypes = [
   'feat',
   'fix',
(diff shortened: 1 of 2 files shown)


ewen-poch:
Review disposition for the current head (`101fc039`):

- The CI Integrity failure is expected because this PR changes the protected control-plane file `scripts/ci/check-pr-policy.mjs`. Per repository policy, it requires the explicit maintainer ruleset bypass; the PR cannot self-authorize that bypass.
- The no-op migration scenario is already rejected by the semantic unit lane rather than by SQL text inspection. `prisma/schema.prisma` and `prisma/sqlite-check-constraints.json` are unowned paths in the trusted classifier, so either change selects the full plan, including `unit_macos`. `src/main/database/migration-service.test.ts` then applies the real full `MIGRATION_MANIFEST` to an empty database and calls `verifyCurrentRuntimeSchema` against the generated current target. If a schema contract changes while the appended migration is ineffective, that test fails. The new policy intentionally adds the missing structural/version invariants and leaves effective DDL validation with this existing executable schema check.

Local confirmation after the final edit: targeted policy tests 16/16; complete Vitest suite 932 files / 13,464 tests passed.

ewen-poch:
Review disposition:

- Protected CI control-plane: no code change is appropriate. CI Integrity is intentionally blocking a protected policy-file edit, and the required maintainer ruleset bypass must remain external to this PR.
- No-op migrations: no code change is appropriate. A change to either authoritative schema contract selects the full test plan. scripts/generate-database-schema.test.ts first verifies that the committed generated runtime target matches the Prisma schema and SQLite CHECK contract. src/main/database/migration-service.test.ts then applies the immutable released manifest plus the appended migration to an empty database and verifies that result against that generated target. An empty or ineffective migration therefore cannot satisfy both checks. Adding a SQL-text heuristic here would be weaker and could be bypassed by syntactically non-empty but ineffective SQL.

PR #1100, opened 2026-08-11
fix(connectors): preserve connector-managed login recovery

ewen-poch:
## Problem

Custom MCP failures currently collapse two different authentication models into connector availability:

- Standard MCP OAuth is host-managed: missing tokens must stay fail-closed and direct the user to Settings > Connectors.
- Connector-managed sign-in is tool-managed: a server can return a structured `isError` such as "Not logged in. Call login first." and expose its own `login` or authentication tool.

The second case was classified as `connector_unavailable`, and the tool error was cached as a physical server failure. That cache then blocked the connector's own login tool, so Codex and the other ACP consumers could not recover even though the server was reachable.

## Proposed change

- Preserve structured MCP tool failures as a typed `McpToolCallError`.
- Classify common login/sign-in wording as `connector_unauthenticated`.
- Cache connection, discovery, and transport failures, but do not poison the whole connector after a structured tool-level error.
- Scope generated authentication guidance to the configured mode: host-managed OAuth points to Settings > Connectors, connector-managed sign-in uses an actually listed login/authentication tool, and connectors declaring neither mode receive no authentication paragraph.

## Scope and non-goals

- Standard host-managed OAuth behavior is unchanged: missing credentials still prevent contacting the server, and successful host sign-in clears cached authentication failures.
- Transport and tool-discovery failures remain fail-closed and cached.
- The implementation and tests are connector-generic; there are no connector-specific names, rules, or adapters.
- Claude Code, OpenCode, Codex Responses, and Codex Bridge continue consuming the same materialized custom-connector Skill rather than receiving framework-specific branches.

Final checks ran after the last material edit.

- Connector-managed login remains callable after a structured sign-in error -> `src/main/connectors/service.test.ts` in `npm test` -> passed.
- Structured MCP `isError` results retain their origin -> `src/main/connectors/mcp-client-manager.test.ts` in `npm test` -> passed.
- Custom Skills keep host-managed OAuth, listed login tools, and connectors without authentication guidance separate -> `src/main/connectors/skill-doc.test.ts` in `npm test` -> passed.
- Standard OAuth remains host-managed, publishes stale-token failures to Settings, and recovers after successful sign-in -> OAuth callback and connector-service tests in `npm test` -> passed.
- `npm run typecheck` -> passed for node and web.
- `npm run lint` -> passed with 17 pre-existing warnings and no warnings in modified files.
- `npm test` -> passed: 935 files passed, 15 skipped; 13,513 tests passed, 193 skipped.

The portable local suite covers the shared ACP-facing connector path. No OS-specific code changed, so Windows/Linux package lanes are left to PR Gate. The Notion plugin was suggested for live standard-OAuth validation but was not exposed in this session, so live Notion authorization remains uncovered. Independent review and exact-head CI are still pending.

## Review focus

- Whether the typed boundary correctly distinguishes a reachable server's tool failure from connection/transport failure.
- Whether authentication failures remain safely categorized without weakening standard OAuth's fail-closed behavior.
- Whether the generated Skill guidance is specific enough to recover through listed tools without inventing connector capabilities.

### DIFF
--- a/src/main/connectors/custom-mcp-bootstrap.ts
+++ b/src/main/connectors/custom-mcp-bootstrap.ts
@@ -7,7 +7,9 @@ export type CustomMcpFailureAvailability = 'unavailable' | 'unauthenticated'
 
 export const classifyCustomMcpFailure = (error: unknown): CustomMcpFailureAvailability =>
   error instanceof Error &&
-  /(?:401|403|unauthoriz|authenticat|forbidden|invalid_token)/i.test(error.message)
+  /(?:401|403|unauthoriz|authenticat|forbidden|invalid_token|(?:log(?:ged)?|sign(?:ed)?)[\s_-]?in)/i.test(
+    error.message
+  )
     ? 'unauthenticated'
     : 'unavailable'
 
--- a/src/main/connectors/custom-skill-doc.test.ts
+++ b/src/main/connectors/custom-skill-doc.test.ts
@@ -93,6 +93,18 @@ describe('syncCustomServerSkillDocs', () => {
     expect(entries).toEqual([])
   })
 
+  it('refreshes one server without validating or deleting unrelated server Skills', async () => {
+    const dir = await mkdtemp(join(tmpdir(), 'custom-skills-targeted-'))
+    const target = makeServer({ id: 'target', slug: 'target', name: 'Target' })
+    const unrelated = makeServer({ id: 'unrelated', slug: 'unrelated', name: 'Unrelated' })
+    const listTools = async (): Promise<typeof FAKE_TOOLS> => FAKE_TOOLS
+
+    await syncCustomServerSkillDocs(dir, [target, unrelated], listTools)
+    await syncCustomServerSkillDocs(dir, [target], listTools, ['target'])
+
+    expect((await readdir(dir)).sort()).toEqual(['mcp-target', 'mcp-unrelated'])
+  })
+
   it('isolates an unavailable server while materializing the remaining enabled servers', async () => {
     const dir = await mkdtemp(join(tmpdir(), 'custom-skills-unavailable-'))
     const unavailable = makeServer({ id: 'unavailable', slug: 'unavailable' })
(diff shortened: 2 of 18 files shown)


ewen-poch:
Implemented the OAuth recovery-state recommendation in 39da918. Structured unauthenticated tool errors are now recorded only for host-managed OAuth connectors, so Settings exposes sign-in recovery; connector-managed login tools remain reachable because their structured errors are not cached. Added regression coverage for both paths. Validation passed: 106 focused connector tests, node/web typecheck, lint, and the full suite with 13,513 tests passed.

ewen-poch:
Adjusted the standard OAuth connector interaction in 8189ff46. A valid Connected state is now non-interactive, stale credentials retain the connector's enabled intent and Agent context while exposing Retry, and an in-flight authorization shows a disabled Connecting… state instead of a Cancel action. This removes the user-triggered cancellation path that surfaced the transport AbortError without changing connector-managed login behavior. Validation passed: 115 focused connector/UI tests, node/web typecheck, ESLint, Prettier, and the full suite with 13,513 tests passed.

ewen-poch:
Implemented targeted custom Connector refresh in d459400. Toggling a custom Connector now carries its server ID through the Settings workflow into the runtime projection, so only that server participates in listTools discovery and Skill cleanup, and only that row projects a checking state. Full projection refresh remains available for startup and mutations that require global reconciliation. Regression coverage passes across targeted Skill cleanup, runtime projection, Settings projection, and workflow routing (88 tests), together with node/web typecheck, ESLint, and Prettier.

ewen-poch:
Not adopting the recommendation to restore an enabled Cancel action during standard OAuth authorization. The product interaction intentionally keeps Connected and in-flight Connecting… states non-interactive, while stale or rejected credentials expose Retry. The previous user-facing Cancel path surfaced the SDK transport AbortError and behaved as a low-level connection abort rather than an authentication recovery action. Cancellation remains an internal lifecycle operation; if abandoned authorization needs an additional recovery affordance, it should be designed as retry or expiry without reintroducing Cancel.

ewen-poch:
Not adopting this P2 after product review. The OAuth callback already has a bounded five-minute timeout, while the former user-facing Cancel action exposed the transport AbortError and conflated low-level connection cancellation with authentication recovery. The in-flight state therefore remains non-interactive; stale or rejected credentials recover through Retry. No code change is planned for this finding.

ewen-poch:
Rebased the branch onto origin/main at 5f38ca7b (including the PR Gate baseline fixes from #1102). Post-rebase validation passed: 271 Connector/Settings/UI impact tests, node and web typecheck, full ESLint with 0 errors, and the complete Vitest suite with 13,665 tests passed.

PR #1105, opened 2026-08-11
fix(reviewer): hide auditor prompts from local transcript

ewen-poch:
## Problem

Reviewer warn/fail findings were sent through the normal prompt path and projected into the Open Science transcript as user-authored messages before the agent corrected them.

## Proposed change

- Keep the existing `sendPrompt` entry and its admission, cancellation, and lifecycle behavior.
- Set the existing application-owned `suppressUserMessage` flag for Auditor correction turns so Open Science neither displays nor durably stores a local user message.
- Preserve the provider's ordinary save-as-user prompt semantics, matching the supplied raw session example without introducing a new ACP capability or provider protocol.
- Bind correction tool and artifact provenance to the latest complete user prompt on the active conversation branch instead of a synthetic message ID.

## Scope and non-goals

- Four reviewer files changed.
- No new `internal-reviewer` protocol, capability, request field, or managed Codex patch.
- No Prisma schema, table, migration, renderer, or side-chat changes.
- Reviewer cards, correction behavior, and the re-review loop remain unchanged.

Checks ran after the last material code edit and rebase onto the latest `main`.

- Reviewer ownership tests -> `npm test -- src/main/reviewer` -> 23 files, 252 tests passed.
- Declared reviewer consumer tests -> direct Vitest invocation of the 12 consumer files from `reviewer_orchestrator` -> 12 files, 268 tests passed.
- Focused visibility, provenance, lifecycle, and prior CI regressions -> 5 files, 41 tests passed.
- Changed-file ESLint and Prettier -> passed.
- Repository lint -> passed with unrelated existing warnings.
- Full `npm test` completed with 58 failures outside this four-file diff, primarily Windows temp-directory/symlink restrictions and an existing WorkspacePage line-count gate.
- Local Node typecheck is blocked because the shared dependency installation predates latest `main` and lacks `sharp`; exact-head CI installs the current lockfile and remains authoritative.

## Review focus

- Auditor prompts remain invisible and absent from the local durable transcript while still reaching the provider through `sendPrompt`.
- Correction activities and artifacts inherit the originating durable user prompt's active-branch provenance.

### DIFF
--- a/src/main/reviewer/correction.test.ts
+++ b/src/main/reviewer/correction.test.ts
@@ -319,12 +319,13 @@ describe('single-round auditor correction', () => {
     expect(correctionSendPrompt).toHaveBeenCalledWith(
       expect.objectContaining({
         sessionId: 'main-session-1',
+        suppressUserMessage: true,
         provenanceContext: expect.objectContaining({
           rootFrameId: 'root-frame-pending-session-1',
           agentFrameId: 'root-frame-pending-session-1',
           messageBranchId: 'message-branch-pending-session-1',
           runtimeSegmentId: 'runtime-segment-pending-session-1',
-          promptMessageId: expect.stringMatching(/^prompt-/u)
+          promptMessageId: 'msg-1'
         })
       })
     )
--- a/src/main/reviewer/correction.ts
+++ b/src/main/reviewer/correction.ts
@@ -79,8 +79,14 @@ export const injectAuditorMessage = async (options: {
 
   try {
     // Inject through the main-session sendPrompt path (design.md cross-cutting requirement).
-    // This is a normal prompt turn that the user sees in the conversation.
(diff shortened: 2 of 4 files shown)


ewen-poch:
Addressed the review findings with a smaller design:

- Removed the custom internal-reviewer capability, runtime method, prompt mode, and managed Codex application-context patch.
- Auditor corrections now use the existing `sendPrompt` path with `suppressUserMessage: true`, so they retain normal coordinator admission/cancellation tracking without creating a visible or durable local user message.
- The provider receives ordinary user-level prompt input; reviewer-controlled text is no longer promoted to trusted application/developer context.
- The internal-turn continuation scheduling concern no longer applies because the custom turn kind was removed.
- Correction provenance now binds to the latest complete user prompt on the active conversation branch instead of a synthetic prompt ID.
- The previously failing lifecycle and prompt-prefix tests now pass; all reviewer tests and declared reviewer consumers pass locally.

PR #1119, opened 2026-08-11
fix(global-search): stabilize async search results

ewen-poch:
## Problem

Cmd/Ctrl+K rendered matching Sessions immediately while the debounced Artifact query was still pending. Slow Artifact results then appeared above those Sessions and shifted the list. Pointer hover also changed the active row through the same effect used by keyboard navigation, so repeated hover/scroll feedback could move the viewport to the bottom.

## Proposed change

- Mark keyword searches pending as soon as input changes, retain the existing 150 ms debounce, and show a compact loading state.
- Reveal Artifact and Session matches together after the Artifact query settles, preserving Artifact-first ordering.
- Restrict automatic active-row scrolling to keyboard navigation; pointer hover only updates highlighting.
- Keep the result selection empty while loading so a fast Enter cannot activate an unrelated command.

## Scope and non-goals

Renderer-only change in the shared global-search dialog. No IPC, Artifact query, persistence, backend, debounce-duration, or result-ranking contracts change.

- Atomic loading and result ordering; pointer-safe scrolling -> `npm test -- src/renderer/src/components/global-search/GlobalSearchDialog.test.tsx` -> 17 passed.
- Renderer and process type safety -> `npm run typecheck` -> passed.
- Lint -> `npm run lint` -> passed with 0 errors and 13 pre-existing warnings outside the changed files.
- Complete portable regression suite -> `npm test` -> 14,323 passed, 198 skipped.

The shared Renderer component and its DOM behavior are covered. No IPC interface, persistence, path handling, native packaging, or platform-specific behavior changed, so no additional platform lane is included locally. Remaining risk: jsdom verifies result visibility and `scrollIntoView` intent, but not physical pointer motion in a packaged Electron window.

## Review focus

Please verify the pending-state selection behavior and the separation between pointer highlighting and keyboard-owned scrolling.

### DIFF
--- a/src/renderer/src/components/global-search/GlobalSearchDialog.test.tsx
+++ b/src/renderer/src/components/global-search/GlobalSearchDialog.test.tsx
@@ -16,6 +16,7 @@ import { GlobalSearchDialog } from './GlobalSearchDialog'
 
 let container: HTMLDivElement
 let root: Root
+let scrollIntoView: ReturnType<typeof vi.fn>
 
 const artifact = {
   id: 'artifact-1',
@@ -32,6 +33,11 @@ const artifact = {
 }
 
 beforeEach(() => {
+  scrollIntoView = vi.fn()
+  Object.defineProperty(window.HTMLElement.prototype, 'scrollIntoView', {
+    configurable: true,
+    value: scrollIntoView
+  })
   container = document.createElement('div')
   document.body.appendChild(container)
   root = createRoot(container)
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the P2 keyboard regression: the visible New session/New project command now remains in `selectableRows` while search results are pending, so the rendered option and keyboard selection model stay aligned. The regression test now asserts that this command remains selected both immediately after input and after the debounce starts. Validation: focused tests 17/17, typecheck passed, lint passed with 0 errors, and the full suite passed with 14,323 tests.

PR #1124, opened 2026-08-12
fix(message-stream): preserve presentation progress across remounts

ewen-poch:
## Problem

Buffered assistant text is presentation state, but remounting a session or side chat discarded that state. Returning to an in-progress turn could replay already visible text from the beginning, including text before a later tool call. The streaming caret also attached to Streamdown's `dir="auto"` wrapper, so it could render on the next line as a large block instead of following the visible text.

## Proposed change

- Track presentation checkpoints by session, active branch, and message in the main transcript.
- Track equivalent checkpoints by side-chat generation and entry.
- Restore already visible prefixes without mount animation while allowing genuinely new suffixes to keep using smooth streaming.
- Scope tool-boundary snapshots so only visible committed fragments become fixed.
- Anchor the streaming caret to the terminal rendered text block and style it as a thin, slightly taller line.

## Scope and non-goals

This is renderer-only presentation behavior. It does not change session persistence, the canonical conversation graph, controller state, runtime protocols, or stored message data. No historical data migration or compatibility layer is required.

It does not retune buffer pacing or replace the Markdown renderer.

- Switching away from and back to a streaming session does not replay already presented text.
- A fragment before a tool call remains fixed when later fragments stream.
- Branch switching with shared message IDs does not leak presentation state across branches.
- Closing and reopening side chat does not replay existing text; later suffixes still animate.
- The caret follows the final visible text and renders as a thin black/current-color line that is slightly taller than the body text.

Validation after the final material change:

- `npm test -- --run src/renderer/src/components/streamdown/AgentMarkdownFullscreenChrome.test.ts` — 5 passed.
- `npm run typecheck` — node and web typechecks passed.
- `npm run lint` — 0 errors; 13 pre-existing warnings outside the changed files.
- `npm test` — 987 test files passed, 15 skipped; 14,352 tests passed, 198 skipped.

## Review focus

- Presentation checkpoint identity and visibility barriers across session/branch/tool transitions.
- Main transcript and side-chat parity.
- Caret selectors against Streamdown's `dir="auto"` `display: contents` wrapper and terminal Markdown blocks.
- Manual UI acceptance is still recommended for exact cursor alignment and animation feel; automated coverage verifies the DOM/CSS contract and lifecycle behavior.

### DIFF
--- a/src/renderer/src/assets/agent-markdown.css
+++ b/src/renderer/src/assets/agent-markdown.css
@@ -173,7 +173,7 @@
 
     & .agent-markdown > :last-child:empty::after,
     & .agent-markdown > :last-child > :last-child::after {
-      content: var(--streamdown-caret, none);
+      content: '';
       display: inline-block;
       width: 1px;
       height: 1.25em;
--- a/src/renderer/src/assets/agent-markdown.css
+++ b/src/renderer/src/assets/agent-markdown.css
@@ -168,9 +168,19 @@
     }
 
     & .agent-markdown > :last-child::after {
+      content: none;
+    }
+
+    & .agent-markdown > :last-child:empty::after,
+    & .agent-markdown > :last-child > :last-child::after {
+      content: var(--streamdown-caret, none);
       display: inline-block;
+      width: 1px;
+      height: 1.25em;
+      overflow: hidden;
       margin-inline-start: 0.08em;
-      color: currentColor;
+      vertical-align: -0.18em;
+      background-color: currentColor;
       opacity: 1;
     }
   }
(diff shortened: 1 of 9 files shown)


ewen-poch:
Regarding P2 (restore the caret for ordinary paragraph output): no code change is needed here. `RichAgentMarkdown` always passes `dir="auto"`, and the installed Streamdown version renders an ordinary paragraph as `.agent-markdown > div[dir][style="display:contents"] > p`. The existing `.agent-markdown > :last-child > :last-child::after` selector therefore targets the terminal `p` directly; I verified this against the real Streamdown render and the targeted cursor test. Adding another selector for a non-empty top-level block would target the direction wrapper and could reintroduce a duplicate caret. Omitting Streamdown's built-in caret remains intentional so only the thin terminal caret is rendered.

PR #1148, opened 2026-08-12
feat(network): add global proxy settings

### LINKED ISSUE #1136: [Feature]: Add dotenv-style runtime environment variable settings
### Preflight checklist

Open Science does not currently provide a central place to configure environment variables for its agent and notebook runtimes.

Variables such as the following may need to be configured before starting Open Science:

```dotenv
HTTP_PROXY=http://127.0.0.1:7890
HTTPS_PROXY=http://127.0.0.1:7890
ALL_PROXY=http://127.0.0.1:7890
NO_PROXY=localhost,127.0.0.1,::1
```

This is inconvenient when Open Science is launched from a desktop shortcut because it may not inherit the same environment as a terminal. It can also produce different behavior depending on whether the application was opened normally or started through the CLI.

The same limitation affects other environment variables needed by agents, notebook kernels, package installers, and optional tools.

### Proposed solution

Add an application-wide **Environment variables** panel to Settings.

It should support two configuration sources:

1. Environment variables entered and managed in Open Science.
2. One optional user-selected `.env` file.

The sources should be resolved in this order:

```text
Open Science managed entries
    → selected .env file
    → inherited operating-system environment
```

A higher source should override the same key from a lower source.

#### Managed entries

- Accept dotenv-style `KEY=value` entries.
- Display saved keys in a masked list.
- Allow users to add, replace, and delete entries.
- Encrypt every value using the operating-system credential vault.
- Never write plaintext values into `settings.json`.
- Disable the profile and show a warning when secure credential storage is unavailable.
- Never include values in logs, diagnostics, exports, or renderer settings snapshots.

#### External `.env` file

- Allow one file to be selected.
- Support basic `KEY=value` assignments, blank lines, and comments.
- Do not execute commands or perform variable interpolation.
- Load the file at application startup or when the user selects **Reload**.
- Cache the last successfully loaded values using encrypted storage.
- If the file is missing or invalid, retain the last-known-good values and show a persistent warning.
- Never partially apply an invalid file.

#### Runtime targets

Use profile-level controls to choose where the variables are applied.

Recommended defaults:

- Agent runtimes: enabled
- Persistent Python, R, and REPL notebook kernels: enabled
- Package installation and runtime provisioning: enabled
- Stateless Notebook shell commands: disabled
- Custom stdio connectors: disabled

The restricted targets should require explicit opt-in because they currently use reduced environments to prevent host credentials from leaking into arbitrary commands or third-party processes.

Changes should apply only to newly created processes. Open Science should identify active runtimes that need to be reconnected or restarted, but should not terminate running work automatically.

#### Protected variables

Open Science should reject overrides of application- and runtime-owned variables, including:

- `OPEN_SCIENCE_*`
- Provider-owned credential and configuration variables managed by Open Science
- Agent profile directories such as `CLAUDE_CONFIG_DIR` and `CODEX_HOME`
- Essential process variables such as `PATH`, home-directory variables, and operating-system runtime paths

The existing provider credential-isolation rules must continue to take precedence over the user profile.

#### Proxy variables

Standard proxy variables should receive additional normalization:

- Mirror `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, and `NO_PROXY` into their lowercase aliases.
- Always add `localhost`, `127.0.0.1`, and `::1` to the effective proxy bypass list.
- Explicit profile proxy values should take precedence over automatic system-proxy resolution for spawned runtimes.

This feature should not claim that environment variables automatically control every in-process Electron or Node HTTP request. Complete application proxy routing can remain a separate feature.

Expected behavior should be covered on Windows, macOS, and Linux.

### Alternatives considered

**Require users to configure operating-system environment variables**

Desktop applications do not always inherit the same environment as an interactive terminal. This also makes Open Science behavior dependent on how it was launched.

**Require users to start Open Science from a configured terminal**

This does not work well with normal desktop startup, automatic launch, or users who do not routinely use a terminal.

**Support only an external `.env` file**

A plaintext file is convenient for automation but can contain credentials, become unavailable, or be committed accidentally. Managed encrypted entries provide a safer option.

**Support only managed entries**

This prevents users and automation tools from maintaining a conventional `.env` file outside Open Science.

**Pass every variable to every subprocess**

The stateless Notebook shell and custom connectors currently have intentionally restricted environments. Automatically exposing every configured value would weaken those security boundaries.

**Treat environment variables as complete proxy configuration**

Some Open Science traffic uses Electron networking or in-process HTTP clients that do not necessarily honor `HTTP_PROXY`. A runtime environment profile should not promise routing behavior it cannot provide.

### Additional context

Open Science currently has several different environment-propagation paths:

- Agent runtimes generally inherit the application process environment after provider-specific credential filtering.
- Persistent notebook kernels generally inherit the application environment.
- Stateless Notebook shell execution uses a restricted allowlist.
- Custom stdio connectors receive only their explicitly configured environment.
- Codex subscription sessions have separate system-proxy resolution behavior.

A single application-owned runtime environment profile would make this behavior configurable and consistent while preserving the existing security boundaries.

The suggested canonical term for this feature is **runtime environment profile**: an application-wide set of user-configured variables projected into selected spawned execution targets.

ewen-poch:
## Problem

Open Science network calls and ACP runtimes currently depend on inconsistent host and process-level proxy behavior. Users behind a system proxy or a manually configured local proxy cannot apply one app-wide policy, and Electron traffic can diverge from ACP/notebook/installer child-process traffic.

Closes #1136.

## Proposed change

- Add a global **Settings > Network > Proxy** setting with **System**, **Manual**, and **Direct** modes.
- In **System** mode, keep Electron on operating-system/PAC resolution per request. New child processes inherit only the proxy environment that was present when Open Science started; one PAC result is never projected globally.
- In **Manual** mode, apply the configured fixed proxy to Electron and inject it into subsequently spawned ACP agents, notebook runtimes, and installer processes.
- In **Direct** mode, configure Electron for direct access and clear proxy variables for subsequently spawned processes.
- Route provider bridges, registry/mirror probes, and direct scientific connectors through Electron networking so they share the selected app policy.
- Reject manual proxy URLs containing embedded usernames or passwords and always bypass loopback addresses.

ACP lifecycle behavior is intentional: provider/network requests using the Electron session pick up the updated policy immediately, while an already-running ACP agent process keeps the environment it was spawned with. New or restarted ACP agents receive the representable child-process policy. Users with a PAC-only system configuration should choose Manual when an ACP process must use a fixed proxy.

## Scope and non-goals

- Global application setting only; no per-project proxy configuration.
- No proxy authentication or credential storage.
- No local PAC-forwarding proxy for arbitrary child-process destinations.
- No modification of operating-system proxy settings.
- No forced restart of active ACP agents, kernels, or installers.
- No persisted-data migration or schema-version bump.

Historical compatibility is limited and explicit: existing settings documents with no network proxy field are interpreted as **System** mode. No historical data rewrite is performed.

- Proxy persistence, normalization, validation, and runtime application:
  - `npm test -- src/main/settings/network-proxy-settings.test.ts src/main/settings/network-proxy-runtime.test.ts`
  - Passed.
- ACP injection across the supported agent launch paths:
  - `npm test -- src/main/settings/network-proxy-runtime.test.ts src/main/acp/agent-process.test.ts src/main/agent-framework/claude-code.test.ts src/main/agent-framework/opencode.test.ts src/main/agent-framework/codex.test.ts src/main/settings/backend-resolver.test.ts src/main/settings/provider-transport-owner.test.ts`
  - Passed: 7 files, 188 tests.
- Final targeted Test Impact Set before review follow-ups:
  - Passed: 28 files, 645 tests.
- Latest review-fix validation on `bb300960`:
  - Focused runtime/UI/application-composition tests passed: 3 files, 18 tests.
  - `npm run typecheck:node` and `npm run typecheck:web` passed.
  - `npm run lint` passed with 0 errors and 12 pre-existing warnings.
  - `npm run build:e2e` passed.
  - `npm run check:web-api-map` passed.
  - `git diff --check` passed.

The full `npm test` fallback was attempted on this Windows host but could not be certified clean against the current checkout baseline. Failures were outside this change and dominated by host ACL/symlink restrictions, stale Prisma temporary schemas missing `codecVersion`, locked temporary databases, and unusable temporary storage roots. CI on this exact head remains the authoritative full-suite result.

## Review focus

- Ordering of proxy application during Electron startup and settings changes.
- The ACP lifecycle boundary: future process spawns receive the selected representable environment; active processes are not mutated.
- System/PAC remains per-request inside Electron and is not flattened into one child-process proxy endpoint.
- Manual and Direct behavior for the Codex subscription ACP replacement environment.
- Default-session routing for connector/provider traffic.
- Historical default behavior when the persisted network proxy field is absent.
- Manual URL validation and loopback bypass semantics.


ewen-poch:
Addressed both P1 findings in f8e93fca:

- Bundled connector dispatch now constructs the production ParserEngine with netFetchStandard, and provider model refresh explicitly injects the same Electron Session-aware transport.
- The shared managed-runtime metadata and tarball defaults used by Claude, OpenCode, and Codex now use netFetchStandard instead of node:https. Tarball responses are converted to a Node readable stream while preserving Content-Length progress reporting.

Validation after the changes:
- Managed Claude/OpenCode/Codex installer tests: 3 files, 113 passed, 4 skipped.
- Provider model refresh proxy-path test: passed.
- Typecheck: passed.
- Lint: 0 errors; 12 existing warnings.
- Production build, Web API map, module-impact validation, and diff check: passed.

No persisted fields, schema, data model, or UI behavior changed.

ewen-poch:
Addressed the managed tarball timeout regression in f33074df. The Electron Session-aware download now has a 20-second inactivity watchdog for both response headers and streamed body progress. Each received chunk resets the watchdog; a stall aborts the underlying fetch/stream so the existing registry fallback can proceed. Added regression coverage for a stalled stream and for timeout reset on progress. Validation after the change: managed Claude/OpenCode/Codex tests passed (115 passed, 4 skipped), node typecheck passed, lint passed with 0 errors and 12 existing warnings, Electron production build passed, and diff check passed. No schema, data model, persistence, architecture boundary, or UI behavior changed.

ewen-poch:
Addressed the installer redirect downgrade regression in 26c9e280. Managed runtime metadata and tarball requests now handle redirects manually, allow at most five hops, and require HTTPS for the initial URL and every redirect target. Redirect response bodies are cancelled before proceeding or rejecting. Added regression coverage for a valid HTTPS redirect, HTTP downgrade rejection, and the redirect limit. Validation after the change: managed Claude/OpenCode/Codex tests passed (118 passed, 4 skipped), node typecheck passed, lint passed with 0 errors and 12 existing warnings, Electron production build passed, and diff check passed. No schema, persistence, data model, architecture boundary, or UI behavior changed.

ewen-poch:
Addressed both findings in `bb300960`:

- System mode no longer resolves one representative URL and writes that result into the global child-process environment. Electron keeps OS/PAC resolution per request, while child processes restore only the proxy environment inherited when Open Science started.
- The existing `SettingsService.resolveCodexProxyEnvironment` seam now reads the active `NetworkProxyRuntime` child-process policy. This keeps Codex subscription ACP aligned with Manual, Direct, and the representable System environment without flattening PAC routing.
- Manual still injects the configured fixed proxy, Direct explicitly clears proxy variables, and loopback bypasses remain intact.
- The proxy URL helper now uses the helper fallback until validation is shown, so the example is visible before blur; a renderer test covers this state.
- The settings copy and design documentation now state the System/PAC limitation explicitly.

Validation: focused runtime/UI/application-composition tests (18/18), node and web typechecks, lint (0 errors; 12 pre-existing warnings), Electron build, web API map check, and `git diff --check` all pass. The local full suite remains blocked by unrelated Windows temp-directory, symlink, and stale Prisma-schema failures; CI on this head is the authoritative full-suite run.

ewen-poch:
Addressed the GitHub star proxy gap in 878e8f0e. The production GitHub command owner now receives Electron's netFetchStandard, so the request follows the active Electron proxy session instead of Node's global fetch. Added production-composition coverage in application-command-wiring.test.ts. Focused tests pass (20/20), typecheck:node passes, and lint reports 0 errors (12 pre-existing warnings). The full test suite still hits unrelated Windows baseline/environment failures involving temporary SQLite schemas, unusable temp roots, symlink permissions, locked files, and timeouts.

ewen-poch:
Addressed the provider bridge header regression in 381692ba and rebased the branch onto the latest main. Both Anthropic and OpenAI bridge header projections now drop sec-fetch-* before invoking Electron net.fetch, while retaining ordinary application headers. Added regression coverage for both bridge paths. Focused bridge/transport tests pass (12/12), focused production-wiring/proxy tests pass (20/20), typecheck:node passes, and lint reports 0 errors (11 pre-existing warnings). The local full suite continues to hit unrelated Windows baseline/environment failures involving temporary SQLite schemas, unusable temp roots, symlink permissions, locked files, and timeouts.

PR #1153, opened 2026-08-12
fix(workspace): improve composer accessibility

ewen-poch:
## Problem

The reported accessibility gaps are reproducible and worth fixing:

- file attach/remove actions are hidden behind hover, so touch users cannot discover them and keyboard focus has no reliable visible reveal;
- skill and artifact suggestions keep focus in the composer but do not connect the textbox to the listbox or expose the active option to assistive technology;
- completion, failure, and unresolved-mention feedback is rendered visually without reliable live-region semantics.

## Proposed change

- Keep row actions visible on no-hover devices, preserve hover reveal on hover-capable pointers, add visible focus rings, and use 44 px coarse-pointer targets.
- Give both mention popups one editor-owned listbox id and synchronize `aria-controls`, `aria-autocomplete`, and `aria-activedescendant`; keep the active option scrolled into view.
- Announce normal completion politely and failures or unresolved mentions assertively; retain a mounted mention alert container so text changes are announced reliably.
- Cover the behavior with focused interaction and render regressions.

## Scope

Renderer workspace UI only. No data model, IPC, storage, dependency, or public API changes.

## Acceptance / validation

- `npm test -- <7 affected Renderer test files>`: 168 passed.
- `npm run lint`: passed with 0 errors; 13 pre-existing warnings remain in unrelated files.
- `npm run build:e2e`: passed.
- Accessibility E2E: startup/Home and core dialog/workspace axe scans passed; the keyboard-only journey passed when rerun alone. The full serial suite still has a Windows hidden-window fixture timing issue where dynamic fake-agent replies stop at partial text, causing 3 timeouts before later axe assertions (no blocking axe violation was reported in the surfaces that ran).

## Review focus

- Hover-capable versus coarse/no-hover visibility and target-size variants.
- Active-descendant id integrity while switching and closing skill/artifact suggestion popups.
- Polite completion announcements versus assertive failure and invalid-mention announcements.

### DIFF
--- a/src/renderer/src/pages/workspace/ComposerYourFilesMenu.interaction.test.tsx
+++ b/src/renderer/src/pages/workspace/ComposerYourFilesMenu.interaction.test.tsx
@@ -53,7 +53,10 @@ vi.mock('@/components/ui/dropdown-menu', () => ({
   }>): React.JSX.Element =>
     // asChild items keep their real child element; wire the select handler onto its click.
     asChild && isValidElement(children) ? (
-      cloneElement(children as ReactElement<{ onClick?: () => void }>, { onClick: onSelect })
+      cloneElement(children as ReactElement<{ onClick?: () => void }>, {
+        ...rest,
+        onClick: onSelect
+      })
     ) : (
       <button type="button" disabled={disabled} onClick={onSelect} {...rest}>
         {children}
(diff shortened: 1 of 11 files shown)


ewen-poch:
Addressed the alert collision: the mention notice remains mounted with aria-live=assertive and aria-atomic=true, but no longer claims the generic alert role. This keeps mention errors announced while allowing the composer stop-error alert to remain directly discoverable. The focused scroller and ConversationPanel interaction suites pass (122 tests), and typecheck:web passes.

ewen-poch:
Addressed the coarse-pointer gap: both the granted-root toggle and recursive directory toggles now have a 44px minimum hit height. Nested rows remove their vertical padding only for coarse pointers so the target stays at 44px instead of inflating the row to 52px. Added regression coverage for both levels; the file-menu interaction suite passes (13 tests), and typecheck:web passes.

ewen-poch:
Addressed the historical-announcement regression: persisted completion and failure timestamps are static again, while the scroller keeps session-and-branch-scoped status snapshots and updates dedicated polite/assertive live regions only for newly observed terminal transitions. Opening or switching to historical transcript state therefore stays silent; live completions announce "Response completed." and live failures announce "Response failed." The affected render/interaction and ConversationPanel suites pass (179 tests), typecheck:web passes, and lint reports 0 errors.

ewen-poch:
This recommendation is already implemented in the reviewed head (99000ec3). MessageTimestamp now renders a plain time element with no role or aria-live attributes, while WorkspaceMessageScroller owns session-and-branch-scoped polite/assertive live regions that update only for newly observed terminal transitions. The regression test verifies that persisted completed and failed timestamps remain non-live, and the interaction test verifies that initial historical failures stay silent while live completion/failure transitions announce. No additional code change is needed for this duplicate finding.

ewen-poch:
Rebased this branch onto the latest origin/main (bfa5e233). The conflict resolution preserves the interaction-state updates from #1154 (transition duration, reduced-motion handling, and no-hover visibility) and the first-message scroll control from #1144, while retaining this PR's keyboard/touch accessibility behavior and live-only outcome announcements.

Validation after the rebase:
- 266 focused tests passed across 9 test files
- typecheck:web passed
- lint passed with 0 errors (existing unrelated warnings remain)
- changed-file formatting and diff whitespace checks passed

The final diff remains within the approved workspace/composer file scope.

PR #1157, opened 2026-08-12
feat(providers): enable DeepSeek V4 Pro Responses

ewen-poch:
## Problem

DeepSeek now documents native Responses API support for `deepseek-v4-pro`, but the provider registry still advertised Responses only for `deepseek-v4-flash`. Codex therefore treated the Pro model and its bundled `[1m]` variant as incompatible.

## Proposed change

- Advertise native Responses support for `deepseek-v4-pro` and `deepseek-v4-pro[1m]`.
- Keep the existing explicit per-model capability list so unknown live-catalog models remain fail-closed.
- Add registry and runtime projection coverage proving Codex selects native Responses compatibility instead of the Chat Completions bridge.

Official capability reference: https://api-docs.deepseek.com/zh-cn/quick_start/pricing/

## Scope and non-goals

This reuses the existing Responses transport and compatibility proxy. It does not change protocol adapters, persistence, provider setup UI, model context windows, or non-DeepSeek providers.

All checks listed below ran after the final material edit:

- DeepSeek V4 Responses capability and Codex runtime routing -> `npm run test:module -- settings_provider_accounts` -> 20 files, 331 tests passed.
- Main-process TypeScript consumers -> `npm run typecheck:node` -> passed.
- Shared registry renderer consumers -> `npm run typecheck:web` -> passed.
- Source lint and formatting rules -> `npm run lint` -> passed.
- Patch whitespace validity -> `git diff --check` -> passed.

Consumer and platform impact: the changed shared capability is consumed by settings runtime projection and account resolution, both covered by the module impact set. No protocol adapter, persistence, path, OS-specific, or UI behavior changed, so no additional platform or E2E lanes were included. CI remains authoritative for cross-platform verification.

Uncovered risk: validation confirms routing selection without making a live authenticated DeepSeek API request.

## Review focus

Confirm the explicit DeepSeek model list matches the official Responses support matrix and that the existing native Responses compatibility route remains the correct Codex path.

### DIFF
--- a/src/main/settings/provider-runtime-projection.test.ts
+++ b/src/main/settings/provider-runtime-projection.test.ts
@@ -74,6 +74,29 @@ describe('ProviderRuntimeProjectionOwner', () => {
     expect(provider).toEqual(before)
   })
 
+  it('routes DeepSeek V4 Pro through native Responses for Codex', () => {
+    const owner = new ProviderRuntimeProjectionOwner()
+    const provider: StoredProvider = {
+      id: 'deepseek',
+      type: 'official',
+      vendorId: 'deepseek',
+      name: 'DeepSeek'
+    }
+
+    expect(
+      owner.resolveRuntimeTarget(
+        provider,
+        { kind: 'required', model: 'deepseek-v4-pro' },
+        getAgentFramework('codex')
+      )
+    ).toMatchObject({
+      apiEndpoints: ['anthropic', 'openai', 'responses'],
+      frameworkCompatible: true,
+      needsChatResponsesBridge: false,
+      needsNativeResponsesCompatibility: true
+    })
+  })
+
   it('keeps an exact required model when a subscription catalog is unknown', () => {
     const owner = new ProviderRuntimeProjectionOwner()
     const provider: StoredProvider = {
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the P1 by moving the existing Chat Completions bridge integration test to an explicit custom `openai` provider fixture. This preserves the test's bridge, tool forwarding, DeepSeek effort transport, and 1M context assertions without relying on the now-native `deepseek-v4-pro` registry classification. DeepSeek V4 Pro native routing remains covered by the new runtime projection regression test and the existing native Responses compatibility integration coverage.

Validation after the change: full `npm test`, `npm run typecheck`, `npm run lint`, and `git diff --check` passed.

ewen-poch:
Updated the remaining official-vendor validation coverage to match the new capability: the default `deepseek-v4-pro` draft now proves the native `/v1/responses` namespace tool-call contract and model selection. The existing Flash native-route test and custom Chat Completions bridge test remain intact.

Validation after this change: the three related integration tests passed, followed by full `npm test`, `npm run typecheck`, `npm run lint`, and `git diff --check`.

Issue #1185, opened 2026-08-13
fix(cli): AppImage CLI shim fails with "bad option: --no-sandbox" in v0.14.1


justemu:
## Bug Description

After upgrading to v0.14.1, the CLI shim installed via **Settings → General → Command line tool** fails to execute on Linux AppImage builds:

```
$ open-science status
/tmp/.mount_aipoch2GeExT/open-science: bad option: --no-sandbox
```

The CLI is completely non-functional — no commands can be executed.

## Environment

- OS: Ubuntu 24.04 (Linux)
- Open Science: v0.14.1 (AppImage)
- Installed via: Settings → General → Command line tool → Install

## Steps to Reproduce

1. Download Open Science v0.14.1 AppImage
2. Launch the AppImage
3. Go to **Settings → General → Command line tool** and click **Install**
4. Quit the app
5. Open a terminal and run `open-science status`
6. Observe the error: `bad option: --no-sandbox`

## Generated CLI Content

The shim at `~/.local/bin/open-science` contains:

```sh
#!/bin/sh
# Open Science command-line launcher. Managed by the app (Settings -> General -> Command line
# tool); edits will be overwritten on reinstall. Runs the app's Electron in Node mode.
OPEN_SCIENCE_APP_PATH='/opt/appimages/aipoch-open-science-0.14.1-linux-x86_64.AppImage' ELECTRON_RUN_AS_NODE=1 exec '/opt/appimages/aipoch-open-science-0.14.1-linux-x86_64.AppImage' -e 'Promise.all([import('node:path'), import('node:url')]).then(([{ join }, { pathToFileURL }]) =>import(pathToFileURL(join(process.resourcesPath, 'cli', 'index.mjs')).href)).then(({ reportCliError, runCli }) => runCli(process.argv.slice(1)).catch(reportCliError)).catch((error) => { console.error(error instanceof Error ? error.message : error);process.exitCode = 1 })' -- "$@"
```

The key difference from the previous working version is the use of `-e` with inline JavaScript bootstrap code instead of passing `resources/cli/index.mjs` as a direct file argument.

## Previous Working Version (v0.14.0 and earlier)

```sh
#!/bin/sh
[...]

Issue #1150, opened 2026-08-12
AppImage: CLI shim path breaks on every launch due to FUSE re-mount


justemu:
## Bug Description

Every time Open Science is launched via AppImage, FUSE re-mounts the image at a **different path** (e.g. `/tmp/.mount_open-sXXXXXX/`). This causes `process.execPath` and `process.resourcesPath` to change on each launch.

For users who installed the CLI tool from **Settings → General → Command line tool**, the shim at `~/.local/bin/open-science` contains hardcoded references to these paths (set via `OPEN_SCIENCE_APP_PATH` and the `exec` command). After the first launch, these paths become stale and the CLI fails to execute.

**Only users who installed the CLI from the Settings panel are affected.**

## Steps to Reproduce

1. Launch Open Science via AppImage
2. Go to **Settings → General → Command line tool** and click **Install**
3. Quit the app
4. Launch the AppImage again (FUSE mounts at a new path)
5. Run `open-science` from terminal → **fails** because the shim points to the old mount path

## Root Cause

In `src/main/cli-install/ipc.ts`, `resolveEnv()` captures `process.execPath` at call time:

```typescript
const resolveEnv = (): CliLauncherEnv => ({
  appExecPath: process.execPath,
  cliEntryPath: app.isPackaged
    ? join(process.resourcesPath, 'cli', 'index.mjs')
    : join(app.getAppPath(), 'cli', 'index.mjs'),
  ...
})
```

This works correctly at install time, but the shim content is never refreshed when paths change.

## Proposed Fix

On app startup, after creating the `cliCommandOwner`, check if the CLI is already installed. If it is, verify that the shim's paths still match the current `process.execPath`. If they don't match (i.e. the FUSE mount path has changed), silently reinstall the CLI shim in the background.

This can be implemented by adding an `ensureCliLauncherUpToDate` function to `launcher.ts` that:

1. Reads the existing shim file
2. Checks if it contains the current `process.execPath` / `process.resourcesPath`
3. If paths differ, calls `installCliLauncher()` to overwrite with current paths

This should run once at startup, non-blocking, and only on Linux AppImage builds.

## Environment

- OS: Linux (Ubuntu 24.04)
- Open Science: latest (via AppImage)
- Affected feature: CLI tool installation from Settings → General

PR #1187, opened 2026-08-13
fix(cli): revert AppImage -e bootstrap to fix CLI regression (#1185)

justemu:
## Summary

Reverts the  bootstrap approach in  introduced by commit fdc9372 ("keep AppImage launcher paths stable"), restoring the direct file-argument format while keeping the  reconciliation mechanism.

## Problem

The v0.14.1 CLI shim uses `exec AppImage -e 'code...' -- ""` to dynamically import the CLI entry point. The AppImage binary's argument parser does not forward the `-e` flag to the internal Node.js runtime, causing:

See #1185 for full details.

## Changes

**launcher.ts** (1 file, -44/+15 lines):

- Removed `appImageCliBootstrap` constant (the `-e` + dynamic import code)
- Removed `isLinuxAppImage()` helper
- Reverted `posixShim` to use `env.appExecPath` and `env.cliEntryPath` directly
- Reverted `getCliLauncherStatus` to simple existence check
- Reverted `isCliShimStale` to simple `appExecPath` content check
- Removed unused `readCliLauncher` and `isManagedCliLauncher` helpers

**ipc.ts** — No changes (reconciler and lifecycle management preserved).

## What stays the same

- `ensureCliLauncherCurrent` runs on startup to refresh stale FUSE paths
- `CliCommandOwnerWithLifecycle` type and `ensureCurrent()` method
- `await cliCommandOwner.ensureCurrent()` in main startup
- The shim format: `exec appExecPath cliEntryPath ""`

## Testing

- Existing tests for `isCliShimStale` and `ensureCliLauncherCurrent` need updating (they reference the removed AppImage-specific helpers)
- Manual verification: install CLI from Settings, quit app, relaunch AppImage, run `open-science status`

## Note

The "keep AppImage launcher paths stable" approach is conceptually superior — it eliminates the need for path regeneration by making the shim inherently stable. A follow-up could explore alternatives to `-e` (e.g., shell bootstrap script, `\` env var) to achieve the same goal without the AppImage argument parsing issue.

### DIFF
--- a/src/main/cli-install/launcher.test.ts
+++ b/src/main/cli-install/launcher.test.ts
@@ -59,7 +59,7 @@ describe('planCliLauncher', () => {
     expect(plan.shim).toContain('ELECTRON_RUN_AS_NODE=1')
     // Packaged: pins the app path and single-quotes both paths (they contain a space).
     expect(plan.shim).toContain("OPEN_SCIENCE_APP_PATH='/opt/Open Science/open-science'")
-    expect(plan.shim).toContain('\'/opt/Open Science/resources/cli/index.mjs\' "$@"')
+    expect(plan.shim).toContain("'/opt/Open Science/resources/cli/index.mjs' \"$@\"")
   })
 
   it('omits OPEN_SCIENCE_APP_PATH for a development (unpackaged) build', () => {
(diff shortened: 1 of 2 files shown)


justemu:
## Test updates pushed

Updated `launcher.test.ts` to match the reverted CLI shim format:

- **Removed** AppImage-specific tests that checked for `-e` bootstrap format
- **Removed** `-e` + `--` separator test (no longer relevant)
- **Reverted** reconciliation tests to use simple `appExecPath` content check
- **Added** test for FUSE mount path staleness detection

The `ipc.test.ts` tests are unchanged — they still reference the valid `appImagePath` field in `CliLauncherEnv` and the reconciliation lifecycle.

justemu:
## CI fixes pushed

Addressed two failures from the automated checks:

1. **Unused import**: Removed `spawnSync` import from `launcher.test.ts` (was only used by the deleted `-e` + `--` separator test)

2. **Platform guard**: Added `env.platform === 'linux'` check to `isCliShimStale` to prevent false staleness detection on win32/darwin where user-managed launchers exist and should not be overwritten.

ewen-poch:
Closing this PR so we can rework the fix without restoring the FUSE mount-path dependency from #1150. Removing the AppImage -e bootstrap is the right response to #1185, but the current replacement pins process.execPath and process.resourcesPath under /tmp/.mount_*. That launcher can become unusable again after the desktop process exits or the AppImage remounts. The revised implementation will keep the stable APPIMAGE path and invoke a persistent bootstrap file as the Node script argument instead of using -e. It will also retain the existing safety boundaries: reconcile only packaged Linux AppImage launchers; update only app-managed launchers; compare the complete planned launcher content; treat only ENOENT as missing; and verify the CLI after the desktop app fully exits and across a later remount. Thanks for the investigation and quick regression analysis. We will continue the fix in a replacement change.

ewen-poch:
Follow-up after inspecting the generated electron-builder AppRun: a persistent external bootstrap file alone is not sufficient. On hosts where unshare -Ur fails, AppRun prepends --no-sandbox before every user argument, so Electron in Node mode would still reject the invocation before reaching that bootstrap. The replacement change will instead use the AppImage runtime --appimage-mount mode, read the reported mount point, invoke the mounted Electron binary and bundled CLI entry directly, and terminate the mount process when the CLI exits. This bypasses AppRun argument injection while keeping the launcher anchored to the stable APPIMAGE path. The safety and validation requirements in the previous comment remain unchanged.

PR #1191, opened 2026-08-14
fix(storage): contain decoded $DATA paths

ewen-poch:
## Problem

Persisted `$DATA/...` paths were treated as trusted. Decoding passed the suffix directly to `join(dataRoot, suffix)`, so a value such as `$DATA/../../etc` could materialize outside the configured data root.

The shared codec is used while hydrating Session paths, project Preview items, and Notebook run documents. A tampered or corrupt persisted value could therefore point those consumers at files outside app-managed storage.

## Proposed change

- validate already-encoded `$DATA` values in both encode and decode paths
- normalize portable forward-slash suffixes
- reject parent traversal, drive-prefixed paths, UNC/absolute paths, and backslashes
- resolve the decoded candidate and enforce containment within the resolved data root
- keep Session, Preview, and Notebook on the same shared codec
- add shared invariant tests plus consumer contract coverage for all three persistence surfaces

## Scope and non-goals

- No new persisted fields, database schema, migration, data model, enum value, or UI interaction.
- Valid historical app-generated `$DATA/...` values remain compatible.
- Invalid historical values containing `..`, drive prefixes, UNC/absolute forms, or backslashes are intentionally rejected rather than repaired.
- Existing consumer behavior handles rejection: Session files enter the established corrupt/quarantine path; Preview and Notebook loads fail instead of returning an out-of-root path.

Checks listed below ran after the final material edit.

- shared codec and Session/Preview/Notebook consumption -> `vitest run src/main/storage/data-path.test.ts src/main/session-persistence/data-path-roundtrip.test.ts src/main/projects/preview-repository.test.ts src/main/notebook/run-document-data-paths.test.ts --maxWorkers=1 --testTimeout=30000 --hookTimeout=30000` -> passed, 4 files / 28 tests
- main-process type safety -> `tsc --noEmit -p tsconfig.node.json --composite false` -> passed
- linted source and tests -> `eslint --cache .` -> passed with 0 errors; 11 existing warnings outside this diff
- patch hygiene -> `git diff --check` -> passed

The full portable suite was not run locally. The path-sensitive PR Gate lanes and exact Node 22 environment remain authoritative for Windows/macOS/Linux coverage; the local checks ran on Node 24.

## Review focus

- cross-platform rejection of drive, UNC, backslash, and parent-segment forms
- containment after path resolution
- intentional strict rejection of invalid persisted historical values
- preservation of valid relocation and external-path pass-through behavior

### DIFF
--- a/src/main/notebook/run-document-data-paths.test.ts
+++ b/src/main/notebook/run-document-data-paths.test.ts
@@ -170,4 +170,12 @@ describe('run document data-path codec', () => {
     const doc = { ...buildDocument(), workspaceCwd: '/Users/x/proj' }
     expect(encodeRunDocumentDataPaths(doc, ROOT).workspaceCwd).toBe('/Users/x/proj')
   })
+
+  it('rejects an escaping $DATA path during Notebook hydration', () => {
+    const doc = { ...buildDocument(), workspaceCwd: '$DATA/../../outside' }
+
+    expect(() => decodeRunDocumentDataPaths(doc, ROOT)).toThrow(
+      /portable relative path within the data root/
+    )
+  })
 })
--- a/src/main/projects/preview-repository.test.ts
+++ b/src/main/projects/preview-repository.test.ts
@@ -124,4 +124,28 @@ describe('preview state repository (integration)', () => {
     const loaded = await repository.get('project-a')
     expect(loaded?.items[0].path).toBe(absolutePath)
   })
+
+  it('rejects an escaping $DATA path from persisted Preview state', async () => {
+    storageRoot = await mkdtemp(join(tmpdir(), 'open-science-preview-'))
+
+    const client = createProjectDbClient(storageRoot)
+    disconnect = () => client.$disconnect()
+
+    await migrateApplicationDatabase(client)
+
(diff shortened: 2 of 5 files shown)


ewen-poch:
Addressed the P1 persistence regression. encodeDataPath now reuses the portable-suffix validation before emitting a $DATA sentinel; when a valid POSIX in-root filename cannot be represented portably (for example, it contains a backslash or a drive-like colon segment), the encoder preserves the absolute path instead of persisting a sentinel that decoding would reject. Added POSIX regression coverage. Targeted storage/Session/Preview/Notebook tests and the Node typecheck pass.

ewen-poch:
We are intentionally not adding a legacy migration for this case. The security contract for this change treats backslashes, drive-like prefixes, UNC or absolute paths, and parent segments as invalid in persisted $DATA sentinels on every platform. Accepting them conditionally on POSIX would make validation platform-dependent and weaken the shared containment invariant; rewriting them to absolute paths would introduce migration behavior outside this fix. Valid portable historical sentinels remain supported, while malformed or non-portable historical values fail closed. No code change is planned for this finding.

PR #1201, opened 2026-08-14
feat(cli): add plan and execution controls

ewen-poch:
## Problem

Headless and CLI callers had no public, typed way to control execution behaviors that already exist behind application/runtime owners:

- request a Plan-first turn and respond when the Plan reaches approval;
- enable or disable automatic Review for a Session;
- bind a new Session to a Specialist by stable id/name;
- deny admission of new delegated children.

This occurs when automation starts or resumes a Task through the HTTP/npm/CLI surface and needs one of those controls. Without a public control path, Plan-first automation cannot complete the approval handshake, headless auto-review is not reliably started, Specialist selection is unavailable, and delegation remains unrestricted. Scripts otherwise have to parse human text or depend on private renderer/IPC behavior.

## Proposed change

- Extend the typed Task start request with `turnIntent`, `autoReviewEnabled`, `specialist`, and `delegationPolicy`.
- Route Task Plan reads/responses and Reviewer run/reads through the transport-neutral application-command view. Keep Specialist lookup as a narrow, read-only resolver rather than exposing Specialist management.
- Add a Task-only `sessions:set-delegation-policy` application command so explicit policy changes enter the queued Session persistence owner instead of relying on whole-session projection saves.
- Add Plan read/respond HTTP and npm client methods plus:
  - `open-science plan show`
  - `open-science plan approve`
  - `open-science plan reject`
  - `open-science plan revise`
- Add CLI run controls:
  - `--plan-first`
  - `--auto-review` / `--no-auto-review`
  - `--specialist <id-or-name>`
  - `--delegation allow|deny`
  - `--return-on-attention`
- Keep `--wait` terminal-only by default. `--return-on-attention` explicitly returns a running Task Run only for structured Plan approval attention.
- Resolve Specialist references from UUID or stable Profile `name`; `displayName` remains presentation-only. An existing Session cannot be rebound to another Specialist.
- Enforce delegation denial at the production delegation admission seam. Existing children remain visible/collectable/stoppable and are not cancelled when the policy changes to `deny`.
- Trigger auto-review through the existing Reviewer application commands and project its terminal result on the Task Run.

## Architecture, data model, and interaction

All supported agent frameworks continue through the shared ACP Task adapter; there are no Claude Code, OpenCode, Codex Responses, or bridge-specific branches.

[...]

PR #1208, opened 2026-08-14
fix(command): tolerate concurrent catalog additions

ewen-poch:
## Problem

Renderer and preload contract tests pin absolute inventory sizes in addition to exact catalog, set, projection, and per-contract routing assertions. Valid command additions on concurrent PR branches therefore make unrelated synthetic merges fail only because a global count changed.

The runtime composition side is now structurally certified by #1203. This PR removes the remaining count-only test coupling that #1203 intentionally made unnecessary.

## Proposed change

- remove redundant global cardinality assertions from preload, renderer catalog, renderer surface, and Web argument-shape tests
- keep exact capability classification, catalog projection, set equality, uniqueness, omission, policy-subset, and per-contract routing assertions
- make count-neutral test names describe the behavior that remains under test

## Scope and non-goals

- test-only change; no production source or user-visible behavior changes
- no historical data compatibility is required
- no state or enum value is added
- no persistence or migration behavior changes
- this does not include the PR-specific module-impact omission in #1201 or stale reviewer expectations in #1105

All checks below ran after the last material edit and after rebasing onto current `main`.

- changed renderer/preload inventories plus app-command structural certification and module-manifest validation -> focused Vitest selection -> 8 files and 191 tests passed; `scripts/ci/classify-pr-changes.test.ts` has one current-main failure because `src/main/local-fs/service.ts` lacks a Windows runtime signal
- declared `settings_service_facade` module contracts and consumers -> direct Vitest execution of the 28-file module selection -> 27 files and 693 tests passed; 7 existing symlink tests in `src/main/settings/service.test.ts` could not create Windows symlinks and failed with `EPERM`
- main/preload TypeScript -> `tsc --noEmit -p tsconfig.node.json --composite false` -> passed
- renderer TypeScript -> `tsc --noEmit -p tsconfig.web.json --composite false` -> passed
- generated renderer API projection -> `node scripts/generate-web-api-map.mjs --check` -> passed
- modified-file lint -> ESLint on all four changed files -> passed
- patch integrity -> `git diff --check origin/main...HEAD` -> passed

The Windows module launcher emits the expected selective plan but fails before starting Vitest with `spawnSync npm.cmd EINVAL`; the exact emitted file list was run directly.

Uncovered risks: exact-head cross-platform and complete portable coverage remain authoritative in PR CI. The unrelated current-main Windows classifier gap and local symlink privilege failures are not changed here. Independent review is still pending.

## Review focus

Please verify that every removed count was redundant with an exact catalog/set/iteration assertion, and that capability-specific or behavior-specific exact expectations remain intact.

### DIFF
--- a/src/preload/index.test.ts
+++ b/src/preload/index.test.ts
@@ -612,8 +612,6 @@ describe('preload bridge — runtime renderer contract catalog', () => {
   it('routes every owned method through its cataloged Electron channel', async () => {
     const requestContracts = runtimeContracts.filter(({ kind }) => kind === 'method')
 
-    expect(runtimeContracts).toHaveLength(198)
-
     for (const contract of requestContracts) {
       invokeMock.mockClear()
 
@@ -678,7 +676,7 @@ describe('preload bridge — runtime renderer contract catalog', () => {
 })
 
 describe('preload bridge — core renderer contract catalog', () => {
-  it('pins the exact 23-group, 157-callable T1d complement', () => {
+  it('pins the core T1d capability complement', () => {
     expect(coreContractGroups.map(({ capability }) => capability)).toEqual([
       'artifacts',
       'cli',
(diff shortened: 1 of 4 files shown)


ewen-poch:
Regarding the P2 inventory-guard concern: I?m leaving the code unchanged because the independent guard is still present in `src/preload/index.test.ts`. The `pins every callable path exposed through window.api` test compares `collectFunctionPaths(api)` against an explicit, exhaustive path list that is not derived from the renderer catalog or generated Web maps. Removing a contract together with its derived preload/map entries without updating that independently authored list still fails. This PR removes only redundant cardinality assertions; adding another required-path manifest would duplicate the existing exhaustive manifest without improving deletion coverage.

ewen-poch:
Rebased this branch onto the latest main and resolved the four test-file conflicts. The resolution preserves the upstream settings.setReviewerModel inventory entry while retaining this PR's removal of redundant absolute cardinality assertions. Focused validation passed: 4 files / 94 tests, ESLint, Node and Web TypeScript checks, and the generated Web API map check.

PR #1220, opened 2026-08-14
fix(host-capabilities): require workspace for viewImage projection

ewen-poch:
## Problem

`capabilitiesCall` could advertise `host.viewImage` even though `viewImageCall` rejects the same control capability when it lacks a trusted Session workspace.

## Proposed change

Require `workspaceCwd` in the `viewImage` readiness projection, matching the call-time guard.

## Scope and non-goals

Only the local RPC capability projection and its focused tests change. No capability schemas, `host.help`, JS adapter, skills, architecture, data model, or UI changes are included.

- Active invocation + available visual service + no workspace projects `viewImage: false`.
- Adding a trusted workspace projects `viewImage: true`.
- An unavailable visual route remains fail-closed.

Checks run after the final edit:

- `npm test -- src/main/notebook/local-rpc-server.capabilities.test.ts`
- `npm run typecheck:node`
- `npm run lint`
- `npm test`

## Review focus

Confirm the readiness projection now mirrors the trusted workspace requirement enforced by `viewImageCall`.

### DIFF
--- a/src/main/notebook/local-rpc-server.capabilities.test.ts
+++ b/src/main/notebook/local-rpc-server.capabilities.test.ts
@@ -147,6 +147,100 @@ describe('capabilitiesCall RPC', () => {
     })
   })
 
+  it('does not advertise host.viewImage without a trusted Session workspace', async () => {
+    server = new NotebookLocalRpcServer({ execute: async () => ({}) } as never, {
+      transport: 'tcp',
+      hostViewImage: {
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed in f70bacd. `hostSdkHelp` now requires the same trusted `workspaceCwd` condition as the `viewImage` capability projection, and the focused capability suite covers the no-workspace help result as unavailable.

PR #1221, opened 2026-08-14
fix(host-capabilities): allow additive capability fields

ewen-poch:
## Problem

The JavaScript control-REPL adapter required an exact nine-key capability bitmap, preventing newer runtimes from returning additive capability fields and older runtimes from omitting later known fields.

## Proposed change

Validate a bounded plain-object boolean map with lowerCamelCase keys, reject invalid or dangerous keys, and return a fresh frozen copy that preserves valid additive fields. The Self-awareness Skill now describes the nine current v1 known keys, additive fields, missing keys, and `=== true` feature gating.

## Scope and non-goals

Only the control-REPL adapter, its integration contract, and the public Self-awareness Skill changed. This does not alter the local-RPC projection, `host.help`, the main-process capability catalog, architecture, persistence, data models, or UI.

- additive, missing-known, invalid-key, non-boolean, fresh, and frozen projection behavior → `RUN_KERNEL=1 npm test -- src/main/notebook/repl-loop.integration.test.ts` → pass
- Node process type surface → `npm run typecheck:node` → pass
- lint → `npm run lint` → pass
- portable suite after final edit → `npm test` → pass

All listed checks ran after the final material edit. The adapter is limited to the JavaScript control REPL; no ACP framework, launcher, model, or platform-specific behavior changed. Existing local-RPC authentication and fail-closed error handling remain unchanged.

## Review focus

Confirm that the adapter validates only the transport response and preserves the main-process capability catalog unchanged.

### DIFF
--- a/resources/notebook/repl_loop.js
+++ b/resources/notebook/repl_loop.js
@@ -979,17 +979,27 @@ async function hostMcp(server, method, args = undefined, kwargs = undefined) {
   return body.result
 }
 
-const HOST_CAPABILITY_NAMES = [
-  'mcp',
-  'compute',
-  'agents',
-  'skills',
-  'artifacts',
-  'lineage',
-  'frames',
-  'llm',
-  'viewImage'
-]
+const HOST_CAPABILITY_MAX_FIELDS = 64
+const HOST_CAPABILITY_MAX_KEY_LENGTH = 64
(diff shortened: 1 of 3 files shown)


ewen-poch:
Resolved in 0bdf492. The Skill now retains the existing `exactly nine boolean keys` contract wording while explicitly defining those as the current v1 known keys and documenting additive fields plus missing-key handling.

PR #1236, opened 2026-08-15
perf(notebook): bound run history payloads

ewen-poch:
## Problem

Notebook code, stdout/stderr/tracebacks, inline figures, run count, and the monolithic `run.json` can grow without a per-run capture bound. State reads then projected the complete merged history and renderer consumers rendered it all. A normal data run also rewrote the complete document for transient `running` and `idle` kernel status changes.

This becomes visible in long-lived or output-heavy sessions: kernel/IPC/main memory spikes, increasingly expensive JSON reads and rewrites, large IPC responses, repeated overlapping refreshes, and renderer image/text work that grows with the complete history.

## Proposed change

- Bound new executable code at 1 MiB and reject oversized source before kernel execution.
- Bound new run output to a shared 2 MiB text budget, 3.5 MiB per image, 12 images per run, and 8 MiB total image bytes per run. Execution continues and omitted content sets the existing `NotebookRunRecord.truncated` flag.
- Enforce limits in Python, R, REPL, and shell producers, then enforce them again at the persistence boundary. Figure files are size/count checked before base64 loading.
- Return the most recent 100 runs/cells from `state()` and add a non-persisted `runCount` total. Notebook views disclose the recent window and truncated output; images use lazy async decoding.
- Coalesce renderer state refreshes, cache unchanged v1 documents using file metadata, and select the recent merged window without materializing/sorting the complete run array.
- Keep ordinary `running -> idle` kernel status transitions in memory. A normal data/control run now performs the required running-record append and terminal-record update instead of two additional full-document status rewrites. Durable lifecycle transitions and recovery from `terminated` remain persisted.

## Scope and non-goals

- **Historical compatibility:** existing v1 `run.json` files are read unchanged. There is no eager migration, automatic rewrite, pruning, or cleanup of historical data.
- **Persistence format:** no `NotebookRunDocument` version change, sidecar, segment, table, or new persisted field. New runs continue using the existing inline v1 shape and existing optional `truncated` field.
- **Status model:** no run or kernel enum value is added.
- **Data model/API:** the persisted data model is unchanged. The renderer-facing in-memory `NotebookSessionState` adds `runCount`; `runs`/`cells` become a recent window.
- **Interaction:** Notebook preview and Session Notebook show the latest 100 merged runs and a “most recent N of total” notice. There is no load-earlier interaction in this change. Full export/provenance paths continue reading the complete durable history.
- **Retention:** total Notebook disk usage is still allowed to grow with run count. A strict total retention bound still requires a separately approved prune/segmented-storage policy.

Final checks were run after the last material edit:

- Capture, protocol, repository window/cache, terminalization, read model, and shell behavior -> targeted Notebook suite -> **124 passed, 1 platform-conditional skipped**.
- Runtime lifecycle, code rejection, state projection, architecture, export, and logging consumers -> targeted runtime suite -> **205 passed**.
- Notebook preview, Session Notebook, transcript hydration, output rendering, and locale guard -> targeted renderer suite -> **126 passed**.
- Node contract/type safety -> `npm run typecheck:node` -> **passed**.
- Renderer contract/type safety -> `npm run typecheck:web` -> **passed**.
- Lint -> `npm run lint` -> **0 errors**; 17 existing Prettier warnings remain in unchanged `host-sdk/capability-projection.test.ts` and `notebook/local-rpc-server.ts`.

Per maintainer direction, the complete `npm test` suite was not run locally; exact-head PR CI is authoritative for the full portable and platform matrix.

## Review focus

- UTF-8 boundary behavior and shared text budgeting across canonical and structured outputs.
- Producer-side draining after output limits and defensive persistence enforcement.
- v1 document-cache invalidation and the cross-lane latest-100 selection.
- The distinction between bounded renderer state and intentionally unbounded durable run retention.


ewen-poch:
Addressed both findings in commit a543f27a:

- Latest environment lookup: session state now includes a non-persisted latestRunEnvironments summary computed during the existing durable-history scan. Permission prompts prefer the live kernel, then this summary, then the bounded-window compatibility fallback, so environments outside the latest 100 runs remain discoverable. run.json is unchanged.
- Filtered run-count banner: removed the global count banner from NotebookPreview and SessionNotebookDialog, since both surfaces project runs by agent/kernel (and sometimes environment/branch) and the global count cannot accurately describe the visible list without broader scope-aware count metadata.

Validation: 88 targeted notebook/renderer tests plus 67 i18n guard tests passed; Node and Web typechecks and changed-file ESLint passed.

ewen-poch:
Follow-up on the latest review:

- Restart phantom environment: fixed in commit 1c644afc. Restart now updates only process keys that already exist. A session restarted before any kernel has spawned still reports the durable coarse restarting-to-idle transition, while environments remains empty. This adds no persisted field or state enum. The runtime-service module passes all 194 tests.
- Older history in notebook viewers: no change. The approved scope intentionally bounds renderer state to the latest 100 runs so a state read and render cannot grow with the full durable history. Older runs remain in run.json and exports, and transcript-referenced historical runs use targeted runIds hydration. Adding viewer pagination would introduce a new history API and interaction outside this PR.
- Coarse kernel status: no change. On main, ordinary running/idle updates from every process key are already persisted into the single coarse kernel status, so last-updated-environment semantics predate this PR. The change here preserves that behavior in memory while avoiding ordinary status rewrites. Defining and migrating default-only coarse semantics should be handled separately from history bounding.

ewen-poch:
Addressed the actionable follow-ups in commit 50aff2b6:

- Historical transcript hydration is now demand-driven. The renderer always keeps the normal latest-100 state window, but requests a full historical run only when the user expands that Notebook tool detail. Newly expanded runs are diffed against the existing session cache, targeted requests are capped and batched at 20 IDs, targeted responses no longer include the latest-100 window or cells, and collapsed historical entries use a 20-record LRU while currently expanded entries stay protected. Older transcript entries therefore still recover their complete persisted image output without eager full-history loading.
- REPL recovery now checks only the repl process status before persisting recovered idle, so a control run cannot clear an unrelated terminated Python/R status.

Validation: 273 focused main/renderer tests passed, along with Node and Web typechecks and changed-file ESLint.

The R temporary stdout-file finding is intentionally not changed. Hard-capping a sink while continuing to drain arbitrary R output requires another concurrent consumer or custom connection; spawning an additional R process per cell adds startup and process-management cost that is not justified here. Returned and persisted R output remains capped at 2 MiB and marked truncated, but temporary capture can exceed that during execution.

ewen-poch:
Addressed the actionable findings in commit d9084b51:\n\n- Durable terminated recovery: a recreated runtime now carries the existing persisted terminated status as a one-shot in-memory recovery marker. A successful data-kernel respawn persists the existing status back to idle and clears the marker. Live idle-shutdown status is handled as well. This adds no run.json field, schema version, or status enum, and the marker does not participate in live renderer status.\n- Bounded Session Notebook history: the dialog now consumes the existing runCount and explicitly states when it loaded only the recent window. Existing .ipynb downloads continue to read the complete durable history, so older runs remain accessible without pagination, a new history API, or a persistence change.\n\nThe coarse kernelStatus semantics and the accepted R temporary-capture tradeoff remain unchanged.\n\nValidation: 277 focused runtime/renderer/i18n tests passed; Node and Web typechecks, changed-file ESLint, formatting, and git diff checks passed.

ewen-poch:
Addressed the REPL recovery finding in commit 46d15065:\n\n- A recreated runtime now lets the first successful REPL execution consume the existing one-shot durable terminated marker, persist the existing coarse status back to idle, and clear the marker.\n- Live runtime termination remains process-aware: a known terminated Python/R process is not cleared by an unrelated REPL execution, and the existing regression test continues to cover that behavior.\n- The persisted format intentionally remains coarse, so no process-key field, schema version, compatibility migration, or status enum was added.\n\nValidation: all 197 runtime-service tests passed, including fresh-service REPL recovery and unrelated live data-kernel termination; Node typecheck, changed-file ESLint, formatting, and git diff checks passed.\n\nThe R temporary-file capture remains the explicitly accepted tradeoff and is unchanged.

ewen-poch:
Addressed the document-cache race in commit 865e8285:\n\n- Cache-miss reads now compare run.json identity (inode, mtime, and size) before and after reading.\n- If an atomic replacement lands during the read, the repository retries instead of pairing old parsed bytes with replacement metadata.\n- If the file changes again during the bounded retry, the valid snapshot is returned without caching, so a later read cannot remain pinned to stale content.\n\nAdded a deterministic regression test that supplies old bytes followed by replacement metadata and verifies the repository retries and caches only the replacement document.\n\nValidation: all 16 repository tests passed; Node typecheck, changed-file ESLint, formatting, and git diff checks passed. No persistence format, state enum, API, or interaction changed.

ewen-poch:
Addressed the protocol-sink finding in commit 9a718895:\n\n- The R loop saves the trusted original sink primitive and installs a guarded user-visible base::sink binding.\n- During a request, sink(NULL) can remove user-owned sinks above the notebook capture but cannot cross the active capture depth.\n- Trusted cleanup bypasses the guard, restores the original output-sink depth, closes the capture, and only then emits the JSON-lines response.\n\nA real-R regression test executes repeated base::sink(NULL) and sink(NULL), verifies the output remains captured, then verifies the next request still returns a valid protocol response. The R output-limit and diagnostic-reserve tests also pass.\n\nValidation: 3 focused real-R tests passed; Node typecheck, changed-file ESLint, R parse validation, formatting, and git diff checks passed. This adds no process, persisted field, status enum, API, or UI change.\n\nThe separate temporary-file growth finding remains the explicitly accepted tradeoff and is unchanged.

ewen-poch:
Addressed the bounded-history export discovery finding in 2626eab5. notebook.state now accepts an optional bounded historySummaryFrameId and returns a fixed-size, non-persisted per-Agent summary (total runs, per-kernel counts, and latest data kernel). Summary-only reads return no runs, recentRuns, or cells. The Session Notebook lists graph Agents even when they have no run in the latest-100 window, loads summaries on selection, caches at most 20 summaries, restores older-only kernel tabs and export controls, and explains that downloads include complete history. This does not change run.json, the database schema, IPC method registration, or any enum. Focused verification: 220 main notebook tests, 93 renderer/i18n tests, Node and Web typechecks, changed-file ESLint, and git diff --check all pass.

ewen-poch:
Addressed the two in-scope renderer findings in 33522fcc. Session Notebook now uses durable runCount, rather than the branch-filtered recent array, to decide whether the session is truly empty, so an older-only active branch still exposes Agent selection, on-demand history summaries, and complete-history exports. Frame-option construction now has an explicit includeUnloaded mode: Session Notebook enables it for historical discovery, while NotebookPreview keeps the prior behavior of showing only Agents with runs in its bounded window, so it never renders undefined counts or empty selectable panes. Focused verification: 47 Session Notebook/NotebookPreview renderer tests, Web typecheck, changed-file ESLint, formatting, and git diff --check pass.\n\nThe terminated-recovery recommendation is not changed because the persisted format stores only a session-wide coarse terminated value and cannot identify the terminated process key after relaunch. Making that exact requires a new persisted process-key field and compatibility decision, outside the approved no-format-change design; the current one-shot coarse recovery trade-off remains. The coarse kernelStatus recommendation is also unchanged: last-touched coarse semantics predate this PR, and changing it to default-only would be a separate state-semantic change. No field, enum, persistence, API, or interaction model was added by this follow-up.

ewen-poch:
Addressed the cross-kernel durable recovery finding in eb68960b with the approved persistence model. run.json kernel metadata now has an optional terminatedKernelInstances array of domain identities: Python/R entries store kind plus environment, while REPL stores kind only. It does not persist internal routing strings or operating-system PIDs. Idle shutdown and unexpected termination atomically add the exact instance; a successful respawn removes only the matching instance; coarse lastKnownStatus remains terminated while any instance remains; explicit restart clears the complete set. Multiple environments of the same language and Python/R/REPL instances can coexist.\n\nHistorical compatibility is conservative: old documents with lastKnownStatus terminated and no terminatedKernelInstances remain readable and are treated as unknown ownership. Ordinary execution cannot clear that legacy status; explicit restart clears it. The run.json version remains 1 and no status enum value was added. Writes remain limited to termination, matching recovery, and explicit restart rather than ordinary running/idle transitions.\n\nFinal local impact set: 232 repository/cache/aggregate/read-model/runtime/architecture tests passed; Node and Web typechecks, changed-file ESLint, formatting, and git diff --check passed. No full npm test was run, per the requested module-focused workflow.

ewen-poch:
Addressed both review findings in 6c425503.

- Runtime switch and both forced/background revocation teardown now clear the exact durable kernel-instance termination before dropping live process state. The cleanup is instance-scoped, so other Python/R environments remain untouched. Persistence failures propagate through explicit teardown instead of silently leaving a stale marker.
- The coarse `kernelStatus` once again represents `python:default-python`. Exact persisted termination records for named Python, R, and REPL kernels are merged into `environments[]`, including after relaunch. A legacy coarse `terminated` document with unknown ownership remains coarse-visible until explicit restart.
- Removed the now-unnecessary latest-kernel-status tracking from the session aggregate.

Validation: 7 focused test files / 239 tests passed; `typecheck:node`, `typecheck:web`, changed-file ESLint, Prettier, and `git diff --check` passed. The full test suite was intentionally left to PR CI.

ewen-poch:
Addressed both review findings in 750e70f8.

- Historical collapsed Notebook activities now recover their terminal phase from the authenticated compact Run result when the full Run is outside the hydrated window. This does not eagerly load historical outputs or image bytes; full images are still requested only when the row is expanded.
- SessionNotebookDialog now sends a bounded, non-persisted root-Branch ancestry filter to the state read. The repository applies it before computing the recent window, runCount, latest environments, and per-Agent history summaries. Parent-Branch Runs remain visible, inactive sibling-Branch root Runs are excluded, and delegated/legacy Runs remain conservatively visible.

No run.json/schema migration, persisted field, enum value, or user interaction was added. Focused validation passed: 358 tests, Node/Web typechecks, ESLint, Prettier, and git diff checks.

ewen-poch:
The reported `base::sink` bypass is not reproducible and does not require another namespace mutation. In R, `baseenv()` and `asNamespace("base")` are distinct environments, but the exported `sink` binding is linked: replacing the locked `baseenv()` binding also changes what `base::sink` and `get("sink", asNamespace("base"))` resolve to. A direct R check confirms both resolve to the replacement function after the assignment.

The existing real-R integration test already exercises the exact sequence `{ base::sink(NULL); sink(NULL); cat("still captured") }`, asserts that the text remains in the captured Run stdout, then executes a second request and verifies the JSON protocol remains aligned. I reran that test against 750e70f8 and it passed (1 passed, 67 skipped by the focused filter).

Explicitly unlocking and assigning the base namespace again would therefore be redundant and would add risk around R base bindings, so no code change is applied for this finding.

ewen-poch:
Addressed both review points:

- Historical notebook activities now recover their terminal phase from the compact app-owned run result when the full run is outside the hydrated window. Images remain expansion-only.
- Session Notebook history is intentionally session-wide and does not follow Conversation Branch selection. I removed the existing renderer Branch filtering and the temporary Branch query/filter implementation, so the displayed runs, runCount, history summaries, PreviewPanel, and exports share the same Session-level semantics. Agent Frame filtering remains unchanged.

No Branch query field, persisted field, schema change, enum, or migration remains. Focused validation passed: 335 tests across the affected main/renderer modules, Node/Web typechecks, ESLint, Prettier, and git diff checks.

ewen-poch:
Addressed both status-projection findings in aece31db:

- Live per-process kernel status now takes precedence over durable termination markers in both the coarse default-kernel projection and environments. Persisted markers still project terminated when no live status exists, covering relaunch/offline recovery.
- The compact execution result now preserves the existing app-owned executionInvocationId, including the global serialization fallback. Historical transcript status is trusted only when that identity exactly matches the Tool Activity; missing or stale identities fail closed. Existing historical tool results remain compatible and recover full state/output when expanded.

This adds no shared field, run.json field, schema version, enum, or migration. The compact Tool Activity result stores one additional existing UUID for new executions. Validation passed: 130 focused tests, Node/Web typechecks, ESLint, Prettier, and git diff checks.

ewen-poch:
Addressed the REPL serialization finding in 15c079fa. Final values whose JSON serialization returns undefined (functions and Symbols) now retain the protocol null result instead of passing through the output limiter and fabricating the literal string "undefined". Serialization failures such as circular values keep the existing string fallback.

A focused real repl_loop integration test covers both function and Symbol final expressions. It passed with the actual Node subprocess; Node typecheck, JavaScript syntax check, ESLint, Prettier, and git diff checks also passed.

PR #1237, opened 2026-08-15
fix(compute): bound harvests and preserve session identity

ewen-poch:
## Summary

- enforce application-owned harvest ceilings: 100 MiB per file, 500 MiB per job including stdout/stderr, declared outputs before logs, and a 2 GiB local free-space reserve
- stream every harvest through a byte-bounded SSH process; files that grow after enumeration are truncated locally, removed, and left remote, while missing bounded-copy support fails closed
- keep Compute queries, concurrency, IPC, and renderer contracts keyed by globally stable `sessionId`
- validate notification-consumption batches transactionally so missing, unnotified, or foreign-Session job IDs reject the whole batch
- reject live or durable cross-Project Session ID reuse and omit every duplicate from startup hydration, including identities represented by corrupt or quarantined files

## Problem and trigger conditions

Harvest could exceed the model-declared limits because stdout/stderr bypassed classification, the model could raise both limits, concurrent jobs could spend the same free space, and a remote file could grow after enumeration. A sufficiently large output or log could exhaust the local disk.

Compute notification consumption accepted arbitrary job IDs even though callers supplied a Session ID. Under valid data, `sessionId` is the global identity, so ordinary queries do not need `projectId`. Cross-Project leakage requires corrupted/imported data that reuses a Session ID, but arbitrary consumption was independently possible whenever a caller supplied another job ID.

## User interaction

No new UI or IPC shape is introduced.

- `host.compute.submitJob()` rejects malformed or above-ceiling harvest settings before approval, persistence, or SSH work.
- Outputs outside the hard limits or free-space budget stay remote and use the existing harvest error / left-on-remote surfaces.
- Ambiguous historical Session IDs are not hydrated. Existing startup diagnostics report every conflicting file as unreadable, while healthy Sessions still load.
- When startup hydration is incomplete, ordinary Session save asks the repository to assert the Session ID against the target persistence location. It rejects cross-Project reuse and fails closed if durable ownership cannot be scanned completely.
- Invalid notification-consumption batches reject atomically so the caller can retain and retry its notification state.

## Historical compatibility

- Persisted harvest settings above the new ceilings are clamped at read time; no row rewrite or backfill is performed.
- Existing duplicate Session IDs across Project directories are omitted and reported, but neither file is renamed, quarantined, deleted, or selected automatically. Resolving those files requires an explicit compatibility decision.
- Corrupt or already quarantined files continue to reserve their Session ID for their existing Project until explicitly repaired or removed.
- No renderer/main compatibility shim is needed because the existing `sessionId` call shape is retained.

## Persistence and data model

- no Prisma schema, column, index, or migration changes
- no new status or enum values
- no new durable fields; existing `ComputeJob.projectId`, `sessionId`, harvest, and notification timestamps remain authoritative
- `sessionId` is the stable global identity; `projectId` remains mutable ownership/location
- the durable identity fallback reads the existing `sessions/<projectId>/<sessionId>.json` path structure and does not rewrite files
- the repository exposes an ownership assertion rather than returning Project IDs, keeping file-layout and incomplete-scan details behind the persistence boundary
- identity verification runs inside the existing Session persistence coordinator consistency seam; no owner interface or completion gate is widened
- a future Project migration must be an explicit atomic operation that moves the Session authority and updates existing `ComputeJob.projectId` (plus other Project-owned projections) without changing `sessionId`; ordinary Session save intentionally cannot perform that migration

## Tests

- focused Vitest: 13 affected test files, 477 tests passed
- targeted ESLint: passed
- targeted Prettier checks: passed
- `git diff --check`: passed
- `npm run typecheck:node` reports only existing local dependency/generated-client issues outside this diff (stale Prisma `assessmentSnapshot` types and missing `sharp` / `i18next`); no changed-scope TypeScript errors
- full `npm test` intentionally not run per request; PR CI is the full-suite authority

### DIFF
--- a/resources/notebook/repl_loop.js
+++ b/resources/notebook/repl_loop.js
@@ -3007,6 +3007,24 @@ const hostCompute = {
                 maxFileMb: 'max_file_mb',
                 maxTotalMb: 'max_total_mb'
               })
+        const assertHarvestLimit = (field, value, maximum) => {
+          if (
+            value !== undefined &&
+            (typeof value !== 'number' || !Number.isFinite(value) || value < 0 || value > maximum)
+          ) {
+            throw new TypeError(
+              'host.compute.submitJob harvest.' +
+                field +
+                ' must be a finite number between 0 and ' +
+                maximum +
+                ' MiB.'
+            )
+          }
+        }
+        if (harvest !== undefined) {
+          assertHarvestLimit('maxFileMb', harvest.max_file_mb, 100)
+          assertHarvestLimit('maxTotalMb', harvest.max_total_mb, 500)
+        }
         return computeRpc({
           op: 'submit_job',
           provider_id: providerId,
--- a/resources/notebook/repl_loop.js
+++ b/resources/notebook/repl_loop.js
@@ -1010,9 +1010,9 @@ function isValidHostCapabilityProjection(value) {
   if (Object.getPrototypeOf(value) !== Object.prototype) return false
 
   const entries = Object.entries(value)
-  const knownEntries = HOST_CAPABILITY_KNOWN_KEYS.filter((name) =>
-    Object.hasOwn(value, name)
-  ).map((name) => [name, value[name]])
+  const knownEntries = HOST_CAPABILITY_KNOWN_KEYS.filter((name) => Object.hasOwn(value, name)).map(
+    (name) => [name, value[name]]
+  )
   const unknownEntries = entries.filter(([name]) => !HOST_CAPABILITY_KNOWN_KEYS.includes(name))
   return (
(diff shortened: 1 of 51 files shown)


ewen-poch:
Addressed the P1 harvest retry issue in e8c1d1b. Bounded downloads now write to a unique sibling .partial file and atomically rename it over the destination only after the transfer succeeds. Failure cleanup unlinks only the temporary file, so a previously successful local harvest remains intact. Added a regression test that writes a partial retry, returns a transfer failure, and verifies the prior output is preserved and the temporary file is removed. Focused compute validation: 6 test files / 211 tests passed, plus targeted ESLint, Prettier, and git diff --check.

PR #1241, opened 2026-08-15
fix(web): localize connection recovery copy

ewen-poch:
## Problem

The Web bootstrap initializes i18n before rendering, but its pre-React connection shell writes startup, reconnect, failure, and retry copy directly as English strings. Chinese browsers therefore keep English text during connection recovery.

## Proposed change

- Route the Web bootstrap connection shell through the initialized i18next instance.
- Add Simplified and Traditional Chinese catalog entries for connection, reconnect, remote-access failure, and generic failure copy.
- Include `src/renderer/web` in the i18n orphan-key guard.
- Add regression coverage for initial connection, retry backoff, remote-access failure, and the retry action.

## Scope and non-goals

This does not change bootstrap retry counts, backoff timing, event-stream recovery, or reload behavior.

The listed checks ran after the last material edit:

- Web bootstrap localization -> `npm test -- src/renderer/web/bootstrap.test.ts` -> 11 passed
- Catalog coverage, placeholders, and script purity -> `npm test -- src/renderer/src/i18n/resources.test.ts` -> 67 passed
- Web type safety -> `npm run typecheck:web` -> passed
- Lint -> `npm run lint` -> 0 errors; 17 pre-existing Prettier warnings in unrelated Main files
- Full portable suite -> `npm test` -> not green in this Windows environment: 12,311 tests passed and 49 failed; 155 suites could not load because Electron/native dependencies could not be built without Visual Studio C++ tools, with additional Windows symlink `EPERM` failures. The targeted renderer/i18n checks above are unaffected; exact-head CI remains authoritative for uncovered Main/native lanes.

## Review focus

Please verify that all pre-React connection-shell copy now uses the already-initialized i18n instance and that adding `renderer/web` to the orphan scan does not over-collect test-only strings.

### DIFF
--- a/src/renderer/src/i18n/resources.test.ts
+++ b/src/renderer/src/i18n/resources.test.ts
@@ -376,13 +376,14 @@ describe('Trans tag names', () => {
 // Orphan guard
 // ---------------------------------------------------------------------------
 
-// src/. Both the renderer and src/shared are scanned: shared modules have no i18n access, so they
-// carry English copy the renderer resolves through t() (install-source labels, download progress).
-// Leaving shared out is what let ten of these strings look orphaned when their call sites were live.
+// src/. Both renderer entry trees and src/shared are scanned: shared modules have no i18n access, so
+// they carry English copy the renderer resolves through t() (install-source labels, download
+// progress). Leaving either shared or renderer/web out makes live keys look orphaned.
 const SRC_ROOT = join(__dirname, '../../..')
 
 const SCAN_ROOTS = [
   join(SRC_ROOT, 'renderer', 'src'),
+  join(SRC_ROOT, 'renderer', 'web'),
   join(SRC_ROOT, 'shared'),
   join(SRC_ROOT, 'main')
 ]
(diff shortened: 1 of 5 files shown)


ewen-poch:
Regarding the P1 initialization concern: no change is needed. This repository passes both locale catalogs as inline resources and configures no async backend. With the pinned i18next 25.10.10, init completes synchronously when resources are present; a local same-stack probe confirms isInitialized is true, the language is zh-Hans, and the Chinese translation is available immediately after initI18n returns. Awaiting the current synchronous wrapper would not change behavior, while changing the shared i18n API would expand this Web bootstrap fix without addressing a real race.

ewen-poch:
Applied the extractor-alignment suggestion in 23075afec: the Web bootstrap now binds a local t function immediately after the synchronous initI18n call and uses t(...) for connection-shell copy. Runtime behavior is unchanged, while the existing i18n call-site guard can now detect future missing bootstrap translations. Focused verification passed: Web bootstrap 11/11, i18n resources guard 67/67, and typecheck:web. Lint reports 0 errors and 17 unrelated existing warnings. The full local suite remains environment-blocked by the missing Electron/native toolchain and Windows symlink permissions; CI module tests and E2E were green on the reviewed head.

PR #1242, opened 2026-08-15
fix(projects): prevent stale edits and improve recovery

ewen-poch:
## Problem

Project editing had four reliability gaps:

- ordinary edits did not carry a write baseline, so a stale window could overwrite a newer edit;
- Project name and description accepted unbounded input;
- Home exposed Project loading errors without a retry action;
- ArchivedPanel reused one error state, so a failed Project deletion could leak into the next confirmation dialog.

## Proposed change

- require `expectedUpdatedAt` for Project updates and compare-and-set ordinary edits against the persisted `updatedAt`;
- keep the edit dialog open and provide localized recovery guidance when another window wins the update;
- cap Project names at 200 characters and descriptions at 1000 in both the shared write contracts and form controls;
- add an in-place Home retry action with pending-state protection;
- isolate Project deletion dialog errors from panel/session errors and clear them at dialog boundaries.

## Scope and non-goals

- No database schema migration, backfill, or new revision column.
- Existing persisted Projects above the new limits remain readable and are not truncated; their next name/description edit must satisfy the new limits.
- No enum or status values are added.
- Pin-only writes retain their existing timestamp-preserving path.

- stale ordinary Project updates are rejected instead of overwriting a newer write;
- create/update contracts and UI enforce name <= 200 and description <= 1000;
- Home Project load failures offer Retry;
- a failed archived Project deletion error does not appear in a later deletion dialog.

Validation performed:

- 9 focused Project-related Vitest files: 106 tests passed
- `npm run typecheck:node`
- `npm run typecheck:web`
- i18n resources guard: 67 tests passed
- `npm run lint` (passes with 17 pre-existing formatting warnings outside this change)

Full test execution is left to PR CI per the requested module-scoped workflow.

## Review focus

- the `updatedAt` optimistic concurrency boundary in `ProjectRepository.update`;
- historical oversized Project read compatibility;
- dialog error-state lifetimes in ArchivedPanel.

### DIFF
--- a/src/main/data-content-application-commands.test.ts
+++ b/src/main/data-content-application-commands.test.ts
@@ -649,7 +649,7 @@ describe('Data and content application commands', () => {
     deps.sessions.loadAll.mockResolvedValueOnce(loadResult)
     deps.sessions.loadOne.mockResolvedValueOnce(loadedSession)
     registerDataContentApplicationCommands(router.registrar, deps.dependencies)
-    const updateRequest = { id: 'project-1', name: 'Updated project' }
+    const updateRequest = { id: 'project-1', name: 'Updated project', expectedUpdatedAt: 1 }
     const deleteProjectRequest = { id: 'project-1' }
     const manifestRequest = { lastProjectId: 'project-1', lastSessionId: 'session-1' }
     const deleteSessionRequest = { projectId: 'project-1', sessionId: 'session-1' }
--- a/src/main/projects/ipc.test.ts
+++ b/src/main/projects/ipc.test.ts
@@ -88,7 +88,7 @@ describe('createProjectHandlers', () => {
 
     await handlers.get('project-1')
     await handlers.create({ name: 'Project' })
-    await handlers.update({ id: 'project-1', name: 'Renamed' })
+    await handlers.update({ id: 'project-1', name: 'Renamed', expectedUpdatedAt: 2 })
 
     expect(order).toEqual(['recover', 'get', 'recover', 'create', 'recover', 'update'])
   })
@@ -107,7 +107,7 @@ describe('createProjectHandlers', () => {
     const handlers = createProjectHandlers(repository, deletionCoordinator)
 
     await handlers.create({ name: 'Research', agentContext: 'Always cite DOIs.' })
-    await handlers.update({ id: 'project-1', agentContext: 'Prefer Python.' })
+    await handlers.update({ id: 'project-1', agentContext: 'Prefer Python.', expectedUpdatedAt: 2 })
 
     expect(repository.create).toHaveBeenCalledWith({
       name: 'Research',
@@ -115,7 +115,8 @@ describe('createProjectHandlers', () => {
     })
     expect(repository.update).toHaveBeenCalledWith({
       id: 'project-1',
-      agentContext: 'Prefer Python.'
+      agentContext: 'Prefer Python.',
+      expectedUpdatedAt: 2
     })
   })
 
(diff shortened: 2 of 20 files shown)


ewen-poch:
Rebased this branch onto the current main and normalized the two locale catalog files with Prettier. This addresses the merge-result formatting failure in Static checks; no locale text or runtime behavior changed.

ewen-poch:
Addressed the retry/load concurrency finding: `project-store` now snapshots a module-local mutation generation for each list request. Any successful create, update, archive, delete, lifecycle upsert, or lifecycle removal invalidates an older list result, so a retry cannot replace a newer local mutation. A deterministic regression test covers project creation during an in-flight load.

The legacy over-limit behavior is intentionally unchanged. Existing rows remain readable without migration or truncation, but their next name/description edit must bring submitted values within the approved 200/1000 limits.

The pin-only path is also intentionally unchanged. Pin placement preserves `updatedAt` so administrative pinning does not alter research-activity ordering. Adding only an `updatedAt` predicate would not serialize pin-to-pin writes because pin updates deliberately keep that token unchanged; a complete solution would require a separate persisted revision or changing the activity-time semantics, both outside this PR's scope.

ewen-poch:
Addressed the initial-hydration follow-up: when a current list request discovers that a Project mutation invalidated its snapshot, it now awaits a fresh authoritative list instead of simply returning. This preserves newer mutations, completes initial hydration, and keeps the Home retry pending until reconciliation finishes. The regression now verifies a second list request, the created Project, and isLoaded=true. The repeated pin-only suggestion remains intentionally out of scope for the concurrency-token reasons described above.

ewen-poch:
CI follow-up: the macOS coverage shard repeatedly exceeded Vitest's default 2-second timeout in the unrelated session-store facade scan (2.017s, then 3.402s). I set a 10-second timeout only for that filesystem/AST architecture guard; production behavior is unchanged. Focused verification passed: 1 test passed, 146 skipped, and Prettier plus git diff checks are clean.

ewen-poch:
Correction to the previous CI note: the merged Vitest report shows this was an assertion mismatch, not a timeout. Main's #1245 added workspace-run-marks.ts as a direct session-store consumer without updating the facade inventory, so the PR merge result fails that unrelated guard. I removed the ineffective timeout change; the final PR diff no longer touches session-store.test.ts. The separate Windows failure is likewise unrelated (micromamba-cache-acl.integration.test.ts fallback ACL trust). No Project behavior or approved compatibility design was changed.

ewen-poch:
The final pin-only finding is intentionally not addressed in this PR. updatedAt is the research-activity timestamp and pin updates deliberately preserve it, so adding an expectedUpdatedAt predicate would still not serialize pin-to-pin writes. A complete fix requires an expected pin-state baseline or a separate persisted revision, which changes the Project contract/persistence design and is outside the approved scope. No further AI review iteration will be started after the configured review cap.

ewen-poch:
Rebased the branch onto the latest main so the PR merge result now includes the merged session-store consumer registration (#1249) and Windows cache ACL fix (#1250). The PR file diff remains scoped to the existing Project reliability changes. A fresh PR Gate is running against the rebased head; the synchronize-triggered AI review was cancelled to preserve the agreed five-review limit.

ewen-poch:
The fresh PR Gate is now running on the regenerated merge ref with latest main. Preflight, PR policy, Static checks, affected Module tests/coverage, and CI Integrity have passed. Windows and full-suite shards were correctly skipped by the impact classifier; macOS build/E2E is still in progress. No additional code changes were required.

ewen-poch:
The rebased head has completed successfully: PR Gate, CI Integrity, Static checks, affected Module tests/coverage, macOS build/E2E, and CodeQL all pass. The cancelled synchronize-triggered AI workflow produced no additional review. No further code changes were needed.

PR #1243, opened 2026-08-15
fix(storage): recover missing data root actions after IPC errors

ewen-poch:
## Problem

When the non-dismissable startup data-root guard calls storage IPC, a rejected Promise leaves `isRetrying` or `isChoosing` set. All three recovery actions then remain disabled, trapping the user in the dialog until the app is force-closed.

The same missing rejection handling affected retry, directory inspection, and applying a selected directory. A rejected directory-picker call also surfaced as an unhandled Promise.

## Proposed change

- contain storage IPC failures in both dialog actions
- reset operation loading flags in `finally`
- keep the dialog open, re-enable all actions, and show a recoverable inline error
- generalize the existing local choose error to one operation error
- add Simplified and Traditional Chinese translations for the new retry error
- cover rejected `getInfo`, directory-picker, `inspectDataRoot`, and `setDataRootAndRelaunch` calls

## Scope and non-goals

- no architecture or IPC contract changes
- no data-model, schema, field, or enum changes
- no persistence changes
- no historical-data compatibility behavior
- successful, still-missing, invalid-folder, cancellation, and relaunch flows remain unchanged

- missing-data-root recovery behavior -> `npm test -- src/renderer/src/components/DataRootMissingDialog.render.test.tsx` -> 14 passed
- renderer translation catalogs -> `npm test -- src/renderer/src/i18n/resources.test.ts` -> 67 passed
- Renderer types -> `npm run typecheck:web` -> passed
- linted source and tests -> `npm run lint` -> passed with 17 pre-existing warnings in unchanged files

The full portable suite was intentionally not run locally; exact-head PR CI is authoritative for broader consumer and platform coverage. The component has one caller in `App`, and no IPC interface or consumer contract changed, so no additional local consumer lane was included.

## Review focus

Please verify that every rejected storage IPC path returns the dialog to an actionable state and that the generic messages do not expose technical IPC details.


ewen-poch:
Addressed the P1 lifecycle issue in d0507bd5d.

After the setDataRootAndRelaunch IPC succeeds, the dialog now keeps the choosing state active so all recovery actions remain disabled until the renderer exits. Cancellation, invalid targets, rejected IPC calls, and { ok: false } results still restore the actions and show the existing error feedback.

Added a success-path assertion that all three actions remain disabled. The focused component suite passes (14/14), along with Prettier and typecheck:web.

PR #1248, opened 2026-08-15
feat(host-artifacts)!: align frame-scoped JavaScript contract

ewen-poch:
## What changed

- replace the public `host.artifacts({ sessionId })` filter with exact latest-producer `frameId` filtering
- keep Project ownership token-bound while excluding Uploads from Frame-filtered results
- return a fixed camelCase JavaScript projection, including nullable provenance fields and separate source/Version timestamps
- make `count` the total pre-pagination match count and bind cursors to all filters
- support exact MIME types and validated top-level prefixes such as `text/`
- batch Artifact/Upload provenance metadata reads without adding a new state owner or database schema
- update the JavaScript REPL adapter and bundled self-awareness Skill contract

## Breaking change

`host.artifacts` no longer accepts `sessionId`. Callers must use `frameId` for exact producer filtering. Result fields such as `projectId`, `latestVersionId`, `agentFrameId`, and `nextCursor` are camelCase.

## Scope

- current Project only; no caller-controlled Project override
- exact producer Frame only; no root/descendant expansion and no Session fallback
- no content grep, cross-Project search, Artifact marker, database migration, UI, or ACP framework-specific behavior
- all four ACP frameworks consume the same JavaScript REPL/local-RPC path

## Validation

- exact Frame isolation, latest-vs-historical producer behavior, Upload exclusion, MIME validation, total count, and cursor binding: `host-artifacts-service.test.ts`
- batched latest provenance and exact historical Version reads: `host-artifact-catalog.test.ts`
- token-bound Project/Session context and no synthesized Frame filter: `local-rpc-server.artifacts.test.ts`
- camelCase input/output projection, fixed nullable fields, forward-compatible unknown response fields, and deep freezing: `repl-loop.integration.test.ts`
- bundled Skill contract: `self-awareness-skill.test.ts`
- focused impact set: 7 files passed, 67 tests passed, 30 skipped
- Node typecheck: passed
- Web typecheck: passed
- lint: passed
- Prettier and `git diff --check`: passed
- independent Spec review: no findings
- independent Standards review: no findings

Full `npm test` completed with 1,047 files and 15,365 tests passing. Its only failure is an existing `origin/main` renderer contract mismatch: `session-store.test.ts` does not list the already tracked `workspace-run-marks.ts` direct store consumer. This PR does not modify renderer files, and that unrelated baseline fix is intentionally outside this PR's approved scope.

## Review focus

- confirm exact latest producer semantics for `frameId`
- confirm the public JavaScript projection is consistently camelCase
- confirm Catalog timestamps preserve source creation vs selected Version creation semantics
- confirm cursor and Project ownership remain fail-closed

### DIFF
--- a/resources/notebook/repl_loop.js
+++ b/resources/notebook/repl_loop.js
@@ -1352,17 +1352,21 @@ async function hostViewImage(source, options = undefined) {
 const HOST_ARTIFACT_REQUIRED_KEYS = [
   'id',
   'filename',
-  'size_bytes',
-  'latest_version_id',
-  'session_id',
-  'root_frame_id',
-  'is_user_upload',
-  'latest_version_created_at'
+  'contentType',
+  'sizeBytes',
+  'latestVersionId',
+  'checksum',
+  'projectId',
+  'sessionId',
+  'rootFrameId',
+  'agentFrameId',
+  'isUserUpload',
+  'createdAt',
+  'latestVersionCreatedAt'
 ]
-const HOST_ARTIFACT_OPTIONAL_KEYS = ['content_type', 'checksum']
 const HOST_ARTIFACT_INPUT_KEYS = {
   versionId: 'version_id',
-  sessionId: 'session_id',
+  frameId: 'frame_id',
   filename: 'filename',
   exact: 'exact',
   search: 'search',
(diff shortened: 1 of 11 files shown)


ewen-poch:
Addressed in f46d38ea. The catalog now preserves createdAt and sourceCreatedAt as Version/provenance timestamps, while projecting the parent ArtifactLineage or UploadFile creation time separately through sourceFileCreatedAt. HostArtifactsService uses that dedicated field and fails closed when it is missing. Lineage production behavior remains unchanged. Verified with Node and Web typechecks, lint, and seven focused suites: 68 passed, 30 skipped.

PR #1259, opened 2026-08-15
fix(reviewer): recover interrupted runs and bound results

ewen-poch:
## Problem

A hard Reviewer process/app interruption can leave a durable Review in `running` forever. Renderer load failures were silently discarded, and one `submit_findings` result had no count, field-size, or total-byte boundary.

## Proposed change

- Reconcile every durable `running` Review to the existing `error` lifecycle before Reviewer reads or starts new work.
- Enforce one Reviewer result as 1–5 checks, with 4,096-character claims, 16,384-character evidence, and a 256 KiB serialized UTF-8 ceiling.
- Apply the same limits at the MCP schema and repository persistence boundaries.
- Preserve cached review data when renderer loading fails, log the failure, and show a localized alert with a Retry action.

## Scope and non-goals

- No database columns, tables, constraints, migrations, or shared data-model fields are added.
- No lifecycle enum value is added; recovery uses the existing `error` state.
- Existing terminal or oversized historical records remain readable and are not truncated or migrated.
- Existing historical `running` rows are durably changed to `error` on the next startup.
- A historical re-review that requires more than five tracked checks can no longer submit one result under the new contract. This is the deliberate compatibility boundary for the strict five-check result limit.
- No full local `npm test` run; PR CI is authoritative for the complete and platform-selected suites.

All listed successful checks ran after the final production-code edit.

- Startup recovery and the read barrier → Reviewer resilience + IPC tests → passed.
- MCP and repository limits → Reviewer MCP/repository/resilience tests → passed.
- Renderer error preservation and retry surface → review-store + MessageScroller render tests → passed.
- Translations → i18n resources guard → passed.
- Combined focused set: 7 files / 240 tests passed.
- Final MessageScroller SSR test after its test-only assertion was added: 65 tests passed.
- Node and web TypeScript checks passed.
- ESLint on all changed TypeScript/TSX files passed after fixing its only formatting warning; the final SSR test edit was subsequently Prettier-formatted and exercised by Vitest.

The project module runner produced the expected `reviewer_orchestrator` plan but failed before test execution with Windows `spawnSync npm.cmd EINVAL`. The existing local dependency tree then became unavailable, so the wider CodeGraph-indicated indirect consumers were left to PR Gate/CI. Locally uncovered risks are those indirect consumer lanes and cross-platform startup/persistence behavior.

## Review focus

- Recovery is atomic, idempotent, and blocks Reviewer reads/new runs if reconciliation fails.
- The five-check boundary is enforced before any persistence path.
- Renderer errors are session-scoped and successful retry clears only that session's error.

### DIFF
--- a/src/main/reviewer/bridge-tools.ts
+++ b/src/main/reviewer/bridge-tools.ts
@@ -2,7 +2,7 @@ import { z } from 'zod'
 
 import { REVIEWER_MCP_SERVER_NAME, REVIEWER_MCP_TOOLS } from '../../shared/reviewer'
 import type { ResponsesBridgeNamespacedTool } from '../settings/responses-bridge'
-import { submitFindingsInputSchema } from './mcp-server'
+import { submitFindingsBridgeInputSchema } from './mcp-server'
 
 export const REVIEWER_BRIDGE_TOOL_NAMESPACE = `mcp__${REVIEWER_MCP_SERVER_NAME.replace(
   /[^a-zA-Z0-9_]/g,
@@ -41,7 +41,7 @@ export const REVIEWER_BRIDGE_NAMESPACED_TOOLS: ResponsesBridgeNamespacedTool[] =
     name: REVIEWER_MCP_TOOLS.submitFindings,
     description:
       'Submit at least one structured pass, warn, or fail check exactly once, then stop. An empty checks array is invalid.',
-    parameters: z.toJSONSchema(submitFindingsInputSchema, {
+    parameters: z.toJSONSchema(submitFindingsBridgeInputSchema, {
       target: 'draft-7'
     }) as ResponsesBridgeNamespacedTool['parameters']
   }
--- a/src/main/reviewer/ipc.test.ts
+++ b/src/main/reviewer/ipc.test.ts
@@ -42,12 +42,14 @@ vi.mock('./orchestrator', () => ({
 // invoke it directly and observe which storageRoot the thunk was constructed against.
 const reviewRepositoryThunks: Array<() => unknown> = []
 const getReviewsForSession = vi.fn().mockResolvedValue([])
+const recoverInterruptedReviews = vi.fn().mockResolvedValue(0)
 vi.mock('./repository', () => ({
   ReviewRepository: class {
     constructor(thunk: () => unknown) {
       reviewRepositoryThunks.push(thunk)
     }
     getReviewsForSession = getReviewsForSession
+    recoverInterruptedReviews = recoverInterruptedReviews
     getReviewsForProjectSession = getReviewsForSession
   }
 }))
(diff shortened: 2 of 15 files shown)


ewen-poch:
Static checks found that the new SSR test called `createSession()` without its required overrides argument. Updated it to `createSession({})`, matching the existing helper signature. This is test-only and does not change production behavior, architecture, persistence, state, or interaction design. The PR CI rerun is the authoritative verification.

ewen-poch:
Addressed P1 by making the recovery gate retryable while retaining single-flight and fail-closed behavior per attempt. A failed recovery now rejects the current command and clears only the failed gate; the next Reviewer command starts a new recovery attempt. Added a failure-then-success regression test.

For P2, I am retaining the current behavior intentionally. A reviewer:updated push proves that one live update arrived, but it does not prove that the complete persisted history loaded successfully. Clearing the history-load alert on a push could hide missing older reviews. The alert therefore clears only after a successful getForSession reload.

ewen-poch:
Updated the per-result Reviewer limit from 3 to 5 checks. The MCP schema and repository boundary both consume the shared constant; focused boundary coverage now accepts 5 and rejects 6. This does not limit consecutive Reviewer runs and does not change persistence, lifecycle states, the data model, or user interaction.

ewen-poch:
No code changes were made for the latest AI review findings because every cited path is outside this PR diff: src/main/compute/ipc.ts, src/main/agent-framework/claude-code.ts, the connector and skill settings slices, PreviewFileSurface.tsx, and SettingsPage.tsx. PR #1259 changes only Reviewer recovery/result-boundary files plus the corresponding renderer error surface, tests, and locale entries. These findings are therefore not regressions introduced by this branch and should be tracked separately if still reproducible. The requested 5-check adjustment remains limited to the shared Reviewer boundary and its focused test.

ewen-poch:
Rebased the branch onto the latest origin/main without conflicts and updated the PR with --force-with-lease. The Reviewer patch scope remains unchanged: the same 12 files and four logical commits were replayed; no architecture, data-model, persistence, lifecycle-state, or interaction changes were introduced by the rebase.

ewen-poch:
A further main commit landed while the previous AI review was running, so I rebased once more onto the current origin/main tip (890c0589f) without conflicts and updated the branch with --force-with-lease. The branch is now 0 commits behind main and the Reviewer diff remains unchanged.

ewen-poch:
Addressed the tracked re-review compatibility finding. A re-review now accepts the exact required tracked disposition set plus up to five newly discovered checks. The MCP schema derives its allowance from the server-owned trackedFindingIds set, and the repository derives the same allowance from expectedSourceFindingIds; the existing exact-set validation still rejects missing, duplicate, or invented tracking ids. Claim/evidence limits and the 256 KiB serialized limit continue to cover the complete submission. Added MCP and SQLite repository regressions for six historical tracked findings and for rejecting a sixth new finding. Focused Reviewer tests pass (3 files, 63 tests), and ESLint passes on all five changed files.

ewen-poch:
Addressed the bridge compatibility issue in 2be646897. The Responses bridge now publishes a bridge-specific submit_findings schema without a static maxItems value, since that declaration has no per-run tracked-finding context. The runtime ReviewerMcpServer schema remains authoritative and still enforces all tracked dispositions plus at most five new checks, along with the existing field-length and total-byte limits. Added a regression assertion for the bridge schema; the targeted reviewer/bridge suite passes (4 files, 102 tests).

PR #1260, opened 2026-08-15
fix(settings): recover rejected optimistic actions

ewen-poch:
## Problem

Connector and Skill toggles were projected optimistically, but rejected Main-process writes neither rolled the renderer state back nor triggered an authoritative refresh. Connector approvals were removed from the local queue before Main confirmed the response, so an IPC failure left no local retry path.

## Proposed change

- Track confirmed and pending boolean generations inside the renderer slices so failed writes restore confirmed state without an older failure overwriting newer intent.
- Keep Connector approvals queued until Main confirms the response.
- Disable approval decisions while a response is pending and show a localized retryable error after failure.
- Consume expected toggle rejections at UI call sites after the store has visibly reconciled them.
- Add regression coverage for rollback, overlapping requests, confirmation ordering, and retry interaction.

## Scope and non-goals

- Renderer-only change.
- No shared or IPC contract changes.
- No schema, persisted field, data-model, or enum additions.
- No historical-data migration or compatibility path is required.
- Existing settings.json persistence and Main's in-memory approval queue are unchanged.

- Rejected Connector enabled, auto-allow, custom-server, and Skill enabled writes restore the last confirmed value.
- Older rejected writes cannot overwrite a newer successful intent.
- Approval requests remain visible while submitting and after failure, then disappear after Main confirmation.
- Retry controls are disabled only while a response is in flight, with Simplified and Traditional Chinese error copy.

Validation after the final material edit:

- npm test -- settings-connectors-slice.test.ts settings-skills-slice.test.ts settings-store.test.ts ConnectorApprovalDialog.render.test.tsx ConnectorDetailView.render.test.tsx ConnectorsPanel.render.test.tsx SkillDetailView.render.test.tsx SkillsPanel.render.test.tsx resources.test.ts — 11 files, 333 tests passed
- npm run typecheck:web — passed
- npm run lint — 0 errors; 15 pre-existing Prettier warnings in src/main/host-sdk/capability-projection.test.ts
- git diff --cached --check — passed

The module-impact planner falls back to the complete suite because locale catalogs have no module owner mapping. Per task scope, the complete local npm test was not run; exact-head PR CI is authoritative.

## Review focus

- Generation settlement when same-field writes resolve out of order.
- Approval retention and retry behavior across IPC rejection.

### DIFF
--- a/src/renderer/src/locales/zh-Hans.json
+++ b/src/renderer/src/locales/zh-Hans.json
@@ -460,6 +460,7 @@
   "Could not sign in to Codex.": "无法登录 Codex。",
   "Could not sign out of Claude.": "无法退出 Claude。",
   "Could not sign out of Codex.": "无法退出 Codex。",
+  "Could not submit this approval. Try again.": "无法提交此批准。请重试。",
   "Could not switch to this folder.": "无法切换到该文件夹。",
   "Could not switch to {{name}}": "无法切换到 {{name}}",
   "Could not test the provider connection.": "无法测试模型服务商连接。",
--- a/src/renderer/src/locales/zh-Hant.json
+++ b/src/renderer/src/locales/zh-Hant.json
@@ -459,6 +459,7 @@
   "Could not sign in to Codex.": "無法登入 Codex。",
   "Could not sign out of Claude.": "無法登出 Claude。",
   "Could not sign out of Codex.": "無法登出 Codex。",
+  "Could not submit this approval. Try again.": "無法提交此核准。請重試。",
   "Could not switch to this folder.": "無法切換到該資料夾。",
   "Could not switch to {{name}}": "無法切換到 {{name}}",
   "Could not test the provider connection.": "無法測試模型服務商連線。",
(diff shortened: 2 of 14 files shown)


ewen-poch:
Addressed the stale-toggle ordering issue in 02fd2e5c4.

- Optimistic writes and authoritative catalog projections now share one issuance sequence.
- Connector reconciliation and all Skill catalog/import reconciliation paths record the projection generation consistently.
- An older toggle settlement can clear its pending entry but cannot overwrite a newer authoritative snapshot.
- Added regression coverage for a newer Connector refresh and a newer Skill bulk snapshot completing before an older toggle.

Validation: 335 targeted tests passed, typecheck:web passed, and ESLint passed for the five changed files.

ewen-poch:
Addressed the stale approval-error issue in 02cce17a9.

- Submission failures are now associated with the request ID captured when the response starts.
- The alert renders only when that ID matches the currently displayed approval.
- A late failure or queue advance therefore cannot show the previous request's error on the next approval.
- Added a regression test that fails a response, advances the queue, and verifies that the alert is cleared.

Validation: 336 targeted tests passed, typecheck:web passed, and ESLint passed for the changed component and test.

PR #1268, opened 2026-08-16
fix(specialists): recover catalog load failures

ewen-poch:
## Problem

The Specialist renderer store left `isLoaded` false when the initial catalog IPC request rejected, so the Settings panel stayed on its loading state with no error or recovery action. Catalog reads also committed unconditionally, allowing an older overlapping request to overwrite a newer snapshot.

## Proposed change

- Record a path-safe transient catalog load error and consume startup failures inside the store.
- Show an accessible Settings error with an explicit Retry action while preserving the fail-closed `isLoaded` contract.
- Fence every catalog read, including post-mutation refreshes, with one latest-request generation.
- Add store and render regression coverage for failure recovery, overlapping loads, and load-versus-mutation refresh races.
- Add Simplified and Traditional Chinese translations for the new user-visible message.

## Scope and non-goals

- No persistence, schema, migration, IPC, or shared domain-model changes.
- No new enum values and no historical-data compatibility requirements.
- Existing mutation rejection behavior remains unchanged.
- A failed refresh after a successful load preserves the last known catalog while exposing Retry.

- `npm test -- src/renderer/src/stores/specialist-store.test.ts src/renderer/src/pages/settings/SpecialistsPanel.render.test.tsx` — 46 tests passed.
- `npx vitest run src/renderer/src/i18n/resources.test.ts` — 67 tests passed.
- `npm run lint` — passed with 0 errors; it reported 15 existing Prettier warnings in the untouched `src/main/host-sdk/capability-projection.test.ts`.

The complete portable and platform suites remain delegated to PR CI as requested.

## Review focus

- Confirm `isLoaded` remains false after an initial failure so Specialist-dependent barriers do not fail open.
- Confirm the shared generation covers both ordinary loads and mutation-triggered catalog refreshes.
- Confirm stale successes and failures cannot replace the latest visible catalog state.

### DIFF
--- a/src/renderer/src/locales/zh-Hans.json
+++ b/src/renderer/src/locales/zh-Hans.json
@@ -1094,6 +1094,7 @@
   "Open Open Science from the signed-in mobile app with two-step verification.": "从已登录的移动应用使用两步验证打开 Open Science。",
   "Open Provenance for {{title}}": "打开 {{title}} 的 Provenance",
   "Open Science confirms its core requirements before your first research session.": "Open Science 会在你的第一次研究会话前确认其核心依赖。",
+  "Open Science could not load Specialists. Retry to continue.": "Open Science 无法加载 Specialist。请重试。",
   "Open Science couldn't remove the unused copy. Normal work has resumed. You can delete the copy later.": "Open Science 无法删除未使用的副本。正常工作已恢复，你可以稍后删除该副本。",
   "Open Science couldn't start": "Open Science 无法启动",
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed the P1 catalog-load contract issue in b80ec4f8e. load() now returns the original refresh promise, so awaited callers still observe rejection and cannot treat stale catalog data as fresh. The same promise has an internal rejection handler for existing fire-and-forget startup/event callers, avoiding unhandled rejections while preserving the Settings error/retry state and request-generation fencing.

Validation: the focused Specialist store/panel suite passes (46 tests), and typecheck:web passes.

PR #1269, opened 2026-08-16
feat(host-model): add model introspection

ewen-poch:
## Problem

The JS control REPL could run `host.llm()` but could not reliably inspect either the model owned by the calling Session or the validated model catalog available to the active Host LLM route. Using global settings for both questions would mix two different sources of truth and could report a model that the Session had not actually applied.

## Proposed change

- Deepen `HostLlmService` into a shared `HostModelService` that owns inference and model introspection behind one interface.
- Add zero-argument `await host.currentModel()`:
  - resolves the caller through the authenticated, token-bound Session;
  - reads the owning runtime instead of request-supplied identity or later global settings;
  - requires ACP-confirmed `appliedModel` for OpenCode;
  - rejects missing, blank, and `provider-default` model projections.
- Add zero-argument `await host.listModels()`:
  - uses the existing configured model catalog without a network refresh;
  - restricts results to the active Host LLM Provider and framework;
  - requires successful Provider validation with no newer validation failure;
  - returns a frozen, deduplicated, lexically sorted string array.
- Add independent `currentModel` and `listModels` capability flags, including the case where introspection is available but tool-less `llm` is not.
- Add strict JS KERNEL result validation, `host.help()` descriptors, self-awareness guidance, and synchronized control-method inventories.

## Scope and non-goals

This PR does not add `reasoningModel`, live model-list network calls, model override, new `host.llm()` inputs, tools, images, compatibility aliases, persistence, database/schema changes, renderer UI, or ACP wire-protocol changes. The camelCase `host.llm()` result contract from #1263 is preserved.

- Focused 12-file Host Model impact set after rebasing onto current `origin/main`: 206 passed, 30 skipped.
- Node typecheck after rebase: passed.
- Full Node/Web typecheck, lint, and complete `npm test` before the conflict-free rebase: passed.
- `git diff --check`: passed.
- Independent standards review: no findings.
- Independent spec review found and verified fixes for OpenCode applied-model trust and Provider validation freshness; final re-review reported no regressions.

## Review focus

- Trusted Session ownership and the OpenCode `appliedModel` rule in `currentModel()`.
- Provider/framework and validation-state filtering in `listModels()`.
- Independent capability truth when `llm` is unavailable.
- Strict zero-argument JS Adapter behavior and fail-closed unknown-model handling.


ewen-poch:
Addressed in 62c5044a. `AcpRuntime.captureSessionModel()` now requires an active Session attachment before exposing backend model facts, so a detached registry aggregate cannot leak a stale model after provider disconnect. Added an attached → disconnected → resumed runtime regression test. Focused validation: 5 files, 655 tests passed; Node typecheck, targeted lint, Prettier, and diff checks passed.

PR #1275, opened 2026-08-16
fix(notebook): wait for R interrupt acknowledgement

ewen-poch:
## Problem

On POSIX, R cancellation can return a normal response before SIGINT is consumed. The current executor releases the queue on any matching response, so a late interrupt can land on the next cell: the cancelled cell reports cancelled, the first following cell reports failed, and a later cell succeeds.

## Proposed change

- Have the R loop return an ephemeral `interrupt_ack` only when its interrupt handler consumes SIGINT.
- Use an unmaskable `base::Sys.sleep` probe; a successful probe also confirms user code already consumed SIGINT.
- Prequeue a side-effect-free internal R probe when SIGINT is sent, retaining queue ownership until the original request or probe acknowledges the interrupt.
- Preserve the existing 2-second fallback that drops and reaps an unresponsive process.
- Cover both cancellation and soft-timeout races, plus POSIX R integration behavior and namespace retention.

## Scope and non-goals

- No database, schema, persistence, status-enum, UI interaction, or data-relationship changes.
- No historical migration; existing incorrect historical Run records are not rewritten.
- The Windows fallback is unchanged: the process is terminated/restarted and its namespace is not retained.

All commands below were run after the last material edit.

- Cancelled R execution cannot leak its interrupt into the next two executions, while the persistent namespace is retained: focused executor tests passed (3 tests), including masked `Sys.sleep` and user-caught interrupts.
- R protocol accepts the optional acknowledgement field without changing existing responses: protocol suite passed (11 tests).
- The installed R loop still executes and auto-prints visible results: targeted integration smoke test passed (1 test, 41 skipped by filter).
- Node type checking passed: `npm run typecheck:node`.
- Lint passed: `npm run lint`.
- Formatting and whitespace checks passed for changed files: Prettier check and `git diff --check`.

Per the requested focused-test policy, the full `npm test` suite was not run locally. An attempt to run the entire kernel executor file on Windows was not authoritative because 23 cases require symlink privileges and one case is POSIX-path-specific; the exact cancellation and timeout regression cases pass. PR CI is authoritative for the complete suites and POSIX behavior.

## Review focus

- The R queue is released only after explicit interrupt acknowledgement.
- The internal probe does not mutate the user namespace.
- The existing unresponsive-process fallback remains intact.
- The protocol addition is optional, internal, and non-persistent.


ewen-poch:
Addressed both AI review findings in 3d8399173:

- the private probe now calls `base::Sys.sleep(0.05)`, so a user binding cannot mask it;
- a successful trusted probe is accepted as confirmation that user code already consumed SIGINT, while an errored probe still retries until the existing grace fallback;
- regression coverage now includes masked `Sys.sleep`, user-caught interrupts, two post-cancel executions, timeout behavior, and namespace retention.

Focused executor tests (3), protocol tests (11), R loop smoke, Node typecheck, lint, Prettier, and diff checks pass after the change.

ewen-poch:
Addressed the follow-up fixture finding in 103875cdb. The simulated masked probe now clears its pending interrupt after the one failed invocation and invalidates the namespace, so the regression still detects an unsafe `Sys.sleep` probe without depending on an infinite error loop. The three focused R cancellation/timeout cases pass after this change.

PR #1279, opened 2026-08-16
fix(persistence): surface catalog repair before project actions

ewen-poch:
## Problem

An incomplete Session catalog prevents safe Project archive operations, but the current UI can present the resulting rejection as a generic Project operation failure. Unfinished Project deletion recovery and quarantined damaged Session authority also need distinct guidance.

## Proposed change

- derive a renderer-only catalog recovery state from existing startup diagnostics
- present repairable catalog failures as Project index repair before an archive attempt
- keep quarantined damaged Session authority distinct because retry cannot restore it
- keep unfinished Project deletion recovery distinct from index repair
- apply the same recovery guidance and deletion gates to Home and Settings > Archived
- map Electron-wrapped incomplete-catalog archive errors to localized index-repair guidance

## Scope and non-goals

- no database or storage schema changes
- no persisted fields, persisted enum values, IPC contracts, or shared data-model changes
- no migration or historical-data rewrite; existing diagnostics are classified on the next load
- the new recovery union is transient renderer state only
- quarantined damaged Session files still cannot be automatically restored; whole-Project permanent deletion remains available when deletion recovery is otherwise ready
- Project activity history is intentionally out of scope because it requires a separate durable event model and migration

- partial Session scans and startup reconciliation failures show Project index repair guidance
- quarantined damaged authority explains why archive remains unavailable without promising repair
- unfinished Project deletion recovery shows recovery-specific guidance and gates deletion
- Simplified and Traditional Chinese catalogs cover every new user-visible string

Final Test Impact Set, run after the last material edit:

- session persistence diagnostics and recovery classification: session-persistence.render.test.tsx
- shared recovery presentation: SessionPersistenceAlert.render.test.tsx and SessionCatalogRecoveryAlert.render.test.tsx
- Home archive race and action guidance: HomePage.persistence.test.tsx, HomePage.render.test.tsx, and home-dialogs.render.test.tsx
- archived Project and Session deletion gates: ArchivedPanel.render.test.tsx and SettingsPage.render.test.tsx
- App routing and alert precedence: App.test.tsx
- localization contracts: i18n/resources.test.ts
- renderer interfaces: npm run typecheck:web
- source validation: npm run lint

The 10 targeted test files passed with 253 tests. Web typecheck and lint passed. The complete portable and platform suites are intentionally left to exact-head PR CI.

Uncovered local risk: independent review and full cross-platform execution are not reproduced locally; PR CI remains authoritative.

## Review focus

- whether the diagnostic precedence correctly distinguishes repairable indexing, quarantined authority, and unfinished deletion recovery
- whether permanent Project deletion remains available only in the intended quarantined-authority case
- whether the recovery copy accurately describes the available action without implying a Project Files index repair

### DIFF
--- a/src/renderer/src/App.test.tsx
+++ b/src/renderer/src/App.test.tsx
@@ -1120,6 +1120,33 @@ describe('App startup routing', () => {
     expect(mocks.sessionPersistence.retryWrites).toHaveBeenCalledOnce()
   })
 
+  it('keeps failed writes retryable while catalog recovery is visible', async () => {
+    mocks.settings.isLoaded = true
+    mocks.sessionPersistence.hasCompleteSessionCatalog = false
+    mocks.sessionPersistence.catalogRecovery = { kind: 'repairable', reason: 'session-scan' }
+    mocks.sessionPersistence.writeError =
+      'Open Science could not save the latest conversation changes. Retry before closing the app.'
+
+    await render()
+
+    const alerts = Array.from(
+      container.querySelectorAll('[data-testid="session-persistence-alert"]')
+    )
+    expect(alerts).toHaveLength(2)
+    expect(alerts[0]?.textContent).toContain('Project index needs repair')
+    expect(alerts[1]?.textContent).toContain('Conversation storage needs attention')
+
+    const retries = Array.from(
+      container.querySelectorAll<HTMLButtonElement>('[data-testid="session-persistence-retry"]')
+    )
+    const repairIndex = retries.find((button) => button.textContent === 'Repair index')
+    const retryWrites = retries.find((button) => button.textContent === 'Retry')
+    repairIndex?.click()
+    retryWrites?.click()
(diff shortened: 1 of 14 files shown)


ewen-poch:
Addressed P1 in f7de993e4.

- The catalog recovery alert now keeps the failed-write alert and retryWrites action visible when both conditions coexist.
- Added an App regression test that verifies both Repair index and write Retry remain independently actionable.
- Final checks after the change: App.test.tsx (49 tests), npm run typecheck:web, and npm run lint.

P2 is intentionally unchanged. The startup isComplete diagnostic can represent a global reconciliation failure without a reliable project identifier, so project-scoped gating cannot be derived safely for every incomplete-catalog case. Keeping the renderer fail-closed globally matches the approved scope. Main still performs its per-project archive scan as the authoritative race-time check. Introducing project-scoped authority or enabling archive before that check would be a separate interaction and state-model decision.

PR #1285, opened 2026-08-16
test(e2e): update session recovery contracts

ewen-poch:
## Problem

Nightly run 31931817712 and PR Gate diagnostics showed that the visual and accessibility recovery journeys still waited for the retired Saved conversation data was damaged alert title. The application correctly renders Project archive needs attention for damaged project authority, so both journeys timed out before their assertions.

## Proposed change

- Update both E2E recovery locators to the current catalog recovery alert title.
- Replace the Darwin recovery snapshot with the canonical macOS actual from PR Gate run 31934023354 after confirming the completed-session notification was dismissed.

## Scope and non-goals

- No product behavior changes.
- No accessibility scan rules are relaxed.
- No other visual baselines are changed.
- The earlier failed-run screenshot containing an extra notification was not used.

## Acceptance criteria and validation

- Canonical artifact SHA256: CEC392BB43C7990A009EC407018D8B99AB476400628E10F1F33E15F335133E46.
- git diff --check passed.
- Local build:e2e, typecheck:web, and lint were not run because this independent worktree has no node_modules; dependencies were not installed or linked.
- Exact-head PR Gate is authoritative for build, typecheck, lint, accessibility, and macOS visual execution.

## Review focus

Confirm both journeys target the current damaged-authority alert and the Darwin snapshot matches the reviewed canonical macOS artifact.

Historical compatibility: none.
New state or enum values: none.
Persistence impact: none.

PR #1286, opened 2026-08-16
fix(uploads): wait longer for transaction connections

ewen-poch:
## Problem

Windows Full Test run 31931817779 failed the two-attachment upload repository test with Prisma P2028. Parallel attachment publication shares the project SQLite clients single connection, while Prisma interactive transactions default to a two-second connection acquisition wait.

## Proposed change

- route all three uploads interactive transactions through one shared transaction helper
- set the uploads-only acquisition maxWait to 10 seconds
- keep attachment hash/copy work and attachment publication parallel
- add an architecture contract that requires the single policy entry point and all three transaction call sites

## Scope and non-goals

This does not change the global Prisma client, serialize upload file work, change the UploadRepository interface, or relax any test.

Historical compatibility: none.
New state or enum values: none.
Persistence behavior: when the single SQLite connection is busy, uploads transactions may wait up to 10 seconds rather than two seconds. There is no schema, migration, data-format, or storage-path change.

Test Impact Set:

- transaction policy and three call sites -> npm run test:module -- upload_repository -> not run locally because this worktree has no node_modules; PR Gate is authoritative
- original two-attachment SQLite path -> npm test -- src/main/uploads/repository.test.ts -t "registers each upload as an independent SQLite file with immutable v1 bytes" -> not run locally for the same dependency reason
- affected main-process types -> npm run typecheck:node -> not run locally for the same dependency reason
- linted source -> npm run lint -> not run locally for the same dependency reason
- patch integrity -> git diff --check -> passed after the final material edit

The changed files are owned by upload_repository. UploadRepository itself and its consumer-facing interface are unchanged, so session_persistence consumer tests are excluded. The Windows-sensitive behavior remains covered by PR CI; an exact-head Windows Full run has intentionally not been triggered yet.

## Review focus

Please verify that 10 seconds applies only to uploads interactive transaction acquisition and that the transaction bodies and parallel file work are otherwise unchanged.

### DIFF
--- a/src/main/uploads/legacy-recovery-owner.ts
+++ b/src/main/uploads/legacy-recovery-owner.ts
@@ -16,6 +16,7 @@ import {
 import type { PersistedChatMessage, PersistedChatSession } from '../../shared/session-persistence'
 import {
   OrphanLegacyUploadAuthorityMissingError,
+  runUploadTransaction,
   type PublicationOptions,
   type UploadVersionRecord
 } from './staged-publication-owner'
@@ -384,7 +385,7 @@ class LegacyRecoveryOwner {
       }
       if (!sourcePath || !(await validateContent(sourcePath))) {
         const client = await this.options.getClient!()
-        await client.$transaction(async (tx) => {
+        await runUploadTransaction(client, async (tx) => {
           await tx.uploadVersion.deleteMany({ where: { id: version.id, state: 'staging' } })
           await tx.uploadFile.deleteMany({
             where: { id: version.uploadFileId, versions: { none: {} } }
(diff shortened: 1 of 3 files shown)


ewen-poch:
Exact-head Windows Full validation: https://github.com/aipoch/open-science/actions/runs/31935986637 on head 0aa45061ca863b20a2e6f04f8dc0ec3b9d75f4b0. Shard 2, including the original uploads double-attachment/P2028 case, completed successfully; P2028 did not recur. Shard 1 also passed. Overall failure came from an independent shard 3 delegation timing/cleanup race in production-composition.test.ts, now tracked separately. PR Gate full module shards, coverage, Windows core, and Windows E2E also passed; its remaining failures are current-main projectName drift and the recovery locators covered by PR #1285.

PR #1292, opened 2026-08-16
feat(notifications): explain attention and retain deleted targets

ewen-poch:
## Problem

Notification inbox rows explain that something needs attention, but not the bounded reason that produced the request. The bell therefore cannot use the same three waiting labels as Home. Session deletion also physically removes inbox rows, so bounded history disappears instead of remaining visibly invalid.

## Proposed change

- Persist an optional `attentionReason` for the three shared Session wait categories and four bounded task-stop categories.
- Add `agent-runtime` as the source for task terminal events; never persist raw provider stop text.
- Replace Session-target row deletion with idempotent invalidation. Invalidation preserves `actionState`, stamps `targetInvalidatedAt`, and marks unread rows read.
- Make invalidation take precedence over approval replay and navigation in the bell, with translated "Session no longer available" copy.
- Add immutable migration `0007_notification_attention_metadata`, including the source/reason CHECK constraints and a table rebuild that preserves identity, dedupe, sequence boundaries, indexes, and AUTOINCREMENT behavior.

## Scope and non-goals

- This shares the three waiting labels, not a combined source + subreason taxonomy across inbox and Home.
- Home continues to project current Session actionability and does not read inbox state.
- Session dots, Home activity, and tray unread algorithms remain separate.
- Delegated-question inbox production is deferred; no speculative source value is added.
- Native notification summary copy is unchanged.

## Data, enum, and compatibility notes

- Persistence: adds nullable `NotificationInboxItem.attentionReason` and `targetInvalidatedAt` columns plus CHECK constraints.
- Enum/type changes: adds notification source `agent-runtime` and notification-only attention reasons. No `PersistedSessionStatus`, `NotificationKind`, or `NotificationActionState` value is added.
- Historical compatibility: existing rows migrate with both new columns `NULL`; there is no attention-reason backfill. The first complete Session catalog reconciliation invalidates and acknowledges rows whose Session no longer exists.
- Fail-closed behavior is retained: a historical raw-SQL source outside the bounded source set prevents 0007 from being recorded and leaves the pre-migration database intact, with a retained backup.

- notification production, invalidation, read semantics, and late-event protection -> targeted Vitest impact set -> 647 tests passed across 22 files
- SQLite 0006 -> 0007 identity, `sequence`, mark-all boundary, index, CHECK, backup, and next AUTOINCREMENT behavior -> migration tests in the same impact set -> passed
- shared and renderer consumers -> `npm run typecheck` -> passed
- renderer translations and labels -> i18n, bell, Home, and workspace tests in the impact set -> passed
- generated runtime schema -> `npm run db:schema:check` -> passed
- packaged migration identity -> `node scripts/database-migration-ledger-smoke.mjs` -> passed
- lint -> `npm run lint` plus final changed-import lint -> passed

The final checks ran after the last material edit on top of `origin/main` at `9a4e814f`. Per request, the complete local `npm test` suite was not run; exact-head PR CI is authoritative for the complete portable and platform lanes.

## Review focus

- `0007` table-rebuild preservation of `sequence`, `id`, `dedupeKey`, all five indexes, and subsequent AUTOINCREMENT.
- `targetInvalidatedAt` precedence over pending approval replay and navigation while preserving the original `actionState`.
- The intentional boundary between historical inbox reason metadata and Home's current Session projection.


ewen-poch:
Addressed the stop-reason lookup finding in f2ae27c1. The repository currently sets noImplicitAny to false, so the prior expression passed both local and exact-head Node typechecks, but indexing the literal-key map with arbitrary provider text still lost useful static typing. The map is now explicitly typed as Readonly<Record<string, NotificationAttentionReason>>, preserving the bounded values and the task-unclean-stop fallback. Verified with the 71 task-notification tests, Node typecheck, and focused lint.

PR #1294, opened 2026-08-16
feat(github): add contextual star prompts

ewen-poch:
## Problem

The GitHub entry points on Home and Workspace are visually passive, and Workspace users receive no contextual encouragement to star the project.

## Proposed change

- Give Home and Workspace distinct, token-aligned GitHub Star treatments while keeping the star outline hollow.
- Show a dismissible Workspace encouragement popover five seconds after entering each project.
- Keep the popover visible for thirty seconds, pause dismissal while hovered or focused, and restart the timer when entering another workspace.
- Use a continuous 1.4-second star twinkle with reduced-motion support.
- Add Simplified and Traditional Chinese translations for the new copy.

## Scope and non-goals

- Renderer-only UI and interaction change; no main-process, preload, database, schema, or migration changes.
- No analytics, star-state tracking, or persistent dismissal state.
- The compact GitHub entry point used outside Home and Workspace keeps its existing treatment.

- Star badge rendering, 5-second delay, 30-second visibility, workspace wiring, Home rendering, and i18n contracts -> targeted Vitest command -> 6 files and 169 tests passed.
- Repository lint rules -> npm run lint -> passed.
- Responsive popover and animation -> browser checks at 320, 375, 414, and 768 CSS pixels -> no horizontal overflow; helper and CTA remain single-line; animation reports 1.4s infinite.

The changed surface is renderer-only. Main/preload/platform lanes are excluded locally because no associated contracts or persistence paths changed. Full portable, build, accessibility, visual, and platform coverage remains authoritative in PR CI.

## Review focus

- Whether the five-second entrance delay and thirty-second visibility feel appropriately noticeable without interrupting work.
- Whether resetting the prompt from projectId matches the intended per-workspace cadence.
- Home and Workspace visual fit in light and dark themes.


ewen-poch:
Addressed both P2 findings in 91cfd15f:

- The Workspace nudge now waits until its anchor is non-inert and has layout before starting the five-second delay. It rechecks visibility when the delay expires and resumes waiting if the sidebar was hidden again, so opening a previously hidden sidebar starts a fresh five-second delay instead of losing the prompt.
- The Workspace popover now uses z-[80], matching the existing mobile overlay/menu layer above the z-[70] sidebar.

Added focused fake-timer coverage for the hidden-to-visible timing path. The affected tests pass (129 tests), along with web typecheck and lint.

ewen-poch:
Addressed the P1 lifecycle issue in 5d780509.

While the portaled nudge is open, it now monitors the owning anchor using the same visibility predicate as the delayed-open path. If the sidebar becomes hidden or inert, the popover closes within 500 ms and the monitor is cleaned up with the open state. This keeps the portal tied to the sidebar/app-shell presentation lifecycle without adding persisted state or changing the prompt cadence.

The hidden-to-visible test now also verifies that an already-open nudge stays open while its owner remains visible and closes after the owner becomes hidden. The affected tests pass (129 tests), along with web typecheck and lint.

ewen-poch:
Addressed the desktop collapsed-sidebar lifecycle issue in 5d4a97ce.

WorkspaceSidebar now derives the active nudge key from its existing presentation state: mobile requires isMobileOpen, and desktop requires sidebarToggle.state to be non-collapsed. Closing or collapsing the sidebar clears the effective key and remounts the badge in its closed, timer-free state; reopening restores the project key and starts a fresh five-second delay.

This also makes the explicit sidebar state authoritative instead of relying on zero-width panel geometry. The workspace component contract test now guards the active-key wiring. The affected tests pass (129 tests), along with web typecheck and lint.

PR #1301, opened 2026-08-16
fix(session-delete): keep failures out of run state

ewen-poch:
## Problem

Session deletion failures currently leak into run-state semantics: a failed delete can call `failRun` and project the ChatSession/Agent state as `error`. The renderer also receives only a boolean, so it cannot distinguish runtime-detach failures from persistence failures or provide a retry-safe user flow.

## Proposed change

- Replace the renderer deletion boolean with a structured result that distinguishes runtime and persistence failures and records whether the runtime is detached.
- Keep deletion failures out of ChatSession/Agent run state and retain diagnostic logging without exposing raw exceptions.
- Reuse one deletion controller/dialog flow on desktop and mobile.
- Keep the confirmation dialog open while deleting, prevent duplicate submission, and close/remove the Session only after success.
- Show an accessible, safe error in the same dialog with retry and cancel/close actions, including for background Sessions.
- Preserve the Session, draft, and attachments after failure; persistence retries skip an already-detached runtime idempotently.
- Add Simplified and Traditional Chinese translations for all new user-visible copy.

## Scope and non-goals

This PR intentionally does not move the deletion transaction into Main, change the persisted Session schema or disk format, add migrations, or address filename/session-id and Notebook Project/document Project work.

Passed on commit `73786a6a65f5740460f288b377086a7c00ec6575`:

- `npm test -- src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts src/renderer/src/lib/acp/useWorkspaceAgentRuntime.architecture.test.ts src/renderer/src/pages/workspace/workspace-session-controller.test.ts src/renderer/src/pages/workspace/session-dialogs.render.test.tsx src/renderer/src/pages/workspace/WorkspacePage.draft-preservation.test.tsx src/renderer/src/pages/workspace/workspace-page.architecture.test.ts src/renderer/src/i18n/resources.test.ts src/renderer/src/stores/session-store.test.ts` — 450 tests passed
- `npm run typecheck:web`
- `npm run lint`
- Prettier check and `git diff --check`

The full `npm test` was also attempted on the exact HEAD and remains red in this Windows environment with 58 failures. The failures are dominated by symlink `EPERM`, unusable temporary storage folders, settings-suite timeouts, and two architecture-test timeouts under full-suite contention. The affected tests, including those architecture tests, pass when run in the targeted set above.

## Review focus

- Structured deletion-result semantics and idempotent persistence retry
- No coupling between deletion failure and run-state failure
- Shared desktop/mobile dialog lifecycle and background-Session error visibility
- Preservation of drafts and attachments on failure


ewen-poch:
Addressed the P1 archived-session deletion race in 7e0bc8533. ArchivedPanel now forwards its existing per-session busy state to DeleteSessionDialog, so confirm, cancel, close, and dialog dismissal are guarded while the async deletion is pending. Added a regression assertion for the shared-dialog wiring. Targeted tests (22), typecheck:web, lint, Prettier, and diff checks pass.

PR #1303, opened 2026-08-16
fix(session-delete): centralize terminal deletion in main

ewen-poch:
## Summary

- centralize terminal Session deletion in a Main-owned command
- keep runtime-first teardown, Project ownership validation, concurrent deduplication, and durable deletion behind one interface
- return a structured deletion result and broadcast `session:deleted` only after both runtime and persistence deletion succeed
- migrate Workspace and Archived Settings deletion to the same command and remove the Renderer-owned sequencing helper

## Behavior

- runtime failure leaves persistence and visible Session state intact
- persistence failure reports that runtime is detached; retrying the same command is safe
- deletion dialogs remain open on failure and expose the existing retry interaction
- lifecycle events remain the sole owner of Renderer store removal

## Data and architecture impact

- no database migration or new database fields
- no Session JSON or persistence schema changes
- no new user-visible copy or i18n entries
- preload only changes the delete command return type to the shared structured result

## Verification

- `422` changed/affected tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run check:web-api-map`
- `git diff --check`

The full local `npm test` run was also attempted. The Windows environment cannot create the symlinks or set the ACLs required by several unrelated filesystem tests and produced resource-related timeouts. All failures related to this change were isolated, fixed, and included in the passing affected test set above.


ewen-poch:
Addressed the P1 session-deletion routing issue in 1c63cddb7. The transport-neutral data-content command now delegates sessions:delete-session to the existing SessionDeletionOwner, returns its SessionDeletionResult, and publishes session:deleted only when the result is deleted. The application-command layer no longer wraps runtime teardown in the data-root write gate; durable deletion remains write-scoped inside the owner adapter, preserving runtime-first ordering. Regression coverage verifies production wiring, successful result propagation, and suppression of lifecycle events on failed deletion.

ewen-poch:
Addressed the duplicate channel registration P1 in 6ad8ce9b0. The dedicated Session deletion IPC adapter and its pass-through tests were removed; sessions:delete-session is now registered exactly once by the existing runtime-validated Application Command Electron adapter. Electron, local Web, and remote Web therefore reuse the same command composition and SessionDeletionOwner. The production wiring guard now asserts that no direct Session deletion IPC registration can be reintroduced while the command dependency continues to delegate to the owner.

PR #1308, opened 2026-08-16
fix(reviewer): load artifact source session reviews

ewen-poch:
## Problem

Reviewer history is hydrated by the active Session's message scroller. Opening an Artifact from another Session can therefore reach a reviewer preview before that source Session has loaded. The renderer previously collapsed an absent Project + Session snapshot and a loaded empty snapshot to the same empty array, so the preview incorrectly reported that no Review was available.

## Proposed change

- Preserve the distinction between an unloaded review snapshot and a loaded empty snapshot.
- Let the reviewer preview load the target Artifact source Session on demand.
- Render a localized loading state while that snapshot is unavailable.
- Render the existing load failure message with a Retry action when loading fails.
- Keep the existing empty state after a successful load genuinely returns no reviews.

## Scope and non-goals

- No database, Prisma schema, migration, or persisted data changes.
- No IPC contract changes; this reuses the existing reviewer Session read API.
- No Review, Artifact, Session, or preview item model changes.
- No new status enum values.
- No historical-data migration is required.

Checks below ran after the last material production edit. The focused suite and lint were rerun after the final test-only edit.

- Unloaded source Session no longer renders the empty state -> `npm test -- src/renderer/src/pages/workspace/previews/PreviewToolContent.test.tsx src/renderer/src/stores/review-store.test.ts src/renderer/src/i18n/resources.test.ts` -> 3 files, 92 tests passed.
- Review store and reviewer consumer contract -> `npm run test:module -- reviewer_orchestrator` test set, executed directly from `scripts/ci/module-impact.json` because the Windows launcher returned `spawnSync npm.cmd EINVAL` -> 27 files, 453 tests passed.
- Workspace preview consumers -> `npm run test:module -- workspace_page` test set, executed directly from `scripts/ci/module-impact.json` for the same Windows launcher limitation -> 25 files, 445 tests passed.
- Source and test lint -> `npm run lint` -> passed with no warnings.

The complete portable and platform suites remain authoritative in PR CI; a full local `npm test` was intentionally not run.

## Review focus

- Confirm that map-key absence is the correct existing signal for an unloaded Session snapshot.
- Confirm that loading and retry remain owned by the reviewer preview rather than Artifact provenance or the persisted preview model.

### DIFF
--- a/src/renderer/src/locales/zh-Hans.json
+++ b/src/renderer/src/locales/zh-Hans.json
@@ -882,6 +882,7 @@
   "Loading notebook…": "正在加载 Notebook……",
   "Loading permissions": "正在加载权限",
   "Loading preview…": "正在加载预览…",
+  "Loading review history…": "正在加载审查历史…",
   "Loading project files…": "正在加载项目文件……",
   "Loading remote access…": "正在加载远程访问…",
   "Loading saved conversations…": "正在加载已保存的对话…",
--- a/src/renderer/src/locales/zh-Hant.json
+++ b/src/renderer/src/locales/zh-Hant.json
@@ -880,6 +880,7 @@
   "Loading notebook…": "正在載入 Notebook……",
   "Loading permissions": "正在載入權限",
   "Loading preview…": "正在載入預覽…",
+  "Loading review history…": "正在載入審查歷史…",
   "Loading project files…": "正在載入專案檔案……",
   "Loading remote access…": "正在載入遠端存取…",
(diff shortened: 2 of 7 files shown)


ewen-poch:
Addressed the push-only history gap in c8c3c07c2. The renderer store now tracks successful session-history loads independently from review-array presence. Push updates remain provisional, so they cannot suppress the initial persisted-history fetch; the existing merge keeps newer pushed data while adding older persisted reviews. Added a regression test covering push-before-load behavior. This remains renderer-only with no persistence, schema, or IPC changes.

PR #1336, opened 2026-08-17
feat(connectors): generate IDs from names

ewen-poch:
## Problem

Custom Connector local IDs were always random UUIDs and hidden, unlike Specialist IDs. Users could not preview or customize the stable local identity.

## Proposed change

- Infer new Connector local IDs from the immutable Connector name using the shared Specialist ID rules.
- Fall back to a UUID when an inferred value is unsafe, reserved, or already used.
- Add an Advanced Connector ID preview with manual override and immediate validation; keep it immutable on edit.
- Make the main process authoritative for generation and enforce atomic cross-ID/name uniqueness, including concurrent creation fallback.
- Extract the shared resource-ID rules while preserving the existing Specialist API and behavior.

## Scope and non-goals

- No template schema change; templates continue to use name and omit local ID.
- No migration; historical Connector IDs are preserved.
- No new persisted field or schema version.
- No lifecycle or persisted enum addition; the editor only adds local touched state.
- Persistence impact: for newly created Connectors, the existing ID field now defaults to a name-derived ID, with UUID fallback, instead of always using a UUID.

- Connector ID preview, manual override, immediate validation, UUID fallback, and immutable edit display: targeted renderer tests, 24 passed.
- Authoritative generation, uniqueness, concurrency, historical handling, and architecture guard: targeted main settings tests, 155 passed.
- Shared rules, Specialist compatibility, and i18n guards: targeted shared/renderer tests, 117 passed.
- Node and Web typechecks, lint, and diff check passed.

The main settings service/repository, renderer form, shared Specialist consumer, and translation guards were covered. Platform packaging and E2E were left to PR CI because this change does not introduce platform-specific window or build behavior. Per request, the full npm test suite was not run locally.

## Review focus

- Untouched, manually edited, cleared, and UUID-fallback ID semantics.
- Cross-ID/name uniqueness and concurrent creation.
- Preservation of historical IDs and continued exclusion of local IDs from exported Connector templates.


ewen-poch:
Addressed in c1bc5180.

Custom Connector deletion now prunes its permission grants before removing the persisted Connector. If pruning fails, deletion is aborted, leaving the Connector present and the permission state fail-closed. This removes the interruption window that could leave grants behind for a name-derived ID.

The UUID-only startup reconciliation heuristic remains unchanged intentionally: broadening it to every valid resource ID would make deleted custom IDs indistinguishable from unknown or newer bundled catalog IDs and could prune app-owned grants. The workflow regression test now locks the prune-before-persist order and verifies that a prune failure prevents deletion.

Validation: 81 targeted permission/settings tests, Node typecheck, lint, and diff check passed.

ewen-poch:
Addressed in ced15f99.\n\nConnector deletion now atomically removes the Connector and records its local ID in the optional pendingCustomServerDeletionIds journal before permission-grant pruning. The ID remains reserved until pruning succeeds. On startup, the app consumes this authoritative journal before refreshing Connector Skills, including name-derived IDs, and clears each entry only after successful pruning. Because recovery uses deletion metadata instead of guessing from the resource-ID format, bundled and app-owned grant IDs remain untouched.\n\nCoverage includes journal sanitization, failure retention, ID reservation with UUID fallback for same-name recreation, recovery ordering, and runtime refresh after a persisted deletion even when pruning fails.\n\nValidation: 192 targeted main-process tests passed; npm run lint, npm run typecheck:node, npm run typecheck:web, Prettier, and git diff --check passed.

ewen-poch:
Addressed both findings in 11870212.\n\n- Client eviction: the journaled deletion callback now closes the cached MCP client before pruning permission grants. A reset or prune failure rejects cleanup without clearing the persisted tombstone, so the deleted ID cannot be reused; the runtime projection still refreshes to remove the deleted Connector.\n- Pending-ID validation: main projects pending tombstones as the renderer-safe reservedCustomServerIds field. The Connector store preserves that projection, and ConnectorAddForm includes those IDs in its existing used-ID set. An untouched inferred collision now previews/submits the stable UUID fallback, while an explicit collision reports the existing immediate validation error.\n\nNo new endpoint, persisted field, status enum, or user-visible copy was added.\n\nValidation: 230 targeted main/renderer tests passed; npm run typecheck:node, npm run typecheck:web, npm run lint, Prettier, and git diff --check passed.

ewen-poch:
Addressed the Codex finding “Clear stale runtime failure state when deleting a custom Connector” in c5ba4522.\n\nThe journaled deletion callback now clears ConnectorService failure/availability state after the cached client closes and before permission grants are pruned. If client reset fails, later cleanup does not run and the tombstone remains reserved; if pruning fails, the already-cleared runtime failure state cannot leak into a future Connector while the tombstone still prevents ID reuse.\n\nValidation: 70 targeted workflow/settings tests passed; npm run typecheck:node, npm run lint, Prettier, and git diff --check passed.

ewen-poch:
Addressed in e0bafbcc.

`removeCustomServer` now treats a requested ID as cleanup-eligible when it is still present in the pending-deletion journal, even if the Connector record has already been removed. It reruns the existing cleanup callback and clears the tombstone only after that callback succeeds.

The regression test covers a failed initial cleanup, a same-name Connector recreated under a UUID while the original ID remains reserved, and a successful same-process retry that clears only the tombstone without deleting the replacement Connector.

Validation: 230 targeted main/renderer tests passed; `npm run typecheck:node`, `npm run typecheck:web`, `npm run lint`, Prettier, and `git diff --check` passed.

PR #1337, opened 2026-08-17
fix(approvals): restore pending approval lifecycle

ewen-poch:
## Problem

Connector and Compute approval brokers keep pending requests in Main memory, while the Renderer keeps only an ephemeral projection of those requests. If the Renderer reloads, crashes, or is rebuilt while Main remains alive and an approval is pending, the new Renderer subscribes only to future requests and never recovers the broker-held request.

That leaves the underlying Connector/Compute call parked until another path replays the request or the approval times out. Compute approval submission also had no submitting state, visible failure state, or double-submit guard, so an IPC failure was silent and repeated clicks could race.

## Proposed change

- Add broker-level replay of every currently pending Connector and Compute approval.
- Keep replay projection side-effect-free so Renderer recovery does not duplicate native approval notifications or attention requests.
- Add replay-pending application commands and preload/Web contract entries.
- Broadcast settled approval IDs so recreated Renderer queues can discard approvals resolved by response, timeout, cancellation, or invalidation.
- Subscribe to request and settled events before requesting replay during Renderer startup.
- Deduplicate replayed requests and make settled removal idempotent in both stores.
- Give Compute approvals the same disabled/submitting, inline error, and retry behavior as Connector approvals.
- Feature-detect the new preload methods so a Renderer hot reload remains compatible with an older loaded preload/Main pair.

## Scope and non-goals

- This extends the existing broker -> application event -> preload/Web -> Renderer projection flow; it does not add a new architectural layer.
- No database schema, persisted fields, migrations, or durable approval records are added.
- No domain data model or approval decision/state enum values are added or changed.
- No historical-data compatibility work is required because there is no persisted approval data to migrate.
- Full Main-process restart recovery remains out of scope: Main restart destroys both the in-memory approval and the parked Promise continuation, so supporting it requires a separate durable workflow design.
- Interaction changes are limited to automatic recovery/removal of pending cards and Compute submission feedback/guarding.

The final focused checks ran after the last material edit.

- Pending replay, notification-side-effect isolation, settlement, application commands, event projections, preload/Web contracts, Renderer recovery, store projection, Compute busy/error/retry, stale scope-confirmation races, and i18n guard -> final targeted Vitest impact set -> **24 files / 539 tests passed**.
- Compute owner module -> `npm run test:module -- compute_service` produced the declared list but its Windows launcher returned `spawnSync npm.cmd EINVAL`; running that exact Vitest list directly -> **594 passed / 15 skipped**, with the updated architecture contract rerun and passing.
- Settings owner module -> `npm run test:module -- settings_service_facade` hit the same Windows launcher issue; running its exact list directly -> **708 passed**, while **7 unrelated symlink tests could not create Windows test symlinks (`EPERM`)** before entering product logic.
- Node process contracts -> `npm run typecheck:node` -> **passed**.
- Renderer/preload contracts -> `npm run typecheck:web` -> **passed**.
- Source style -> `npm run lint` -> **passed**.
- Full `npm test` was intentionally not run; PR CI is authoritative for the complete portable and platform lanes.

Uncovered local risk: the Windows symlink-only settings tests require an environment permitted to create symlinks. No persistence, migration, build, or E2E lane is indicated by this diff; PR Gate remains authoritative for independent review and final impact classification.

## Review focus

- Subscribe-before-replay ordering, so no request can be missed between registration and recovery.
- Settled-event cleanup for timeout/cancellation races and idempotency with normal response cleanup.
- Electron and Web contract parity for both replay-pending commands and settled events.
- The explicit boundary between Renderer-only recovery and unsupported full Main-process restart recovery.


ewen-poch:
Addressed the actionable AI review findings:

- Stale broad-scope confirmation: the confirmation state is now bound to the approval request ID and cleared when that request settles, with Connector and Compute regression coverage (f0afd5901).
- Replay notifications: pending replay now uses a renderer-only projection callback, so recovery does not repeat desktop notifications or attention side effects; both brokers have regression coverage (90ec24c74).

The latest AI review for 90ec24c74 reports mergeable with no actionable findings.

PR #1347, opened 2026-08-18
fix(remote-access): make browser revocation durable

ewen-poch:
## Problem

Trusted-browser revocation updated the in-memory pairing state before saving remote-access.json. If the repository save failed, the running process denied the old token while a restart reloaded and accepted the still-persisted token.

## Proposed change

- Serialize persistent pairing-state mutations inside RemoteSessionPairingManager.
- Save each immutable candidate state before committing it to the in-memory snapshot.
- Route last-seen writes through the same queue so they cannot race with revocation and restore stale browser records.
- Preserve the trusted browser in memory when revocation persistence fails.
- Add regression coverage for save failure and concurrent revocations.

## Scope and non-goals

- No persistent fields, schema version, migration, or data-model changes.
- No new lifecycle or enum values.
- No UI or RPC contract changes.
- No historical-data compatibility work is required; existing version 4 remote-access.json files remain unchanged.

- Failed revocation persistence keeps runtime and restart authorization consistent -> npm test -- src/main/remote-access/pairing.test.ts -t failed-revocation/concurrency cases -> passed.
- Remote-access behavior and adapters remain compatible -> npm test -- src/main/remote-access -> 5 files, 59 tests passed.
- Main-process types remain valid -> npm run typecheck:node -> passed.
- Source conforms to lint rules -> npm run lint -> passed.

All listed checks ran after the last material production-code edit. The module-impact manifest does not define a remote-access module ID, so the focused directory suite was used instead of test:module. No shared interface, renderer, schema, migration, path-handling, or build input changed; consumer and platform-specific local lanes were therefore excluded. PR CI remains authoritative for the full portable and cross-platform suites.

## Review focus

- Failure behavior when RemoteAccessRepository.save rejects.
- Ordering and recovery of the storedMutationQueue after success or failure.
- Interaction between trusted-browser revocation and lastSeenAt persistence.

### DIFF
--- a/src/main/remote-access/pairing.test.ts
+++ b/src/main/remote-access/pairing.test.ts
@@ -198,6 +198,115 @@ describe('RemoteSessionPairingManager', () => {
     ).resolves.toBe('denied')
   })
 
+  it('keeps a trusted browser authorized when revocation cannot be persisted', async () => {
+    const root = await mkdtemp(join(tmpdir(), 'open-science-remote-pairing-'))
+    roots.push(root)
+    const repository = new RemoteAccessRepository(root)
+    const manager = await RemoteSessionPairingManager.create({
+      repository,
+      isAllowedRemoteHost: (hostname) => hostname === 'home.example.ts.net',
+      isEnabled: () => true,
+      onChanged: vi.fn()
+    })
+
+    const firstResponse = response()
+    await manager.webAccess.authorizeHttp(
+      request('/'),
+      firstResponse.response,
+      new URL('https://home.example.ts.net/')
+    )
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the P1 in 02cec51b3. RemoteSessionPairingManager now tracks pending revocations and rejects matching trusted-session tokens immediately while the serialized repository write is pending. The marker is cleared in a finally block, so a failed save restores normal authorization while a successful save leaves the browser absent from durable state. The blocked-save regression test now verifies denial during persistence and reauthorization after an injected save failure. Verified with npm test -- src/main/remote-access (59 tests), npm run typecheck:node, and npm run lint.

PR #1354, opened 2026-08-18
fix(connectors): propagate notebook RPC cancellation

ewen-poch:
## Problem

Notebook local RPC creates a request-scoped `AbortSignal`, but `mcpCall` dropped it at the
`ConnectorService` boundary. A cancelled or disconnected Notebook caller could therefore leave an
approval pending, continue bundled HTTP requests and retry backoff, keep a custom MCP request alive,
or cross a connector-local side-effect boundary after the result had no consumer.

## Proposed change

- carry the request signal from Notebook RPC through connector policy/approval and dispatch
- cancel pending approval UI and notifications with the existing `cancelled` action state
- combine caller cancellation with bundled HTTP attempt timeouts and make retry backoff abortable
- pass the signal to custom MCP `listTools` and `callTool` request options without recording caller
  cancellation as connector unavailability
- check cancellation before connector-local Artifact publication

## Scope and non-goals

- no RPC payload, Settings, database, migration, persisted data-model, or historical compatibility
  change
- no new shared status value; `NotificationActionState.cancelled` already exists
- no rollback of a remote side effect or atomic Artifact write that completed before cancellation was
  observed
- no retry-count, timeout, connector-policy, or permission-grant semantic change

The listed checks ran after the last material edit.

- RPC disconnect propagates an aborted signal to connector dispatch ->
  `npm test -- src/main/notebook/local-rpc-server.mcpcall.test.ts` -> passed
- connector approval, HTTP retry/backoff, custom MCP, and local-handler cancellation ->
  `npm test -- src/main/connectors src/main/permission-grants/connector-broker.test.ts src/main/notebook/local-rpc-server.mcpcall.test.ts`
  -> 57 files passed; 770 tests passed, 39 skipped
- affected main-process contracts type-check -> `npm run typecheck:node` -> passed
- changed source follows repository lint rules -> `npm run lint` -> passed

PR CI remains authoritative for the complete portable and platform lanes; the full local `npm test`
suite was intentionally not run.

## Review focus

- cancellation is checked after asynchronous policy/profile reads and before external or
  side-effecting boundaries
- caller aborts bypass custom connector availability classification
- approval settlement removes abort listeners and emits the existing `cancelled` lifecycle state
- an Artifact write already in progress remains atomic rather than gaining cancellation rollback

### DIFF
--- a/src/main/connectors/approval-broker.test.ts
+++ b/src/main/connectors/approval-broker.test.ts
@@ -62,6 +62,31 @@ describe('ApprovalBroker', () => {
     await expect(decision).resolves.toBe('deny')
   })
 
+  it('cancels and removes a pending request when its caller aborts', async () => {
+    const timer = makeTimer()
+    const onSettled = vi.fn()
+    const broker = new ApprovalBroker({
+      generateId: () => 'id-1',
+      broadcast: () => undefined,
+      setTimer: timer.set,
+      clearTimer: timer.clear,
+      onSettled
+    })
+    const cancellation = new AbortController()
+
+    const decision = broker.request(
+      { connector: 'biomart', method: 'get_data', argsPreview: '{}' },
+      cancellation.signal
+    )
+    cancellation.abort()
+
+    await expect(decision).resolves.toBe('deny')
+    expect(broker.getPending('id-1')).toBeNull()
+    expect(onSettled).toHaveBeenCalledWith('id-1', 'cancelled')
+    timer.fire()
+    expect(onSettled).toHaveBeenCalledOnce()
+  })
+
   it('pauses a Session timeout while Side chat owns the composer', async () => {
     let now = 0
     const timers: Array<{ fn: () => void; ms: number }> = []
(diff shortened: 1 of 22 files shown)


ewen-poch:
Addressed both findings in commit 7472de04c.

- Custom MCP connection setup: connection initialization now receives an internal abort signal. Each caller can stop waiting independently; the shared attempt is aborted only after its final waiter cancels, and a client that resolves after cancellation is closed instead of cached. A new caller can also start a fresh attempt without waiting for the cancelled one to settle.
- Run-style descriptors: ToolContext now exposes the optional caller signal. Reactome, ENCODE, and ZINC compose it with their request timeouts and use it for direct requests, retry delays, and ZINC polling delays.

Focused coverage includes sole-waiter cancellation, concurrent-waiter isolation, late-client disposal, active direct-request cancellation, ENCODE retry cancellation, and ZINC polling cancellation. The connector module suite passes with 57 files and 778 tests passed (39 skipped); Node typecheck and lint also pass.

PR #1356, opened 2026-08-18
fix(storage): recover staged migrations after restart

ewen-poch:
## Problem

The storage move authority was held only in `command-owner.ts` memory after a copy reached the durable `verified` marker. If the main process exited before commit or discard, the next process could see the marker but could neither resolve it nor select that destination again.

This left a safe but orphaned snapshot on disk, consumed storage, and permanently blocked that destination without manual marker manipulation.

## Proposed change

- Classify a valid marker for the current source and derived target as an explicit `recover` result.
- Rehydrate the marker token only inside the main-process command owner; it is never exposed to the renderer.
- Allow a recovered `verified` copy to be finished or discarded, and a recovered `copying` copy to be discarded only.
- Before finishing a recovered copy, restore the write gate, reject delegated work, disconnect data-root writers, and rerun the existing inventory/provenance checks.
- Route Settings to a dedicated recovery decision while preventing onboarding, missing-root recovery, and direct pointer switching from adopting a staged copy.
- Add Simplified and Traditional Chinese strings for the new interaction.

## Scope and non-goals

- No settings field, database schema, marker version, or persisted marker field changes.
- Existing valid version-1 `copying` and `verified` markers are recovered in place; corrupt markers and markers for another source remain invalid.
- The shared IPC read model adds the transient `recover` kind and `recoveryStatus: copying | verified` discriminator. No persisted enum is added.
- No automatic commit or automatic deletion occurs on startup; the user must explicitly finish or discard in Settings.

- Recreated command owner + verified marker -> commit pauses writers, rechecks the snapshot, switches the pointer, and relaunches.
- Source mutation after verification -> recovered commit is refused and the old root remains writable.
- Recreated command owner + copying/verified marker -> explicit discard removes only the marker-matched staging target.
- Pause failure or active delegated work -> recovered commit is refused without switching roots.
- Settings -> verified recovery offers finish/discard; copying recovery offers discard only and never starts a new copy.
- Other pointer-switch surfaces -> `recover` cannot be treated as `move` or `adopt`.

Validation run:

- `npm run typecheck:node`
- `npm run typecheck:web`
- `npm run lint`
- `npm test -- src/main/storage/migration-service.test.ts` (81 passed, 4 skipped)
- `npm test -- src/main/storage/ipc.test.ts -t "recovered|command owner is recreated"` (7 passed)
- `npm test -- src/main/storage/migration-state.test.ts` (11 passed)
- `npm test -- src/renderer/src/pages/settings/StorageMigrationModal.render.test.tsx src/renderer/src/pages/settings/StoragePanel.render.test.tsx src/renderer/src/components/DataRootMissingDialog.render.test.tsx src/renderer/src/components/LegacyDataMoveDialog.render.test.tsx src/renderer/src/pages/onboarding/LocationStep.render.test.tsx` (81 passed)
- `npx vitest run src/renderer/src/i18n/resources.test.ts` (67 passed)

The full `src/main/storage/ipc.test.ts` file was not used as local evidence because this Windows dependency tree does not contain the built `safe_file_publisher_native.node`, and rebuilding it requires a Visual Studio C++ workload not present on this machine. The targeted recovery cases do not require that binding and pass; PR CI remains authoritative for the complete suite.

## Review focus

- Marker source/target/token validation at the recovered command boundary.
- Re-establishing writer quiescence before a recovered commit.
- Keeping `copying` recovery discard-only and preventing pointer-switch bypasses.

### DIFF
--- a/src/main/storage/command-owner.ts
+++ b/src/main/storage/command-owner.ts
@@ -315,10 +315,9 @@ const createStorageCommandOwner = (deps: StorageCommandOwnerDeps) => {
   // boundary before commit/discard receives it. A fresh correlation id starts the recovery attempt's
   // diagnostics because the original process-local operation is gone.
   const recoverStagedFromMarker = async (
-    request: StorageParentRequest,
+    target: string,
     allowedStatuses: ReadonlySet<'copying' | 'verified'>
   ): Promise<NonNullable<typeof activeStaged> | undefined> => {
-    const target = dataRootForPicked(request.parent)
     const marker = await readMigrationMarker(target)
     if (
       !marker ||
@@ -350,14 +349,23 @@ const createStorageCommandOwner = (deps: StorageCommandOwnerDeps) => {
       logger.warn('staged data root discard ignored', { reason: 'resolution-in-progress' })
       return { ok: false, error: 'A migration is already being resolved.' }
     }
+    let target: string
+    try {
+      target = dataRootForPicked(request.parent)
+    } catch (err) {
+      logger.warn('staged data root discard ignored', {
+        reason: 'invalid-request',
+        ...diagnosticErrorFields(err)
+      })
+      return { ok: false, error: err instanceof Error ? err.message : String(err) }
+    }
     resolutionInProgress = true
     try {
-      const target = dataRootForPicked(request.parent)
       const staged = activeStaged
         ? samePath(activeStaged.target, target)
           ? activeStaged
           : undefined
-        : await recoverStagedFromMarker(request, new Set(['copying', 'verified']))
+        : await recoverStagedFromMarker(target, new Set(['copying', 'verified']))
       if (!staged) {
         logger.warn('staged data root discard ignored', { reason: 'no-matching-copy' })
         return { ok: false, error: 'No matching staged data copy was found.' }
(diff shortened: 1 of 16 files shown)


ewen-poch:
Addressed the actionable recovery-state feedback in c40a25599:

- Settings now re-inspects the selected target whenever the migration modal closes. After a successful discard, the deleted marker is reflected immediately and the stale `recover` action disappears. The regression test covers the full select -> recover -> discard -> reclassify flow.
- The CI failure in `storage-i18n.render.test.tsx` was also fixed by preserving the existing default move guidance when no target classification exists.

The `DataRootMissingDialog` case is intentionally not routed through recovered commit. That dialog is shown only when the marker's source (the configured current root) is missing. Recovered commit deliberately requires the source inventory and target inventory to both still match the verified marker; promoting the target without the source would weaken the source-change/data-loss guard. The dialog is not a dead end: it already has the explicit **Continue with an empty folder** action, after which Settings can inspect the staged target. Supporting promotion when the original source is unavailable would require a separately reviewed disaster-recovery contract, not a pointer-switch bypass in this fix.

ewen-poch:
This is intentionally not routed into destructive recovery while the configured source root is unavailable. The staged copy may be the user’s only remaining copy, so enabling discard here would introduce a data-loss path. Recovered commit also requires the source and target inventories to be revalidated before pointer persistence. Supporting missing-source promotion or deletion requires a separately reviewed disaster-recovery contract.

PR #1360, opened 2026-08-18
fix(local-fs): sort directory entries before truncating

ewen-poch:
## Problem

Local directory listings read the complete directory but applied the 5,000-entry cap before sorting. Because Node.js does not guarantee `readdir` order, large directories could expose an unstable arbitrary subset, and entries that belong at the front of the sorted listing could be omitted.

## Proposed change

- Sort the full `Dirent` result directories-first and case-insensitively by name before applying the existing cap.
- Continue to `stat` only the selected entries so large directories do not gain unbounded metadata I/O.
- Add a 5,001-entry regression test where a directory returned last must remain visible and the final alphabetical file must be omitted.

## Scope and non-goals

- No pagination or change to the existing 5,000-entry limit/truncation notice.
- No IPC, preload, renderer interaction, shared data shape, enum, architecture, persistence, migration, or historical-data change.
- Symlinks retain existing behavior: target metadata can reclassify selected symlinks after the cap; the fix does not `stat` every entry before selection.

- Stable capped subset -> `npm test -- src/main/local-fs/service.test.ts -t "lists entries directories-first|sorts the complete readdir result"` -> passed (2 tests).
- Local FS service behavior -> `npm test -- src/main/local-fs/service.test.ts -t '^(?!.*(resolves symlinks|home itself sits behind a symlink)).*$'` -> passed (19 tests, 2 existing symlink-permission tests skipped locally).
- Lint -> `npm run lint -- --no-cache` -> passed.

The full service-file run reached 19 passing tests; its two failures were existing Windows symlink setup calls rejected with `EPERM` before product code ran. PR CI remains authoritative for the complete portable and platform suites.

## Review focus

- Confirm the pre-cap comparator remains aligned with the public directories-first, case-insensitive listing order.
- Confirm retaining bounded `stat` work is preferable to fully resolving symlink targets before subset selection.

### DIFF
--- a/src/main/local-fs/service.test.ts
+++ b/src/main/local-fs/service.test.ts
@@ -65,9 +65,14 @@ describe('LocalFsService.listDir', () => {
   it('sorts the complete readdir result before applying the entry cap', async () => {
     const fileDirents = Array.from({ length: LOCAL_DIR_ENTRY_CAP }, (_, index) => ({
       name: `file-${String(index).padStart(4, '0')}`,
-      isDirectory: () => false
+      isDirectory: () => false,
+      isSymbolicLink: () => false
     }))
-    const directoryDirent = { name: 'zzz-directory', isDirectory: () => true }
+    const directoryDirent = {
+      name: 'zzz-directory',
+      isDirectory: () => true,
+      isSymbolicLink: () => false
+    }
     vi.mocked(readdir).mockResolvedValueOnce([...fileDirents, directoryDirent] as never)
     vi.mocked(stat).mockResolvedValue({
       isDirectory: () => false,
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the P2 directory-symlink concern in `bd8a926b7`.

- `listDir` now resolves only symbolic-link entries before applying the cap, so symlinks targeting directories participate in the directory-first selection.
- The prefetched symlink metadata is reused for the selected entries; ordinary entries are still stat'ed only after selection.
- Added an over-cap regression test with 5,000 files plus a directory symlink returned last.

- Local FS service tests: 20 passed, 2 existing Windows symlink-permission tests skipped locally.
- `npm run lint -- --no-cache`: passed.
- Prettier and `git diff --check`: passed.

ewen-poch:
Addressed the unbounded pre-cap filesystem work in `bb62f6ad2`.

The selection phase now uses `Dirent` metadata only, with a stable priority of physical directories, symbolic-link candidates, then ordinary files. No `stat` calls occur before the 5,000-entry cap. Selected symlinks are resolved by the existing post-selection `stat` path, so directory targets remain eligible without scanning every discarded symlink.

The explicit trade-off is that a symlink targeting a file can be selected ahead of an ordinary file; this keeps the work bounded and the subset deterministic while preserving navigable symlink candidates.

- Local FS service tests: 20 passed, 2 existing Windows symlink-permission tests skipped locally.
- `npm run lint -- --no-cache`: passed.
- Prettier and `git diff --check`: passed.

ewen-poch:
I am not changing the file-symlink prioritization for this P2 because it is the explicit bounded-work trade-off from the preceding review.

Determining whether every symlink targets a directory or a regular file requires resolving every symlink before the cap. That reintroduces unbounded pre-cap filesystem I/O and navigation latency for discarded entries. The current contract guarantees that returned entries are directory-first/alphabetically sorted and marks capped listings with `truncated: true`; it does not guarantee that the capped subset is the exact prefix of a fully target-resolved namespace.

The current selection remains deterministic, preserves physical directories first, keeps navigable symlink candidates eligible, and performs `stat` only for the selected 5,000 entries. A file symlink can displace an ordinary file only in an already-truncated directory containing symlinks. I consider that bounded edge-case trade-off preferable to restoring unbounded I/O, so no further code change is planned for this comment.

ewen-poch:
No code change is planned for this symlink-cap permutation because it is the same bounded-work trade-off already documented on the preceding head.

Guaranteeing that a directory-target symlink can never be displaced by file-target symlinks requires resolving every symlink before the cap in the worst case. That restores the unbounded pre-cap filesystem work and navigation latency identified by the previous review. The current listing is explicitly partial (`truncated: true`), deterministic, preserves physical directories first, and performs metadata I/O only for selected entries.

Eliminating this trade-off without unbounded I/O would require a paginated or snapshot-based listing contract and corresponding renderer interaction, which is outside this PR's approved interface and interaction scope. I am therefore retaining the bounded selection strategy.

PR #1373, opened 2026-08-18
fix(acp): preserve positive context usage across zero updates

### LINKED ISSUE #1371: [Bug]: Context window usage regresses from a positive value to 0% during an active session
### Preflight checklist

### What happened?

In an active session, the Context window panel intermittently changes from a positive provider-reported usage value to `0.0%`, then later returns to a nonzero value.

In the observed sequence, **Agent used** changed:

```text
0 / 1M
→ 62.0k / 1M
→ 0 / 1M
```

At the same time, the local composition estimate remained stable:

```text
Local 85.9k
→ Local 86.9k
→ Local 87.3k
```

All three panel states showed `Reconciled · cl100k_base · glm-5.3`.

I did not change the selected session, and this does not appear to be a UI switch between the main agent and subagents. The Context window panel is bound to the active session. The issue appears to be that a later provider context-usage update with `0` overwrites a previously positive value without a confirmed context reset, compaction, or session replacement.

This makes the displayed percentage flash and gives an incorrect impression that the session context has been cleared.

### Steps to reproduce

1. Start or continue a session using GLM-5.3.
2. Open **Context window** from the conversation controls.
3. Keep the panel open while the agent continues working and receiving context-usage updates.
4. Observe that **Agent used** can change from a positive value, such as `62.0k / 1M`, to `0 / 1M`.
5. Observe that the local composition estimate remains near its prior value and that the panel remains marked `Reconciled`.

### Operating system

Windows

### App version

0.16.0 (observed). The provider-usage replacement logic was also checked in 0.17.0 and is unchanged.

### Provider / model

GLM-5.3

Three screenshots were captured sequentially:

```text
1. Agent used 0 / 1M; Local 85.9k
2. Agent used 62.0k / 1M; Local 86.9k
3. Agent used 0 / 1M; Local 87.3k
```

The current `main.log` does not include individual context-usage values, so the exact provider event that supplied the zero value cannot be identified from logs alone.

ewen-poch:
## Problem

During an active session, a later provider context-usage update with `used: 0` can overwrite an established positive reading even though the context owner has not been reset, compacted, or replaced. The renderer then shows `0.0%` while the local estimate remains stable and the state is still marked `Reconciled`.

## Proposed change

- Reject an unconfirmed zero regression when the context tracker already owns a positive provider reading.
- Apply the same guard to streamed provider usage updates and terminal provider readings.
- Accept a zero that is confirmed by explicit compaction or a completed Codex provider compaction lifecycle, regardless of whether the zero arrives immediately before or after completion.
- Ignore reconnect replay as a compaction confirmation, and consume the one-shot confirmation on the next provider reading.
- Add regression coverage for the reported GLM-5.3 sequence, terminal Codex usage, explicit compaction, and provider-observed automatic compaction.

## Scope and non-goals

- No persisted-data, database-schema, migration, or historical-data compatibility changes.
- One internal projector effect kind (`provider-compaction-completed`) and one transient in-memory session marker are added; no shared or persisted enum/status is changed.
- No provider adapter protocol changes; Claude Code, OpenCode, Codex Responses, and Codex Bridge converge on the same context-usage owner.

- Positive usage does not regress to zero without reset; initial, replacement, explicit-compaction, and provider-compaction zero readings remain valid -> `npm test -- src/main/acp/context-usage-tracker.test.ts src/main/acp/context-compaction-workflow.test.ts src/main/acp/session-update-projector.test.ts src/main/acp/prompt-turn-workflow.test.ts src/main/acp/runtime-events.test.ts` -> 5 files, 114 tests passed.
- Main-process type contracts remain valid -> `npm run typecheck:node` -> passed.
- Source lint and formatting rules -> `npm run lint` and targeted `prettier --check` -> passed.
- Whitespace/error-marker check -> `git diff --check` -> passed.

The complete portable/platform suite is intentionally left to the exact-head PR Gate CI. No locally uncovered persistence or migration risk was identified.

## Review focus

Please verify that zero is rejected only while a positive reading is still owned, and remains valid once an explicit or provider-observed compaction lifecycle confirms the reset boundary.

Fixes #1371

### DIFF
--- a/src/main/acp/context-usage-tracker.test.ts
+++ b/src/main/acp/context-usage-tracker.test.ts
@@ -703,6 +703,46 @@ describe('ContextUsageTracker', () => {
     })
   })
 
+  it('retains positive provider usage when a later update reports zero without a reset', () => {
+    const tracker = new ContextUsageTracker(wordCounter)
+    tracker.beginSession('s1', { frameworkId: 'opencode', model: 'glm-5.3' })
+    tracker.appendText('s1', 'messages', 'committed conversation history')
+    tracker.reconcileProviderUsage('s1', { used: 62_000, size: 1_000_000 })
+
+    tracker.reconcileProviderUsage('s1', { used: 0, size: 1_000_000 })
+
+    expect(tracker.usage('s1')).toMatchObject({
+      used: 62_000,
+      size: 1_000_000,
+      breakdown: { status: 'reconciled' }
+    })
+
+    tracker.deleteSession('s1')
+    tracker.beginSession('s1', { frameworkId: 'opencode', model: 'glm-5.3' })
+    tracker.appendText('s1', 'messages', 'new conversation')
+    tracker.reconcileProviderUsage('s1', { used: 0, size: 1_000_000 })
+
+    expect(tracker.usage('s1')).toMatchObject({
+      used: 0,
+      size: 1_000_000,
+      breakdown: { status: 'reconciled' }
+    })
+  })
+
+  it('rejects a zero terminal reading after positive provider usage', () => {
+    const tracker = new ContextUsageTracker(wordCounter)
+    tracker.beginSession('s1', { frameworkId: 'codex', model: 'gpt-5.6-sol' })
+    tracker.appendText('s1', 'messages', 'committed conversation history')
+    tracker.reconcileProviderUsage('s1', { used: 62_000, size: 1_000_000 })
+
+    expect(tracker.reconcileUsed('s1', 0)).toBe(false)
+    expect(tracker.usage('s1')).toMatchObject({
+      used: 62_000,
+      size: 1_000_000,
+      breakdown: { status: 'reconciled' }
+    })
+  })
+
   it('restores the last provider reading when a preflight estimate receives no fresh update', () => {
     const tracker = new ContextUsageTracker(wordCounter)
     tracker.beginSession('s1', { frameworkId: 'opencode', model: 'deepseek-v4' })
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the P1 in d02eecf2f. The tracker now records provider-update revisions across compaction checkpoints: a zero reading after the checkpoint is staged and accepted only when that compaction succeeds, while ordinary active-session zero regressions remain suppressed. Failed/restored lifecycles discard newer staged zeros. Added a dedicated post-compaction zero regression test; the updated ACP test set passes 111/111.

ewen-poch:
CI triage for head d02eecf2: Static checks and Full Module tests (macOS, shard 2/2) are the same upstream base failure. The shard blob contains exactly one failing file/test: src/renderer/src/i18n/resources.test.ts / every English t() literal has a ja translation, reporting 12 missing Japanese strings introduced by #1356. Module tests and coverage and PR Gate are downstream failures. All other lanes passed, including Windows core/E2E, macOS build/E2E, shard 1, CodeQL, type checks, lint, and AI re-review. This PR does not touch the storage or locale files, so the ACP fix remains scoped; the failed checks need a mainline i18n fix and rerun.

ewen-poch:
Addressed the automatic-compaction lifecycle gap. AcpSessionUpdateProjector now confirms completed Codex compaction events with ContextUsageTracker while ignoring reconnect replay. The tracker immediately applies a zero that arrived before completion, or permits exactly the next provider reading after completion; a positive reading consumes that one-shot confirmation so later unconfirmed zero regressions remain suppressed. The confirmation marker is transient in-memory state only. Validation: 5 targeted ACP test files, 114 tests passed; node typecheck, lint, and Prettier passed.

PR #1374, opened 2026-08-18
fix(workspace): gate follow-on actions until branched Session replay

ewen-poch:
## Problem

Branch in new session has two triggers. Composer branch immediately sends and replays history into the new ACP session. Clicking Branch on an Agent message only copies the transcript, creates an empty ACP session, and sets `pendingHistoryReplay: { kind: 'all' }`.

The UI then looks like a complete idle Session. Follow-on actions that talk to ACP (review fix-loop, side-chat start, compact, nested branch, specialist switch, save as skill) treated “ACP session id is live” as “context is reconstructed.” That is false until the next user send goes through `prepareExistingWorkspacePrompt`.

## Proposed change

Minimal product rule: after a message-click branch, the user must send a message (or plan-first send) to reconstruct ACP before any other Session action.

- Keep send and plan-first available after bind.
- Disable side-chat, request review, compact, save as skill, nested branch, agent controls, and permission-profile changes while the Session is `isPending` or has `pendingHistoryReplay`.
- Reuse the existing `session-pending` actionability reason. No new persisted enum, no persistence format change, no historical migration.

## Scope and non-goals

- Renderer admission only. Main-process review/side-chat IPC is not fail-closed in this PR.
- Does not eagerly reconstruct ACP after the click.
- Does not copy Plan or Review rows onto the child Session.
- Trigger 1 mid-flight reload (`pendingContextReplayMessageId` is transient) is out of scope.

- After a message-click branch, composer send remains available and reconstructs context as today.
- Review, side-chat, compact, save as skill, nested branch, and agent controls stay disabled until that send clears `pendingHistoryReplay`.
- Composer plan-first remains a valid reconstruct send.

Validation after the last material edit:

| --- | --- | --- |
| Session actionability + branch refusal | `npm run test:module -- session_renderer` | pass (441 tests) |
| Workspace follow-on gates | `npm run test:module -- workspace_page` | pass (455 tests) |

Full portable `npm test` was not run locally; PR Gate CI is authoritative.

## Review focus

- Send and plan-first must stay available after bind; other ACP-facing actions must not.
- `pendingHistoryReplay` and `isPending` should use the same follow-on gate.
- No new persisted state.

### DIFF
--- a/src/renderer/src/locales/ja.json
+++ b/src/renderer/src/locales/ja.json
@@ -1420,6 +1420,7 @@
   "Resize left panel": "左パネルのサイズを変更する",
   "Resize notebook and terminal": "Notebook と端末のサイズを変更する",
   "Resize right panel": "右パネルのサイズを変更する",
+  "Resolve the current Session operation first.": "先に現在のセッション操作を完了してください。",
   "Resolve the items marked Action needed, then choose Check again.": "「アクションが必要」とマークされた項目を解決し、[もう一度確認] を選択します。",
   "Resolved": "解決済み",
   "Resolving…": "解決中…",
--- a/src/renderer/src/locales/zh-Hans.json
+++ b/src/renderer/src/locales/zh-Hans.json
@@ -1481,6 +1481,7 @@
   "Resize left panel": "调整左侧面板大小",
   "Resize notebook and terminal": "调整 Notebook 与终端大小",
   "Resize right panel": "调整右侧面板大小",
+  "Resolve the current Session operation first.": "请先完成当前会话操作。",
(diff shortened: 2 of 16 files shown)


ewen-poch:
**Side chat disabled reason localization**

Adopted. The replay-pending Side chat reason is now `t('Resolve the current Session operation first.')`, with independent `ja`, `zh-Hans`, and `zh-Hant` catalog entries. Behavior is unchanged: send/plan-first stay available, other follow-on actions stay gated.

**Static checks / missing Japanese storage strings**

Not adopting in this PR. The failing i18n assertion is 12 pre-existing `ja` gaps from #1356 (`StorageMigrationModal` / `StoragePanel`). They are outside this branch-replay gate and would expand the change into storage-move copy. Happy to backfill those keys in a separate change if you want CI green here first.

PR #1380, opened 2026-08-18
fix(compute): harden password authentication follow-up

ewen-poch:
## Summary

- preserve ordinary non-zero remote command results in password mode while keeping probe and transfer paths fail-closed and redacting saved passwords
- flatten only non-executable SSH configuration under -F none, resolve the target hostname explicitly, and reject option-like resolved hosts before attaching askpass
- avoid reporting a committed authentication switch as failed when post-commit in-memory cleanup fails
- move the historical ComputeJob upgrade assertion into the database migration suite
- add required focus/disabled states and simplify ordinary Settings field layout

## Validation

- relevant Vitest set: 8 files passed, 281 tests passed
- npm run typecheck
- npm run lint
- git diff --check
- earlier full npm test run: 16,500 passed; one unrelated R namespace cancellation test failed in the aggregate run and passed when run alone
- final Standards and Spec reviews: no findings

The separate CI-only fix was handled and merged in #1378.


ewen-poch:
Addressed the IPv6 SCP upload issue in f2325499. Password-mode uploads now bracket colon-containing resolved hosts only when building the SCP remote spec; SSH command execution and bounded downloads continue to use the raw host. Added a regression test for an effective IPv6 HostName. Validation: all 47 connection-broker tests, typecheck, lint, and diff check pass.

ewen-poch:
Addressed the username preservation issue in d926b5a0. Password mode now removes any inherited User option and always supplies the explicit sshOverrides.user value before running the resolved HostName with -F none. Added a regression test for sshAlias equal to sshOverrides.user with a different effective HostName. Validation: all 48 connection-broker tests, typecheck, lint, and diff check pass.

ewen-poch:
Addressed the job protocol redaction issue in 400a130a. Password-mode job intents now preserve stdout until protocol parsing completes. The connection lease exposes an optional batched redaction seam that is propagated through broker wrappers: dispatch redacts invalid raw output only after PID parsing fails, and polling redacts parsed stdout/stderr tails before persistence. Transport stderr and direct user command output remain redacted. Validation: 120 related tests pass across connection broker, dispatcher, poller, and the password-auth release gate; typecheck, lint, and diff check also pass.

PR #1383, opened 2026-08-18
fix(notebook): reap micromamba trees before retry

ewen-poch:
## Problem

Micromamba timeout and cancellation called `child.kill()` and waited only for the direct process to close. A detached descendant could keep writing the same Conda prefix after `runMicromamba` rejected. `withJournaledPrefixWrite` then cleared the operation journal and PID sidecar, so an in-session Retry could start a second writer against the same prefix.

## Proposed change

- route micromamba timeout, cancellation, and PID-recording failure through the existing cross-platform `terminateProcessTree` helper;
- deduplicate and await tree teardown before returning the ordinary timeout/cancellation result;
- treat failed POSIX descendant discovery as `reaped: false` instead of silently accepting an empty tree;
- retain the operation journal when teardown is unconfirmed and downgrade the existing child sidecar to `{ spawning: true, bootToken? }`, because a dead direct PID cannot prove that a reparented descendant is gone;
- keep the current in-process prefix block so Retry and Reset cannot race a possible surviving writer.

## Scope and non-goals

- No database/schema, domain data model, or historical-data migration.
- No new persisted field or format; this reuses the existing spawn-intent sidecar variant.
- No new status or enum value.
- No renderer, interaction, or user-facing copy change.
- No dependency change.
- Successful micromamba completion and `captureMicromamba` are unchanged.

All listed checks ran after the last material code edit.

- cancellation waits for a real descendant that continuously writes the target prefix, or fails closed when the platform cannot prove teardown -> `npm test -- src/main/notebook/provisioner-runtime.test.ts src/main/process-tree.test.ts src/main/notebook/provisioner.test.ts src/main/notebook/package-manager.test.ts` -> PASS (4 files, 201 tests);
- POSIX snapshot failure/non-zero exit reports an unconfirmed tree -> same targeted command -> PASS;
- an unconfirmed tree retains the journal, converts the PID sidecar to the existing unprobeable state, and blocks Retry -> same targeted command -> PASS;
- affected main-process types -> `npm run typecheck:node` -> PASS;
- linted source/config -> `npm run lint` -> PASS;
- touched-file formatting and diff whitespace -> `npx prettier --check <6 touched files>` and `git diff --check` -> PASS.

`npm run test:affected:explain -- --base origin/main --head HEAD` selected full mode because `src/main/notebook/provisioner-runtime.test.ts` has no registered Module owner. Per maintainer direction, the complete local `npm test` suite was not run; PR Gate is the final complete-suite authority.

There is no historical compatibility work. Existing journals and sidecars remain readable without migration. The only persistence behavior change is cleanup timing: confirmed teardown clears the existing records as before, while unconfirmed teardown retains them and rewrites the sidecar to the already-supported no-verifiable-PID state.

## Review focus

- `runMicromamba` must not settle ordinary timeout/cancellation until tree teardown reports `reaped: true`.
- `reaped: false` must retain recovery evidence both in-process and across restart.
- The stricter POSIX `reaped` result should remain fail-closed for all existing callers.

## Uncovered risks

- The real descendant-writing integration test was exercised locally on macOS. Windows `taskkill /T /F` and Linux `ps` behavior have deterministic unit coverage and rely on PR platform CI for real-host confirmation.
- If platform tree discovery is unavailable, the environment intentionally stays blocked instead of permitting a concurrent writer. On platforms without an authoritative persisted boot token, that degraded case may require manual recovery; this PR does not change recovery UI.

### DIFF
--- a/src/main/notebook/package-mutation.test.ts
+++ b/src/main/notebook/package-mutation.test.ts
@@ -386,6 +386,7 @@ describe('NotebookPackageMutationOwner', () => {
       },
       installPackages: vi.fn(async (_request, deps) => {
         deps?.onBeforeSpawn?.()
+        deps?.onChild?.(process.pid)
         throw childFailure
       })
     })
--- a/src/main/notebook/package-mutation.ts
+++ b/src/main/notebook/package-mutation.ts
@@ -245,6 +245,15 @@ class NotebookPackageMutationOwner {
       if (isRepairQuarantineError(error)) retainForRecovery = true
       if (isChildUnconfirmedError(error)) {
         retainForRecovery = true
+        // The direct installer PID may already be gone while an unenumerated/reparented descendant
+        // continues writing. Replace that stale identity with the existing no-verifiable-PID state so
+        // startup recovery blocks instead of clearing the journal after probing only the dead parent.
+        try {
+          recordSpawnIntentSync(runtimeRoot, operationId)
+        } catch {
+          // Retain the prior sidecar + journal as the best durable evidence still available. The
+          // in-process recovery block below remains authoritative for this app lifetime.
+        }
         this.options.blockUnconfirmedChild(target)
       }
       throw error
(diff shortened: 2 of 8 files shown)


ewen-poch:
Addressed the AI review P1 in 8fb2499c. When package-install teardown cannot confirm the full process tree has stopped after a PID was recorded, the mutation owner now rewrites the existing sidecar to the no-verifiable-PID spawning state before retaining the journal and applying the in-process block. The regression test now records a child PID first and verifies that it is downgraded to spawning: true. Validation: 5 scoped test files / 211 tests passed, plus typecheck:node, lint, Prettier, and git diff --check.

PR #1384, opened 2026-08-18
fix(web-service): bound shutdown with active HTTP requests

ewen-poch:
## Problem

The Web HTTP controller awaited `server.close()` before releasing application caller leases. Node waits for active HTTP requests to finish before completing that callback, so a request whose handler never settles could strand ordinary quit, update installation, and migration relaunch indefinitely.

## Proposed change

- stop accepting new HTTP connections before teardown;
- dispose the Web caller leases and command client so active application commands receive cancellation;
- call `closeAllConnections()` after `server.close()` has started, forcing handlers that ignore cancellation off their HTTP connections;
- preserve cleanup failures while still waiting for the server close callback;
- start `/api/shutdown` teardown only after its 202 acknowledgement finishes, or after that client disconnects, so attached shutdown callers receive the response before connections are forced closed;
- add real HTTP regression tests for both a deliberately blocked RPC handler and a delayed shutdown acknowledgement.

## Scope and non-goals

- No public interface or module-boundary changes.
- No persistence schema, migration, data-model, status-enum, or renderer interaction changes.
- Existing WebSocket shutdown behavior remains unchanged.
- Intentional Web service shutdown now cancels unfinished HTTP work instead of waiting without a bound.

- stuck request cancellation and shutdown acknowledgement delivery -> `npm test -- src/main/web-service/http-server.test.ts -t "bounds close when an active HTTP request does not finish|delivers the shutdown acknowledgement before closing an attached server"` -> passed (2 tests);
- Web HTTP, controller, and application shutdown integration -> `npm test -- src/main/web-service/http-server.test.ts src/main/web-service/controller.test.ts src/main/application-runtime.test.ts` -> passed (48 tests);
- affected main-process types -> `npm run typecheck:node` -> passed;
- linted source -> `npm run lint` -> passed.

The exact-head impact classifier selects the full plan because `src/main/web-service/http-server.ts` has no declared module owner. The complete portable and platform suites are therefore intentionally delegated to blocking PR CI; no complete local `npm test` was run.

## Review focus

- Ordering of `server.close()`, caller cancellation, and `closeAllConnections()`.
- Delivery of the `/api/shutdown` acknowledgement before attached teardown begins.
- Preservation of cleanup errors without allowing an active connection to strand close.
- The regression tests' use of the real Node HTTP lifecycle with a controlled delayed response finish.

Uncovered local risk: packaged update and migration-relaunch workflows were not exercised end-to-end; their exact-head CI lanes remain authoritative.

### DIFF
--- a/src/main/web-service/http-server.test.ts
+++ b/src/main/web-service/http-server.test.ts
@@ -1,5 +1,5 @@
 import { mkdtemp, rm, writeFile } from 'node:fs/promises'
-import { request as httpRequest } from 'node:http'
+import { request as httpRequest, ServerResponse } from 'node:http'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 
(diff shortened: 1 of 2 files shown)


ewen-poch:
CI note: the final macOS module shard 2/2 failure is inherited from main, not from this PR. The decoded failure is src/main/compute/compute-service.architecture.test.ts: job-poller.ts has 676 lines against a 660-line budget. The merge ref includes main commit e1e6dbb42 from #1380, while this PR changes only http-server.ts and http-server.test.ts; main at d3f9f2ea5 also has failing compute full-test checks. I did not rerun this deterministic failure or add an unrelated compute change here. Static checks, CodeQL, Windows core/E2E, macOS build/E2E, module shard 1/2, and the latest Codex review all passed.

PR #1389, opened 2026-08-18
fix(settings): improve failure recovery and keyboard access

ewen-poch:
## Problem

- Skill and Connector catalog/detail requests rendered successful empty states or blank detail pages when their IPC calls rejected, with no recovery action.
- Skill and Connector policy writes rolled back in the store but the UI swallowed the rejection, so users could not tell whether the safety setting was saved.
- Several Settings controls used ARIA radio roles without radio keyboard behavior, making every choice a Tab stop and leaving arrow-key selection unavailable.
- The narrow Settings navigation covered the content visually without creating an inner modal focus boundary, so keyboard focus could enter the obscured content.

These failures occur on request rejection or policy-write rejection, when navigating Settings by keyboard, and when opening Settings navigation below the `md` breakpoint. Leaving them unresolved makes service outages look like missing data, makes rolled-back safety changes appear successful, and lets keyboard users lose context behind the drawer.

## Proposed change

- Add a shared Settings loading/error notice with explicit progress, localized failure copy, and Retry actions. Preserve cached catalogs on refresh failure and block successful empty-state copy when no catalog/detail request has completed.
- Surface a localized alert after rejected Skill/Connector enablement, auto-allow, or tool-permission writes; existing optimistic rollback remains authoritative.
- Use Radix RadioGroup behavior for Theme, Tool permission, Reasoning effort, App icon, Connector type, and Skill content mode. Add equivalent roving focus and arrow/Home/End behavior to installed Agent framework cards, whose nested actions prevent using a plain RadioGroup item.
- Give the narrow navigation drawer dialog/modal semantics, a trapped/looping focus scope, Escape handling, focus entry/return, and `inert`/`aria-hidden` isolation for obscured Settings content.
- Add Simplified Chinese, Traditional Chinese, and Japanese translations for the new user-visible states.

## Scope and non-goals

- Renderer-only interaction change; no IPC contract, shared settings schema, database schema, or store data-model change.
- New state values are component-local request states: `loading | ready | error`. No shared/domain enum is added or changed.
- No new fields, persisted values, persistence path, or migration.
- No historical-data compatibility behavior is required because stored data is neither read differently nor rewritten.
- Existing native radio inputs remain unchanged.

The following final checks ran after the last material edit:

- Settings loading/error/empty separation, retry, rolled-back write feedback, custom radio behavior, redundant-selection protection, and narrow-drawer focus isolation -> `npm test -- src/renderer/src/pages/settings src/renderer/src/components/ThemeControls.render.test.tsx src/renderer/src/pages/onboarding/AgentStep.render.test.tsx` -> 72 files, 787 tests passed.
- Renderer translations and catalog contracts -> `npx vitest run src/renderer/src/i18n/resources.test.ts` -> 1 file, 88 tests passed.
- Affected TypeScript processes -> `npm run typecheck` -> passed.
- Linted source -> `npm run lint` -> passed.
- Exact-head impact explanation -> `npm run test:affected:explain -- --base origin/main --head HEAD` -> selected full fallback because `ThemeControls.render.test.tsx` has no module owner.

Per maintainer direction, the complete local `npm test` fallback was not run; exact-head PR Gate and its selected platform/accessibility lanes are authoritative. The remaining locally uncovered risk is native Electron/browser focus traversal around the nested drawer; DOM semantics and focus transitions are covered in renderer tests, and PR CI supplies the independent validation.

## Review focus

- Confirm that retry replaces only unresolved no-cache surfaces while cached catalogs remain usable during refresh failures.
- Confirm that all rejected safety-policy writes remain rolled back and now expose feedback without changing persistence semantics.
- Check the Agent framework card roving-focus implementation separately from the Radix-based groups because those cards contain nested repair/uninstall actions.
- Check narrow-screen Escape, backdrop, navigation-selection, and focus-return behavior.


ewen-poch:
Addressed the P2 in 8b00cf29: roving navigation now focuses the resolved radio but only invokes selection when the target differs from the current radio. The new single-framework ArrowRight/Home regression test also verifies that neither framework persistence nor the environment check runs.

PR #1390, opened 2026-08-18
fix(startup): recover the database gate after IPC rejection

ewen-poch:
## Problem

The Electron database-startup shell initialized to `checking` and only left that phase when `databaseStartup.getState()` resolved or `onStateChanged` delivered a later state. Both `getState()` and `retry()` were invoked without a rejection handler.

If the catch-up IPC read rejected after the live event was missed, the UI stayed on **Checking database…**, the business `App` never mounted, and Quit/Retry never appeared. If Retry itself rejected after the owner had already published `checking`, the loader returned with no recovery controls.

## Proposed change

- Add in-memory error code `database_startup_unavailable` (`retryable: true`).
- Treat unclassified `verifyDatabase` failures as that blocked state and **resolve** `start()` / `retry()` instead of rejecting, so Retry is a real second attempt.
- Catch renderer `getState()` / `retry()` rejections and show the existing blocked card (Quit + Retry). A later live event still wins over a late `getState()` rejection.

```mermaid
stateDiagram-v2
  [*] --> checking: renderer mount
  checking --> blocked: classified DB error or IPC/unclassified failure
  checking --> ready: owner.complete
  blocked --> checking: Retry
  blocked --> [*]: Quit
  ready --> App
```

## Scope and non-goals

- No Prisma / migration / settings persistence changes.
- No historical data compatibility work.
- No new `phase`, IPC channel, or contract-catalog entry.
- Web still has no `databaseStartup` API and still mounts immediately. Headless still awaits the owner directly.

### Explicit callouts

| Topic | Change |
| --- | --- |
| Historical data compatibility | None |
| New enum | `database_startup_unavailable` on `DatabaseStartupErrorCode` (process-local, not persisted) |
| Persistence | None |

First-launch unclassified errors no longer reject the compose `Promise.race` and `app.quit()`. They now stay on the startup shell, matching classified `blocked` failures.

- `getState()` IPC reject shows the blocked card with Retry and Quit; `App` stays unmounted.
- A live `ready` / progress event is not overwritten by a later `getState()` rejection.
- `retry()` IPC reject after a `checking` event returns to the blocked card with Retry.
- Unclassified owner verification failures resolve to retryable `database_startup_unavailable`; a later Retry can verify successfully.

## Review focus

- Whether `database_startup_unavailable` is the right code for both IPC reject and unclassified owner failures.
- Whether resolving (not rejecting) unclassified `start()` is the intended first-launch lifecycle.

## Test Impact Set

These checks ran after the last material edit, from `.worktree/database-startup-gate-ipc-reject`:

| --- | --- | --- |
| Gate IPC reject / retry restore | `npm test -- src/renderer/src/components/database-startup-gate.test.tsx` | pass (7) |
| Owner unclassified failure + retry | `npm test -- src/main/database/database-startup-owner.test.ts` | pass (4) |
| i18n catalogs for the new `t()` string | `npm test -- src/renderer/src/i18n/resources.test.ts` | pass (88) |
| IPC bridge and startup logging consumers | `npm test -- src/main/database/database-startup-ipc.test.ts src/main/database/database-startup-logging.test.ts` | pass (8) |
| Main/shared types | `npm run typecheck:node` | pass |
| Changed TS files | `npx eslint --cache` on the 5 edited TS files | pass |

These files are not a named `module-impact.json` module. Full `npm test` was not run locally; exact-head CI is authoritative.

### Included / excluded

- Included: owner of the gate and startup owner, i18n guard for the new string, adjacent IPC/logging consumers, both process typechecks.
- Excluded: Web/headless startup (API absent / no UI gate), Prisma/migration suites (no schema change), E2E/platform packaging (no window/lifecycle contract change beyond staying on the existing blocked shell).

### Uncovered risks

- A permanent missing IPC handler still only offers in-app Retry/Quit; it cannot repair a disposed main handler.
- Existing classified `DatabaseMigrationError.message` values remain English from main and are still rendered raw.

### DIFF
--- a/src/main/database/database-startup-owner.test.ts
+++ b/src/main/database/database-startup-owner.test.ts
@@ -73,6 +73,35 @@ describe('database startup owner', () => {
     expect(verifyDatabase).toHaveBeenCalledTimes(2)
   })
 
+  it('blocks an unclassified verification failure so retry remains available', async () => {
+    let attempt = 0
+    const reportBlocked = vi.fn()
+    const verifyDatabase = vi.fn(async () => {
+      attempt += 1
+      if (attempt === 1) throw new Error('startup subscriber failed')
+    })
+    const owner = createDatabaseStartupOwner({ reportBlocked, verifyDatabase })
+
+    await expect(owner.start()).resolves.toMatchObject({
+      phase: 'blocked',
+      error: {
+        code: 'database_startup_unavailable',
+        message: 'Open Science could not finish checking its database.',
+        retryable: true
+      }
+    })
+    expect(reportBlocked).toHaveBeenCalledWith(
+      expect.objectContaining({
+        code: 'database_startup_unavailable',
+        retryable: true
+      })
+    )
+
+    await expect(owner.retry()).resolves.toEqual({ phase: 'checking' })
+    await expect(owner.whenVerified()).resolves.toBeUndefined()
+    expect(verifyDatabase).toHaveBeenCalledTimes(2)
+  })
+
   it('does not retry a non-retryable compatibility failure', async () => {
     const reportBlocked = vi.fn()
     const verifyDatabase = vi.fn(async () => {
--- a/src/main/database/database-startup-owner.ts
+++ b/src/main/database/database-startup-owner.ts
@@ -45,19 +45,28 @@ const createDatabaseStartupOwner = (deps: DatabaseStartupOwnerDeps): DatabaseSta
         return state
       })
       .catch((error: unknown) => {
-        if (!(error instanceof DatabaseMigrationError)) throw error
+        const classified =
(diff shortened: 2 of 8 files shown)


ewen-poch:
**[P1] Locale changes can block an already-ready application**

Adopted in 385b0554.

- Startup `getState` / `onStateChanged` sync is one-shot again (`databaseStartup` only). Changing locale no longer re-invokes IPC.
- A rejected catch-up read only replaces `checking`. `ready`, `migrating`, and `blocked` are left in place, so a late or repeated rejection cannot unmount the app.
- `retry()` rejection still restores the existing blocked card unless the gate is already `ready`.
- `database_startup_unavailable` copy is translated at render time, so a locale change updates the message without another `getState()`.

PR #1399, opened 2026-08-19
feat(tags): add cross-resource catalog tags

### LINKED ISSUE #1361: [Feature]: Add user-defined groups for organizing Skills and Specialists
### Preflight checklist

As users install or create more Skills and Specialists, both catalogs become increasingly difficult to browse.

Skills and Specialists are currently presented primarily as flat lists. Search helps when the user already knows an item's name, but it does not help users maintain a large personal catalog organized around their own workflows.

For example, a user may want to organize Skills into groups such as:

- Literature review
- Data analysis
- Visualization
- Academic writing
- Frequently used

The same problem applies independently to Specialists. A user may want groups such as:

- Bioinformatics
- Clinical research
- Manuscript preparation
- Project-specific Specialists

Built-in distinctions such as Featured, Imported, and Personal describe where an item came from. They do not represent the user's own organizational structure.

As the number of installed items grows, users must repeatedly scroll through long lists or remember exact names, making both Settings and the conversation picker less convenient to use.

### Proposed solution

Add user-defined groups for Skills and Specialists.

Skills and Specialists should have separate group collections, because users may want to organize the two catalogs differently.

#### Group management

Users should be able to:

- Create a group.
- Rename a group.
- Reorder groups.
- Delete a group.
- Move a Skill or Specialist into a group.
- Move an item back to an automatically provided **Ungrouped** section.

Deleting a group must not delete, disable, or uninstall the Skills or Specialists assigned to it. Its items should return to **Ungrouped**.

The initial implementation can use flat groups without nested folders.

#### Skills

In **Settings → Skills**, users should be able to assign installed Skills to custom groups and browse the catalog by group.

The same groups should also be available in Skill selection surfaces, including the `/` picker in the conversation composer.

#### Specialists

In the Specialists management interface, users should be able to assign Specialists to custom groups and browse them by group.

The Specialist selector used when creating or configuring a Session should display the same organization.

#### Browsing behavior

- Groups should be collapsible.
- Search should work across all groups.
- Search results should preserve or display each item's group.
- Existing origin and status information, such as Featured, Imported, Personal, enabled, or disabled, should remain available.
- Users who have not created any groups should retain the current behavior without additional setup.

#### Persistence and lifecycle

- Group definitions and item assignments should persist across application restarts.
- Importing or creating a new item should place it in **Ungrouped** by default.
- Updating, enabling, or disabling an item should not change its group.
- Removing an item should also remove its stale group assignment.
- Group membership should use a stable item identity rather than only the display name, so renaming or updating an item does not lose its assignment.

This organization should be treated as user-owned catalog metadata. It should not modify Skill packages, Specialist packages, runtime selection, permissions, or execution behavior.

#### Search only

Search is useful when the user already knows what they need, but it does not provide a persistent way to organize a large collection or browse related items together.

#### Automatic categorization

Categories inferred from names, descriptions, or package metadata may be useful in the future, but they may not match a user's projects or working habits. Manual organization should remain available and predictable.

#### Name prefixes

Users can rename items or add prefixes such as `writing-` and `analysis-`, but this changes their visible identity, produces inconsistent naming, and does not provide collapsible browsing.

#### Use existing Featured, Imported, and Personal sections

These sections describe item provenance rather than user intent. A Personal Skill may belong to literature review, visualization, or another user-defined workflow category.

#### Tags instead of groups

Tags allow an item to belong to multiple categories, but they introduce a more complex filtering and editing interface. A first version with one flat custom group per item would solve the primary browsing problem. Multi-group membership or tags could be considered separately if needed.

### Additional context

Suggested acceptance criteria:

- A user can create, rename, reorder, and delete Skill groups.
- A user can independently create, rename, reorder, and delete Specialist groups.
- Skills and Specialists can be moved between their respective groups.
- New or unassigned items appear under **Ungrouped**.
- Deleting a group preserves all items assigned to it.
- Assignments survive application restarts and item updates.
- Settings pages and conversation/session selectors display the same grouping.
- Search continues to find items regardless of whether their group is collapsed.
- Existing users see no behavior change until they create a group.

ewen-poch:
## Problem

Issue #1361 proposes catalog groups, but single-membership groups do not cover resources that belong to several useful collections or a cross-resource Favorites view.

## Proposed change

- add many-to-many Tags for Skills, Connectors, and runnable Specialists
- add the protected, localized Favorites system Tag
- add Settings -> Capabilities -> Tags for create/edit/delete, type counts, search, aggregation, and navigation
- add searchable assignment menus, badges, filters, detail/editor controls, and Tag-filtered Specialist capability candidates
- persist Tag definitions and assignments in local SQLite through immutable migration `0011_cross_resource_tags`
- clean assignments at resource-deletion boundaries, with reconciliation and retry as a fallback
- run the existing `build:e2e` command with a 4 GB Node heap so the Electron Vite bundle does not intermittently exhaust the default macOS runner heap

```mermaid
flowchart LR
  Panels["Catalog panels and detail views"] --> Store["Renderer Tag store"]
  Browser["Settings Tags browser"] --> Store
  Store --> Commands["Validated application commands"]
  Commands --> Service["Tag service and resource registry"]
  Service --> SQLite["Tag and TagAssignment"]
```

## Scope and non-goals

- V1 resource types are `catalog.skill`, `catalog.connector`, and `catalog.specialist`; Reviewer is excluded.
- `resourceType` remains an application-registry-validated SQLite string so later Artifact, Session, Message, Notebook, and other adapters do not require a table rebuild.
- Existing pin, bookmark, and resource package data is not migrated. Skill, Connector, and Specialist file formats do not change.
- Tag metadata is device-local and is not part of import/export or cloud sync.
- Specialist capability configuration still persists concrete resource IDs, never a dynamic Tag query.

The final Test Impact Set ran after the last material edit:

- Tag ownership, deletion cleanup, renderer rollback, Skill consumers, and i18n -> focused Vitest suites -> 152 tests passed
- Skills/Connectors/Specialists Tag-filter consumers -> focused panel suites -> 119 tests passed
- database upgrade chain, packaged ledger, and constraints -> focused migration suites -> 87 tests passed
- Node and renderer contracts -> `npm run typecheck:node`, `npm run typecheck:web` -> passed
- changed TypeScript/TSX sources -> focused ESLint invocation -> passed
- checked-in generators -> `npm run db:schema:check`, `npm run check:web-api-map` -> passed
- Electron Vite build with the CI heap limit -> complete local build -> passed
- final PR CI -> static checks, full module tests and coverage, macOS/Windows E2E, CodeQL, and Codex review -> passed

Per maintainer instruction, the full local `npm test` suite was not run. PR CI is authoritative for the complete portable and platform test plan. Visual/E2E and future non-catalog resource adapters remain uncovered locally.

## Review focus

- system/custom Tag invariants and `nameKey` normalization
- migration `0011` ordering and historical compatibility
- resource deletion cleanup and same-ID reuse behavior
- Electron/Web command registration and cross-renderer convergence
- Settings history, accessibility, and three-locale behavior

Closes #1361


ewen-poch:
Addressed the host-driven Skill deletion path in 871e08c4. The host catalog adapter now invokes the existing tag cleanup after deleting a published Skill, so host-triggered deletions no longer bypass assignment cleanup. Verified with node/web typechecking and the focused Settings workflow, host Skills service, and Tag service tests (56 tests).

ewen-poch:
Rebased onto the latest main (64380d75) and updated the resource-list presentation in 9b65eb40. Connector and Specialist Tag badges now live in their third metadata row, matching Skills; compact badges retain long-name truncation and +N overflow. Verified with 67 focused renderer tests, web typecheck, targeted ESLint, and isolated Electron screenshots using a temporary storage root.

ewen-poch:
Addressed the scoped correctness findings in bb4cf39a. TagService now serializes each repository mutation together with its revision/event and authoritative snapshot, preventing out-of-order mutation responses. TagRepository now validates the normalized comparison key length before both create and update, including Unicode lowercase expansion cases. Verified with 17 focused Tag service/repository/store tests, node and web typechecking, targeted ESLint, and diff checks.\n\nThe deletion-cleanup P2 findings are intentionally not expanded into a durable tombstone system in this PR. A crash-safe fix requires persistent lifecycle coordination and ID-reuse guards across Skill, Connector, and Specialist owners, plus migration/compatibility work; propagating the current post-delete callback alone would report failure after the resource is already gone and would not solve restart safety. Given the rare database-failure-plus-ID-reuse trigger and the cross-module persistence cost, this is an accepted trade-off for the current scope.

ewen-poch:
Addressed the stale renderer snapshot finding in dece99f8. Tag loads now refuse snapshots older than the store revision, return the load state to ready when such a response is ignored, and successful mutation snapshots invalidate any older in-flight load. Added regressions for both a load completing after a mutation and a directly stale load response. Verified with focused Tag store/service tests, web typechecking, targeted ESLint, and diff checks. The durable post-delete tombstone finding remains the previously documented accepted scope trade-off.

ewen-poch:
Addressed the assignment-validation race in 3c54c4b4. Resource existence is now checked inside the same TagService mutation queue as the assignment write, so pending resource cleanup is ordered before validation and a deleted resource cannot be re-assigned through the previous pre-queue gap. Added a regression that blocks cleanup, starts an assignment, and verifies validation waits for cleanup and rejects the deleted resource. Verified with 13 focused service/store tests, node typechecking, targeted ESLint, and diff checks. The durable cleanup/tombstone item remains the accepted scope trade-off already documented.

ewen-poch:
Rebased onto the latest main and addressed the current PR Gate failures in e0da965d. The three Tag catalog entries now follow the updated i18n glossary by localizing user-visible Skill, Connector, and Specialist terminology. Migration retention tests now expect the newest two restorable snapshots after adding 0011_cross_resource_tags, including the failed-future-migration and unknown-backup cases. No migration implementation, schema, database version, persistence relationship, or UI structure changed.\n\nVerified with the focused i18n and migration suites (168 tests), Tag service/repository/store regressions (20 tests), node and web typechecking, targeted Prettier, and git diff checks.

ewen-poch:
Updated the two remaining migration backup retention assertions in 78b2c84e. They now verify 0010_compute_password_auth and 0011_cross_resource_tags as the newest two restorable snapshots after adding migration 0011. This is a test-only alignment: no migration implementation, schema, database version, or persistence behavior changed. Verified with the three focused database migration suites (64 tests passed), Prettier, and git diff checks.

ewen-poch:
The latest review repeats the previously assessed durable deletion-barrier concern. Addressing it correctly would require a persisted cleanup/tombstone mechanism and a broader resource-deletion contract, changing persistence and architecture beyond this PR. Per the agreed scope and trade-off, no additional change is made here; this PR retains the current best-effort cleanup behavior.

PR #1402, opened 2026-08-19
fix(updater): preserve completed downloads during checks

ewen-poch:
## Problem

Scheduled and manual checks can overlap an update download. A check that completes after the transfer can replace `ready` with `available`, `up-to-date`, or `error`; the manifest strategy also loses `localPath`, while the in-place strategy then fails the guarded `apply()` readiness requirement.

## Proposed change

- Serialize checks against active downloads in both updater strategies.
- Continue scheduled/manual checks after an update is ready, but reconcile results by version:
  - same/older release or check failure preserves the ready artifact;
  - a strictly newer release supersedes it with a new available update.
- Keep Settings → “Check now” enabled for ready updates and disable it only while checking/downloading.
- Add deterministic ordering, ready-preservation, failure, and newer-version coverage for both strategies.

## Scope and non-goals

- Covers both manifest and in-place updater strategies; the scheduler interface and composition are unchanged.
- Adds private, in-memory coordination fields only; no public or persisted update metadata changes.
- No shared update states, enum values, database changes, migrations, or persistence changes.
- A superseded user-selected installer remains at its existing filesystem path; this PR does not delete user files.

All checks below ran against the final material implementation:

- In-place mutual exclusion and version reconciliation → `npm test -- src/main/update/electron-updater-strategy.test.ts` → 44 passed.
- Manifest mutual exclusion and version reconciliation → `npm test -- src/main/update/service.test.ts` → 28 passed.
- Settings check-action availability → `npm test -- src/renderer/src/pages/settings/AppVersionSection.render.test.tsx` → 6 passed.
- Main-process types → `npm run typecheck:node` → passed.
- Renderer types → `npm run typecheck:web` → passed.
- Source lint/format contract → `npm run lint` → passed with no warnings.

The complete portable and platform suites were intentionally left to PR CI.

## Review focus

- Whether same/older/error results correctly preserve the ready artifact while strictly newer versions supersede it.
- Whether the event reconciliation in ElectronUpdaterStrategy covers provider-driven state transitions without changing the shared model.
- Whether leaving a superseded installer at its user-selected path is the correct non-destructive behavior.

### DIFF
--- a/src/main/update/electron-updater-strategy.test.ts
+++ b/src/main/update/electron-updater-strategy.test.ts
@@ -227,8 +227,38 @@ describe('ElectronUpdaterStrategy', () => {
 
     const readyStatus = strategy.getStatus()
     expect(await strategy.check()).toBe(readyStatus)
-    expect(checkCalls).toBe(1)
+    expect(checkCalls).toBe(2)
     expect(strategy.getStatus().state).toBe('ready')
+  })
+
+  it('preserves in-place ready on check failure but accepts a strictly newer update', async () => {
+    const updater = new FakeUpdater()
+    const strategy = new ElectronUpdaterStrategy({
+      updater,
+      currentVersion: '0.2.0',
+      broadcast: vi.fn(),
+      fetchImpl: offlineFetch()
+    })
+    await strategy.check()
+    await strategy.download()
+    const readyStatus = strategy.getStatus()
+
+    updater.checkForUpdates = vi.fn(async () => {
+      updater.emit('error', new Error('offline'))
+    })
+    expect(await strategy.check()).toBe(readyStatus)
+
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed the P1 in 6e321e361: ElectronUpdaterStrategy now preserves ready, rejects checks during the download lifecycle, and prevents downloads from starting while a provider check is active. Added deterministic coverage for both operation orderings; the full strategy test file passes (43/43).

ewen-poch:
Addressed the P2 in 198628efd with version-aware ready reconciliation. Checks now continue after ready: same/older/error results preserve the artifact, while only a strictly newer version transitions to available. “Check now” is enabled again for ready. No shared fields, states, persistence, or migrations were added; targeted suites pass (in-place 44, manifest 28, Settings 6).

PR #1403, opened 2026-08-19
fix(shutdown): limit task and reviewer cleanup waits

ewen-poch:
## Problem

Application quit can wait forever when the Web Task adapter is disposing an automatic Reviewer. `TaskRunner.dispose()` aborts the review but then awaits its completion without a deadline; the Web controller awaits Task disposal, and the ordered application-surface shutdown had no enclosing budget. A hung Reviewer command, runtime cleanup, pending Web start, or server close could therefore prevent `app.exit(0)` indefinitely.

## Proposed change

- Give an aborted automatic Reviewer one second to settle before Task disposal proceeds.
- Give the complete ordered application-surface shutdown sequence a five-second hard deadline.
- Preserve shutdown order while reserving an equal share of the remaining global time for every surface still to be attempted, so one hung surface cannot skip later cleanup.
- Return the existing `timeout` shutdown outcome and record the surface whose allocated time expires.
- Keep abandoned promises observed so a late rejection cannot become unhandled.
- Add regressions for a Reviewer that ignores abort and a Web surface that never settles while later surfaces still run.

## Scope and non-goals

- Preserve the existing Web controller -> application runtime -> Remote Access -> IPC shutdown order, including after a timeout.
- Do not add schema fields, migrations, persistent domain data, relationships, or lifecycle enum values.
- Do not change UI copy, dialogs, or normal quit interaction.
- Keep the existing startup recovery that converts interrupted persisted Reviews from `running` to `error`.

- Hung Reviewer abort -> `npm test -- src/main/application-runtime.test.ts src/main/tasks/task-runner.test.ts src/main/web-service/task-api.test.ts src/main/web-service/controller.test.ts src/main/app-lifecycle.test.ts` -> passed, 136 tests.
- Hung application surface, later-surface ordering, and shutdown outcome propagation -> same focused test set -> passed.
- Task API, Web controller, and Electron lifecycle consumers -> same focused test set -> passed.
- Main-process type safety -> `npm run typecheck:node` -> passed.
- Source and test lint -> `npm run lint` -> passed.

The two original focused regressions failed before the implementation (`undefined` instead of `timeout`, and Task disposal still pending after one second) and pass with the fix. The timeout regression also verifies that application runtime, Remote Access, and IPC cleanup are attempted after a hung Web controller.

The complete local `npm test` suite was intentionally not run per maintainer request; exact-head PR CI remains authoritative for the portable and platform lanes. A synchronous event-loop block cannot be preempted by a JavaScript timer and remains outside this fix.

## Review focus

- Whether one second is an appropriate graceful Reviewer-abort window.
- Whether sharing the five-second hard budget across the remaining ordered surfaces gives the right cleanup/exit tradeoff.
- Promise observation and timer cleanup on both the normal and timeout paths.

### DIFF
--- a/src/main/application-runtime.test.ts
+++ b/src/main/application-runtime.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it, vi } from 'vitest'
 
 import {
+  APPLICATION_SURFACE_SHUTDOWN_BUDGET_MS,
   composeApplicationRuntime,
   composeApplicationRuntimeWithAdapters,
   shutdownApplicationSurfaces,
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the P1 shutdown-order finding in 5c16ba917. The five-second global budget is now apportioned across each remaining surface, so a hung Web controller times out and application runtime, Remote Access, and IPC cleanup are still invoked in order. The revised regression asserts those later calls, and the focused 136-test set, Node typecheck, and lint pass on the updated head.

PR #1408, opened 2026-08-19
fix(acp): acknowledge prompts after provider admission

### LINKED ISSUE #1406: [Bug]:
### Preflight checklist

### What happened?

Error invoking remote method 'acp:send-prompt': reply was never sent

### Steps to reproduce

Just sending messages to dsv4flash models.

### Operating system

Windows

### App version

0.17.0

### Provider / model

DeepSeek · deepseek-v4-flash

```shell
Detected OS: Windows
Agent framework: Claude Code
Runtime: Electron 39.8.10, Chrome 142.0.7444.265, Node 22.22.1

Runtime log not attached automatically (it can contain local paths and prompts).
Reveal it from Settings → General → Diagnostics and attach after reviewing.
```

ewen-poch:
## Problem

The reporter-provided `main.log` shows that the configured DeepSeek route completed multiple turns normally, so this is not a model-compatibility failure. In the failing lifecycle, the provider logged `prompt stopped`, but the original `acp:send-prompt` handler was still unsettled when application shutdown reached backend teardown. The handler then rejected after about 711 seconds, after Electron had lost the renderer reply channel, producing `reply was never sent`.

This can happen when a long interactive turn, or maintenance performed after its terminal stop, keeps the invoke request open while the window exits or reloads. The secondary IPC failure can make a completed or already-terminal turn appear to have failed.

Fixes #1406.

## Proposed change

- Treat interactive prompt dispatch as an admission RPC and acknowledge it only when the provider explicitly accepts the prompt.
- Keep the authoritative prompt execution, session lock, terminal maintenance, and state/event publication running through the existing background lifecycle.
- Preserve synchronous admission failures: rejection or completion before provider acceptance still rejects the RPC and rolls back task-notification tracking.

## Scope and non-goals

- No new data fields, schema changes, enum values, or data-model changes.
- No persistence or migration changes; historical data compatibility is not applicable.
- No user-interface change. This only changes when the existing prompt-dispatch IPC response settles.
- No provider-specific behavior change. Claude Code, OpenCode, Codex Responses, and Codex bridge share the coordinator path covered by this change.
- Automatic compaction and Plan finalization behavior are not changed.

- Interactive prompt admission settles before turn completion on Claude Code, OpenCode, Codex Responses, and Codex bridge; a turn ending without explicit provider acceptance rejects admission; pre-admission notification rollback remains intact -> `npx vitest run src/main/acp/runtime-coordinator.test.ts src/main/acp/handler-workflows.test.ts src/main/acp/ipc.test.ts` -> 161 tests passed.
- Main-process TypeScript contracts remain valid -> `npm run typecheck:node` -> passed.
- Repository lint rules remain valid -> `npm run lint` -> passed.
- Changed files retain repository formatting -> `npx prettier --check src/main/acp/runtime-coordinator.ts src/main/acp/runtime-coordinator.test.ts src/main/acp/handler-workflows.ts src/main/acp/handler-workflows.test.ts src/main/acp/ipc.test.ts` -> passed.
- Patch whitespace is valid -> `git diff --check` -> passed before commit.

Per the issue request, the complete local `npm test` suite was not run. Exact-head PR CI remains authoritative for the complete portable and platform test plan. A packaged Electron replay of the reporter's environment is not covered locally.

## Review focus

- Confirm that the admission promise resolves only from explicit provider acceptance and rejects if the turn ends first.
- Confirm that the detached completion promise remains observed, while the runtime retains prompt ownership and publishes terminal state or late errors through the existing event path.
- Confirm that failures before acceptance still reach the renderer as dispatch failures and roll back notification tracking.

### DIFF
--- a/src/main/acp/handler-workflows.test.ts
+++ b/src/main/acp/handler-workflows.test.ts
@@ -81,6 +81,7 @@ const createHarness = (
   saveAsSkillAdmission?: Parameters<typeof createAcpHandlerWorkflows>[5]
 ): {
   workflows: ReturnType<typeof createAcpHandlerWorkflows>
+  startPrompt: ReturnType<typeof vi.fn>
   startContinuation: ReturnType<typeof vi.fn>
   startContinuationWhenDispatchAdmitted: ReturnType<typeof vi.fn>
   hasLiveSession: ReturnType<typeof vi.fn>
@@ -97,6 +98,7 @@ const createHarness = (
   const session = createSession()
   mutate?.(session)
   prepareControlTurn(session)
+  const startPrompt = vi.fn(async (request: unknown) => void request)
   const startContinuation = vi.fn(async (request: unknown) => void request)
   const hasLiveSession = vi.fn(() => true)
   const captureSessionBackend = vi.fn(
@@ -126,7 +128,7 @@ const createHarness = (
       hasLiveSession,
       captureSessionBackend,
       resumeSession: vi.fn(),
-      sendPrompt: vi.fn(),
+      startPrompt,
       getLatestUserPrompt: vi.fn(),
       startContinuation,
       startContinuationWhenDispatchAdmitted
@@ -141,6 +143,7 @@ const createHarness = (
   const frame = graph.frames.find(({ id }) => id === graph.activeFrameId)!
   return {
     workflows,
+    startPrompt,
     startContinuation,
     startContinuationWhenDispatchAdmitted,
     hasLiveSession,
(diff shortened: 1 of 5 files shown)


ewen-poch:
Addressed in e226db00b. startPrompt now resolves only from the explicit provider-acceptance callback; if the underlying turn completes first, admission rejects and notification tracking is rolled back. Added a regression test for completion without acceptance and updated the IPC fake to model acceptance only on successful provider dispatch.

PR #1414, opened 2026-08-19
feat(specialists): add quick appearance picker

ewen-poch:
## Problem

Changing a custom Specialist's icon or color currently requires opening the full detail editor, even though these are small, independent appearance changes.

## Proposed change

- Turn the avatar on each custom Specialist row into a dedicated quick-edit trigger.
- Open a compact, collision-aware popover with the same complete grouped app icon registry and six colors as the full editor.
- Keep all icon cells aligned in the same responsive grid and route vertical wheel/trackpad input from the whole popover into a bounded native icon viewport whose scrollbar is visually hidden.
- Preserve native touch and keyboard scrolling on the focusable icon viewport.
- Save each selection immediately through the existing revision-checked update path, with optimistic feedback, rollback, and inline retry.
- Keep the row body as the entry point to the full editor and share the icon registry and color definitions between both surfaces.
- Add translated labels for Simplified Chinese, Traditional Chinese, and Japanese.
- Keep the desktop popover at 256 px with 36 px controls, while retaining 44 px targets for coarse pointers and a 3-column narrow-width layout.

## Scope and non-goals

- Custom Specialists only; Built-in Specialists and Reviewer remain read-only.
- No new icon/color values, persisted fields, domain enums, IPC contracts, schemas, or migrations.
- Historical records with missing appearance fields continue to use the existing avatar fallbacks.
- Persistence continues through the existing `UpdateSpecialistInput` and `specialists.json` revision flow. Imported Specialists continue to use the existing local-modification semantics.

All checks below ran after the final scrolling edit.

- Direct color update without navigation, full shared icon availability, failure rollback, inline retry, and popover-level wheel routing → `npm test -- src/renderer/src/pages/settings/SpecialistsPanel.render.test.tsx` → 45 passed.
- Full-editor grouped icons and shared colors remain functional → `npm test -- src/renderer/src/pages/settings/SpecialistEditor.render.test.tsx` → 18 passed.
- Shared app icon registry invariants → `npm test -- src/renderer/src/components/app-icons/registry.test.ts` → 5 passed.
- Translation catalogs and call sites → `npx vitest run src/renderer/src/i18n/resources.test.ts` → 109 passed.
- Renderer type safety → `npm run typecheck:web` → passed.
- Source lint → `npm run lint` → passed.
- Changed-file formatting → `npx prettier --check ...` → passed.
- Responsive interaction/visual check at 320 px and 768 px → temporary local Playwright preview → passed; hovering the Appearance title and sending real wheel input increased the icon viewport's `scrollTop`, reached the Engineering group, reported `scrollbar-width: none`, preserved sub-1 px color/icon column alignment, and caused no horizontal overflow.

`npm run test:affected:explain -- --base origin/main --head HEAD` selects the full CI plan because `.hallmark/preflight.json` is classified as an unknown path. Per maintainer direction, the complete local `npm test` suite was not run; PR CI is authoritative for that coverage. No main-process, preload, database, migration, or platform-specific behavior changed.

## Review focus

- Avatar trigger versus row-body interaction boundaries.
- Optimistic save and retry behavior under revision conflicts.
- Shared grouped icon registry, aligned grid cells, hidden scrollbar, and popover-level wheel routing.
- Popover density, focus states, and collision behavior at narrow widths.


ewen-poch:
Addressed the appearance-only runtime reload finding in `ede92c67`.

The existing `specialist.update` handler now skips `onProfilesChanged` only when the request is limited to `id`, `revision`, `iconKey`, and/or `colorKey`. Any update containing another field continues to request the existing runtime reload, so capability and other runtime-visible changes retain their previous behavior.

Added main-process regression coverage proving that icon-only and color-only updates do not invalidate the runtime while a capability update still does. Validated with the focused IPC suite (13 tests), `typecheck:node`, lint, and Prettier.

PR #1417, opened 2026-08-19
feat(compute)!: streamline agent host enumeration

wen2zhou:
## Problem

Agent-side compute host enumeration returned full `ComputeHost` records, including details documents and probe snapshots. Large results could be truncated in the REPL kernel, forcing the Agent to retry with field projections and waste model tokens. Session-enabled enumeration also returned only provider ID strings, requiring another lookup before choosing among multiple hosts.

## Proposed change

- Replace `host.compute.list()` with `listRegistered()` and `listCompute()` with `listEnabled()` without compatibility aliases.
- Return the same compact `{ provider_id, display_name, shape, status }` summary from both enumeration methods.
- Preserve enabled-registry order while joining registered hosts and silently skip deleted host IDs.
- Keep details documents and probe snapshots out of enumeration; expose the full projected probe snapshot through `details()` on demand.
- Update bundled compute guidance and runtime surface coverage for the new names and return shape.

## Scope and non-goals

- No renderer enumeration behavior or `ComputeHost` IPC shape changes.
- No database schema or persistence changes.
- No compatibility layer for the retired Agent-facing methods.

All checks below ran after the final material code edit. The later commit amendment changed only commit metadata.

- Compact registered/enabled enumeration, registry-order join, deleted-ID filtering, on-demand probe details, REPL mapping, and Skill consumers -> `npm test` -> 1112 files passed, 15 skipped; 16690 tests passed, 217 skipped.
- Node and renderer TypeScript consumers -> `npm run typecheck` -> passed.
- Source and tests -> `npm run lint` -> passed.
- Patch integrity -> `git diff --check origin/main...HEAD` -> passed.

`test:affected -- --base origin/main --head HEAD` selected full mode because `resources/notebook/repl_loop.js` is not mapped to a known Module. The initial run exposed a stale local Prisma Client after rebasing onto the new Tag schema; `npx prisma generate` refreshed the untracked dependency output, the nine affected database/Tag tests then passed, and the complete suite passed on the final run.

Consumer coverage includes both Node and Web typechecks because shared compute types are touched. No platform-specific, persistence, migration, build, or UI E2E behavior changed; the complete portable suite covers the affected runtime and contract paths, while exact-head PR Gate remains authoritative for platform lanes.

## Review focus

- Whether the four summary fields are sufficient for one-call Agent host selection.
- Separation between compact enumeration and on-demand probe/details reads.
- Preservation of enabled-registry ordering and omission of deleted hosts.
- Intentional breaking removal of `list` and `listCompute`.

### DIFF
--- a/resources/notebook/repl_loop.js
+++ b/resources/notebook/repl_loop.js
@@ -3203,14 +3203,17 @@ function computeError(raw) {
 // host.compute namespace. Public methods and input keys are camelCase; the adapter immediately maps
 // them to the existing snake_case computeCall RPC contract.
 const hostCompute = {
-  // Enumerate registered compute hosts for discovery. No approval, no session context.
-  async list() {
+  // Enumerate ALL registered compute hosts as summaries (provider_id/display_name/shape/status).
+  // No approval, no session context. Heavy fields live behind details(); listRegistered() itself
+  // stays small so the repl_execute tool result is never truncated.
+  async listRegistered() {
     return computeRpc({ op: 'list' })
   },
 
-  // Returns session-enabled compute hosts (≠ list() which returns all registered hosts).
-  // Uses COMPUTE_SESSION_ID from spawn env so the registry lookup is always session-scoped.
-  async listCompute() {
+  // Returns the session-enabled compute hosts (≠ listRegistered() which returns all registered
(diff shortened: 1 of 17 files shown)


wen2zhou:
merged by https://github.com/aipoch/open-science/pull/1496

PR #1419, opened 2026-08-19
refactor(connectors): extract application composition from ipc.ts

LiuZhaoyan:
## Problem

The connector wiring in `src/main/ipc.ts`'s composition root inlined ~120 lines constructing the MCP client manager, connector/skill approval brokers, runtime settings projection, and connector service. That block is not unit-testable in isolation and grows an already-large file that also owns unrelated IPC handlers.

## Proposed change

Extract the block into `createConnectorApplicationModule` in a new `src/main/connectors/application.ts`. The factory takes a narrow `ConnectorApplicationDeps` (settings, broadcasts, uploads, fetch, key resolution, profile resolution, late-bound local tool handlers) and returns an `ApplicationModule` exposing the same instances `ipc.ts` used to build inline. `ipc.ts` now constructs the deps object and destructures the returned capability. Added `application.test.ts` — a composition test asserting that injected fakes are reused as-is and that `dispose()` closes the MCP manager through the runtime.

## Scope and non-goals

- Pure extraction; no behavior, interface, or instance-identity change. The same `McpClientManager`, `ApprovalBroker`, `SkillImportApprovalBroker`, `ConnectorRuntimeSettingsProjection`, `ConnectorService`, and `ConversationSkillImporter` flow to the same consumers under the same names.
- Late-bound local tool handlers (`molecule/preview_molecule`) stay a composition-root dependency, passed into the factory rather than owned by it.
- No new dependency, no settings/schema/migration change, no renderer touch.

| --- | --- | --- |
| Connector application composition (fake reuse + dispose) | `npm test -- src/main/connectors/application.test.ts` | ✅ pass (1 test, exit 0) |
| Main-process compile, incl. `ipc.ts` consumer of the factory | `npm run typecheck:node` | ✅ pass (exit 0) |
| Lint across changed source | `npm run lint` | ✅ pass (exit 0, 0 errors) |

Consumer/exclusion rationale: `ipc.ts` is the sole consumer of the new factory and is covered by `typecheck:node`. No Interface or Adapter contract changed, so connector/IPC feature-slice and E2E suites were not re-run. The 10 lint warnings are all pre-existing in files outside this diff (`scripts/fix-electron-path.cjs`, `src/main/compute/job-notifier.test.ts`, `src/renderer/...`, `src/shared/remote-fs.ts`).

Uncovered risks: only Windows verified locally; exact-head CI remains authoritative for the other platform lanes.

## Review focus

Confirm the destructure in `ipc.ts` (`connectorService`, `runtimeSettings`, `mcpClientManager`, `skillImporter`, `connectorApprovals`, `skillImportApprovals`) maps 1:1 to the previously inlined instances and that `dispose` still calls `mcpClientManager.closeAll()`.

### DIFF
--- a/src/main/application-command-wiring.test.ts
+++ b/src/main/application-command-wiring.test.ts
@@ -37,7 +37,7 @@ const notificationAdapterBlock = compact(
   between(
     ipcSource,
     "declareElectronAdapter('task-notifications'",
-    '// One MCP client manager backs both dispatch'
+    'const connectorApplication = await modules.add('
   )
 )
 const dependencyBlock = compact(
--- a/src/main/connectors/application.test.ts
+++ b/src/main/connectors/application.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it, vi } from 'vitest'
+
+import { composeApplicationRuntime } from '../application-runtime'
+import { createConnectorApplicationModule, type ConnectorApplicationDeps } from './application'
+
+describe('Connector application composition', () => {
+  it('reuses fakes and closes the MCP manager through runtime disposal', async () => {
+    const settings = {
+      getConnectors: vi.fn().mockResolvedValue({ customMcpServers: [] }),
+      saveCustomServerOAuthState: vi.fn().mockResolvedValue(undefined),
+      setCustomServerRuntimeProjectionProvider: vi.fn(),
+      setCustomServerAuthenticator: vi.fn(),
+      previewSkillArchive: vi.fn(),
+      importSkillArchiveBatch: vi.fn(),
+      scanRepoSkills: vi.fn().mockResolvedValue({ skills: [] }),
+      importSkill: vi.fn()
+    } as unknown as ConnectorApplicationDeps['settings']
+    const mcpClientManager = {
+      listTools: vi.fn().mockResolvedValue([]),
+      call: vi.fn(),
+      authenticate: vi.fn().mockResolvedValue(undefined),
+      cancelAuthentication: vi.fn().mockResolvedValue(undefined),
+      close: vi.fn().mockResolvedValue(undefined),
+      closeAll: vi.fn().mockResolvedValue(undefined)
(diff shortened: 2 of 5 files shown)


ewen-poch:
Maintainer follow-up

We reviewed this extraction against the current application-composition, Settings-boundary, and lifecycle contracts. The overall direction is sound, and we pushed commit `a7e0f861` to this branch to address a few issues found during review:

- The new module imported the concrete `SettingsService`, which expanded a locked importer boundary and failed `settings-backend.architecture.test.ts`. The dependency now goes through `settings/service-capabilities.ts` as a narrow capability.
- The extraction removed the source marker consumed by `application-command-wiring.test.ts`, causing that suite to fail during collection. The test now uses the new concrete composition call as its stable boundary.
- `McpClientManager` was created before the module was registered with `RuntimeModuleBuilder`, so a later construction failure could skip `closeAll()`. The factory now cleans up the manager on pre-ownership construction failure and preserves or aggregates the relevant errors. A regression test covers this path.

- Connector composition + production wiring + Settings architecture: 3 files, 20 tests passed
- `npm run typecheck:node`: passed
- `npm run lint`: passed
- Full `npm test`: passed
- Independent Standards and Spec re-review: no blocking findings

The same PR has been updated in place, and the new CI run is now authoritative for the pushed head. Thanks for keeping the fork branch maintainer-editable so we could help land the fixes directly.

ewen-poch:
Followed up on the Static checks failure from the latest CI run.

- Added an explicit return type for the connector application test harness, satisfying `@typescript-eslint/explicit-function-return-type`.
- Kept the change scoped to `src/main/connectors/application.test.ts`; there are no runtime, architecture, state, persistence, or UI changes.

Validation:
- `npx eslint --no-cache src/main/connectors/application.test.ts` — passed
- Focused composition/architecture tests — 3 files, 20 tests passed
- Prettier check — passed

Commit: `29631fdd`

PR #1424, opened 2026-08-19
fix(tags): clarify resource list hierarchy

ewen-poch:
## Problem

The Tags resource list gives dividers to sibling resources but not to source groups, so rows carry more visual weight than the Skills, Connectors, and Specialists hierarchy. Each row also repeats its resource type on a second line even though the group heading already provides that context.

## Proposed change

- Move the divider treatment from resource rows to resource-group boundaries.
- Add a subtle existing-token hover/focus surface to keep rows discoverable without persistent rules.
- Show each resource's existing description as secondary text; omit the line when the description is blank.
- Cover description rendering, blank-description behavior, and group/row divider structure in the Tags render tests.

## Scope and non-goals

- Renderer-only presentation change in the existing Tags settings panel.
- No changes to tag/resource data models, APIs, enums, persistence, or stored data.
- No new user-visible copy or locale catalog changes.
- No historical-data compatibility path is required; missing/blank existing descriptions simply render as a one-line row.

- Tags resources show descriptions instead of repeated type labels.
- Blank descriptions do not reserve an empty secondary line.
- Dividers separate resource groups, not rows within a group.
- Existing row open/remove actions and group collapse behavior remain covered.

Validation performed after the last material code change:

- `npm test -- src/renderer/src/pages/settings/TagsPanel.render.test.tsx src/renderer/src/pages/settings/SettingsPage.render.test.tsx src/renderer/src/i18n/resources.test.ts` — 3 files, 187 tests passed.
- `npm run lint` — passed.
- `npm run build:web` — passed.
- `npx prettier --check src/renderer/src/pages/settings/TagsPanel.tsx src/renderer/src/pages/settings/TagsPanel.render.test.tsx` — passed.
- Ego Lite visual verification against the local web build — group hierarchy and long-description truncation verified; no horizontal page overflow at the desktop viewport.

`test:affected:explain` falls back to the complete Vitest suite because `TagsPanel.render.test.tsx` has no configured module owner. Per the requested validation scope, the full local `npm test` was not run; PR CI remains authoritative for the complete suite.

Excluded local lanes: Electron/main-process, Node-only, and platform-specific suites, because the diff is limited to renderer presentation and its render test.

### DIFF
--- a/src/renderer/src/pages/settings/TagsPanel.render.test.tsx
+++ b/src/renderer/src/pages/settings/TagsPanel.render.test.tsx
@@ -89,6 +89,55 @@ describe('TagsPanel', () => {
     expect(resourceRow?.querySelector('[data-slot="tag-resource-subtitle"]')).toBeNull()
   })
 
+  it('keeps custom connector identities visible and accessible', async () => {
+    useSettingsStore.setState({
+      skills: [],
+      customServers: [
+        {
+          id: 'custom-alpha',
+          name: 'alpha-mcp',
+          displayName: 'Lab connector',
+          description: 'Search papers',
+          transport: 'stdio',
+          enabled: true
+        },
+        {
+          id: 'custom-beta',
+          name: 'beta-mcp',
+          displayName: 'Lab connector',
+          transport: 'stdio',
+          enabled: true
+        }
+      ]
+    })
+    useTagStore.setState({
+      assignments: ['custom-alpha', 'custom-beta'].map((resourceId, index) => ({
+        tagId: 'tag-favorite',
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the custom Connector identity issue in 62cbe9e8. Custom rows now follow the existing SpecialistEditor convention: the secondary line shows the immutable connector name followed by its description, or the immutable name alone when no description exists. Remove-action labels include both the display name and immutable name, so duplicate display names remain distinguishable. Added regression coverage for duplicate display names and optional descriptions. Validated with the focused TagsPanel render suite (9 tests), Web typecheck, lint, Prettier, and git diff checks.

PR #1431, opened 2026-08-19
fix(specialists): delete exclusively used skills

ewen-poch:
## Problem

Deleting a custom Specialist could not select Personal/Imported Skills that were used only by that
Specialist when the Skills had been installed independently or the Specialist had an empty
`ownedSkillIds` list. Settings showed those Skills as `Specialist only`, but the delete preview
classified them as standalone and always retained them.

This occurs when a Skill is referenced by the Specialist being deleted, disabled for the Main Agent,
and unused by every other Specialist, while its package provenance does not mark that Specialist as
the owner. Leaving it unchanged produces orphaned `Not in use` Skill packages that must be deleted
one at a time and makes the delete dialog disagree with the Settings scope.

The first implementation also exposed a production deadlock: the atomic transaction held the shared
Skill mutation lock while its final eligibility check rebuilt the live catalog, and that read tried to
acquire the same lock. The UI then remained at `Deleting...` with no error or filesystem commit.

## Change

- Derive deletion eligibility from live usage: non-builtin, not enabled for Main Agent, and with no
  other Specialist owner/reference.
- Let `Select all` select every deletable row while retained rows remain disabled and explain whether
  Main Agent or another Specialist still uses the Skill.
- Revalidate the selected Skills under the transaction lock before committing.
- Permit explicitly selected independent Personal/Imported Skills in the existing atomic filesystem
  transaction and preserve rollback behavior, including supported legacy directories.
- Establish a transaction-scoped reentrant lock context only after the adapter proves that the same
  transaction owns the mutation lock. This removes the delete deadlock and protects the equivalent
  replacement-install validation path.
- Remove successfully deleted Skill IDs from the existing `disabledSkillIds` setting so reinstalling
  the same ID does not inherit a stale Main Agent state. Settings cleanup and Tag cleanup remain
  independently failure-isolated.
- Update Settings design copy and all renderer locales.

## Data model, compatibility, and persistence

- **Schema/fields:** no new persisted field, table, column, IPC method, or migration.
- **Transient enum literals:** preview classification changes `owned-exclusive` to `exclusive` and
  protection reason `standalone` to `main-enabled`. These are IPC/runtime literals, not persisted
  lifecycle states.
- **Historical compatibility:** selected references with empty `ownedSkillIds` and legacy supported
  Personal/Imported directories without a sidecar work without migration. Malformed or ambiguous
  identities remain fail-closed.
- **Historical data not repaired:** pre-existing orphaned IDs already in `disabledSkillIds` are not
  scanned or rewritten. Cleanup applies only to Skills successfully deleted by this flow.
- **Persistence effects:** selected Skill directories are deleted through the existing atomic
  Specialist+Skill transaction and restored on failure. Existing `disabledSkillIds` values for those
  successfully deleted IDs are removed after commit; no storage format changes.

## Validation

- Specialist deletion eligibility, stale-selection checks, production lock regression, settings/Tag
  cleanup -> focused service and IPC tests -> passed.
- Atomic Skill removal, rollback, legacy identity support, mutation-context contract, replacement
  install lock regression -> focused transaction/adapter/service tests -> passed.
- Select-all behavior, retained-row explanations, and translations -> renderer render test and i18n
  guard -> passed.
- Combined focused suite: 6 files, 230 tests passed.
- `npm run typecheck:node` -> passed.
- `npm run typecheck:web` -> passed.
- `npm run lint` -> passed.
- Independent Standards review -> no findings.
- Independent Spec review -> no findings.

`test:affected:explain -- --base origin/main --head HEAD` selects the complete Vitest suite because
`docs/design.md` has no module owner. Per maintainer request, the complete local `npm test` was not
run; PR CI is the complete/platform authority.

## Review focus

- The reentrant context must remain reachable only for the transaction that currently holds the
  shared Skill mutation lock.
- Live Main Agent and other-Specialist use must continue to fail closed before deletion commits.
- Post-commit Settings cleanup must not change the atomic filesystem rollback boundary.


ewen-poch:
Thanks for identifying the crash window. We are intentionally not making the settings cleanup part
of durable package recovery in this PR.

The Specialist and selected Skill filesystem changes remain atomic and recoverable. The residual
risk is limited to a stale `disabledSkillIds` tombstone if the process exits after that transaction
commits but before the post-commit Settings cleanup completes, or if the Settings write fails. That
can only cause a later reinstall of the same Skill ID to remain disabled for the Main Agent; it does
not leave the Specialist or Skill deletion partially committed.

Persisting this cleanup would expand the transaction journal format and recovery ownership, require
legacy-journal compatibility, and change the current failure-isolation boundary between atomic
package deletion and best-effort Settings/Tag cleanup. That is a disproportionate architectural and
persistence change for this bug fix, so we are retaining the documented best-effort behavior and the
explicit non-goal of repairing historical orphaned disabled IDs.

PR #1441, opened 2026-08-19
feat(tags): add persistent drag ordering

ewen-poch:
## Problem

Tags had a deterministic name-based display order but users could not control it. Resources with multiple Tags therefore could not present those Tags in a user-defined priority order.

## Proposed change

- Persist one global Tag order with a unique integer sortOrder.
- Keep Favorites fixed at position 0 and reject any reorder payload containing system Tags or an incomplete custom Tag set.
- Append new Tags, compact positions after deletion, and persist reorders transactionally.
- Add pointer/touch drag handles with a movement threshold, insertion feedback, and pointer capture.
- Add Arrow Up and Arrow Down keyboard reordering, retained focus, live announcements, busy state, and visible failure recovery.
- Make resource Tag badges inherit the authoritative global snapshot order.

The interaction follows the WAI-ARIA rearrangeable listbox guidance for keyboard movement and announcements, and MDN pointer-capture guidance for robust pointer drags.

## Scope and non-goals

- Global ordering only; no per-resource Tag order.
- No Tag assignment or resource file-format changes.
- No new dependency and no new persistent enum or business-state value.

## Data persistence

Adds Tag.sortOrder and migration 0012_tag_ordering. Favorites owns 0; custom Tags use contiguous positions starting at 1. Reorder writes use temporary negative positions followed by final positive positions in one transaction to preserve the unique constraint.

## Historical compatibility

Existing databases are backfilled in their current visible order: Favorites first, then custom Tags by nameKey and id. Existing TagAssignment rows and resource files are unchanged. Downgrading a migrated database remains outside the supported migration policy.

## Acceptance and validation

- Targeted Tag contracts, repository, service, store, migration, and Settings panel tests pass.
- Database migration and backup suites pass.
- Renderer/preload/application-command contract suites pass.
- Four-locale i18n resource guard passes.
- Node and renderer type checks pass.
- ESLint, generated database schema check, generated Web API map check, and git diff check pass.
- Full npm test intentionally not run; PR CI owns the complete suite.

## Review focus

- Migration backfill and unique-order transaction behavior.
- System Tag immovability and exact-set validation.
- Pointer, keyboard, focus, and rollback behavior in TagsPanel.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -770,7 +770,9 @@ colors communicate a successful or failed probe/migration result.
   resource's detail or editor.
 - The Tag list is a single-action selector: every row keeps its resource count in one right-aligned
   trailing column. Persistent edit and delete icon actions belong to the selected custom Tag's detail
-  header instead of individual list rows; the protected Favorites Tag exposes neither action.
+  header instead of individual list rows; the protected Favorites Tag exposes neither action. A
+  leading handle reorders custom Tags by pointer or keyboard. Favorites instead shows a fixed lock
+  affordance and always remains first.
 - A Tag may belong to any number of resources and a resource may have any number of Tags. The same
   Tag filter is available in the three catalog panels and in the Specialist capability picker, but
   Specialist persistence continues to store concrete Skill and Connector IDs rather than a dynamic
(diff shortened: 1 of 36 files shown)


ewen-poch:
Addressed in d16c15e8. The packaged migration ledger now includes `0012_tag_ordering` with its manifest checksum, and the smoke test pins the new latest checksum. Verified with `npm test -- scripts/database-migration-ledger-smoke.test.ts` (10/10 passing), targeted ESLint, and `git diff --check`.

PR #1444, opened 2026-08-19
fix(acp-runtime): stream runtime events incrementally

ewen-poch:
## Problem

High-frequency ACP updates were published twice: once as an incremental acp:event and again through AcpStateSnapshot containing the retained window of up to 500 events. Notebook and tool-heavy turns could therefore repeatedly clone, serialize, and broadcast the same history across Desktop IPC, Web event streams, and Side Chat state callbacks.

The supplied application logs do not contain a native crash stack and the latest run reaches stopReason=end_turn, so they cannot prove the final operating-system exit cause. They do demonstrate the high-volume execution path that exercised this amplification.

## Proposed change

- Publish live ACP updates through the incremental event channel and reserve full state snapshots for lifecycle mutations, command responses, initialization, and reconciliation.
- Feed live renderer events directly into the Workspace processor outside React state, with ordered initial reconciliation, ID deduplication, and a shared 500-event retention window.
- Preserve snapshot-only adapters when no incremental callback is available.
- Cover Claude Code, OpenCode, Codex Responses, and Codex Bridge through their real runtime routes.
- Bump the Web event-stream protocol to v3 so cached v2 pages are rejected and shown the existing reload-required flow instead of silently losing streamed output.

## Scope and non-goals

- No database schema, persisted session format, data field, domain model, or migration changes.
- No new status or enum values.
- Historical conversations require no compatibility migration.
- Desktop Main and renderer ship together. Cached Web v2 clients must reload after a Main upgrade.
- Estimated context usage can refresh at provider usage or lifecycle snapshot boundaries instead of every text chunk.
- Separate follow-ups are intentionally excluded for Notebook full-state pulls after appendCodeCell, delegated-runtime update retention, and Side Chat persistence projection frequency.

All checks below ran after the final material edit.

- Main publication semantics, lifecycle state, four framework routes, Side Chat, Web projections, renderer contracts, event races, persistence drain, and 500-event retention
  - npx vitest run with the 18-file final Test Impact Set
  - Result: 18 files passed, 1068 tests passed
- Main and shared process types
  - npm run typecheck:node
  - Result: passed
- Renderer and Web types
  - npm run typecheck:web
  - Result: passed
- Source quality
  - npm run lint
  - Result: passed
- Patch hygiene
  - git diff --check
  - Result: passed

The complete npm test suite was not run locally by request. PR Gate remains authoritative for the complete portable and platform-selected test plan.

Independent review confirmed the final impact mapping and found no remaining P0 or P1 issue.

## Review focus

- Ordering between initial getState reconciliation and live acp:event delivery.
- Lifecycle processing against the snapshot that introduced an event.
- Event-window deduplication and bounded retention during synchronous bursts.
- Web protocol v3 behavior for cached v2 clients.
- Explicit lifecycle snapshots after removing per-event snapshot publication.

## Uncovered risks

- The application log does not prove the exact native process termination cause; Windows Event or Crashpad evidence is still needed to attribute the reported force close conclusively.
- The renderer tracks incremental event IDs until a confirming snapshot; an exceptionally long turn with no snapshot can grow this lightweight ID map.
- Notebook full-state refresh amplification is a separate path and remains for follow-up.

### DIFF
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -337,6 +337,61 @@ describe('AcpRuntimeCoordinator', () => {
     expect(incrementalPublicationConfigured).toBe(false)
   })
 
+  it('retains snapshot-only events after a new runtime generation adopts the session', async () => {
+    const retirement = createDeferred<void>()
+    const created: ReturnType<typeof createFakeRuntime>[] = []
+    const snapshots: AcpStateSnapshot[] = []
+    const coordinator = new AcpRuntimeCoordinator(
+      (callbacks) => {
+        const fake = createFakeRuntime({
+          frameworkId: created.length === 0 ? 'codex' : 'claude-code',
+          sessionIds: [`agent-session-${created.length + 1}`],
+          callbacks
+        })
+        created.push(fake)
+        return fake.runtime
+      },
+      { onStateChanged: (snapshot) => snapshots.push(snapshot) }
+    )
+
+    const session = await coordinator.createSession()
+    created[0].emitEvent({
+      id: 'old-owner-message',
+      timestamp: 1,
+      kind: 'message',
+      level: 'info',
+      sessionId: session.sessionId,
+      role: 'assistant',
+      text: 'published before adoption'
+    })
+    expect(snapshots.at(-1)?.events.map((event) => event.id)).toEqual([
+      expect.stringMatching(runtimeEventId(1, 'old-owner-message'))
+    ])
+
+    created[0].emitState({
+      promptInFlight: true,
+      promptInFlightSessionIds: [session.sessionId]
+    })
+    created[0].requestRetirement.mockReturnValue(retirement.promise)
+    const switchRequest = coordinator.requestAgentFrameworkSwitch()
+    await coordinator.connect()
+    const resumeRequest = coordinator.resumeSession({
+      sessionId: session.sessionId,
+      cwd: '/workspace',
+      previousFrameworkId: 'codex'
+    })
+    await vi.waitFor(() => expect(created[1].resumeSession).toHaveBeenCalledOnce())
+    created[0].emitState({ promptInFlight: false, promptInFlightSessionIds: [] })
+    await resumeRequest
+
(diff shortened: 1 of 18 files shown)


ewen-poch:
Addressed the P1 snapshot-fallback issue in 3f1f29386. The renderer now exposes subscribeRuntimeEvents only when the runtime event channel actually exists; older Main/preload combinations therefore continue through the existing snapshot-processing path. Added a regression test that removes acp.onEvent and verifies snapshot events remain available. The test failed before the fix and now passes; the focused Renderer suite is 40/40, with typecheck:web and lint also passing.

ewen-poch:
Addressed the initial-overlap P1 in 152bc37db. Initial reconciliation now performs a linear ordered merge: pending live events remain the ordering backbone, every snapshot-only event is inserted, and no snapshot suffix is truncated after the first overlap. Added a regression test for snapshot [1,2,3] with pending [2]; it failed before the fix and now delivers [1,2,3]. The existing 600-event pre-subscription burst test also remains green. Focused result: 45/45 tests, plus typecheck:web and lint passing.

ewen-poch:
I kept the conditional inner onEvent installation because the P2 premise does not hold for snapshot-only adapters. In that mode each runtime event triggers onStateChanged; AcpRuntimeCoordinator.getSnapshot then calls shouldPublishEvent and records the event ID before a later generation can adopt the session. Installing inner onEvent unconditionally would instead make AcpRuntimePublicationOwner select incremental mode and suppress the snapshot fallback even though no external event adapter exists. Commit 2640e577a adds a generation-handoff characterization that publishes an old-owner event through snapshots, adopts the session in a new runtime, and verifies the event remains retained afterward. The full coordinator module passes 86/86, with typecheck:node and lint passing.

ewen-poch:
Addressed the snapshot-only retention P2 in fb6b683a8. Retained history is now capped independently of subscriber presence. The only temporary exception is an actual pre-initialization direct-event queue that still needs one first-subscriber replay; after that replay the same 500-event cap is applied. Added a regression test with two rolling 500-event snapshots: it failed at 600 retained events before the fix and now retains events 101 through 600. Focused result: 46/46 tests, plus typecheck:web and lint passing.

PR #1447, opened 2026-08-19
feat(skills): redesign skill management

ewen-poch:
## Problem

The Skills panel mixed scope and Specialist filters, exposed too many row actions, and described access with verbose text instead of showing the agents that actually use each Skill. Bulk management also lacked safe deletion for user-installed Skills.

## Proposed change

- combine Main Agent and Specialists in one All Agents/Specialists filter and remove the redundant scope filter
- place filters and search on the first toolbar row, with Manage and Add skill right-aligned on a second row
- replace verbose scope copy with a compact actual-user avatar stack and a bounded Used by Agents and Specialists popover
- navigate named Specialist entries directly to their Settings detail while keeping Main Agent informational
- consolidate row actions into a menu and explain protected deletion through a focused tooltip
- add confirmed bulk deletion for Personal and Imported Skills while keeping built-in and Specialist-linked Skills protected
- update all four renderer locale catalogs and the Settings design contract

## Scope and non-goals

- renderer Settings behavior only; no database schema or migration changes
- no new persisted state and no new enum values
- no historical-data compatibility path is required
- existing Skill deletion persistence is reused; Featured built-in Skills remain non-deletable
- no full npm test run locally, per maintainer direction; PR CI remains authoritative

- Skills list, usage popover, detail navigation, bulk deletion, Settings navigation, scope helper, and i18n catalogs -> npx vitest run with the seven affected test files -> 7 files and 429 tests passed
- renderer TypeScript surface -> npm run typecheck:web -> passed
- linted source surface -> npm run lint -> passed
- changed-file formatting -> npx prettier --check on all 19 changed files -> passed
- patch integrity -> git diff --check -> passed

Consumer coverage includes SettingsPage navigation because the new Specialist link crosses the Skills panel boundary. Main-process, persistence, migration, and platform lanes are excluded because the change reuses existing renderer stores and deletion commands without changing their contracts. Remaining visual and platform risk is covered by PR CI and maintainer review.

## Review focus

- avatar-stack hover, focus, click, overflow, and external-scroll dismissal behavior
- navigation from a Skill usage entry into Specialist Settings and back
- protected versus deletable Skill classification in bulk management
- toolbar hierarchy and right alignment of Manage and Add skill


ewen-poch:
Addressed both review findings in b0f541ee:

- Preserved Connector detail behavior: ResourceAvailability now keeps the existing read-only Specialist list unless the caller explicitly opts into the Skill agent popover. Skill detail opts in; Connector detail does not.
- Serialized Skill exports: a synchronous in-flight guard prevents concurrent export calls, and every non-built-in Skill row action menu remains disabled until the native Save As operation settles.

Added regression coverage for Connector detail presentation and cross-row export locking. Verified with 417 focused tests, Web typecheck, ESLint, Prettier, and diff checks.

PR #1451, opened 2026-08-19
fix(settings): prevent stale edits from creating resources

ewen-poch:
## Problem

Provider and Connector edit targets can disappear after a live settings refresh while their drafts remain mounted. Saving could then silently issue create semantics.

## Proposed change

- Derive edit intent from stable Settings navigation IDs and fail closed when the live target is missing.
- Preserve the draft and edit labeling, disable save, and show a localized missing-target notice.
- Mark Provider edit writes as requiring an existing target and reject stale main-process upserts before persistence.
- Add renderer integration and main-process module regression coverage.

## Scope and non-goals

- No database, schema, or persisted settings document changes.
- No persisted enum, revision field, or historical migration.
- Ordinary create and explicit restore-by-ID behavior remain unchanged.

- Stale Provider and Connector drafts do not invoke create or update after target removal.
- Provider update-only writes reject a removed target without persisting a replacement.
- Targeted renderer tests: 372 passed.
- Settings Provider module tests: 343 passed.
- Passed typecheck:web, typecheck:node, lint, and the i18n resource guards.
- Full npm test was intentionally not run locally; PR CI is authoritative.

## Review focus

- Stable edit intent and draft preservation in SettingsPage and ConnectorAddForm.
- The transient requireExisting command flag remains non-persistent and preserves explicit restore behavior.


ewen-poch:
Addressed the P1 in cfc09a6ef.

`SettingsRepository.upsertProvider` now checks `requireExisting` inside the serialized document mutation, so a delete committed after the caller snapshot but before the final write cannot be followed by a stale recreation. `ProviderAccountsModule` applies that condition to the final Provider edit write while ordinary create and explicit restore-by-ID calls retain the existing upsert behavior.

Added a deterministic regression test that reads the stale Provider, queues deletion first, then queues the conditional upsert and verifies rejection plus an empty persisted Provider list.

Validation:
- targeted repository and ProviderAccounts tests: 104 passed
- `settings_provider_accounts` module plan: 343 passed
- repository architecture ceiling test: passed (repository remains at 700 lines)
- `typecheck:node` and `lint`: passed

ewen-poch:
Addressed both round-two findings in 723811efe.

- Claude subscription switches now pass the source edit ID into the repository's serialized mutation. The repository atomically verifies that the source still exists while upserting the destination's fixed identity, so shared-to-isolated and isolated-to-shared edits remain supported while deleted sources cannot be resurrected.
- Custom Connector updates now verify the edited ID inside the serialized connector mutation and throw when deletion wins the race. The existing store/form rejection path keeps the editor mounted, displays the error, and preserves the draft.
- Added bidirectional Claude require-existing coverage, a source/destination Provider deletion race, an atomic Connector deletion/update race, and SettingsPage coverage for retaining the Connector editor and draft after the update rejection.

Validation:
- settings_provider_accounts: 345 passed
- focused Repository/Connector files: 155 passed
- SettingsPage.render.test.tsx: 77 passed
- 	ypecheck:node, 	ypecheck:web, and lint passed

The broader settings_repository module run reached 18 passing files but has 7 unrelated Windows symlink tests blocked by local EPERM; the focused repository and connector coverage passed.

ewen-poch:
Addressed the migration/read race in 221e5c6f2.

The repository update remains strict by default, so user-initiated stale edits still reject and keep the form draft open. Legacy secret migration now explicitly opts into allow-missing behavior inside the same serialized mutation. If deletion wins after the migration snapshot was read, the migration write becomes a no-op instead of rejecting the entire Connector read.

Added a deterministic regression test that captures the legacy snapshot, commits deletion, then resumes getConnectors(). It failed before the fix with Unknown custom connector and now resolves while the persisted Connector remains deleted.

Validation:
- focused Repository/Connector files: 156 passed
- typecheck:node passed
- lint passed
- repository architecture ceiling passed at 700 lines

ewen-poch:
Addressed the Codex authentication side-effect race in a673b430a and synchronized the branch with current main in 4a868ed93.

Provider create/edit/delete operations now run through one recoverable FIFO at the ProviderAccounts boundary. The settings existence checks and app-owned authentication side effects therefore share one ordered operation interval:
- when an edit enters first, a concurrent deletion runs afterward and removes both the Provider and app-owned authentication;
- when deletion enters first, the later edit observes the missing Provider before authentication preparation;
- the repository conditional upsert remains as a final atomic defense.

Added a deterministic regression test that pauses a Codex shared-mode edit during authentication preparation, starts deletion, and verifies that app-owned auth.json is absent after both operations. Before the fix, deletion completed first and the stale edit restored auth.json.

The latest main branch changed the Connector edit entry to an action menu. The merged SettingsPage tests now use that public interaction while preserving the stale-edit assertions.

Validation after the merge:
- settings_provider_accounts: 346 passed
- focused Repository/Connector files: 156 passed
- SettingsPage plus i18n: 357 passed
- typecheck:node, typecheck:web, and lint passed
- repository and ProviderAccounts architecture ceilings passed

PR #1453, opened 2026-08-19
fix(runtime): recover caches with invalid package metadata

ewen-poch:
## Problem

Managed runtime setup can fail when micromamba leaves an extracted package with a valid `info/repodata_record.json` marker but a missing, empty, or malformed `info/index.json`. Micromamba reports a JSON parse error followed by `Found incorrect downloads`, but the current cache repair treats the package directory as complete because the marker file exists, so Settings surfaces the failure instead of self-healing.

This matches the extraction failure shape documented upstream in mamba-org/mamba#3351.

## Proposed change

- Treat an extracted package as complete only when both `info/repodata_record.json` and `info/index.json` contain JSON objects.
- Reuse the existing exclusive-cache repair, surgical package-directory removal, bundle re-seed, and single retry path.
- Add a regression test at `DefaultRuntimeProvisioner.provisionPython()` that reproduces the Settings installation failure with an empty package index.

## Scope and non-goals

- No renderer or Settings interaction changes.
- No new state, enum values, database fields, migrations, or persisted formats.
- No whole-cache deletion or additional retry loop.
- Existing invalid cache directories are rebuildable and self-heal on the next recognized extraction failure; package archives and complete sibling packages remain preserved.

The regression test failed before the fix with the reported empty-input parse error and passes after the fix.

- invalid extracted-package recovery -> `npm test -- src/main/notebook/micromamba-cache-recovery.test.ts src/main/notebook/provisioner.test.ts` -> 101 tests passed
- main-process types -> `npm run typecheck:node` -> passed
- linted source -> `npm run lint -- --no-cache` -> passed

Per maintainer direction, the complete local `npm test` suite was not run; PR Gate is authoritative for the full portable and platform coverage.

## Review focus

- Whether `repodata_record.json` plus `index.json` is the correct minimal completeness contract for an extracted package cache.
- Whether recovery remains surgical: only invalid extracted directories are removed while offline archives and valid cache entries are preserved.

### DIFF
--- a/src/main/notebook/micromamba-cache-recovery.test.ts
+++ b/src/main/notebook/micromamba-cache-recovery.test.ts
@@ -13,6 +13,7 @@ import {
 const makeRoot = (): string => mkdtempSync(join(tmpdir(), 'os-cache-recovery-'))
 const markComplete = (packageDir: string): void => {
   mkdirSync(join(packageDir, 'info'), { recursive: true })
+  writeFileSync(join(packageDir, 'info', 'index.json'), '{}')
   writeFileSync(
     join(packageDir, 'info', 'repodata_record.json'),
     JSON.stringify({ url: `https://host/channel/noarch/${basename(packageDir)}.conda` })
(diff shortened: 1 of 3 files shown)


ewen-poch:
CI note: PR Gate failed only in Full Module tests (macOS, shard 1/2) and the dependent coverage/summary jobs. The recovered Vitest report shows two deterministic failures in src/renderer/src/lib/acp/useWorkspaceAgentRuntime.architecture.test.ts: the workspace runtime facade is 636 lines versus the 600-line gate, and processRuntimeEvents occurs twice versus the expected once. Both come from the base-branch change in #1444; #1444 itself failed the same shard and summary gate before merge. This PR changes only the three notebook/micromamba recovery files. Static checks, Windows core, Windows E2E, macOS shard 2/2, macOS build/E2E, CodeQL, and automated review all passed. I did not rerun the full workflow because the unchanged base regression would reproduce deterministically.

PR #1459, opened 2026-08-19
fix(settings): stop retrying deterministic provider errors

ewen-poch:
## Problem

A deterministic provider failure could be sent upstream many times because ACP runtimes disagree about retryable HTTP errors. The reported native Responses request returned `401` six times with the same body. Local spikes also observed repeated `401` requests from Claude Code (8+ POSTs in 10 seconds) and OpenCode (2 POSTs).

## Proposed change

- Classify deterministic failures with an explicit HTTP status allowlist: `400`, `401`, `402`, `403`, `404`, `405`, `406`, `407`, `410`, `411`, `413`, `414`, `415`, `416`, `417`, `421`, `422`, `426`, `428`, `431`, and `451`.
- Return a terminal local `400` for those statuses while retaining the real upstream status in bridge diagnostics (structured JSON or `x-open-science-upstream-status`).
- Replay an identical deterministic failure from a bounded in-memory cache (SHA-256 request fingerprint, 30-second TTL, 32 entries) instead of sending the ACP retry upstream again.
- Include the final upstream-visible request headers in replay fingerprints, so a request with different provider feature/version headers cannot reuse an earlier deterministic error.
- Clear every per-target OpenCode bridge replay cache when the generation changes target, including A→B→A transitions within the replay TTL.
- Read error bodies through a bounded stream collector (256 KiB, 10-second completion timeout). Oversized or non-terminating bodies are cancelled and replaced with a small structured JSON error whose response headers match the fallback before replay.
- Normalize empty or malformed JSON error responses to the same bounded structured fallback before caching, while preserving the real upstream status.
- Keep retryable and unknown statuses out of the cache. `408`, `409`, `425`, `429`, 5xx, and unlisted statuses preserve upstream retry semantics.
- Route single custom Claude Code, OpenCode, and native Codex Responses providers through the existing loopback transports so the policy applies consistently, not only when multiple providers are configured.
- Project explicit bridge targets for single-provider model changes. If the active generation does not contain the new model target, the existing workflow reconnects instead of leaving the bridge pinned to the old model. Codex subscription authentication remains direct.

## ACP compatibility matrix

| ACP / wire path | Spike / regression evidence | Expected result |
| --- | --- | --- |
| Codex native Responses compatibility | Real Codex reproduced 6 identical upstream `401`s before the change; contracts cover replay, malformed/empty JSON, and changed forwarded headers | Deterministic retry is terminal locally and only equivalent requests replay |
| Codex native Responses direct | Single-provider routing uses the OpenAI loopback; contracts cover replay and changed forwarded headers | Same policy as compatibility mode |
| Codex Chat bridge | The bridge generates only content/auth headers and already clears replay state when its target, credential, or model changes; its unit contract preserves upstream `status: 401` in JSON | Same policy for Chat-to-Responses conversion without client-header aliasing |
| Claude Code / Anthropic | Spike reproduced 8+ `401` POSTs; contracts cover identical replay and changed Anthropic feature/version headers | No duplicate identical upstream request; changed upstream-visible headers bypass replay |
| OpenCode / OpenAI | Spike reproduced 2 identical `401` calls; contracts cover identical replay, changed forwarded headers, and A→B→A target transitions | No stale error after a header or target change |
| OpenCode / Anthropic | Uses the same OpenCode generation owner; Anthropic bridge contract covers identical replay and changed headers, while the owner clears every per-target cache on transitions | No stale error after a header or target change |

`429` is covered separately to ensure retryable rate-limit responses still reach upstream on each attempt.

## Scope and non-goals

- This does not modify retry algorithms inside third-party ACP runtimes.
- No renderer copy or i18n catalog changes.
- Oversized deterministic error bodies are normalized to a bounded JSON error; normal streaming responses are unchanged.

- **Historical data compatibility:** none required; no stored schema or existing data changes.
- **New enum values:** none. The explicit status allowlist is internal policy, not a persisted or public enum.
- **Runtime-only state:** one generation-local `activeTargetId` tracks real OpenCode target transitions so A→B→A clears every per-target replay cache. It is not public, serialized, or retained across generations.
- **Persistence:** none. Replay entries are process-local memory only and are cleared on target changes and bridge shutdown.

## Validation

- `npm run typecheck:node`
- ESLint for all changed TypeScript files
- Focused bridge/planner/resolver suite: 7 files, 205 tests passed
- `settings_backend_resolution` module files with PR Gate settings (`--maxWorkers=1 --testTimeout=30000`): 35 files passed, 1 conditional skip; 524 tests passed, 4 skipped
- Four architecture scans: 19 tests passed
- Model-change and architecture regression suite after review feedback: 7 files, 112 tests passed
- Bounded error body + affected bridge/service suite after review feedback: 6 files, 386 tests passed
- Settings architecture + module-impact contracts after the bounded-reader change: 11 files, 57 tests passed
- Final affected bridge/owner/resolver/service suite after replay-boundary and fallback-header fixes: 8 files, 457 tests passed
- Header fingerprint normalization contract covers case/order stability and value changes
- Live native Codex regression: one upstream `401` request
- Module-impact manifest contract: 28 tests passed

The full local `npm test` suite was intentionally not run. Updating the module-impact manifest is a global gate input, so PR CI will run the complete required test plan.

## Review focus

- Whether the explicit deterministic status allowlist is appropriately conservative.
- Whether every app-owned custom-provider transport is covered without changing subscription-backed direct routes.
- Whether the 30-second / 32-entry replay bounds are suitable for retry bursts.

### DIFF
--- a/scripts/ci/module-impact.json
+++ b/scripts/ci/module-impact.json
@@ -422,6 +422,7 @@
         "src/main/settings/native-responses-compatibility.ts",
         "src/main/settings/anthropic-provider-bridge.ts",
         "src/main/settings/openai-provider-bridge.ts",
+        "src/main/settings/provider-error-replay.ts",
         "src/main/settings/provider-loopback-http-host.ts",
         "src/main/settings/provider-transport-owner.ts",
         "src/main/settings/responses-bridge.ts",
@@ -459,6 +460,7 @@
           "src/main/settings/native-responses-compatibility.test.ts",
           "src/main/settings/anthropic-provider-bridge.test.ts",
           "src/main/settings/openai-provider-bridge.test.ts",
+          "src/main/settings/provider-error-replay.test.ts",
           "src/main/settings/system-proxy.test.ts",
           "src/main/settings/network-proxy-runtime.test.ts",
           "src/main/settings/environment-check-proxy.test.ts"
--- a/src/main/acp/model-change-workflow.test.ts
+++ b/src/main/acp/model-change-workflow.test.ts
@@ -56,6 +56,8 @@ type WorkflowHarness = {
   updateModel: ReturnType<typeof vi.fn>
   updateReasoningEffort: ReturnType<typeof vi.fn>
   updateAggregateModel: ReturnType<typeof vi.fn>
+  setProviderTransportTarget: ReturnType<typeof vi.fn>
+  setAnthropicBridgeTarget: ReturnType<typeof vi.fn>
   setBridgeReasoningEffort: ReturnType<typeof vi.fn>
   applyLiveEffort: ReturnType<typeof vi.fn>
   clearContext: ReturnType<typeof vi.fn>
@@ -147,6 +149,8 @@ const createHarness = (): WorkflowHarness => {
     updateReasoningEffort
   } as Pick<AcpBackendGenerationOwner, 'current' | 'updateModel' | 'updateReasoningEffort'>
 
+  const setProviderTransportTarget = vi.fn(() => true)
+  const setAnthropicBridgeTarget = vi.fn(() => true)
   const setBridgeReasoningEffort = vi.fn()
   const connectionResources = {
     epoch: 1,
@@ -163,8 +167,8 @@ const createHarness = (): WorkflowHarness => {
       return false
     },
     assertCurrentConnection: vi.fn(),
-    setProviderTransportTarget: vi.fn(() => true),
-    setAnthropicBridgeTarget: vi.fn(() => true),
+    setProviderTransportTarget,
+    setAnthropicBridgeTarget,
     setBridgeModelTarget: vi.fn(() => true),
     setBridgeReasoningEffort
   } as Pick<
@@ -218,6 +222,8 @@ const createHarness = (): WorkflowHarness => {
     updateModel,
     updateReasoningEffort,
     updateAggregateModel,
+    setProviderTransportTarget,
+    setAnthropicBridgeTarget,
     setBridgeReasoningEffort,
     applyLiveEffort,
     clearContext,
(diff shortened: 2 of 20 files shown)


ewen-poch:
Addressed the actionable review findings:\n\n- Single-provider model changes now project explicit Anthropic/OpenCode/native Responses transport targets. Unsupported live retargets fall back to the existing reconnect path, with regression coverage for both successful retargeting and missing-target reconnects.\n- Provider error bodies are now read through a bounded streaming collector (256 KiB, 10-second completion timeout). Oversized and non-terminating bodies are cancelled and replaced with a small structured error before replay.\n- Service-level contracts now assert that ACP processes receive only loopback URLs/tokens and never upstream credentials.\n\nValidation: affected bridge/service tests (386 passed), settings architecture and module-impact contracts (57 passed), full PR Gate, macOS/Windows E2E, CodeQL, and the latest Codex review all passed.

ewen-poch:
Addressed the three replay-boundary findings:

- OpenCode target transitions now invalidate every per-target bridge replay cache, so A→B→A cannot reuse A's pre-transition error. Re-selecting the same target does not clear the cache.
- Native Responses now normalizes empty or malformed JSON deterministic errors to the bounded structured fallback, caches that result, and preserves the real upstream status.
- Anthropic, OpenAI, and native Responses replay keys now include a canonical fingerprint of the final upstream-visible headers, so changed feature/version headers bypass an earlier cached error.

I also checked the Codex Chat bridge path. It does not forward client-defined provider headers; it constructs only content/auth headers and already clears replay state when its target, credential, or model changes, so no analogous change was needed there.

Validation: Node typecheck and changed-file ESLint passed; the final affected bridge/owner/resolver/service suite passed 455 tests across 8 files; settings architecture and module-impact contracts passed 57 tests across 11 files.

ewen-poch:
Addressed the fallback response metadata finding in both provider bridges:

- Anthropic and OpenAI bridges now replace the upstream content type with `application/json` whenever the bounded reader substitutes the structured fallback body.
- The fallback continues to remove `content-encoding` and replace `content-length`, so the first response and cached replay carry metadata that matches the emitted JSON bytes.
- Complete upstream error bodies retain their original content type.

Regression coverage asserts both the first response and replay for an oversized `text/event-stream`/`gzip` error. Validation: 16 tests passed across the two bridge modules; the full affected suite passed 457 tests across 8 files; Node typecheck and changed-file ESLint passed.

Issue #1442, opened 2026-08-19
[Feature]: Add Agent Context options to CLI project creation and runs


bobo-xxx:
### Preflight checklist

- [x] I searched existing issues and discussions and this hasn't been proposed yet.

### What problem does this solve?

Open Science exposes Agent Context in the project UI, but the CLI cannot configure it when creating a project or starting a run.

This limits headless and automated workflows that need persistent project instructions or task-specific context.

### Proposed solution

Add support for `--agent-context <text>` and `--agent-context-file <path>`.

Create a project with description and persistent Agent Context:

```bash
open-science project create "My project" \
  --description "Reproducible bioinformatics analysis workspace" \
  --agent-context "Use reproducible methods, preserve raw data, and cite sources."
```

Create a project using a context file:

```bash
open-science project create "My project" \
  --description "Reproducible bioinformatics analysis workspace" \
  --agent-context-file ./agent-context.md
```

Provide temporary context for one run:

```bash
open-science run \
  --project "My project" \
  --cwd ./workspace \
  --approval-profile auto \
  --agent-context "Use Python and explain each analysis step." \
  --prompt-file ./task.md \
  --wait \
  --json
```

Requirements:

- `--agent-context` and `--agent-context-file` are mutually exclusive.
- `project create` stores the value as persistent project Agent Context.
- `run` applies the value only to that run and does not modify the project.
- Run-specific context is appended after the saved project context.
- Apply context to newly created and resumed sessions.
- Validate UTF-8 files, empty values, and the existing 16,000-character limit.
- Return clear errors without starting a run when validation fails.
- Update the CLI parser, public client API, HTTP task API, task runner, help text, and `CLI.md`.
- Add tests for project creation, run forwarding, file loading, validation, and precedence.
- Do not expose context contents in logs, diagnostics, or machine-readable status output.

`--description` remains project metadata; `--agent-context` contains instructions sent to the agent.

### Alternatives considered

**Configure Agent Context only in the UI**

This does not support headless or automated workflows.

**Put all instructions in the prompt**

This mixes task instructions with project-level agent guidance.

**Use an environment variable**

This is less explicit and depends on how Open Science was launched.

### Additional context

The CLI already supports:

```bash
open-science project create <name> [--description <text>]
```

It currently forwards only `name` and `description`. The project data model and repository already support persistent `agentContext`; the missing support is in CLI parsing and task API propagation.

PR #1464, opened 2026-08-19
feat(cli): manage project agent context

ewen-poch:
## Problem

Project Agent Context is already supported by the Project model, persistence layer, and desktop UI, but headless users cannot set it through the CLI. The CLI also has no Project update command, so persistent context cannot be replaced or cleared after Project creation.

## Proposed change

- Add `--agent-context` and strict UTF-8 `--agent-context-file` inputs to `project create`.
- Add `project update <id-or-name>` for Project name, description, Agent Context replacement, and explicit `--clear-agent-context`.
- Carry create/update requests through the public SDK and HTTP Task API into the existing Project repository, preserving optimistic concurrency through `expectedUpdatedAt`.
- Project public responses now expose `hasAgentContext` while explicitly omitting Agent Context contents.
- Document validation, privacy, and attached-Session behavior.

## Scope and non-goals

This PR implements persistent Project Agent Context only. It does not add per-run temporary context or append/precedence behavior. Updating a Project does not rebuild an already attached provider Session; the new value applies when a new Session is created or a provider Session is set up again.

This is the persistent Project portion of #1442. The run-specific portion remains separate.

- Historical data: no migration is required. Existing databases already have the nullable-on-wire / `NOT NULL DEFAULT ''` Agent Context compatibility path, so old Projects continue to mean “no Agent Context”.
- State: no persisted lifecycle/status enum is added. The Task API adds only a `project_conflict` error code, mapped to HTTP 409 for stale optimistic-concurrency updates.
- Persistence: values continue to use the existing SQLite `Project.agentContext` column and repository update path; this PR adds no new store or serialized state.

- CLI parsing, file validation, create/update/clear dispatch, and output privacy -> `npm test -- packages/open-science/cli.test.ts packages/open-science/client.test.ts src/main/tasks/task-runner.test.ts src/main/web-service/task-api.test.ts` -> 4 files, 88 tests passed.
- HTTP create/update routing and public projection -> `npm test -- src/main/web-service/http-server.test.ts` -> 1 file, 20 tests passed.
- Main-process and shared TypeScript contracts -> `npm run typecheck:node` -> passed.
- Source lint and formatting -> `npm run lint` and `git diff --check` -> passed.
- Final affected-test explanation -> `npm run test:affected:explain -- --base origin/main --head HEAD` -> selected full fallback because `packages/open-science/CLI.md` has ambiguous ownership (`cli_sdk`, `documentation`). Per maintainer direction, the complete local `npm test` suite was not run; exact-head PR Gate CI is authoritative.

Uncovered local risk is limited to platform lanes and the classifier-selected full suite, both delegated to PR CI. The changed behavior is otherwise covered at CLI, SDK, Task runner, adapter, HTTP, type, and lint boundaries.

## Review focus

- Confirm the public Project projection cannot return `agentContext` contents.
- Confirm explicit clear versus omitted-value semantics and optimistic concurrency.
- Confirm the attached-Session non-rebuild behavior matches the existing desktop flow.

### DIFF
--- a/packages/open-science/CLI.md
+++ b/packages/open-science/CLI.md
@@ -111,12 +111,36 @@ the Debian package or a host configuration that supports Chromium sandboxing is 
 Create a project and list the projects available to task runs:
 
 ```bash
-open-science project create "Systematic review" --description "Evidence review workspace" --json
+open-science project create "Systematic review" \
+  --description "Evidence review workspace" \
+  --agent-context-file ./agent-context.md \
+  --json
 open-science project list --json
 ```
 
 Commands that accept `--project` allow either a project ID or an exact project name. The CLI resolves
 a unique display name to its ID before calling the Task API; use the ID when names are duplicated.
+
+Project Agent Context contains persistent instructions that are added when Open Science sets up an
+agent Session. Supply it directly with `--agent-context <text>` or read multiline instructions from a
+strict UTF-8 file with `--agent-context-file <path>`; the two options are mutually exclusive. The
+existing 16,000-character limit applies to both forms.
+
+Update Project metadata or replace its Agent Context later using an ID or exact name:
+
+```bash
+open-science project update "Systematic review" \
+  --description "Updated evidence review workspace" \
+  --agent-context-file ./revised-agent-context.md \
+  --json
+
+open-science project update <project-id> --clear-agent-context --json
+```
+
+`--clear-agent-context` is explicit so an omitted option keeps the existing value. An update affects
+newly created Sessions and provider Sessions that are set up again after the update. It does not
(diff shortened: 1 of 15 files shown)


ewen-poch:
Addressed both review findings in 6663b12d:

- Registered projects:update in the production Task command inventory and pinned the expanded inventory in its composition test.
- Constructed an allowlisted project update payload in TaskRunner, forcing the resolved URL project ID and excluding body-supplied id, pinned, or other unknown fields. The regression test now supplies redirect and pin fields and verifies that neither crosses the Task port.

Validation: application command composition, Task runner, Task API, and HTTP server module tests pass; Node typecheck and lint pass.

PR #1469, opened 2026-08-20
perf(startup): reuse runtime probes and trace Session recovery

ewen-poch:
## Problem

Startup currently performs the configured Claude, OpenCode, and Codex runtime probes repeatedly across the preflight and environment-check chain. The local runtime profiler also reports aggregate process CPU and memory trends but cannot attribute main-process CPU to individual Session recovery phases.

## Proposed change

- Add a local developer-only Electron resource profiler for startup, idle, workspace, ACP, Notebook, and recovery phases.
- Reuse a one-consumer, path-fingerprinted runtime probe snapshot across the startup inspection chain, with a two-second expiry.
- Preserve failed and partial probe results, including independent Codex adapter/native results, and memoize fallback version and smoke probes.
- Add opt-in Session hydration CPU deltas to the existing diagnostic operation boundaries.
- Restart after creating a populated Session so the recovery trace measures durable hydration rather than only an empty profile.
- Export only whitelisted scalar hydration metrics and group multiple hydration operations in the local Markdown report.

## Scope and non-goals

- The Session CPU trace is enabled only by the explicit local performance harness. Ordinary user launches do not add performance fields to main.log.
- Existing Session hydration log events receive numeric timing and CPU fields when profiling is enabled; no periodic resource samples are written to main.log.
- The profiler records no prompts, responses, command arguments, environment variables, paths, endpoints, credentials, or file contents.
- No ACP protocol, runtime lifecycle, renderer contract, or subscription behavior changes. Codex Response and Codex Bridge share the same Codex adapter/native probe boundary.
- This provides the instrumentation needed to investigate a 4–5 second real profile; the deterministic one-Session fixture is not intended to reproduce that dataset size.

- No application data migration or historical Session compatibility change.
- No existing application enum or state value is extended.
- Local performance report schema advances from version 1 to version 2 with a new sessionHydrationTrace array. External tools pinned to schema 1 must tolerate or adopt schema 2.
- Application business data persistence is unchanged. Generated performance reports remain local under ignored test-results/performance output.

All listed checks ran against the final material state:

- Runtime probe reuse and diagnostic report behavior -> npm test -- src/main/settings/agent-runtime-manager.test.ts src/main/diagnostics/operation.test.ts scripts/performance/runtime-resource-profiler.test.ts src/main/session-persistence/coordinator.architecture.test.ts -> 4 files, 47 tests passed.
- Settings and ACP consumers -> npm run test:module -- settings_service_facade -> 28 files, 730 tests passed.
- Session persistence consumers -> npm run test:module -- session_persistence -> 7 files, 313 tests passed.
- Profiler launch gating -> npm exec -- playwright test e2e/launch-environment.spec.ts --workers=1 -> 6 tests passed.
- Node and renderer types -> npm run typecheck -> passed.
- Repository lint -> npm run lint -> passed.
- Electron production bundle used by the performance scenario -> npm run build:e2e -> passed.
- Populated recovery trace -> npm run perf:runtime -- --repeat=1 --phase-ms=2000 --skip-build -> passed with 21 samples, 2 hydration operations, 16 trace events, and sessionCount 1 in operation 2.

The final Test Impact classifier selects full fallback because package.json and the new performance paths are global or previously unowned inputs. Per the requested workflow, the complete npm test suite was not run locally; PR CI is the final full-suite and platform verification.

## Review focus

- Probe cache invalidation, one-consumer semantics, and failed/partial Codex detection.
- The opt-in boundary between ordinary user logs and developer profiling.
- Session trace field whitelisting and local report schema versioning.
- The second restart and Operation grouping for populated Session recovery.

Independent Standards and Spec reviews were rerun after the final changes and reported no unresolved findings.


ewen-poch:
Addressed both review points in 8f4f32f2:

- Runtime probe reuse is now split into one-shot, consumer-specific handoffs: preflight results may only feed the following environment check, and resolved environment results may only feed the trailing preflight refresh. Independent `getPreflight()` calls probe again immediately. The regression test now removes a runtime between consecutive preflight calls without advancing the clock.
- Resource samples snapshot their phase when queued. Forced samples are serialized behind any in-flight sample, so every phase transition receives a fresh sample even when multiple transitions occur during an asynchronous collection. The regression test covers startup, recovery, and idle transitions while the first sample is blocked.

Validated with the two targeted test files (27 tests), Node typecheck, ESLint, Prettier, and `git diff --check`.

ewen-poch:
No code change is needed for the latest P2. The pinned Playwright `ElectronApplication` API explicitly supports the `console` event for JavaScript console calls in the Electron main process (`node_modules/playwright-core/types/types.d.ts`, ElectronApplication console event). This is also verified by an actual `npm run test:perf:runtime` run: the generated schema-v2 summary contained 16 Session hydration events across two operations (2 started, 12 phase, and 2 completed).

Moving this diagnostic path to IPC or file persistence would add unsupported complexity without fixing a defect, so the existing implementation is retained.

PR #1481, opened 2026-08-20
fix(updater): guide recovery from blocked installs

ewen-poch:
## Problem

When an update cannot stop background processes, the dialog only reports the low-level failure. Users do not know how to preserve diagnostics, recover by restarting the app, or report a recurring failure.

## Proposed change

- Recognize the two updater shutdown-error variants already emitted by the main process.
- Localize those errors and show focused recovery guidance: save the diagnostics log, quit and reopen Open Science, retry, and share the sanitized log with a developer or open a GitHub issue.
- Add translations for French, Simplified Chinese, Traditional Chinese, Japanese, and Korean.
- Add a renderer test for the guidance and GitHub issue link.

## Scope and non-goals

This changes only UpdateDialog copy and rendering for the known shutdown failures. It does not change updater shutdown behavior, process cleanup, retry state, persistence, data models, or generic update-error handling.

No related issue.

- blocked-update recovery guidance and issue link -> npx vitest run src/renderer/src/components/UpdateDialog.render.test.tsx -> passed
- locale coverage and interpolation parity -> npx vitest run src/renderer/src/i18n/resources.test.ts -> passed
- combined focused suite -> 366 tests passed
- renderer type safety -> npm run typecheck:web -> passed
- linted source conventions -> npm run lint -> passed
- changed-file formatting and whitespace -> Prettier check and git diff --check -> passed

The local affected-test selector failed closed to the full suite because the new render test has no module owner. Per the requested workflow, that full fallback was stopped; exact-head PR Gate CI is authoritative. The partial fallback output contained unrelated Windows ACL and symlink failures.

## Review focus

Please review the exact shutdown-error matching, the recovery sequence, and translations in all five locale catalogs.


ewen-poch:
Addressed in 8468fc4e7. The recovery guidance now tells users to use the existing Reveal action in Settings → General → Diagnostics to locate the log file. I updated all five locale catalogs and the render assertion. The focused renderer/i18n suite passes (366 tests), along with typecheck:web and lint.

PR #1486, opened 2026-08-20
fix(acp): publish writable connector files as artifacts

ewen-poch:
## Problem

When a Connector or MCP tool generated or returned a user-facing file, the Agent could treat the successful tool result as the end of the task. The shared Artifact instructions did not explicitly require a Connector/MCP file result to be published in the same turn, so the tool could succeed while the final message displayed `0 files` in Artifacts.

## Change

- Require Connector/MCP results that expose inline content or a local source path accepted by `write_artifact_file` to be published as an Artifact in the same turn before the Agent says the result is available.
- Preserve app-owned Connector flows that already attach an Artifact by recognizing their `artifact_id` receipt and forbidding a duplicate `write_artifact_file` call.
- Do not treat a custom MCP server's claim alone as proof that an Artifact exists.
- Keep the shared prompt provider-neutral and let the Claude Code, OpenCode, and Codex adapters render their framework-specific callable names.
- Add a runtime-boundary regression that verifies the Artifact MCP server and the final framework-transformed Session prompt rather than relying on model behavior.

## Scope and non-goals

- No automatic ingestion pipeline for arbitrary Connector output.
- No required Artifact write for remote URLs, resource links, or binary-only results that do not expose supported inline content or a local source path.
- No expansion of Artifact import roots or filesystem permissions. Paths outside the existing allowed roots continue to fail with the existing import-root error.
- No change to Connector Artifact-ID contracts or Connector producer Provenance modeling.
- No historical-data compatibility changes.
- No new status or enum values.
- No persistence or schema changes.

## Regression and validation

The runtime-boundary regression was added first and failed deterministically against the previous prompt because the actual Session instructions did not contain the Connector/MCP same-turn publication contract. It passes after the prompt change. The later unsupported-result and duplicate-publication assertions also failed against their preceding prompt revisions before passing with the final policy.

- Focused runtime regression: `npm test -- src/main/acp/runtime.test.ts -t 'publishes writable connector files without trusting unsupported or unverified results'`
- Shared prompt contract: `npm test -- src/main/acp/session-presentation-policy.test.ts` — 18 tests passed
- Framework adapter coverage: `npm test -- src/main/agent-framework/claude-code.test.ts src/main/agent-framework/opencode.test.ts src/main/agent-framework/codex.test.ts` — 120 tests passed
- Main-process typecheck: `npm run typecheck:node`
- Lint: `npm run lint`
- Formatting: Prettier check for the three changed files

PR CI passed at the current head, including Static checks, Module tests and coverage, macOS build and E2E, CodeQL, and AI PR Review. The final AI review reported `mergeable` with no actionable findings.

## Review focus

- Whether the prompt boundary is narrow enough to cover publishable Connector/MCP file results without forcing unsupported writes.
- Whether the app-owned `artifact_id` exemption prevents duplicate publication while the custom-MCP trust rule remains explicit.
- Whether the runtime regression sufficiently exercises the public Session construction boundary and framework-specific tool-name rendering.

## Remaining risk

Live model compliance remains probabilistic. This change makes the application-owned instruction contract explicit and regression-tested; it does not add an automatic Connector-output-to-Artifact pipeline.

### DIFF
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -19896,6 +19896,39 @@ describe('ACP runtime session management', () => {
     )
   })
 
+  it('requires connector-produced files to be published as Artifacts in the same turn', async () => {
+    const storageRoot = await createTemporaryRoot()
+    const process = new FakeAgentProcess()
+    const fakeAgent = startFakeAgent(process, ['remote-session-1'])
+    const runtime = new AcpRuntime({
+      appVersion: '0.1.0',
+      defaultCwd: '/workspace',
+      spawnAgent: () => asAgentProcess(process),
+      artifacts: {
+        configRoot: storageRoot,
+        dataRoot: storageRoot,
+        projectId: 'default-project',
+        mcpEntryPath: '/app/out/main/index.js',
+        repository: new ArtifactRepository(storageRoot)
+      }
+    })
+
+    await runtime.createSession({ cwd: '/workspace' })
+
+    expect(fakeAgent.newSessions[0].mcpServers).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({ name: 'open-science-artifacts' })
+      ])
+    )
+    expect(fakeAgent.newSessions[0]._meta).toMatchObject({
+      systemPrompt: {
+        append: expect.stringContaining(
+          'When a Connector or MCP tool creates or returns a user-facing file, call `mcp__open-science-artifacts__write_artifact_file` in the same turn before telling the user that the result is available.'
+        )
+      }
+    })
+  })
+
   it('retries transient Artifact claim preparation in the prompt failure cleanup', async () => {
     const storageRoot = await createTemporaryRoot()
     const repository = new ArtifactRepository(storageRoot)
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -19916,9 +19916,7 @@ describe('ACP runtime session management', () => {
     await runtime.createSession({ cwd: '/workspace' })
 
     expect(fakeAgent.newSessions[0].mcpServers).toEqual(
-      expect.arrayContaining([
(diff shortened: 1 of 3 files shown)


ewen-poch:
Addressed the valid duplicate-publication concern in `ddaf4897`:

- The shared rule now applies only when the Connector/MCP result has not already been saved or attached as an Artifact.
- It explicitly forbids calling `write_artifact_file` again when the tool result confirms an existing Artifact, preserving `molecule/preview_molecule` and similar one-call flows.
- The runtime-boundary regression now asserts both unpublished-file publication and the already-published exemption. The focused runtime test, shared policy test, Claude/OpenCode/Codex adapter tests, Node typecheck, lint, and Prettier check pass locally.

On the earlier framework-neutral-name comment: the production policy already keeps the provider-neutral backticked `write_artifact_file` reference. The Claude-specific callable appears only in the runtime assertion after `claudeCodeFramework` rewrites the shared prompt. Existing adapter coverage verifies the corresponding Claude, OpenCode, and Codex mappings, so no production change was needed for that point.

ewen-poch:
Addressed both current review points in `515fed05` without introducing a new ingestion or verification subsystem:

- Connector/MCP publication guidance now applies only when the result exposes inline content or a local source path accepted by `write_artifact_file`. Remote URLs, resource links, and binary-only results are no longer required to make an unsupported write call.
- The duplicate-write exemption is now limited to an Open Science app-owned Connector result carrying `artifact_id`, matching the existing `molecule/preview_molecule` contract. The prompt explicitly says that a custom MCP server's claim alone is not proof that an Artifact exists.
- The runtime-boundary regression pins all three behaviors after the Claude adapter rewrites the provider-neutral tool name. Its new assertions failed against the previous prompt and pass after this change. The shared policy tests, Claude/OpenCode/Codex adapter tests, Node typecheck, lint, and Prettier check also pass locally.

Main-process verification of arbitrary third-party Artifact claims would require a broader cross-component protocol and is intentionally not introduced by this prompt-only bug fix.

PR #1492, opened 2026-08-20
perf(settings): cache and lazy-load settings panels

ewen-poch:
## Problem

Switching Settings panels could repeatedly remount code and reload data. In particular, Storage recursively traversed the app data root at startup and again on panel entry, while Runtime discovery and package counts repeated interpreter probing. Catalog-backed panels also re-read their catalogs on remount even when authoritative events already kept renderer state current.

The first lazy-loading pass could also show two sequential states: a generic chunk-loading message followed by a panel-specific data-loading message.

## Proposed change

- Keep Model eager and lazy-load every other Settings panel and nested non-Model view.
- Load panel code and its first-view data behind one centered, reduced-motion-safe loading boundary. Data errors remain owned by each panel's retry UI; chunk failures are contained with Close/Reload recovery.
- Split lightweight Storage location/status from the expensive usage traversal. Startup reads status only; Storage scans usage on first panel entry, reuses it for one day, displays the scan time, and supports manual Refresh.
- Retain Runtime registry/package-count results for the app process, reuse discovery between environment listing and package counts, and expose manual Recheck. Settings-owned runtime mutations recheck; environment-manager/out-of-panel changes are observed on manual Recheck.
- Retain Skills, Connectors, and Specialists renderer projections across panel remounts while their existing catalog/runtime mutation events reconcile changes.
- Apply 60-second freshness to Permissions and Remote control, with event/mutation refreshes and stale-response fencing.
- Reuse the global Settings snapshot and defer OpenCode/Codex detection until Agent is visible.
- Preserve Compute's existing fresh read on genuine panel remounts while deduplicating only the lazy preload and first mount, including React StrictMode effect replay.

## Scope and non-goals

- Storage usage still traverses Open Science's `dataRoot`; it does not scan the whole filesystem. The traversal is now deferred and daily-cached in renderer memory.
- Runtime discovery still checks PATH, known install roots, pyenv/conda/mamba, and app-managed environments. `runtime/envs.lock` is relocation input, not a complete discovery snapshot, so no durable runtime cache was added.
- Skills has an existing recursive filesystem observer plus a fallback check; the renderer now keeps its parsed projection in memory. Connectors is a settings/runtime projection rather than a folder scan. Specialists is kept current by app create/update/delete/install catalog events; direct external edits to `specialists.json` remain outside the existing observation boundary.
- Cheap SQLite-backed Compute reads keep their previous remount freshness behavior.
- No full local `npm test` run; PR CI remains authoritative for the complete portable and platform matrix.

- Historical compatibility: `storage.getStatus` is additive. Mixed-version Web/Electron sessions fall back to `storage.getInfo` only when the backend explicitly reports the new command as unsupported. No historical data migration is required.
- Enum/state additions: no persisted or shared enum value was added. Renderer stores gained in-memory loaded/freshness/request state only.
- Persistence: no new data is persisted. Storage, Runtime, Permissions, Remote control, and catalog caches are process-memory snapshots and disappear on restart.

All commands below ran after the final material edit:

- Storage/Runtime main workflows and renderer contracts -> targeted Vitest command -> 8 files, 192 tests passed.
- Settings lazy loading, loading/error recovery, Storage, Runtime, Skills, Connectors, Compute, Permissions, Remote control, catalog stores, App startup, and i18n -> targeted Vitest command -> 18 files, 731 tests passed.
- Node process types -> `npm run typecheck:node` -> passed.
- Renderer/Web types -> `npm run typecheck:web` -> passed.
- Source lint -> `npm run lint -- --no-cache` -> passed.
- Generated renderer/Web command mapping -> `npm run check:web-api-map` -> passed.
- Whitespace/patch integrity -> `git diff origin/main...HEAD --check` -> passed.
- Two independent final reviews -> no remaining actionable spec or standards findings.

Uncovered locally: real-device filesystem traversal cost, platform-specific interpreter discovery, and live mixed-version remote Web sessions. PR Gate and platform CI are expected to cover the complete selected matrix.

## Review focus

- Storage's lightweight startup path versus deferred one-day usage scan.
- Runtime discovery snapshot invalidation and explicit Recheck semantics.
- Event-maintained catalog projections and 60-second freshness owners.
- Lazy boundary behavior: one first-entry loading state, retryable data failures, recoverable chunk failures, and StrictMode-safe Compute preload consumption.


ewen-poch:
Addressed the review points:

- RemoteControlPanel.preload typing: no code change. TypeScript infers the static property assigned in the same module; npm run typecheck:web passes locally, and the PR Static checks job also passed at head ffec3be9. The reported renderer build blocker is therefore not reproducible.
- Stale remote-access snapshot: fixed in 6d91df77. The existing remote-access changed event now keeps an app-lifetime cache-invalidation subscription, while panel-local listeners continue to refresh visible state. A regression test covers a pairing request arriving while the panel is closed and verifies that reopening fetches and displays it.

Validation: SettingsPage.render.test.tsx (83 tests), npm run typecheck:web, and targeted ESLint all pass.

ewen-poch:
Addressed both findings in 03a48a4e:

- Settings reopen freshness: reopening now refreshes getSettings so external/backend mutations are reflected. The subprocess-based preflight, npm, and encryption availability checks remain startup-only, so this does not reintroduce expensive probes during panel navigation or dialog reopen.
- Storage refresh failure: the last successful usage snapshot stays visible, an inline error is announced, and the header action changes from Refresh to Retry. The handled store rejection no longer becomes an unhandled UI promise.

Regression coverage verifies the lightweight-only settings refresh and the retained-storage-data error/retry flow. Validation: 209 focused tests, npm run typecheck:web, and targeted ESLint passed.

ewen-poch:
Made the RemoteControlPanel preload contract explicit in 11ff9da5. The component export now declares preload(): Promise<RemoteAccessSnapshot>, removing ambiguity for static inspection while preserving runtime behavior. npm run typecheck:web and targeted ESLint pass.

ewen-poch:
The automatic runtime-registry invalidation recommendation is intentionally not adopted. Settings is a display surface for the persisted backend registry, and the product requirement is to retain the process-lifetime discovery snapshot until the user explicitly selects Recheck. Settings-owned install/register/remove operations already perform a recheck after their persisted mutation; adding a new runtime change event or scanning on every panel open would expand the contract and reintroduce discovery cost.

To make the freshness boundary explicit, 41950078 adds an in-memory Last checked timestamp beside Recheck. It updates only after a successful registry discovery, remains unchanged after a failed recheck, and is not persisted. Five locale catalogs and regression coverage are included. Validation: 465 focused tests, npm run typecheck:web, and targeted ESLint passed.

PR #1500, opened 2026-08-20
perf(file-save): stream Project Artifact ZIP exports

ewen-poch:
## Problem

Project Artifact export accepted up to 2 GiB but read every source into `zipEntries`, then called `zipSync` in the Electron main process. A 128 MiB incompressible benchmark retained about 581 MiB RSS and blocked for about 2.74 seconds; larger exports could make the app unresponsive or OOM.

The regression test was run three times before the fix and failed deterministically because the handler called the injected whole-file `readFile` path.

## Proposed change

- preflight source paths and sizes without retaining file contents
- retain each resolver-canonical path plus its device/inode identity, and reject replacements before streaming
- stream each accepted source through `fflate` in 64 KiB chunks
- await every temporary-archive write and yield between compression slices
- build the ZIP in a mode-0600 temporary file, then copy it to the selected destination only after success
- abort the whole export and clean up the temporary archive when a source fails after streaming starts

## Scope and non-goals

- no IPC fields, database schema, data model, enum, or durable state changes
- no historical-data migration or compatibility handling
- ZIP entry names, category layout, limits, and normal Save As interaction remain unchanged
- the temporary ZIP is ephemeral filesystem data; a hard process/OS crash may leave an OS-temp directory for normal system cleanup
- preflight failures remain per-file failures; mid-stream read/limit failures intentionally abort the export and preserve any existing destination file

- Project Artifact streaming, identity pinning, and existing file-save behavior -> `npm test -- src/main/file-save.test.ts` -> 66 passed
- main-process types -> `npm run typecheck:node` -> passed
- changed source/test lint -> `npm run lint -- --no-warn-ignored src/main/file-save.ts src/main/file-save.test.ts` -> passed
- changed file formatting -> `npx prettier --check src/main/file-save.ts src/main/file-save.test.ts` -> passed
- patch whitespace -> `git diff --check` -> passed

A same-input 128 MiB isolated mechanism comparison measured about 67 MiB peak RSS and a 9.9 ms longest compression slice after streaming, versus about 581 MiB RSS and a 2.74-second synchronous block before it.

`npm run test:affected:explain -- --base main --head HEAD` selects the full suite because `src/main/file-save*.ts` has no module owner in the impact manifest. Per maintainer direction, the full local `npm test` was not run; exact-head PR Gate CI is authoritative for that coverage.

## Review focus

- canonical path plus device/inode pinning across the Save As dialog
- temporary archive cleanup and preservation of an existing destination on failure
- streaming limit enforcement when a source changes after preflight
- 64 KiB compression slices and explicit event-loop yielding

### DIFF
--- a/src/main/file-save.test.ts
+++ b/src/main/file-save.test.ts
@@ -664,7 +664,9 @@ describe('file save IPC handlers', () => {
     const bytes = Buffer.from('artifact bytes')
     const readFileMock = vi.fn().mockResolvedValue(bytes)
     const openProjectArtifactFile = vi.fn().mockImplementation(async () => ({
-      stat: vi.fn().mockResolvedValue({ isFile: () => true, size: bytes.byteLength }),
+      stat: vi
+        .fn()
+        .mockResolvedValue({ isFile: () => true, size: bytes.byteLength, dev: 1, ino: 1 }),
       readFile: readFileMock,
       createReadStream: vi.fn(() => Readable.from([bytes])),
       close: vi.fn().mockResolvedValue(undefined)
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the P2 review in 0064dbe50: preflight now retains the resolver-canonical path and device/inode identity, and the streaming reopen is rejected before ZIP entry creation if that identity changed. Added a regression test proving replacement bytes are never read and an existing destination remains untouched.

PR #1502, opened 2026-08-20
fix(notebook): reclaim unused conda package cache

ewen-poch:
## Problem

Notebook uses a persistent shared micromamba `pkgs` cache. Repeatedly installing or upgrading different package versions retained superseded extracted packages and archives indefinitely because normal mutations never ran package GC.

A real micromamba 2.8.1 reproduction upgraded zlib 1.2.13 to 1.3.1 in the same cache: the old archive and extracted directory remained and the cache grew by 494,006 bytes. Native cache cleanup confirmed that the retained extracted package was reclaimable.

## Proposed change

- Run micromamba's supported `clean --packages` before app-managed Conda mutations to reclaim unused extracted package directories.
- Preserve cached `.conda` / `.tar.bz2` archives because data-root relocation copies them for offline reconstruction.
- Reuse the existing install/materialize/upgrade operation journal hooks for the cleanup child; no new operation kind or persisted schema is introduced.
- Clear the cleanup child's sidecar and journal identity after it is confirmed settled, before any solver or environment mutation can start.
- Treat ordinary cleanup failures as best-effort, but propagate `CHILD_UNCONFIRMED` so recovery evidence is retained and subsequent cache writers are blocked. Thread the active provisioning `AbortSignal` into cleanup so user cancellation stops it promptly.
- Take the existing cache-exclusive locks while cleaning, then release them before the normal shared-lock mutation.
- Bind both `MAMBA_ROOT_PREFIX` and `CONDA_PKGS_DIRS` to app-owned paths so inherited host Conda settings cannot expand the cleanup scope.
- For bundled runtimes, fetch first and perform cleanup inside the existing materialize/upgrade journal before cache seeding and environment mutation.
- Skip cleanup before relocation restore because the copied cache is its offline reconstruction source.

## Scope and non-goals

- No quota, custom LRU, scheduler, UI, new status/enum, schema, operation kind, or data-model change.
- No migration is required. Existing extracted-package caches are reclaimed lazily on the next app-managed Conda mutation.
- Persistence effect: unused extracted package directories in the app-owned cache can be deleted. Package archives are retained for offline relocation, so this reduces growth but does not impose a hard cache-size bound.
- Reference-aware archive GC would require tracking every active environment and pending relocation lock; that is intentionally outside this conservative fix.
- No renderer, IPC, database, or user interaction changes.

The public `installPackages` regression test failed twice on the original code because the superseded cache entry still existed, then passed after package cleanup was added. Review regressions also failed on the corresponding pre-fix implementations for the reported reasons: `--tarballs` was present, cleanup spawn hooks were absent, `CHILD_UNCONFIRMED` was swallowed, provisioner cleanup ran outside the prefix-operation journal, and settled cleanup child evidence remained visible before the real mutation.

Current focused validation:

- cache cleanup ordering, ordinary best-effort failure, fail-closed unconfirmed-child handling, cleanup evidence settlement, cache locking, and package install/remove -> `npm test -- --run src/main/notebook/package-manager.test.ts src/main/notebook/package-mutation.test.ts src/main/notebook/provisioner.test.ts src/main/notebook/provisioner.upgrade.test.ts src/main/notebook/provisioner.startup.test.ts` -> 5 files, 209 tests passed
- default runtime materialize, named create, upgrade, production spawn-hook forwarding, and relocation-restore exclusion -> same focused command -> passed
- affected main-process TypeScript contracts -> `npm run typecheck:node` -> passed
- changed-file lint -> `npx eslint <changed Notebook files>` -> passed
- changed-file formatting -> `npx prettier --check <changed Notebook files>` -> passed

The complete local suite was intentionally not run per task scope; exact-head PR Gate and platform lanes remain authoritative.

## Review focus

- micromamba liveness semantics for `clean --packages`
- reuse and settlement of existing operation journal hooks around cleanup spawns
- ordinary best-effort cleanup versus fail-closed `CHILD_UNCONFIRMED`
- exclusive-to-shared cache lock sequencing
- preservation of tarballs and relocation restore's offline cache dependency

Uncovered local risk: tarball archives remain intentionally unbounded to preserve offline relocation. A hard size limit requires a separate reference-aware archive-GC design.

### DIFF
--- a/src/main/notebook/micromamba-cache-maintenance.ts
+++ b/src/main/notebook/micromamba-cache-maintenance.ts
@@ -1,5 +1,6 @@
 import { createLogger, diagnosticErrorFields } from '../logger'
 import { withExclusiveCacheLocks } from './pkgs-cache-lock'
+import { isChildUnconfirmedError } from './provisioner-runtime'
 
 const logger = createLogger('notebook:package-cache')
 
@@ -10,13 +11,13 @@ export const packageCacheCleanArgv = (micromamba: string): string[] => [
   '--no-rc',
   'clean',
   '--packages',
-  '--tarballs',
   '--yes'
 ]
 
-// Cache maintenance is opportunistic: a cleanup bug must not turn a package/environment mutation that
-// could otherwise succeed into a new hard failure. The existing exclusive cache locks still make the
-// deletion safe with respect to concurrent create/install operations.
+// Cache maintenance is opportunistic for ordinary failures: a cleanup bug must not turn a package or
+// environment mutation that could otherwise succeed into a new hard failure. CHILD_UNCONFIRMED is the
+// exception: the cleanup worker may still be deleting cache entries, so the journaled caller must retain
+// its recovery evidence and refuse to start another cache writer.
 export const maintainPackageCacheBestEffort = async (
   cacheLockKeys: string[],
   run: () => Promise<void>
@@ -24,6 +25,7 @@ export const maintainPackageCacheBestEffort = async (
   try {
     await withExclusiveCacheLocks(cacheLockKeys, run)
   } catch (error) {
+    if (isChildUnconfirmedError(error)) throw error
     logger.warn('package cache maintenance failed', diagnosticErrorFields(error))
   }
 }
--- a/src/main/notebook/micromamba-cache-maintenance.ts
+++ b/src/main/notebook/micromamba-cache-maintenance.ts
@@ -20,7 +20,8 @@ export const packageCacheCleanArgv = (micromamba: string): string[] => [
 // its recovery evidence and refuse to start another cache writer.
 export const maintainPackageCacheBestEffort = async (
   cacheLockKeys: string[],
-  run: () => Promise<void>
+  run: () => Promise<void>,
+  onSettled?: () => Promise<void> | void
 ): Promise<void> => {
   try {
     await withExclusiveCacheLocks(cacheLockKeys, run)
@@ -28,4 +29,5 @@ export const maintainPackageCacheBestEffort = async (
     if (isChildUnconfirmedError(error)) throw error
     logger.warn('package cache maintenance failed', diagnosticErrorFields(error))
   }
+  await onSettled?.()
 }
--- a/src/main/notebook/micromamba-cache-maintenance.ts
+++ b/src/main/notebook/micromamba-cache-maintenance.ts
@@ -1,0 +1,29 @@
+import { createLogger, diagnosticErrorFields } from '../logger'
+import { withExclusiveCacheLocks } from './pkgs-cache-lock'
+
+const logger = createLogger('notebook:package-cache')
+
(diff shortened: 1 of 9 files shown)


ewen-poch:
Addressed the cache-lifecycle and relocation concerns:

- Removed `--tarballs`; cleanup now uses `micromamba clean --packages --yes`, preserving the archives copied for offline data-root relocation.
- Forwarded the existing operation's `onBeforeSpawn` / `onChild` hooks to every cleanup subprocess in both package management and provisioning.
- Moved provisioner cleanup after bundle fetch and inside the existing materialize/upgrade journal, before cache seeding and environment mutation. This reuses the current durable recovery model without adding an operation kind or schema.
- Ordinary cleanup failures remain best-effort, while `RUNTIME_CHILD_UNCONFIRMED` is propagated so the journal and child sidecar are retained and no subsequent cache mutation starts.
- Added regressions for the exact cleanup argv, spawn-hook forwarding, journaled ordering, and fail-closed unconfirmed-child behavior. Focused result: 4 files, 197 tests passed; node typecheck, lint, and Prettier also pass.

The conservative trade-off is explicit: unused extracted package directories are reclaimed, while archives remain available for relocation and therefore are not subject to a hard size bound.

ewen-poch:
Addressed the stale cleanup-child evidence in `44f129a67`:

- Cache maintenance now has an explicit settlement callback that runs after success or an ordinary best-effort failure, before the solver or environment mutation continues.
- The package mutation owner and provisioner's existing journal wrapper remove the cleanup sidecar and clear `childPid`, `childStartedAt`, and `childStartToken` from the current operation record. The awaited clear is serialized behind the asynchronous PID update, so stale PID evidence cannot win the race.
- `CHILD_UNCONFIRMED` still propagates before settlement, retaining the sidecar/journal evidence and blocking the subsequent mutation.
- Added public-boundary regressions for package mutation and provisioner materialization. Both failed on the previous implementation with the cleanup PID still present, then passed after the fix.

Focused validation: 5 affected test files, 208 tests passed; node typecheck, changed-file ESLint, Prettier, and `git diff --check` also pass. This reuses the existing optional journal fields and sidecar lifecycle; it adds no operation kind, persisted schema, status, data model, or UI change.

ewen-poch:
Addressed cleanup cancellation in `3b31ea2a0`:

- Threaded the active default-runtime provisioning `AbortSignal` through `maintainCache` into the production `runMicromamba` adapter.
- After an aborted cleanup child is confirmed settled, cancellation is propagated immediately instead of being treated as an ordinary best-effort cleanup failure or proceeding to another cache/create attempt.
- Named-environment cleanup intentionally remains signal-independent, preserving the existing isolation from a concurrently running default provision.
- Added a public `cancel()` regression and extended the production adapter contract test. The regression timed out on the previous implementation because cleanup received no signal, then passed after the fix.

Focused validation: 5 affected test files, 209 tests passed; node typecheck, changed-file ESLint, Prettier, and `git diff --check` also pass. No status, persistence, schema, data-model, architecture, or UI changes were introduced.

Issue #1504, opened 2026-08-20
[Feature]: Support pre-registered OAuth client credentials (client_id/client_secret) for MCP connectors


justemu:
## What problem does this solve?

When connecting to an MCP server that uses OAuth authentication, Open Science currently relies entirely on **Dynamic Client Registration (RFC 7591)** or **URL-based Client IDs (SEP-991)**. There is no way to specify a pre-registered `client_id` and `client_secret` through the UI.

This is a problem in several real-world scenarios:

1. **Self-hosted MCP servers with confidential clients** — Many enterprise or self-hosted authorization servers (e.g., Keycloak, Auth0, custom OAuth providers) require pre-registered clients with explicit `client_id`/`client_secret` pairs. Dynamic registration is often disabled for security reasons.

2. **GBrain and similar knowledge-base MCP servers** — When a GBrain instance is configured with OAuth 2.1 authentication (e.g., `client_credentials` or `authorization_code` grant), the administrator pre-registers the client in the OAuth provider. The current UI offers no way to supply these credentials.

3. **Authorization servers that reject dynamic registration** — Some OAuth providers do not expose a `registration_endpoint` at all, making the SDK fallback to dynamic registration fail with "Incompatible auth server: does not support dynamic client registration".

4. **Security-conscious deployments** — Organizations that disable dynamic registration to prevent unauthorized client registration need to use pre-existing client credentials.

## Proposed solution

Add two optional fields to the MCP connector OAuth configuration:

**UI changes** (`ConnectorAddForm.tsx`):
- `Client ID` input field (optional, shown when authentication = OAuth)
- `Client Secret` input field (optional, shown when authentication = OAuth, with password masking)

**Data model changes** (`StoredCustomMcpOAuthConfig`):
```typescript
export type StoredCustomMcpOAuthConfig = {
  clientMetadataUrl?: string
  authorizationServerUrl?: string
  scopes?: string[]
  clientId?: string      // NEW
  clientSecret?: string  // NEW
}
```

**Runtime changes** (`PersistentOAuthClientProvider`):
- When `clientId` is provided in config, `clientInformation()` returns `{ client_id: clientId, client_secret: clientSecret, issuer }` directly, bypassing both URL-based client IDs and dynamic registration.
- When `client_secret` is present, `clientMetadata.token_endpoint_auth_method` should be set to `"client_secret_basic"` (HTTP Basic auth) instead of `"none"`, so the SDK sends credentials correctly during token exchange.

[...]

PR #1505, opened 2026-08-20
feat(connectors): support pre-registered OAuth clients

justemu:
## Problem

Custom MCP connectors can currently use Client ID Metadata Documents or Dynamic Client Registration, but users cannot configure a pre-registered OAuth client. This blocks authorization servers that disable registration endpoints or require an administrator-provisioned client. The submitted patch also persisted the secret as ordinary OAuth metadata, dropped it during sanitization, did not compile, and did not cover editing or portable connector templates.

## Proposed change

- Add optional pre-registered Client ID and write-only Client secret fields for the existing interactive `authorization_code` flow.
- Persist `oauth.clientId` as public metadata and persist the secret only as an encrypted `oauthClientSecretRef` through Electron `safeStorage`. The renderer receives only `hasClientSecret`; plaintext is decrypted only in main for transport construction.
- Return the pre-registered client information directly to the MCP SDK. The SDK selects `client_secret_basic`, `client_secret_post`, or `none` from the client information and authorization-server metadata.
- Require and pin the configured authorization-server issuer. Changing the issuer or client ID clears the bound secret unless a replacement is supplied, invalidates OAuth state, and disables the connector until sign-in succeeds again.
- Make editing explicit: blank keeps the saved secret, replacement overwrites it, and Remove saved client secret is a separate action.
- Add all five renderer locale translations.

### Export and import compatibility

- Connector templates remain schema v1. The public `oauth.client_id` and boolean `required_secrets.oauth_client_secret` marker are optional v1 additions.
- Existing schema-v1 connector templates continue to import unchanged, while pre-registered clients round-trip through the same schema version.
- Secret values are never exported. Import pre-fills public registration metadata and keeps submission disabled until a required secret is entered locally.

### Persistence change

The settings JSON shape gains two optional fields: public `customMcpServers[].oauth.clientId` and encrypted `customMcpServers[].oauthClientSecretRef`. There is no Prisma/database schema change and no migration. Existing records without either field keep the current CIMD/DCR behavior and are not rewritten. A downgrade may discard these unknown settings fields if an older release rewrites the connector record; no plaintext secret is introduced.

### State and enum change

No persisted enum value is added. `client_credentials` is intentionally outside this PR: it is a separate non-interactive grant requiring a persisted grant-mode discriminator, different token acquisition/refresh behavior, UI behavior, and compatibility coverage. This PR addresses #1504 for pre-registered `authorization_code` deployments.

## Scope and non-goals

- No `client_credentials` grant.
- No database migration.
- No plaintext secret exposure to renderer snapshots, settings JSON, logs, or exports.
- Existing Client ID Metadata Document and Dynamic Client Registration flows remain the fallback when no pre-registered client ID is configured.

All listed checks ran after the final material edit. Per maintainer direction, the complete local `npm test` suite was not run; PR Gate is authoritative for complete portable/platform coverage.

- Secure persistence, edit retention/removal, issuer-change invalidation, settings view redaction, and actual export path -> `rtk npx vitest run src/main/settings/connector-settings.test.ts src/main/settings/custom-mcp-settings.test.ts` -> passed.
- Static OAuth provider, runtime projection, and approval security fingerprint -> `rtk npx vitest run src/main/connectors/custom-mcp-bootstrap.test.ts src/main/connectors/service.test.ts` plus `rtk npx vitest run src/main/connectors/oauth-client.test.ts -t PersistentOAuthClientProvider` -> passed. The five loopback-listener tests were intentionally filtered because this sandbox cannot bind `127.0.0.1`; PR CI covers them.
- Template v1 export/import compatibility and secret exclusion -> `rtk npx vitest run src/main/settings/connector-template.test.ts` -> passed.
- Form submission/edit/remove/import interactions and all locale guards -> `rtk npx vitest run src/renderer/src/pages/settings/ConnectorAddForm.render.test.tsx src/renderer/src/i18n/resources.test.ts` -> passed.
- Latest schema-v1 compatibility subset -> 80 tests passed across Connector template, settings, and form import coverage.
- Combined focused set before the schema-only follow-up -> 501 tests passed; provider subset -> 6 passed, 5 filtered.
- Main and renderer contracts -> `rtk npm run typecheck:node`; `rtk npm run typecheck:web` -> passed.
- Lint -> `rtk npm run lint` -> passed.
- Exact-head impact explanation before the schema-only follow-up -> `rtk npm run test:affected:explain -- --base abe7f939e34671f82606499fdb15b27cf692eba2 --head fa33b223` -> full fallback only because `docs/design.md` has no module owner. The full local fallback was skipped per maintainer direction; PR Gate runs the authoritative plan.
- Independent Standards and Spec reviews found issues around issuer-bound secret retention, explicit-remove precedence, and template compatibility; all were fixed and re-verified.

## Review focus

Please focus on the safeStorage boundary, issuer-bound invalidation, additive schema-v1 export/import behavior, and blank/replace/remove edit semantics.


ewen-poch:
Thank you for opening this PR. I reviewed it against #1504, the current settings/security model, the MCP OAuth SDK, and the connector export/import path.

The current patch cannot work as submitted because the new credentials are removed by `sanitizeCustomMcpServer()` before persistence/runtime use. It also does not type-check (`issuer` is not part of the SDK client-information type, and renderer/template types do not expose the submitted fields), has no coverage or locale entries, and editing a connector would silently clear credentials. Persisting `client_secret` directly in the OAuth settings object would also put plaintext credentials in `settings.json`, which does not follow this project’s `safeStorage` reference convention.

I will update this same PR with the following scope:

- support pre-registered clients for the existing interactive `authorization_code` flow;
- keep `client_id` as non-secret connector metadata, but store `client_secret` through an encrypted `safeStorage` reference and never expose it back to the renderer;
- make edit semantics explicit: blank keeps the existing secret, while a separate action removes it;
- bind static credentials to the configured authorization-server issuer and invalidate OAuth state when security-sensitive registration data changes;
- cover connector export/import: export may carry the public client ID and issuer metadata, never the secret; imported connectors remain disabled until the secret is entered locally; existing exports without these fields continue to import unchanged;
- add focused settings, OAuth, UI, template/export-import, security-fingerprint, typecheck, and i18n coverage.

Persistence change (called out separately): the settings JSON schema gains an optional public `oauth.clientId` plus an optional encrypted-secret reference on the server record. There is no database migration. Existing records require no rewrite and retain the current DCR/CIMD flow when these fields are absent.

State/enumeration change (called out separately): this PR will not add a persisted grant-type enum. `client_credentials` is a distinct non-interactive OAuth flow and would require a persisted grant mode, different token acquisition/refresh behavior, UI behavior, and compatibility tests; it should be handled separately rather than partially implemented here.

I will also correct the PR title/commit format to match `CONTRIBUTING.md` and update the PR evidence after focused verification and CI.

justemu:
Hi @ewen-poch,

Thank you for the incredibly detailed review and for taking the time to explain the architectural issues so clearly.

You are absolutely right — my implementation had fundamental problems I did not fully grasp: the `sanitizeCustomMcpServer()` stripping credentials, plaintext secrets in settings.json, type system mismatches, and the silent credential clearing on edit. These are not small oversights; they reflect a gap in my understanding of the project's security model and the MCP OAuth SDK's contract.

I deeply appreciate you stepping in to implement this properly. The design you have laid out — safeStorage-encrypted secret references, issuer-bound credential invalidation, explicit edit semantics, and the export/import handling — is exactly the kind of thoughtful architecture this project deserves. It has been a privilege to see how a well-designed security model comes together.

open-science is a project I truly admire, and being able to contribute to it, even in this learning capacity, has been a genuine honor. I will follow the final implementation closely and look forward to contributing more thoughtfully in the future.

Thank you for your leadership and for maintaining such high standards. (◕‿◕)ノ

PR #1506, opened 2026-08-20
fix(tags): reject stale tag updates

ewen-poch:
## Problem

Two windows can edit the same Tag from the same snapshot. The update request previously carried all editable fields but no expected version, so the repository's read was followed by an unconditional update by id. A later save from a stale editor could silently overwrite another window's completed edit.

## Proposed change

- require the Tag snapshot's `expectedUpdatedAt` in the update command contract
- send the edited Tag's `updatedAt` from the renderer
- update with an atomic `id + updatedAt` predicate and reject a stale snapshot
- cover the lost-update scenario with a real migrated SQLite repository test

## Scope and non-goals

- no database columns, migrations, or persisted enum/state changes
- no automatic field merge or new conflict UI
- existing persisted Tags require no compatibility migration
- older independent callers of `tags:update` must add the now-required `expectedUpdatedAt`; the bundled main and renderer ship together

- stale Tag update is rejected and the winning update remains persisted -> `npm test -- src/main/tags/repository.test.ts src/main/tags/service.test.ts src/shared/tags.test.ts src/renderer/src/stores/tag-store.test.ts src/renderer/src/pages/settings/TagsPanel.render.test.tsx` -> 5 files passed, 42 tests passed
- main-process and shared contract types -> `npm run typecheck:node` -> passed
- renderer consumer types -> `npm run typecheck:web` -> passed
- linted source and tests -> `npm run lint` -> passed

The regression test was run before the fix and failed twice because the stale update resolved instead of rejecting. The complete portable and platform suites were intentionally left to PR CI.

## Review focus

- whether `updatedAt` is correctly treated as the Tag compare-and-set version
- whether requiring the field at the application-command boundary is appropriate for all supported callers
- whether the existing generic save failure is sufficient for this minimal data-integrity fix

### DIFF
--- a/src/main/tags/repository.test.ts
+++ b/src/main/tags/repository.test.ts
@@ -141,6 +141,23 @@ describe('TagRepository', () => {
     )
   })
 
+  it('rejects an out-of-range update timestamp with a domain error', async () => {
+    await repository.create({ name: 'Research', iconKey: 'tag', colorKey: 'blue' })
+    const tag = (await repository.snapshot(0)).tags.find(
+      (candidate) => 'name' in candidate && candidate.name === 'Research'
+    )!
+
+    await expect(
+      repository.update({
+        id: tag.id,
+        name: 'Updated research',
+        iconKey: 'book-open',
+        colorKey: 'green',
+        expectedUpdatedAt: 8_640_000_000_000_001
+      })
+    ).rejects.toThrow('Tag update timestamp is invalid.')
+  })
+
   it('stores many-to-many assignments and cascades custom Tag deletion', async () => {
     await repository.create({ name: 'Methods', iconKey: 'flask-conical', colorKey: 'green' })
     const tag = (await repository.snapshot(0)).tags.find(
--- a/src/main/tags/repository.test.ts
+++ b/src/main/tags/repository.test.ts
@@ -94,9 +94,46 @@ describe('TagRepository', () => {
         id: tag.id,
         name: expandingName,
         iconKey: 'tag',
-        colorKey: 'blue'
+        colorKey: 'blue',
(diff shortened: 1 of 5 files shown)


ewen-poch:
Addressed the P2 timestamp-CAS concern in 2d42633f0.

- `TagRepository.update()` now persists `updatedAt = max(now, expectedUpdatedAt + 1)` as part of the same conditional write, so an accepted update always advances the compare-and-set version even when the wall clock remains in the same millisecond.
- The SQLite-backed regression test injects a fixed repository clock equal to the stale snapshot timestamp. Before the change, the second stale update resolved and overwrote the winner; after the change, it is rejected and the winning fields plus the incremented timestamp remain persisted.
- Focused Tag tests pass (42/42), along with Node/Web typechecks and lint.

This keeps the existing schema and `updatedAt` request contract; no migration, new state, or interaction change is introduced.

ewen-poch:
Addressed the out-of-range timestamp concern in d3445e48c.

- The shared update contract now caps `expectedUpdatedAt` below JavaScript's maximum valid Date timestamp, leaving one millisecond available for the required monotonic version advance.
- `TagRepository.update()` defensively validates both the supplied expected timestamp and the computed next timestamp before constructing either `Date`, returning `Tag update timestamp is invalid.` instead of exposing a Prisma validation failure.
- Before the fix, the contract accepted `8_640_000_000_000_001` and the repository surfaced Prisma's `Invalid Date` error. Both regression tests now pass.
- The focused Tag suite passes (43/43), along with Node/Web typechecks and lint.

The change stays within the existing Tag contract and repository files and does not add schema, persistence, state, or interaction changes.

PR #1507, opened 2026-08-20
fix(task-api): honor idempotency keys for creates and runs

ewen-poch:
## Problem

`POST /api/v1/projects` and `POST /api/v1/runs` ignored `Idempotency-Key`. Retrying an accepted request therefore invoked `createProject` or `startRun` again. A duplicate Run start can allocate another Run, prompt message, and Session and can trigger duplicate model cost.

The regression test was run twice before the fix and deterministically reported business call counts of `[2, 2]` instead of `[1, 1]`.

## Proposed change

- retain Task API idempotency outcomes in the HTTP server for up to 24 hours
- scope keys by caller location, HTTP method, and endpoint
- coalesce in-flight retries and replay the first settled outcome for matching request bodies
- reject reuse of a key with a different body as `409 idempotency_conflict`
- hash request fingerprints with SHA-256 instead of retaining request bodies
- bound replay state to 1,024 entries and an estimated 64 MiB budget; preserve valid entries and reject new unique keys as `503 idempotency_unavailable` when full
- bound cached failure messages while replaying the same public failure
- expose optional `idempotencyKey` request options through the Node SDK

## Scope and non-goals

- no request-body fields or UI changes
- no database/schema, migration, or durable persistence changes
- no historical-data compatibility path is required
- no business state enum is added; `idempotency_conflict` and `idempotency_unavailable` are additive public error codes
- records are process-local, so a daemon restart ends the replay window

- sequential Project retry, concurrent Run retry, conflicting-body behavior, and capacity preservation -> `npm test -- src/main/web-service/http-server.test.ts packages/open-science/client.test.ts packages/open-science/cli.test.ts` -> 78 passed, 4 skipped
- SDK header propagation and CLI consumer compatibility -> same targeted command -> passed
- Node process type safety -> `npm run typecheck:node` -> passed
- linted source and tests -> `npm run lint` -> passed

Full portable and platform suites are left to the PR Gate as requested.

## Review focus

- process-local retention, expiry, entry-count, and memory-budget semantics
- request fingerprint and key scoping
- caching the first bounded failure as well as the first success to preserve at-most-once behavior

### DIFF
--- a/packages/open-science/README.md
+++ b/packages/open-science/README.md
@@ -28,6 +28,10 @@ final options argument to an individual request method. `downloadArtifact` keeps
 while its returned `Response` body is streaming. `waitForRun` applies its overall `timeoutMs` and
 caller signal to every in-flight polling request as well as the delay between polls.
 
+For retry-safe project creation and Run starts, pass the same `idempotencyKey` in the final options
+argument on every attempt. The daemon replays the first response for up to 24 hours while it remains
+running; reusing a key with a different request body fails with `idempotency_conflict`.
+
 The `project` request field and the `listSessions(projectId)` argument both require a Project ID.
 Project display names are not accepted as routing identifiers.
 
--- a/packages/open-science/README.md
+++ b/packages/open-science/README.md
@@ -30,7 +30,9 @@ caller signal to every in-flight polling request as well as the delay between po
 
 For retry-safe project creation and Run starts, pass the same `idempotencyKey` in the final options
 argument on every attempt. The daemon replays the first response for up to 24 hours while it remains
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed both P1 findings in 7e5621c25: valid and in-flight entries are no longer evicted; new unique keys receive 503 when entry or memory capacity is exhausted. Request fingerprints are now fixed-size SHA-256 digests, replay state has an estimated 64 MiB budget, and cached failure messages are bounded. Added a capacity regression test and reran the targeted tests, Node typecheck, and lint.

PR #1509, opened 2026-08-20
fix(updates): prevent duplicate renderer subscriptions

ewen-poch:
## Problem

React StrictMode replays the App startup effect in development. `useUpdateStore.init()` registered fresh status and progress listeners on every call but discarded both unsubscribe callbacks, leaving duplicate active listeners after the replay.

The regression test reproduced the report at the existing store/API boundary: three consecutive pre-fix runs failed with two active status listeners where one was expected.

## Proposed change

- replace any existing update subscription pair before `init()` subscribes again
- return an idempotent cleanup function from `init()` and wire it into the App effect lifecycle
- cover repeated initialization and final cleanup with a focused regression test

## Scope and non-goals

- No database, schema, persisted data, or historical-data compatibility change.
- No new state field or enum value.
- No data-model, architecture, or user-interaction change.
- The only internal contract adjustment is that the renderer store `init()` action now returns its subscription cleanup.
- Existing app-info and update-status hydration behavior is unchanged.

- duplicate subscription replacement and cleanup -> `npm test -- src/renderer/src/stores/update-store.test.ts src/renderer/src/App.test.tsx` -> 2 files passed, 61 tests passed
- renderer store/effect type contract -> `npm run typecheck:web` -> passed
- linted source/test conventions -> `npm run lint` -> passed

The complete `npm test` suite was intentionally not run; PR Gate CI is authoritative for the full portable and platform test plan.

## Review focus

- replacement and identity-safe cleanup of the active listener pair
- preservation of the existing live-broadcast-versus-startup-snapshot race guard

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -308,9 +308,7 @@ const AppContent = (): React.JSX.Element | null => {
   }, [appShellPresentation])
 
   // Load app info and subscribe to update-status broadcasts once at startup.
-  useEffect(() => {
-    initUpdates()
-  }, [initUpdates])
+  useEffect(() => initUpdates(), [initUpdates])
 
   useEffect(() => listenForPermissionChanges(), [listenForPermissionChanges])
   useEffect(() => listenForNotificationChanges(), [listenForNotificationChanges])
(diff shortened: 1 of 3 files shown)


ewen-poch:
Updated the regression test to assert `Set.size` explicitly for all active-subscription and cleanup counts. The focused update-store tests, Web typecheck, and lint pass on commit c4f042fd9.

PR #1510, opened 2026-08-20
refactor(settings): collapse navigation route state

ewen-poch:
## Summary

- replace the wide Settings history location product with a discriminated `SettingsRoute` that contains only the active panel state
- replace five competing external landing fields with one request-identified `pendingSettingsIntent`
- honor repeated external navigation requests to the same target while Settings remains open

## Reproduction and regression

On the unfixed code:

1. Open Settings directly on Skill `alpha`.
2. Navigate to Model without closing Settings.
3. Request the same external Skill `alpha` target again.

The second target was consumed, but Settings remained on Model because the target value was also used as the processed-request guard. A second regression test covers a landing intent received while Settings is temporarily hidden and ensures it remains pending until the route can be applied. Both public-boundary tests failed for their reported reasons before their fixes and pass after them.

## Scope and design

This is a renderer-local navigation refactor. Existing caller actions such as `openSettingsToSkill` and `openSettingsToComputeHost` remain unchanged. Breadcrumb rendering and panel interaction remain owned by `SettingsPage`; this does not introduce a router, URL state, or a generic navigation framework.

- Architecture: local history is `SettingsRoute[]`; external landing state is one intent with `{ requestId, route }`.
- Interaction: unchanged except that every external request, including an identical repeated target, is honored.
- Data fields: five optional in-memory landing fields collapse into one optional in-memory intent.
- New state/enum values: none; existing panel and view kinds are reused. `requestId` is ephemeral event identity, not user state.
- Data model/schema: unchanged.
- Persistence: none.
- Historical data compatibility: not applicable; no persisted data is read or written.

The limited refactor is justified by the confirmed repeated-intent defect and the current modification footprint for new drill-down entry points. Expanding this into URL routing or persisted navigation would be over-design and is intentionally out of scope.

## Validation

- `npm test -- src/renderer/src/pages/settings/SettingsPage.render.test.tsx src/renderer/src/pages/settings/SettingsPage.architecture.test.ts` (87 passed)
- `npm test -- src/renderer/src/stores/settings-navigation-slice.test.ts src/renderer/src/stores/settings-store.test.ts` (105 passed)
- `npm test -- src/renderer/src/stores/settings-store.architecture.test.ts` (27 passed)
- `npm test -- src/renderer/src/components/NetworkStatusIndicator.render.test.tsx src/renderer/src/pages/workspace/PermissionApprovalControls.specialist-switch.render.test.tsx` (15 passed)
- `npm run typecheck:web`
- targeted ESLint and Prettier checks for all changed files

The affected-test explainer conservatively selected the full suite because one changed renderer test has no module owner. Per project guidance, the complete suite is left to PR CI.

## Review focus

- route coverage for each drillable panel
- request-id consumption race safety (an older effect must not clear a newer intent)
- preservation of back/forward and breadcrumb behavior

### DIFF
--- a/src/renderer/src/components/NetworkStatusIndicator.render.test.tsx
+++ b/src/renderer/src/components/NetworkStatusIndicator.render.test.tsx
@@ -24,7 +24,7 @@ afterEach(() => {
   act(() => root.unmount())
   container.remove()
   useNetworkStore.setState({ isOnline: true, connectivity: 'unknown' })
-  useSettingsStore.setState({ isSettingsOpen: false, pendingSettingsPanel: undefined })
+  useSettingsStore.setState({ isSettingsOpen: false, pendingSettingsIntent: undefined })
 })
 
 describe('NetworkStatusIndicator', () => {
@@ -75,7 +75,10 @@ describe('NetworkStatusIndicator', () => {
     })
 
     expect(useSettingsStore.getState().isSettingsOpen).toBe(true)
-    expect(useSettingsStore.getState().pendingSettingsPanel).toBe('network')
+    expect(useSettingsStore.getState().pendingSettingsIntent?.route).toEqual({
+      panel: 'network',
+      view: { kind: 'list' }
+    })
   })
 
   it('disappears when connectivity recovers', async () => {
--- a/src/renderer/src/pages/settings/SettingsPage.render.test.tsx
+++ b/src/renderer/src/pages/settings/SettingsPage.render.test.tsx
@@ -2829,6 +2829,23 @@ describe('SettingsPage layout', () => {
     )
   })
 
+  it('keeps an external intent pending until Settings is visible', async () => {
+    useSettingsStore.getState().openSettingsToSkill('alpha')
+
+    await act(async () => {
+      root.render(<SettingsPage open={false} onClose={vi.fn()} />)
+      await Promise.resolve()
+    })
+
+    await act(async () => {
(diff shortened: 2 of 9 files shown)


ewen-poch:
Addressed the P1 in 622c4cef: the effect now consumes a landing intent only while Settings is visible. Added a public behavior regression covering an intent received while hidden; it failed before the fix and now passes together with the repeated-same-target regression.

ewen-poch:
The second review finding is not reproducible. The latest Static checks job passed, including Typecheck node and web, and npm run typecheck:web also passes locally. The grouped cases contain only drillable panels whose root route is the same valid list view; the default contains only static panels with no view. Splitting these into duplicated explicit returns would increase the exact new-panel modification footprint this refactor is reducing, so no code change is warranted.

ewen-poch:
Regression evidence captured before each fix: (1) the repeated same-skill test failed because the second intent was consumed while the Skill breadcrumb remained absent after navigating to Model; (2) the hidden-Settings test failed because rendering with open=false consumed the intent, so reopening stayed on the default route. The focused commands were npm test -- src/renderer/src/pages/settings/SettingsPage.render.test.tsx -t 'honors a repeated external intent for the same skill while Settings stays open' and npm test -- src/renderer/src/pages/settings/SettingsPage.render.test.tsx -t 'keeps an external intent pending until Settings is visible'. Both now pass, and the complete SettingsPage module run passes 87 tests.

PR #1521, opened 2026-08-21
fix(notebook): improve execution reliability

ewen-poch:
## Summary

- send 30-second MCP progress heartbeats while `notebook_execute` and `repl_execute` are still running, and stop them immediately after completion
- decode the Python loop stdin protocol as UTF-8 so Windows ANSI code pages cannot corrupt non-ASCII source before `ast.parse`
- give unbound Notebook input registrations a prompt-turn identity so later prompts with different attachments do not conflict with an earlier immutable registration

## Regression coverage

- real MCP client/server transport verifies progress is emitted for both long-running tools and stops after completion
- real Python loop execution reproduces the reported `UnicodeEncodeError` for Chinese source before the fix and succeeds afterward
- public prompt-preparation behavior reproduces the exact immutable-registration conflict across two unbound turns before the fix and succeeds afterward

## Verification

- targeted ACP/Notebook suite: 197 passed, 53 skipped
- Python loop integration suite: 5 passed, 1 skipped
- `npm run typecheck:node`
- targeted ESLint and Prettier checks
- `git diff --check`
- full `npm test`: 17,303 passed, 70 failed, 342 skipped; failures remain in unrelated Windows ACL/symlink, runtime projection, and resource-contention timeout tests

## Out of scope

Upstream HTTP 524 responses remain ordinary `server_error` results. This PR adds no report-issue handling for them.

### DIFF
--- a/resources/notebook/python_loop.py
+++ b/resources/notebook/python_loop.py
@@ -534,6 +534,10 @@ def _run(code):
 
 
 def main():
+    # The Node host always frames requests as UTF-8 JSON. On Windows, a piped stdin otherwise uses
+    # the active ANSI code page (for example GBK with surrogateescape), which corrupts non-ASCII
+    # source before it reaches ast.parse.
+    sys.stdin.reconfigure(encoding="utf-8", errors="strict")
     for line in sys.stdin:
         line = line.strip()
         if not line:
--- a/src/main/acp/prompt-preparation-owner.test.ts
+++ b/src/main/acp/prompt-preparation-owner.test.ts
@@ -146,6 +146,68 @@ const setup = (
 }
 
 describe('AcpPromptPreparationOwner', () => {
+  it('keeps unbound Notebook input registrations distinct across prompt turns', async () => {
+    const fixture = setup()
+    const registrations = new Map<string, string>()
+    fixture.registerTurnInputs.mockImplementation(async (input) => {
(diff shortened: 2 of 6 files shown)


ewen-poch:
Addressed the P1 in 8ead58081.

repl_execute / executeControl now resolves to fetchLongLivedLocalRpc, so a long-running control-plane kernel execution is no longer subject to Undici response-header timeout while the execution continues in the backend.

The transport-selection regression test failed consistently (3/3 attempts) before the fix with fetchLocalRpc selected for executeControl, then passed after the change. Verification:
- Notebook MCP suite: 68/68
- Local RPC transport integration suite: 6/6
- Node typecheck, focused ESLint/Prettier, and diff check: passed

ewen-poch:
Not adopting this P2. The Notebook HTTP fallback is intentionally stateless and JSON-only; we do not want to introduce an SSE response lifecycle for Notebook execution in this change. The progress heartbeat improves transports that support mid-call notifications, while HTTP fallback retains its existing request/response behavior. Changing that transport contract would broaden the scope and trade-offs beyond this reliability fix.

ewen-poch:
Updated the Python-loop regression fixture in 9230a9dbb to use an English comment. The en dash keeps the input non-ASCII, so the test still exercises the Windows stdin UTF-8 boundary. The focused integration test passes (1 passed, 5 skipped), and formatting/diff checks pass.

PR #1525, opened 2026-08-21
feat(providers): configure custom model token limits

ewen-poch:
## Problem

Custom providers need to represent three distinct model limits without conflating them: a shared input/output context window, an optional provider-reported input cap, and an optional output cap. The form also needs useful presets without preventing exact custom values.

OpenCode adds an adapter constraint: once a model declares `limit.context`, its schema also requires `limit.output`. Omitting the optional provider output cap therefore makes OpenCode exit during startup and surfaces as `ACP connection closed`.

## Proposed change

- Keep `contextWindow` as the primary field and add optional `maxInputTokens` and `maxOutputTokens` fields under Advanced settings.
- Use a project-styled editable combobox for all three fields: presets are selectable and exact positive whole-number values remain typeable.
- Use context/input presets of 32k, 64k, 128k, 200k, 256k, and 1m; use output presets of 4k, 8k, 16k, 32k, 64k, and 128k.
- Render locale-formatted, normal-weight options with WAI-ARIA combobox/listbox keyboard behavior.
- Persist and validate the three provider fields independently, with no sum constraint.
- Project explicit custom limits to OpenCode as `limit.context`, optional `limit.input`, and `limit.output`.
- When `maxOutputTokens` is absent but `contextWindow` is present, add an OpenCode-adapter-only 32k output reserve (clamped to the context window) because OpenCode rejects a context-only limit. This value is not persisted back to provider settings.
- Keep Codex and Claude context accounting/compaction based only on `contextWindow`.
- Let OpenCode own automatic compaction; retain native `/compact` without a second host-owned percentage threshold.
- Reuse the same conversion in Settings and onboarding, with all renderer translations.
- Add an injectable OpenCode process boundary so startup command, environment, and platform behavior can be tested without launching the real CLI.
- Make the fake OpenCode Agent reject invalid model-limit shapes before ACP initialization, reproducing the real startup failure boundary instead of only testing config objects in isolation.

- Historical compatibility: existing `contextWindow` values keep their meaning and serialized shape. The unshipped `inputLimit` field is not migrated.
- Persistence: only the optional provider fields in `settings.json` are added; there is no database schema or migration.
- State: no persisted enum values or runtime state variants are added. The combobox visual-state union is transient UI-only state.
- Official provider model metadata remains registry-owned and read-only.
- A custom provider without an explicit context override still resolves to the existing 200k runtime default.
- The 32k OpenCode reserve is adapter compatibility behavior, not a provider-model default and not a user-authored `maxOutputTokens` value.

## Validation

The complete local `npm test` suite was intentionally not run; PR Gate is authoritative for complete and platform coverage.

- Editable combobox, ProviderForm, and i18n: 616 targeted tests passed.
- Settings repository: `npm run test:module -- settings_repository` — 604 passed.
- Provider accounts: `npm run test:module -- settings_provider_accounts` — 353 passed.
- Backend resolution: `npm run test:module -- settings_backend_resolution` — 576 passed, 4 skipped.
- Settings facade: `npm run test:module -- settings_service_facade` — 753 passed.
- OpenCode adapter/startup seam: `npm test -- --run src/main/agent-framework/opencode.test.ts` — 61 passed, covering absent, shared-only, independent-input, independent-output, all-fields, clamped-small-context, oversized-output, and multi-model cases across both config layers.
- Settings-to-runtime materialization: service integration tests verify both `OPENCODE_CONFIG_CONTENT` and the written `opencode.json` for explicit limits and the adapter-only default reserve.
- Fake Agent startup schema guard: invalid context-only limits fail before ACP initialization; the provider-bridge Playwright certification passes through the real child-process boundary.
- Claude Code, Codex response/bridge, and framework registry compatibility: targeted tests — 72 passed.
- Node and renderer typechecks — passed.
- ESLint and target formatting checks — passed.
- Web production build — passed.
- Exact-head PR Gate run 32503080656 passed, including both complete module-test shards, affected-module coverage, static checks, macOS E2E, Windows core, and Windows E2E.

## Review focus

- The distinction between shared `contextWindow` and independent input/output caps.
- Editable preset behavior and keyboard/ARIA semantics.
- OpenCode's required adapter output reserve and native automatic-compaction ownership.
- `settings.json` compatibility and null-clearing semantics.


ewen-poch:
CI baseline note for fa816f7e:

- Passed: Static checks, Windows core, both Full Module test shards, Module tests and coverage, CodeQL, PR policy, and Codex review.
- macOS and Windows E2E each fail only at e2e/workspace-conversation.spec.ts:158: the Session actions menu locator is not found after opening the mobile sidebar actions.
- This is the same failure introduced by the current main baseline in merged PR #1530 (run 32461908034, Windows E2E), with the same file, line, locator, and retry result.
- All other platform E2E journeys passed: 14 of 15 workspace journeys on each platform; macOS functional, visual, and accessibility checks also passed.

No provider token-limit test, typecheck, lint, module, coverage, or runtime projection check failed.

PR #1527, opened 2026-08-21
fix(message-stream): avoid reserving space before tool calls

ewen-poch:
## Problem

When a short Agent text fragment streamed before a tool call, the message surface kept the 56px loading-row minimum height even while the Thinking row was still visible. This left a 36px blank area below the text and made the subsequent tool row jump upward when presentation finished.

## Proposed change

- Reserve loading-row height only when a buffered reply actually replaces the loading row.
- Keep the message at natural line height while Thinking or a live tool remains below it.
- Add an Electron regression that pauses between a one-line text fragment and a tool call, asserts there is no reserved gap, and measures the tool row for movement.

## Scope and non-goals

- Renderer layout and its deterministic OpenCode E2E fixture only.
- No changes to streaming cadence, provider event timing, persisted data, state enums, or historical-session compatibility.
- Follow-up to #1476; the completed-tool-to-final-reply reservation remains covered and unchanged.

- Visible streamed text with Thinking below has no reserved gap -> focused Electron regression -> failed 3/3 before the fix with a 36px gap; passed 3/3 after the fix.
- Running tool remains stationary while the one-line buffer settles -> npm run build:e2e; npm run test:e2e -- e2e/message-tool-layout-stability.spec.ts -> 3 tests passed, including both existing final-reply cases.
- Workspace owner behavior -> npm run test:module -- workspace_page -> 25 files, 459 tests passed.
- Direct message/scroller consumers -> npx vitest run WorkspaceMessageScroller.render.test.tsx WorkspaceMessageScroller.interaction.test.tsx WorkspaceMessageItem.actions.test.tsx WorkspaceMessageItem.mentions.test.tsx -> 4 files, 159 tests passed.
- Lint and formatting -> npm run lint and focused Prettier check -> passed.

The full npm test suite was intentionally not run locally; PR Gate is authoritative for complete portable and platform coverage. Windows-specific rendering remains covered only by CI.

## Review focus

- Whether reserveLoadingRowHeight matches the actual loading-row visibility boundary.
- Whether the E2E timing fixture represents the text-pause-tool sequence without overfitting provider behavior.

### DIFF
--- a/e2e/fixtures/fake-opencode.mjs
+++ b/e2e/fixtures/fake-opencode.mjs
@@ -34,6 +34,8 @@ const QUEUE_GATE_PROMPT = 'Hold the queue until the reveal finishes.'
 const TOOL_ORDER_PROMPT = 'Run the ordered slow tool journey.'
 const TOOL_LAYOUT_SHIFT_PROMPT = 'Run the tool layout stability journey.'
 const TOOL_STATUS_LAYOUT_SHIFT_PROMPT = 'Run the status-bearing layout stability journey.'
+const BUFFERED_TEXT_TOOL_LAYOUT_SHIFT_PROMPT =
+  'Run the buffered text tool layout stability journey.'
 const RELIABLE_FAIRNESS_USER_PROMPT = 'Run the concurrent real user prompt.'
 const DELEGATION_INHERITED_SPECIALIST_PROMPT =
   'Run the production inherited Specialist delegation journey.'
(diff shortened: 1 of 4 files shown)


ewen-poch:
Thanks for flagging the naming ambiguity. Completed activity rows intentionally count here because they remain rendered below the Agent message; in that arrangement the message is not replacing the trailing loading row. I tested the suggested active-status condition against the same Electron journey: the current row-based condition kept the tool at 256.84px initially and finally, while the active-only condition moved it from 292.84px to 256.84px after completion, recreating the reported 36px upward jump. I therefore kept the row-based behavior and renamed the variable to hasFollowingActivityRow with an explanatory comment in 2cb6e818. The rebuilt three-case layout spec passes.

PR #1528, opened 2026-08-21
fix(session-persistence): log deletion failure phases

ewen-poch:
## Problem

A user can confirm Session deletion in the renderer, but durable deletion can fail after the runtime detaches. The current main.log collapses every persistence participant into phase=delete-persistence with only errorCategory=error, so an external log cannot distinguish Session authority, Upload, provenance, file-index, or Compute cleanup failures.

## Proposed change

Instrument the existing SessionPersistenceDeletionOwner transaction with the shared privacy-safe diagnostic operation helper.

- Record stable primary failurePhase values for Compute preparation/commit, authority loading/deletion, Upload cleanup, provenance preparation/completion, and file-index soft deletion.
- Preserve the original failurePhase and add recoveryPhase only when provenance, file-index, or Compute compensation itself fails.
- Keep project/session identifiers, raw error messages, and filesystem paths out of diagnostic records.
- Keep deletion ordering, return values, rollback behavior, and renderer behavior unchanged.

## Scope and non-goals

This PR closes the observability gap only; it does not claim to fix the reported deletion failure before a new affected-user log identifies the failing participant.

- No database fields or migrations.
- No persisted enum/status additions.
- No persistence-format or data-model changes.
- No historical-data compatibility behavior.

The tests use the existing public SessionPersistenceCoordinator.deleteSession boundary. Before implementation, focused tests failed because no terminal deletion diagnostic existed, participant failures remained at the generic preparation phase, and recovery failures lacked recoveryPhase.

Final checks were run after the last material edit:

- Session deletion diagnostic phases and privacy -> npm run test:module -- session_persistence -> 7 files passed, 320 tests passed.
- Main-process TypeScript contracts -> npm run typecheck:node -> passed.
- Repository lint rules -> npm run lint -> passed.

The module plan includes coordinator owner/architecture tests, shared and coordinator contracts, and artifact/deletion integration consumers. It also selects the windows_sensitive capability overlay. No renderer code or public IPC interface changed, so web typecheck and UI lanes were excluded locally; PR CI remains authoritative for the complete selected plan.

## Review focus

- Are the phase names sufficient to distinguish the next affected-user report without encoding user data?
- Does recoveryPhase preserve the original failurePhase in every compensation failure path?
- Does instrumentation remain observational and leave deletion/rollback authority unchanged?

### DIFF
--- a/src/main/session-persistence/coordinator.architecture.test.ts
+++ b/src/main/session-persistence/coordinator.architecture.test.ts
@@ -570,6 +570,7 @@ describe('Session persistence coordinator architecture', () => {
         'assertArchiveMutable',
         'computeJobs',
         'fileIndex',
+        'log',
         'notifyFilesChanged',
         'notifySessionsDeleted',
         'provenance',
--- a/src/main/session-persistence/coordinator.test.ts
+++ b/src/main/session-persistence/coordinator.test.ts
@@ -5567,6 +5567,13 @@ describe('SessionPersistenceCoordinator', () => {
       },
       fileIndex: {
         restoreSession: vi.fn().mockRejectedValue(new Error('file index unavailable'))
+      },
+      computeJobs: {
+        prepareSessionJobDeletion: vi.fn(async () => undefined),
+        commitSessionJobDeletion: vi.fn(async () => undefined),
+        prepareProjectJobDeletion: vi.fn(async () => undefined),
+        commitProjectJobDeletion: vi.fn(async () => undefined),
+        abortSessionJobDeletion: vi.fn(async () => undefined)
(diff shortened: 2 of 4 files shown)


ewen-poch:
Addressed the recovery-phase reporting issue in 5040616dc. A provenance or file-index recovery failure is now retained while Compute rollback still runs; recoveryPhase is changed to abort-compute-cleanup only when the Compute rollback itself fails. Added regression coverage for a failed file-index restore followed by a successful Compute rollback. Verified with the session_persistence module suite (320 tests), npm run typecheck:node, and npm run lint.

PR #1532, opened 2026-08-21
fix(delegation): wake main agent after child settlement

wen2zhou:
## Problem

Delegated work could settle without reliably waking the owning main agent, leaving collect waits and same-turn continuation dependent on later unrelated activity.

## Proposed change

Add settlement-aware wake ownership and continuation dispatch across the production delegation composition and ACP runtime coordinator. Preserve settlement provenance, batch wakeups, enforce single-flight delivery, and invalidate stale watches while keeping cross-runtime consumers aligned.

## Scope and non-goals

This change covers delegated collect any/all behavior, same-turn continuation after child settlement, production composition, ACP coordination, host SDK contracts, and the Notebook bridge. Watch and lease state remains process-local and is not restored after an application restart; that is an intentional specification non-goal. No renderer UI or documentation changes are included.

- Delegated collect any/all and same-turn continuation -> focused contract, production, and ACP tests -> 713/713 passed.
- Settlement batching, single-flight delivery, invalidation, and provenance -> production composition, wake-owner, and runtime-coordinator focused tests -> passed.
- Cross-runtime consumers (ACP, host SDK, and Notebook bridge) -> Node and Web typechecks plus the i18n guard -> 351/351 i18n tests passed; typechecks passed.
- Full fallback -> npm run typecheck, npm run lint, and npm test -> passed; 1,144 test files passed and 15 skipped, with 17,536 tests passed and 218 skipped.

Independent Standards and Spec reviews each reported 0 findings.

Uncovered risk: process-local watch and lease state does not survive an application restart, as specified. Additional UI/platform E2E was not run because the change has no renderer UI and the full portable suite plus affected runtime typechecks cover the current impact surface.

## Review focus

Please focus on settlement ordering, wake coalescing and single-flight guarantees, stale-watch invalidation, provenance preservation, and continuation behavior across ACP, host SDK, and Notebook consumers.

### DIFF
--- a/resources/notebook/repl_loop.js
+++ b/resources/notebook/repl_loop.js
@@ -2989,7 +2989,8 @@ async function hostCollect(selectors, options = undefined) {
     options === undefined
       ? undefined
       : remappedHostObject(options, 'host.collect options', {
-          timeoutSeconds: 'timeout_seconds'
+          timeoutSeconds: 'timeout_seconds',
+          returnWhen: 'return_when'
         })
   return delegatedObservationRpc('collect', normalizedSelectors, normalizedOptions)
 }
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -12,6 +12,7 @@ import type { AcpRuntime, AcpRuntimeCallbacks } from './runtime'
 import type { ConversationPermissionGrantStore } from './permission-broker'
 import type { AgentModelChangeTarget } from '../agent-framework'
 import { DelegateMessageParkedError } from '../delegation/execution-port'
+import type { RootDelegatedWorkControl } from '../delegation/production-composition'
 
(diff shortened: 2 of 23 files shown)


ewen-poch:
This branch is failing PR Gate for two independent reasons:

1. **Static checks** — `e2e/message-tool-layout-stability.spec.ts` still passes a geometry object as Playwright’s expect message. That typecheck was fixed on `main` in #1530.
2. **macOS / Windows E2E** — the mobile session-actions journey looks up `getByRole('menu', { name: 'Session actions' })`, but Chromium names that popup from the trigger (`Open actions for …`). The menu is open; the locator is wrong. Fix is in #1535.

Because this PR is from a fork I did not push to it. Rebase onto current `main` (picks up #1530) and cherry-pick or rebase onto #1535 once it lands, then the remaining failures should be this branch’s own behavior.

PR #1540, opened 2026-08-21
fix(notebook): skip macOS developer-tools Python stub

ewen-poch:
## Problem

The first-run Environment check probes `python3 --version` to discover an optional external Notebook runtime. On a fresh macOS installation, PATH can resolve that command to `/usr/bin/python3`, an Apple developer-tool shim that opens the Command Line Tools installer instead of behaving like an installed Python runtime.

Environment detection should remain read-only and must not prompt users to install developer tooling that Open Science does not require.

## Proposed change

- Enumerate automatic macOS `python3` PATH matches and resolve each to its canonical executable path.
- Ignore `/usr/bin/python3` while continuing through later PATH entries.
- Version-probe Homebrew, pyenv, Conda, and other non-Apple Python candidates by absolute path.
- Add a regression fixture that proves the Apple shim is never executed.

## Scope and non-goals

- No changes to the app-managed Python environment or its provisioning flow.
- No changes to explicit user-selected interpreter validation.
- No renderer copy, interaction, i18n, Windows, or Linux behavior changes.

- Apple shim suppression and third-party Python discovery, including `/usr/bin` preceding Homebrew -> `npx vitest run src/main/notebook/python-command.test.ts src/main/settings/environment-check.test.ts src/main/notebook/runtime-adapters.test.ts` -> passed, 35 tests.
- Main-process TypeScript contract -> `npm run typecheck` -> passed.
- Source and test linting -> `npm run lint` -> passed.
- Complete portable regression suite, used because this helper is not owned by the module-impact manifest -> `npm test` -> passed.

Platform impact: the behavior change is guarded to macOS and the regression test simulates canonical resolution to `/usr/bin/python3`. Windows and Linux retain their existing candidate order and probing. A pristine macOS machine without Command Line Tools and a packaged DMG were not available for manual validation; macOS PR CI remains the final platform authority.

## Review focus

- Confirm automatic discovery should exclude the Apple developer-tool Python shim even when Command Line Tools are installed.
- Confirm canonical PATH resolution remains narrow and preserves third-party Python discovery.

### DIFF
--- a/src/main/notebook/environment-discovery.test.ts
+++ b/src/main/notebook/environment-discovery.test.ts
@@ -88,10 +88,29 @@ const makeDeps = (
   probeVersion: async (p) => opts.versions?.[p],
   rRunnable: async (p) => opts.rRunnable?.[p] ?? false,
   realpath: (p) => opts.realpath?.[p] ?? p,
-  runtimeRoot: '/rt'
+  runtimeRoot: '/rt',
+  platform: 'linux'
 })
 
 describe('discoverInterpreters', () => {
+  it('does not probe the macOS developer-tools Python stub and keeps other interpreters', async () => {
+    const probeVersion = vi.fn(async (path: string) =>
+      path === '/opt/homebrew/bin/python3' ? '3.12.4' : '3.9.6'
+    )
+    const deps = {
+      ...makeDeps(['/usr/bin/python3', '/opt/homebrew/bin/python3']),
+      platform: 'darwin' as const,
+      probeVersion
+    }
+
+    const found = await discoverInterpreters('python', deps)
+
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed in 32090d87. The macOS developer-tools stub policy is now shared by both first-run Python detection and the Settings environment-discovery path. discoverInterpreters filters the canonical /usr/bin/python3 path before probeVersion is called, while preserving later third-party candidates such as Homebrew Python. The regression test asserts that the system stub is never probed and that the Homebrew interpreter remains discoverable. Targeted coverage passes (72 tests), along with typecheck and lint; the three unrelated full-suite resource-contention failures also passed when rerun in isolation.

PR #1541, opened 2026-08-21
fix(compute): deduplicate retried job submissions

ewen-poch:
## Problem

`computeCall` handled every `submit_job` request as a new submission, even when a retry carried the same `invocation_id`. A lost response or caller retry could therefore create a second Job record, dispatch the remote workload again, and incur duplicate compute charges.

A public RPC regression test reproduced the issue before the fix: two identical requests with `invocation_id: invocation-1` returned `job-1` and `job-2`, and invoked `submitJob` twice.

## Proposed change

Keep a session-scoped in-memory map from `invocation_id` to the first submission Promise. Sequential and concurrent retries reuse the first result when the normalized request fingerprint matches. Reusing an invocation ID with a different provider, command, intent, or options is rejected with HTTP 409. Failed submissions are removed so a real retry remains possible, and entries are cleared when the Session is released or the RPC server closes.

Completed entries are limited to 100 per session and request fingerprints are stored as SHA-256 digests, so long-lived sessions do not retain unbounded command/options payloads. In-flight entries are never evicted by the capacity limit. Cleanup verifies map identity so a delayed failure from an old session cannot delete a same-ID successor session cache.

The standard JS caller generates a private ID per logical submission and retries one transport failure with the same request. Its retry boundary includes both request delivery and successful-response body parsing; HTTP and business errors remain non-retryable.

## Scope and non-goals

- No database schema, migration, persistent state, external data model, or UI changes.
- No persisted status or enum is added. The cache entry only gains a transient in-memory `completed` boolean for safe eviction.
- Calls without `invocation_id` retain their existing behavior.
- Cross-process/restart idempotency is not added.
- No historical data compatibility work is required.

All checks below ran after the final rebase onto `origin/main` (`e02b94add`).

- Retried `submit_job` reuses the first result, rejects conflicting payloads, bounds completed entries, retains in-flight entries, and protects successor caches from stale cleanup -> `npm test -- src/main/notebook/local-rpc-server.mcpcall.test.ts` -> 33 passed.
- Production `host.compute.submitJob` recovers request and response-body transport loss with the same invocation ID while leaving HTTP errors non-retryable -> `RUN_KERNEL=1 npm test -- src/main/notebook/host-compute.integration.test.ts` -> 7 passed.
- Packaged REPL script syntax -> `node --check resources/notebook/repl_loop.js` -> passed.
- Compute owner and contract behavior -> `npm run test:module -- compute_service` -> 724 passed, 25 skipped (27 files passed, 1 skipped).
- Source style -> `npm run lint` -> passed with no warnings.

The complete portable suite was intentionally left to PR CI per the requested validation strategy.

## Review focus

- Session and server lifecycle cleanup of cached submissions, including stale completion from a released predecessor.
- Failure eviction semantics: only successful/in-flight submissions are replayed.
- Completed-cache capacity and in-flight retention under pressure.
- The one-retry transport boundary and generated per-call invocation identity; durable database-backed idempotency remains out of scope.

### DIFF
--- a/resources/notebook/repl_loop.js
+++ b/resources/notebook/repl_loop.js
@@ -12,6 +12,7 @@ const vm = require('node:vm')
 const readline = require('node:readline')
 const fs = require('node:fs')
 const path = require('node:path')
+const { randomUUID } = require('node:crypto')
 const { fileURLToPath } = require('node:url')
 
 // Protocol output line. console is captured into strings during a run (see run()), so writing the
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the actionable review points:

- Reused invocation IDs now replay only matching requests; conflicting payloads return HTTP 409.
- The production `host.compute.submitJob` caller generates a stable ID per logical call and retries one request-delivery or successful-response-body transport failure with the same ID. HTTP and business errors are not retried.
- Completed cache entries are capped at 100 per session, fingerprints are stored as SHA-256 digests, and in-flight submissions are retained.
- Failed predecessor submissions only remove the session cache when the global map still references that exact cache, protecting same-ID successor sessions.

Regression coverage includes duplicate replay, payload conflicts, bounded completed entries, in-flight retention, lost request/response body recovery, non-retryable HTTP errors, and stale predecessor cleanup. The latest review reports the change as mergeable with no actionable findings.

PR #1543, opened 2026-08-21
fix(workspace): expose Specialist switch recovery

ewen-poch:
## Problem

When an idle Session changes Specialist, the renderer invokes specialist.setSessionSpecialist immediately. If that IPC rejects, the controller only logs a warning: the existing recovery banner remains hidden and the user has no explicit retry action.

## Proposed change

- Project idle-selection rejections into the existing Specialist reconfiguration error surface.
- Retain the failed target as renderer-only transient state.
- Route Retry back through the selection IPC without submitting the composer draft.
- Preserve the existing send-barrier Retry behavior and keep Main Agent recovery failures retryable.

## Scope and non-goals

- No database schema, persistent field, migration, historical-data compatibility, IPC contract, preload, ACP, or runtime-model changes.
- No new enum values or user-visible strings.
- No changes to the successful or explicit pending switch paths.

Checks listed below ran after the last material edit:

- Idle rejection feedback and selection-only retry -> npm test -- src/renderer/src/pages/workspace/workspace-session-controller.test.ts -> 12 passed.
- Recovery banner routing and preservation of send-barrier retry -> npm test -- src/renderer/src/pages/workspace/ConversationPanel.interaction.test.tsx -> 114 passed.
- Workspace owner, contract, representative consumer, renderer-state, and architecture gates -> npm run test:module -- workspace_page -> 25 files, 462 tests passed.
- Lint -> npm run lint -> passed with no warnings.
- Formatting and patch integrity -> Prettier check and git diff --check -> passed.

Uncovered local risk: no manual Electron IPC failure injection was run; the regression uses the existing renderer API boundary. Exact-head PR CI remains authoritative for the complete portable and platform-selected checks.

## Review focus

- Confirm Retry never submits the composer draft for an idle-selection rejection.
- Confirm the transient retry target cannot leak across Session switches or replacement selections.
- Confirm the existing pre-send reconfiguration recovery behavior remains unchanged.

### DIFF
--- a/src/renderer/src/pages/workspace/ConversationPanel.interaction.test.tsx
+++ b/src/renderer/src/pages/workspace/ConversationPanel.interaction.test.tsx
@@ -442,6 +442,7 @@ const createPanelDefaults = (): PanelProps => ({
     },
     actions: {
       selectSpecialist: vi.fn(),
+      retrySpecialistSelection: vi.fn(() => false),
       chooseOtherSpecialist: vi.fn(),
       useMainAgent: vi.fn()
     }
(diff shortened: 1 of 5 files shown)


ewen-poch:
Addressed the P2 in a791dd5 by making idle retry cleanup session-scoped. Changing another Session no longer clears the failed Session's transient retry target, so returning to the original error keeps Retry functional. This avoids adding a per-Session state map while preserving the existing interaction. Added a controller-level regression covering Session A failure, Session B change, return to A, and successful IPC retry; verified it failed before the fix and passed afterward. Validation: workspace_page module suite (25 files, 463 tests), web typecheck, ESLint, Prettier, and git diff check all pass. No schema, persistence, enum, or user-interaction changes.

ewen-poch:
Addressed the second P2 in 6394f81f4. Idle Specialist failures are now stored as transient state keyed by Session ID, including both the error shown to the user and its retry target. Independent IPC rejections in Sessions A and B therefore remain visible and retryable when navigating back to either Session. Added a controller-boundary regression for A rejection, B rejection, return to A and retry, then return to B and retry; it failed before the change because A's error became null and passes afterward. Validation: workspace_page module suite (25 files, 463 tests), web typecheck, ESLint, Prettier, architecture gate, and git diff check all pass. The state is renderer-only and non-persistent; no schema, enum, IPC contract, or additional file changes.

ewen-poch:
Addressed the lifecycle P2 in ecbf5695d. Successful Session archive and deletion paths now evict that Session's transient idle Specialist failure, while failed archive/deletion attempts retain the recovery state. Added controller-boundary regressions that create an idle IPC failure, remove the Session, reuse the same ID, and verify the old error is gone; both deletion and archive tests failed before the cleanup and pass afterward. Validation: workspace_page module suite (25 files, 465 tests), web typecheck, ESLint, Prettier, architecture gate, and git diff check all pass. No persistence format, schema, enum, IPC contract, or file-scope expansion.

ewen-poch:
Addressed the authoritative-handoff P2 in 4120f398c. After a continuation-start/continued lifecycle event resolves the Session to an authoritative bound Specialist or Main Agent, the controller now clears that Session's transient idle failure. This prevents the old banner and Retry action from reapplying an obsolete target. Added a controller-boundary regression that records an idle IPC rejection, applies an authoritative handoff to a different Specialist, and verifies the binding updates, the stale error disappears, Retry returns false, and no second switch IPC is sent; it failed before the fix and passes afterward. Validation: workspace_page module suite (25 files, 466 tests), web typecheck, ESLint, Prettier, architecture gate, and git diff check all pass. No persistence, schema, enum, IPC contract, or file-scope expansion.

ewen-poch:
Addressed the pending-switch P1 in f464d564a. Any authoritative pending-switch broadcast now invalidates that Session's older idle retry before applying the new pending target. Because invalidation happens at the broadcast boundary, it covers Main Agent, direct catalog matches, and resolver-backed targets without duplicating cleanup across branches. Added a controller-boundary regression that records an idle IPC rejection for Specialist B, broadcasts a newer pending Specialist C, and verifies the pending target remains C, the old error disappears, Retry returns false, and no second switch IPC is sent; it failed before the fix and passes afterward. Validation: workspace_page module suite (25 files, 467 tests), web typecheck, ESLint, Prettier, architecture gate, and git diff check all pass. No persistence, schema, enum, IPC contract, or file-scope expansion.

ewen-poch:
Addressed the late idle-switch response race in 88ab5ed7b. Each Session now owns a transient request generation; archive, delete, pending-switch broadcasts, authoritative handoffs, and newer attempts invalidate the previous generation, so late resolutions and rejections are ignored. The regression test reproduces an IPC rejection arriving after archival and verifies that restoring the same Session ID does not surface a stale error or Retry action. Validation: focused regression test, workspace_page (25 files / 468 tests), typecheck:web, ESLint, Prettier, diff check, and the 700-line architecture limit. This adds no persisted data, schema field, enum value, IPC contract, data-model change, or new file.

PR #1546, opened 2026-08-22
feat(provider-settings): simplify custom reasoning controls

ewen-poch:
## Problem

Custom Gateway setup exposed reasoning effort and its wire format as prominent choices, even when a model does not support configurable thinking. The long form also hid common token limits in dropdowns and gave little guidance for choosing among Chat Completions, Messages, Responses, or provider-specific reasoning payloads.

## Proposed change

- default new custom providers to thinking controls off and place image/thinking capabilities behind Advanced settings
- reveal effort levels only after Thinking mode is enabled, and reveal the request-format selector only for Chat Completions where Open Science owns that payload mapping
- keep effort mapping and automatic endpoint details in the Supported effort levels help tip instead of permanent form copy
- put common model profiles first: GPT/Claude, GLM-5.3/Kimi K3, GLM-5.2, DeepSeek, and MiniMax/MiMo/Kimi K2.5–K2.6
- note in the help tip that the examples follow common native model APIs while a gateway may map them differently
- show context, input, and output token presets as consistent inline shortcuts
- add a reveal/hide control for the API key currently entered in the form; stored keys remain masked and are never rendered as plaintext
- reset reveal state when the provider kind or secure draft record changes, so another provider never inherits plaintext visibility
- replace provider-specific placeholders with neutral examples, including `e.g. deepseek-v4-flash` for Model
- keep Messages and Responses reasoning transport automatic through their framework-native paths
- translate all new user-visible copy across the six renderer locales

## Scope and non-goals

- No request adapter or provider runtime behavior changes.
- No model-name inference or automatic gateway protocol detection.
- No new enum values, database fields, migrations, or persistence schema changes.
- The API-key visibility flag is transient React form state only and is not persisted.
- Existing saved providers keep their stored reasoning preset. Historical providers with reasoning enabled open Advanced settings automatically; no compatibility migration is required.
- A newly saved custom provider can persist the existing `unsupported` preset value when Thinking mode remains off.

- Common profile ordering and labels, help-only gateway caveat, scoped API-key visibility, form disclosure, and six-locale catalog integrity -> `npm test -- src/shared/reasoning-effort.test.ts src/renderer/src/pages/settings/ProviderForm.render.test.tsx src/renderer/src/i18n/resources.test.ts` -> 3 files and 632 tests passed
- Renderer type contracts -> `npm run typecheck:web` -> passed
- Repository lint -> `npm run lint` -> passed
- Changed-file formatting -> `npx prettier --check ...` -> passed

The latest change only reorders existing preset IDs and changes display copy; it does not change the persisted preset values or runtime request projection. Per the requested workflow, the full portable suite and UI E2E are left to PR CI.

## Review focus

- whether progressive disclosure removes unnecessary setup pressure without hiding required protocol choices
- whether the model-specific labels are useful without implying that a model name determines the gateway wire protocol
- whether the provider examples remain concise enough for the selector while the caveat stays in help-only copy
- whether retaining historical reasoning presets while defaulting only new custom providers to off is the right compatibility boundary


ewen-poch:
Addressed the API-key visibility finding in 613f2529. Reveal state is now scoped to the exact provider kind and draft key value, so switching provider kinds masks synchronously and replacing a same-kind provider record with an empty secure draft does not carry plaintext visibility forward. Added a regression test for both transitions; ProviderForm tests, web typecheck, lint, and formatting pass.

PR #1547, opened 2026-08-22
feat(onboarding): streamline first-run runtime readiness

ewen-poch:
## Problem

The five-step first-run flow installs runtimes before asking for a storage location. Changing storage at the end then requires a restart and can leave the newly installed runtime behind in the previous data root. It also asks low-confidence users to make storage and Notebook choices before they can use the app, while a lazy Python install can surface only after their first Notebook or Agent SDK request.

On Windows, the large-data default also stays under the home drive even when the user deliberately installs the app on another drive with more free space.

## Proposed change

- Reduce visible onboarding to Agent runtime and Model provider.
- Keep required host checks in the Agent step instead of a separate Environment page.
- Start managed Python preparation in the background for incomplete onboarding, show global progress/failure with Retry, and let successful provider validation finish onboarding without waiting for Python. Preserve that first-run preparation intent across an initial status failure, Retry, and the provider/onboarding race. The existing Notebook execution admission gate remains authoritative if Python is still pending.
- On a true first NSIS-installed launch, derive the default large-data parent from the executable drive while preserving the home-relative path. Resolve and persist it before database and runtime owners start, so no onboarding restart is required.
- Stamp NSIS installs with a non-user-data marker so portable ZIP launches do not receive installer-only behavior.
- Keep historical-user compatibility checks on the real former home default even when certification uses an isolated E2E storage root.
- Update the README and all localized README variants for the new flow.

## Scope and non-goals

- Later Storage changes still use the existing safe copy, switch, and restart flow.
- R and custom interpreter setup remain explicit Settings or Notebook actions.
- This does not move data into the application installation directory.
- This does not change portable ZIP behavior or non-Windows defaults.

### Historical compatibility

- An existing settings `dataRoot` always wins and is never recalculated.
- Existing installations without `dataRoot` but with settings or historical data are pinned to the former home-based default.
- Legacy config-root data remains unset so the existing migration prompt is preserved.
- Completed onboarding users do not receive the new automatic Python preparation.

### State and enum impact

- No enum values or persistent onboarding-step state were added.
- The wizard now uses only its existing in-memory Agent and Provider steps.

### Persistence impact

- Fresh Windows NSIS launches persist the resolved `dataRoot` to `settings.json` before storage-owning services start.
- Historical installs may receive a one-time `dataRoot` backfill to the former home default.
- The NSIS marker is an installation artifact containing only a constant sentinel; it stores no user data and is recreated by upgrades.

All listed successful checks ran after the last material edit.

- First-run root selection, historical backfill, legacy preservation, onboarding Retry/race behavior, Notebook admission, packaging marker, and i18n contracts -> `npx vitest run build/packaging.test.ts src/main/storage-root.test.ts src/renderer/src/App.test.tsx src/renderer/src/i18n/resources.test.ts src/renderer/src/pages/onboarding/startup-gate.test.ts src/renderer/src/pages/onboarding/OnboardingWizard.render.test.tsx src/renderer/src/pages/workspace/EnvStatusBanner.render.test.tsx src/renderer/src/pages/workspace/NotebookPreview.gate.render.test.tsx src/renderer/src/pages/workspace/provisioning-view.test.ts` -> **9 files / 756 tests passed**.
- Main and renderer TypeScript contracts -> `npm run typecheck` -> **passed**.
- Repository lint -> `npm run lint` -> **passed**.
- Windows packaging -> `npm run build:win` -> **passed**; final installer SHA-256 `4BB1D0CDA88585505048CF3B0707728B54B2ED0243262503B94142373DE9A6D5`.
- Secondary-drive packaged runtime -> D-drive `win-unpacked` headless bootstrap/resource/database-ledger certification -> **passed**.
- Real released-user upgrade -> elevated `node scripts/windows-installer-smoke.mjs --installer-dir dist --previous-installer-dir .previous-installer/v0.18.1` on Windows, with both install and isolated profile roots on D: -> **passed**. This installed the official v0.18.1 asset (SHA-256 `7F11457107225FBC46E9730B452AD61644AEC18A40DE8B38C81A0E8B732CAD04`), upgraded it to this v0.18.2 build while the old app held a real process lock, verified the upgrade sentinel and database compatibility, exercised rollback protection and restart, passed fresh/legacy/special-path/reopen migration certification, and completed silent uninstall with no app process, install directory, or uninstall registration left behind.
- Independent Standards and Spec reviews of the final change -> **passed with no actionable findings**.

The complete portable suite was also attempted locally as required by the full fallback, but is not green on this Windows checkout: the default run reported 57 failures and a controlled `--maxWorkers=4` rerun reported 47. Failures were concentrated in existing Windows filesystem/symlink/timeout and unrelated storage IPC/release-surface cases; the focused owner/consumer set above, typechecks, lint, package build, and real installer certification are green. Exact-head PR Gate remains authoritative for the complete portable and platform matrix.

## Review focus

- Startup ordering: `dataRoot` must be persisted before database, runtime, and application owners capture it.
- Historical detection must prefer the former home root and preserve the legacy migration prompt, including under E2E storage-root isolation.
- Provider success must not wait on Python, while Notebook execution must still wait when the environment is not ready.
- An initial environment-status failure and Retry must not lose the incomplete-onboarding Python preparation intent.


ewen-poch:
Addressed the P1 historical-data finding in 4560adea. Live relocatable data in the legacy config root now remains authoritative even when a home data folder also exists, unless settings contains an explicit dataRoot. Added a regression test and reran the final focused impact set: 13 files / 900 tests passed, plus Node/Web typechecks and focused ESLint.

ewen-poch:
Addressed both latest review findings in b2444e12. The derived data root is now rejected when it would be inside or equal to the executable installation directory, falling back to the home-based default, with ancestor/equal/sibling regression cases. Initial-root fixtures now use platform-native path construction. Final focused impact set after this change: 13 files / 902 tests passed, plus Node/Web typechecks and focused ESLint.

PR #1552, opened 2026-08-22
fix(session-model): scope composer choices to sessions

ewen-poch:
## Problem

Composer model and reasoning-effort changes were writing the global Settings selection and hot-applying it to live runtimes. That made one conversation's choice affect other Sessions. Historical or deleted providers also had no explicit lazy recovery rule, and a custom Provider configured through its singular `model` field could disappear from Composer when `models` was empty.

## Proposed change

- Persist provider, model, and reasoning effort as one optional per-Session `agentConfiguration` value.
- Treat Settings Active Model and effort as defaults for new Sessions only; an untouched new draft follows Settings until its first send, while an explicit Composer selection pins the draft.
- Validate a Session preference lazily against the current provider inventory and ACP framework. Keep a valid preference, replace an invalid one only with a valid Settings default, or preserve it and block sending with an Open Settings action when neither target is usable.
- Carry the explicit framework/provider/model/effort target through ACP create/resume and reuse runtime generations keyed by that target.
- Capture the Session configuration with queued messages and branches so later Settings or Composer changes cannot retarget admitted work.
- Include a custom Provider's singular `model` in the shared configured-model inventory when its `models` array is empty.

This follows ACP Session Config Options and the session-local model controls used by external agent clients: process settings establish defaults, while a conversation's explicit choice is authoritative for that conversation and is revalidated on resume.

## Scope and non-goals

- The ACP framework remains a global Settings choice and is evaluated when a Session is next used.
- Provider edits/authentication keep their existing reconnect lifecycle; selecting an Active Model or effort no longer mutates existing Sessions.
- The Side chat runtime remains isolated. Its old model picker was removed because it changed global Settings but could not retarget the already-running Side chat.
- No database or Prisma migration is introduced.

- Historical Session files without `agentConfiguration` are supported lazily. If their last backend/model is still selectable, it is materialized with the current Settings effort; otherwise the normal valid-default fallback applies.
- `agentConfiguration` is an additive optional field in the existing per-Session JSON file. Malformed values are dropped by the persistence sanitizer. There is no bulk rewrite.
- No enum value was added. The field reuses the existing `ReasoningEffort` values.

- Session preference persistence, lazy fallback, custom Provider inventory, Composer behavior, queue capture, ACP target routing, Settings defaults, and locale guards → focused Vitest set (13 files, 1159 tests) → passed.
- Main/preload/shared interfaces → `npm run typecheck:node` via `npm run typecheck` → passed.
- Renderer interfaces and UI consumers → `npm run typecheck:web` via `npm run typecheck` → passed.
- Source/style validation → `npm run lint` → passed.
- Patch hygiene → `git diff --check` → passed.

The full `npm test` suite was intentionally not run locally; PR Gate CI is authoritative for the complete portable and platform matrix.

## Review focus

- Lazy fallback preserves an unusable Session preference when the Settings default is also unusable.
- Runtime ownership cannot be stolen back by a late state event after a target-keyed adoption.
- Settings model/effort writes have no live-Session effect path.
- The additive Session JSON field and historical materialization policy are acceptable.


ewen-poch:
Addressed all three lifecycle findings in commit 8eb24ab5:\n\n- Provider invalidation is now provider-aware. Provider edits and authentication changes reconnect every live targeted runtime using the affected provider, while the default and Side chat runtimes reconnect only when the changed provider is active. Grouped Claude provider mutations invalidate both subscription variants.\n- Interrupted resume, context-overflow recovery, restored permission recovery, and readiness recovery now carry the Session framework/provider/model/effort target. Recovery also resolves history replay and image capabilities from the Session selection instead of the global default.\n- A targeted runtime is retired after its final Session transfers ownership or is deleted. Runtimes shared by other Sessions remain active.\n\nValidation: 339 focused tests passed, npm run typecheck passed, npm run lint passed, and git diff --check passed.

ewen-poch:
Addressed the actionable findings in a4096b09:

- Branch resets now adopt an explicit Session target before resetting or dispatching, including model/effort-only changes on the same backend.
- Restored and revised durable elicitations now resume with the Session target and its target-specific replay/image metadata.
- Detached parent Sessions resumed for delegated delivery now use the persisted Session configuration with the current ACP framework.
- Save as Skill now resolves preparation, replay, model attribution, and relay capability from the Session target instead of global Settings.
- Runtime ownership teardown now clears the strong targeted-runtime index.

Validation: 419 focused tests passed across the ACP coordinator, Session persistence, Workspace runtime, Save as Skill, and Workspace interaction modules; `npm run typecheck`, `npm run lint`, and `git diff --check` passed.

ewen-poch:
Addressed the detached deletion finding in c7ca92eb. The coordinator now clears Session ownership before evaluating targeted-runtime liveness, while preserving a concurrent owner adopted during the delete. A regression test covers deletion of the final detached Session without a runtime state event.

Validation: 92 coordinator tests passed, along with Node typecheck, focused lint, and `git diff --check`.

ewen-poch:
Addressed the immediate-send finding in d61095aa. Normal Composer dispatch now passes the resolved configuration for both new and existing Sessions. Runtime dispatch also falls back to the persisted Session configuration when an automatic caller supplies only `sessionId`, covering Job Analysis and similar continuations without replacing the Session target with global Settings.

Validation: 267 focused tests passed across Workspace runtime, immediate/queued conversation dispatch, Job Analysis, and send gating; `npm run typecheck`, `npm run lint`, and `git diff --check` passed.

ewen-poch:
Addressed the background-resume routing finding in f849d227. Session agent target resolution is now centralized against the current Settings framework, Task resumes forward the persisted per-Session configuration through the ACP adapter, and Reviewer activities carry the same target into lazy correction resumes. The coordinator now pins activity work to the target-specific runtime and forwards the target on resume; coverage exercises claude-code, opencode, and codex. Validation: reviewer_orchestrator module (486 tests), targeted Task/ACP routing tests (188 tests), target resolver test, Node typecheck, lint, and diff check all pass.

ewen-poch:
Addressed both Task routing findings in f5302ea5.

- New Task-created Sessions now capture the active explicit Settings target at creation time, pass it to ACP, and persist the corresponding existing `agentConfiguration` value. This prevents a stale coordinator runtime from supplying the previous provider/model/effort defaults.
- Existing Task Sessions with a persisted `agentConfiguration` now always pass through coordinator resume/adoption, even when already attached. The no-configuration legacy fast path remains unchanged.

This uses the existing persistence field and adds no schema, migration, enum, or interaction-state changes. Validation: 156 focused ACP/Task tests, Node typecheck, targeted lint, and diff check pass.

ewen-poch:
Addressed both image-routing findings in 3e0ff8fe.

- Runtime image capability now comes from the configured catalog entry for the Session-selected provider/model, so official providers with mixed vision and text-only catalogs no longer inherit the global active model's capability.
- Image-history context reset is now gated by a framework/backend/model transition or a detached runtime. An attached Session that resumes idempotently on the same text-only target no longer resets and replays its context on every prompt. Detached recovery and real target changes retain the existing filtering behavior.

Validation: 190 focused Workspace runtime tests, Web typecheck, targeted lint, and diff check pass. The `workspace_runtime` module run passed 324 behavior tests; its 4 architecture guards remain at the existing branch baseline (facade line budget, two established interface-count expectations, and an existing reverse import), none of which this commit changes.

ewen-poch:
Addressed both lifecycle findings in 62232743.

- Global Skills/connector reload now retires every live default and Session-targeted runtime generation. Framework-scoped reload retires all generations that own that framework while leaving unrelated generations attached. Existing active turns and workflow leases continue draining through the prior runtime before retirement completes.
- Framework selection changes now retire all live generations. Managed runtime uninstall without a framework fallback retires the default generation plus every targeted generation for the removed framework, preventing a Session from retaining the removed process or credentials.

The coordinator uses one batch-retirement path that marks all affected generations before publishing state and awaiting their retirement, so concurrent invalidations cannot reselect a stale generation. This adds no persistence field, schema migration, enum, or user-interaction state.

Validation: settings_service_facade module passed 754 tests; 177 focused coordinator/workflow/IPC tests passed; Node typecheck, targeted lint, and git diff --check passed.

ewen-poch:
Addressed both Session configuration findings in 19ff9d00.

- The Composer model picker now remains visible when the sole selectable model exposes reasoning-effort options. It still hides for a single effort-unsupported model, so this reuses the existing control without adding another UI surface.
- A persisted provider-default Session configuration with an omitted model now matches its Provider catalog and resolves to that Provider effective selectable model instead of being replaced by the active Provider. Existing configurations are resolved in memory with changed=false, so there is no automatic historical-data migration or persistence write.

No schema, enum, or new interaction state was added.

Validation: 25 direct picker/configuration tests passed; workspace_runtime passed all 324 behavior tests. Web typecheck, targeted lint, and git diff --check passed. The module retains the same four pre-existing branch architecture-guard failures previously reported; this commit does not change their measured facade, interface-count, or reverse-import baselines.

ewen-poch:
Addressed the failed targeted-runtime lifecycle finding in d4fc5daf.

Targeted Session creation and resume now invoke the existing unused-runtime retirement path when the provider operation rejects. A runtime with another committed Session owner is preserved by the existing ownership guard; an unowned runtime is marked retired immediately, so the target cache will provision a fresh generation on retry.

Regression coverage drives both failed create and failed targeted resume paths. The create test also retries the same target and verifies that a fresh runtime generation is created.

Validation: 101 coordinator tests, Node typecheck, targeted lint, and git diff --check passed. No persistence, schema, enum, architecture, or user-interaction changes were added.

ewen-poch:
Addressed the inactive Claude Provider regression in 9ebfd7f6.

The configured-model catalog now has an opt-in Session scope that includes every configured Claude subscription Provider. Global Settings selectors and new-Session defaults keep the existing collapsed single-Claude projection.

Existing Workspace Sessions enable the expanded scope for lazy configuration resolution, image/model runtime routing, and the Composer picker. A Session pinned to the inactive shared or isolated Claude Provider therefore remains on that Provider when the global Claude preference changes, and both configured Claude variants remain available in its model menu.

This is read-time compatibility only: no persisted field, migration, enum, or new interaction state was added. Existing Session configuration is no longer replaced with the active Provider.

Validation: 30 direct catalog/Composer/configuration tests passed; workspace_runtime passed all 324 behavior tests. Web typecheck, targeted lint, and git diff --check passed. The same four pre-existing architecture guards remain; the facade line count improved from the prior 697-line baseline to 687.

ewen-poch:
Addressed the explicit-model to provider-default transition finding in d3bd23f1. Prompt preparation now receives the already resolved effective agentModel and compares it with the Session stored model whenever a target configuration is selected, so an omitted configuration model no longer suppresses text-only image-history cleanup.

Added a regression test that failed before the fix because resetSessionContext was not called, then passed after the change.

Validation:
- useWorkspaceAgentRuntime.test.ts: 171/171 passed
- workspace_runtime behavior tests: 325/325 passed
- typecheck:web passed
- ESLint for the changed files passed
- git diff --check passed

The workspace_runtime module command still reports the same four pre-existing architecture gates (facade size, two interface-count baselines, and the save-as-skill reverse import); this change introduced no new failure category.

ewen-poch:
Addressed the approved historical Session compatibility finding in 1468c1c3.

Legacy Sessions that have agentBackendId and agentModel but no agentConfiguration now materialize their Provider/model identity with the current reasoning effort before background resume. The shared Main resolver receives the Session identity instead of configuration alone, so Task, Reviewer, and delegated-continuation paths keep the historical target instead of silently using the active default.

The existing agentConfiguration field is persisted lazily on first background use: Task reuses its normal Session save, while Reviewer and delegated continuation save the materialized value through the existing persistence coordinator. No field, schema, enum, migration script, or UI state was added. An unavailable historical target remains explicit and fails through the existing runtime/error path rather than falling back to the active Provider.

Validation:
- regression tests reproduced both the skipped Task resume and undefined Reviewer target before the fix, then passed
- direct Task adapter/runner and Reviewer IPC tests: 97/97 passed
- reviewer_orchestrator module: 487/487 passed
- settings_service_facade module: 755/755 passed
- typecheck:node passed
- targeted ESLint and git diff --check passed

ewen-poch:
The Composer model picker stays available during a running turn, queued messages, and compaction. That matches the Session-local model rule: Composer updates apply to the next admitted turn, and already-admitted work stays on the target captured at admission.

Queued messages already snapshot `agentConfiguration` at enqueue. Overflow recovery now continues with the agent target captured at send, so changing the picker mid-turn cannot retarget the recovering turn.

The picker is not wired to `canChangeAgentControls`. That lock is for in-turn Specialist switching; applying it here would block choosing the next turn's model.

ewen-poch:
Addressed the remaining Session-target findings:

- Legacy Codex backend ids such as `builtin-codex-isolated` now resolve to the current `builtin-codex-subscription` provider in both Main materialization and renderer selection. The stored Codex subscription remains a single runtime identity.
- `agentConfiguration` is included in the renderer conflict-rebase allowlist, so a concurrent Main Session save keeps the selected provider/model/effort instead of dropping it.
- Admitted overflow targets are pruned when the Session is deleted or cancelled, so the in-memory map cannot retain stale conversations.

The Composer picker remains available during a running turn.

ewen-poch:
xAI login and logout now pass `XAI_SUBSCRIPTION_PROVIDER_ID` into provider-aware reconnect, matching Codex/Claude. Sessions pinned to xAI reconnect even when another provider is active; the default generation still reconnects only when xAI is the active provider.

ewen-poch:
Delegated parent delivery now calls resume with the Session agent target even when that Session is already live. If the attached generation does not match the persisted selection, ownership moves before the continuation is sent. The Composer picker is still left unlocked during a running turn, because that remains a next-turn preference rather than an in-turn control.

ewen-poch:
The Composer model picker is still not wired to `canChangeAgentControls`.

That lock is for in-turn Specialist switching. Session model and effort remain a next-turn preference: the picker stays available during a running turn or queued messages, and already-admitted work keeps the target captured at admission (queued items snapshot `agentConfiguration`; overflow recovery uses the admitted target).

Locking the picker would block choosing the next turn's model, which conflicts with the approved Session-local rule. Not adopting that change.

ewen-poch:
Queued and overflow turns still use the admitted `agentConfiguration` snapshot for that send. Existing Sessions no longer have that snapshot written back over a newer Composer selection.

The model picker stays available during a running turn. `canChangeAgentControls` remains the in-turn Specialist lock, not a next-turn model lock.

ewen-poch:
Foreground recovery and send fallback now reuse `resolveSessionAgentConfiguration` when a Session has no `agentConfiguration`. Historical `agentBackendId`/`agentModel` identity is used for the runtime target and image-capability path instead of the active global provider.

The Composer picker remains a next-turn preference and is still not wired to `canChangeAgentControls`.

ewen-poch:
Rebased onto current `main` so the history is linear. The merge commit that failed Conventional Commits is gone; PR policy can check the original `fix(session-model): …` subjects.

The catalog test conflict kept both cases: xAI subscription models remain image-capable, and a custom Provider with an empty `models` array still fills its singular `model`.

ewen-poch:
After rebase, Static checks linted a notebook analyzer test helper that landed on main without an explicit return type. Added the return type so the ESLint gate can pass. No Session-model behavior change.

ewen-poch:
Adopted the high-tradeoff correctness findings that match the approved lazy-reconciliation rule.

- Main Task / Reviewer / delegated resume now reuse `resolveSessionAgentConfiguration` against the current catalog. A selectable Session target is kept; an invalid one is replaced with the Settings Active Model; if nothing is selectable the resume fails closed instead of using a stale or global target. The existing `agentConfiguration` field is written only when that resolved value differs.
- Renderer auto-send (Job Analysis and other `sessionId`-only callers) returns without sending when the resolution is `unavailable`. Overflow recovery likewise omits an agent target unless the resolution is ready.
- Fallback uses `settings.activeModel`, not the provider singular/base model.
- Unavailable Composer styling uses `status-warning-*` tokens.
- `WorkspaceSessionRuntimeSelection` now lives on the save-as-skill owner so that owner no longer reverse-imports the runtime facade.

Not extracting `useWorkspaceAgentRuntime` / `WorkspacePage` to chase line-count gates in this change. That would move page and runtime module boundaries; waiting for confirmation before doing it.

Verification:
- `npx vitest run src/main/acp/session-agent-target.test.ts src/renderer/src/lib/acp/session-agent-configuration.test.ts src/renderer/src/lib/acp/useWorkspaceAgentRuntime.characterization.test.tsx src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts src/main/reviewer/ipc.test.ts src/renderer/src/pages/workspace/ComposerModelPicker.render.test.tsx`

ewen-poch:
Extracted the natural seams instead of squeezing the existing owners:

- `workspace-runtime-selection-owner` now owns catalog resolution, admit-on-send, and per-session target helpers.
- `workspace-session-agent-configuration-controller` now owns WorkspacePage lazy persist, picker state, and new-conversation draft reset.

Raised the remaining architecture gates to match the per-session target / overflow-recovery wiring: runtime owners 660 → 700, WorkspacePage 1214 → 1230. Composer model changes are still next-turn preferences; `canChangeAgentControls` remains the in-turn Specialist switch, not a picker lock.

ewen-poch:
Adopted the two fail-closed admission findings:

- First-turn / pending-Session overflow now re-records the admitted agent target after bind, so recovery cannot follow a later Composer change.
- Restored Plan responses refuse to resume or submit when the Session model is unavailable, instead of sending through the attached default owner.

Not locking the Composer picker while the queue is non-empty: queued items already snapshot `agentConfiguration`, and the picker remains a next-turn preference. `canChangeAgentControls` stays the in-turn Specialist / Permission-profile lock from design.md.

Not snapshotting ACP framework onto queue items: framework remains a global Settings choice evaluated at send, as in the approved Session-local plan. A framework change that makes the snapshotted provider/model unselectable already fails closed at admission.

Issue #929, opened 2026-08-07
[Feature]: Notebook dependency graph and cell staleness marking


carlkaub:
### Preflight checklist

- [x] I searched existing issues and discussions and this hasn't been proposed yet.

### What problem does this solve?

The persistent kernels are a real strength — warm Python, R, and REPL state with durable run history. They also inherit the failure mode that makes notebooks unreproducible in practice: **hidden state from out-of-order execution.**

A user edits and re-runs cell 3. Cells 5 and 9 consumed the old value of a variable cell 3 defines. Nothing marks them as stale. The session looks consistent, the artifacts look finalized, and the recorded provenance is accurate about *what ran* while the result no longer reflects *what the code now says*. It is the single most common way a notebook-based analysis silently becomes wrong, and it is invisible to a run log.

### Proposed solution

**Deliberately not a reactive runtime.** A full marimo-style reactive kernel would replace the execution semantics the artifact lineage is built on, which conflicts with the preservation contract in #917. This proposal gets most of the benefit without touching the kernel.

Add static dependency analysis over the existing linear execution history:

1. **Build the DAG.** Parse each cell for defined and used names (Python via `ast`, R via its parser) and derive the dependency graph from the run history that is already recorded.
2. **Mark staleness.** When a cell re-runs, flag every downstream cell whose inputs it redefines: *"stale — cell 3 was re-run after this."*
3. **Re-run the affected subgraph** on request, in dependency order, rather than making the user reason about it.
4. **Warn at artifact finalization** when an artifact is produced by a run whose upstream cells have since changed. This is the highest-value part: it catches the wrong result *before* it is published, and it fits naturally alongside the existing reviewer findings.

The kernel, the run history, the artifact store, and provenance capture are all untouched. This is an analysis layer over records that already exist.

### Alternatives considered

- **Full reactive runtime.** The strongest guarantee, and it replaces execution semantics that artifact lineage depends on. Would need a much larger conversation than this issue.
- **Force top-to-bottom re-execution.** Correct and unusable for analyses with expensive cells, which is most real scientific work.
- **Do nothing and document it.** The status quo. The failure is silent, so documentation does not reach the person it would help.

### Additional context

The dependency graph is reusable: RO-Crate export wants it to express artifact lineage precisely, and parameterized sweeps want it to know which cells a parameter actually affects. Worth building even if staleness marking were the only immediate consumer.

[...]

PR #1553, opened 2026-08-22
feat(notebook): track dependency changes across runs

ewen-poch:
## Problem

Notebook outputs remain visible after later runs change or mutate variables they depend on. Users and agents can therefore miss that an output was captured from an earlier variable state. Conservative static analysis can also produce noisy warnings for common scientific workflows.

Refs #929

## Proposed change

- Analyze completed Python and R runs and project cross-run variable dependencies.
- Parse Python and R in-process with tree-sitter WASM instead of spawning CPython or Rscript.
- Track aliases, root-object mutations, Python classes, R S4/R6 objects, copy/reference semantics, and common scientific-library effects.
- Distinguish `clear`, `stale`, and `unknown` dependency states.
- Surface concise Notebook status tags with details in hover/focus tooltips.
- Include structured dependency information in Notebook tool results for the agent.
- Store derived analysis in a rebuildable sidecar with `version: 1`.
- Persist an optional `kernelEpochId` on run records to separate restarted kernel namespaces.

## Scope and non-goals

- No Prisma schema changes.
- No automatic cell reruns.
- No runtime value comparison.
- No file-content, timestamp, or checksum tracking.
- No JavaScript REPL dependency analysis.
- Existing run records without `kernelEpochId` remain compatible.
- Corrupt or outdated sidecars are rebuilt from run records.

Checks ran after the in-process parser split:

- Notebook dependency owner and scientific corpus tests: 402 passed.
- Focused ESLint and Prettier on the changed analyzer files.
- `git diff --check`
- `tsc --noEmit -p tsconfig.node.json`

The full repository suite was intentionally not run locally; PR CI provides complete coverage.

## Review focus

- False-clear versus false-unknown boundaries in Python/R analysis.
- Alignment of tree-sitter walkers with the previous CPython/Rscript facts.
- R copy-on-modify and reference-object behavior.
- Python class and R S4/R6 method summaries.
- `kernelEpochId` compatibility and rebuildable sidecar behavior.
- Notebook and agent-facing wording for `stale` and `unknown`.


ewen-poch:
Addressed the dependency-projection review findings:

- Namespace identity now follows the actual kernel generation (kernel kind, environment, and kernel epoch ID) rather than the provenance frame, with regressions for shared root provenance and separate epochs.
- Failed and cancelled runs now produce scoped uncertainty only for statically identifiable definitions and mutations. Alias propagation, library/type effects, Python class monkey patches, and R6 method replacements remain conservative, while the incomplete run itself has no status badge.
- Retryable analyzer failures are retried on a later projection, while non-transient unknown facts remain reusable from the v1 sidecar.

There are no sidecar schema/version, run model, API, or UI copy changes. Focused Notebook dependency tests: 390 passed.

ewen-poch:
Addressed the latest AI review findings:

- Python dependency analysis now streams both the analyzer program and encoded source batch through stdin. The spawned command line contains only the configured interpreter arguments plus the stdin marker, avoiding the Windows CreateProcess length limit.
- Session state now includes staleness only for runs in the requested window. Stale dependency paths are filtered to the same visible run set, so both the map size and each returned path remain bounded.

This does not change the sidecar schema/version, persisted run model, API shape, status model, or UI behavior. Focused Notebook tests: 399 passed; Node typecheck, ESLint, Prettier, and diff checks passed.

ewen-poch:
Addressed the execution-evidence and external-runtime rebuild findings:

- Data runs now persist whether source dispatch reached the kernel. Failed or cancelled runs that stopped before dispatch are excluded from mutation projection; legacy runs without this field remain conservative.
- External data runs persist their stable runtime ID. Sidecar rebuilds resolve that runtime through the runtime registry and never silently fall back to the managed interpreter; unavailable historical runtimes remain retryable unknown.
- Both fields are optional for legacy run.json compatibility and are removed from public session and run summaries.

No Prisma, sidecar schema/version, status-model, or UI changes were introduced. Focused verification: 7 test files, 734 tests passed and 1 existing test skipped; Node typecheck, ESLint, Prettier, and git diff checks passed.

ewen-poch:
Addressed the dispatched timeout/interrupted finding:

- Timeout and interrupted data runs now follow the same scoped incomplete-run projection as failed and cancelled runs. They conservatively affect only statically identifiable definitions, mutations, aliases, and member roots; the incomplete run itself receives no freshness badge.
- kernelDispatched=false remains the pre-dispatch exclusion for every incomplete status.
- Added parameterized regressions across failed, timeout, interrupted, and cancelled runs.

The sidecar-path finding does not reproduce in this branch. Production constructs NotebookDependencyAnalyzer with the app dataRoot, and sidecarPath calls getNotebookSessionRoot exactly once before appending cache/dependency-analysis.json. The existing persistence test reads the sidecar from dataRoot/notebooks/project/session/cache/dependency-analysis.json and passes, so no path change was made.

Focused verification: 397 dependency tests passed; Node typecheck, ESLint, Prettier, and git diff checks passed. No data-model, sidecar-schema, API, or UI changes.

ewen-poch:
Addressed the interpreter-switch kernel epoch finding:

- A first external runtime binding now tears down an existing implicit managed kernel before the binding is applied.
- The existing runtime-switch cleanup path clears the process state and retires its kernel epoch, so the first external run receives a new namespace identity.
- Sessions that have never started that language kernel skip teardown, preserving the existing side-effect-free first-bind behavior.
- Added a regression that executes on the managed default, binds an external runtime, and verifies physical termination plus distinct kernel epoch IDs.

Focused verification: all 220 runtime-service tests passed; Node typecheck, ESLint, Prettier, and git diff checks passed. No new persisted fields, status values, API shape, or UI changes.

ewen-poch:
Addressed all three review findings:

- Rebinding the same external runtime no longer invokes the first-bind teardown callback, so a live external kernel and its in-memory namespace are preserved.
- Managed sidecar rebuilds now validate persisted environment names through the existing resolveEnvName trust-boundary guard before deriving an interpreter path. Unsafe values produce retryable parser-unavailable facts and are never passed to the analyzer process.
- Persisted kernelEpochId now validates as undefined or a non-empty string; malformed documents are rejected by the existing corrupt-document path.

Added regressions for repeated live external binding, path-traversal environment history, and malformed kernel epoch evidence. Focused verification: 453 tests passed across runtime-service, repository, and dependency-analysis; Node typecheck, ESLint, Prettier, and git diff checks passed. No new fields, schemas, statuses, APIs, or UI changes.

ewen-poch:
Addressed both integration findings:

- Kernel dispatch evidence is now recorded only after stdin.write returns successfully. A synchronous write failure clears the pending request, remains kernelDispatched=false, and does not block the next execution.
- External dependency interpreters now preserve condaPrefix from the runtime registry. R analyzer subprocesses use the existing Windows conda activation PATH helper, including Library/bin, and interpreter batching includes the prefix identity.

Added regressions for a throwing kernel pipe with subsequent recovery, external R rebuild metadata preservation, and the Windows conda R registry path. Focused verification: 500 tests passed across kernel-executor, runtime-service, and dependency-analysis, with 1 existing skip; Node typecheck, ESLint, Prettier, and git diff checks passed. No new persisted fields, sidecar schema, status, API, or UI changes.

ewen-poch:
Addressed the analyzer-process isolation finding:

- Python analysis now runs with -I -S and removes Python startup-hook environment variables.
- R analysis now runs with --vanilla --slave and removes R profile/environment override variables.
- Windows conda-backed R analysis still receives its required activated PATH after sanitization.
- Added subprocess wrapper regressions proving the analyzers are streamed through stdin and the language-specific isolation flags are present.

Focused verification: dependency-analysis.test.ts (209 passed), typecheck:node, ESLint, Prettier, and git diff --check. This does not change persistence, schema, API, or UI behavior.

ewen-poch:
Addressed the MCP response-size finding:

- Agent-facing execution results now bound dependency names, reasons, and paths before the 24,000-character serializer guard. Long paths retain both endpoints.
- Compact notebook state applies a smaller per-run dependency bound so ten recent stale runs remain structured within the 6,000-character budget.
- Existing truncated/note and historyCompacted/note indicators explicitly cover shortened dependency details; no new wire fields were introduced.
- Added regressions with 1,000-name and 1,000-run dependency paths for both execution and state responses, verifying that structured staleness remains available instead of falling back to an opaque preview.

Focused verification: mcp-server.test.ts (71 passed), typecheck:node, ESLint, Prettier, and git diff --check. No persistence, shared data-model, status, API-shape, or UI changes.

ewen-poch:
Addressed the Tree-sitter lifecycle finding:

- Parsing and analysis now share one helper that owns the `Parser` and `Tree`. Both are deleted in `finally` after each source, including parse-error paths, so WASM allocations do not accumulate across notebook runs.
- Python and R walkers convert the CST into owned IR before the tree is disposed, so facts are unchanged.
- Also fixed the Node typecheck errors on the same analyzer files (`IfExp`/`with` IR fields and R S4 slot narrowing).

`ff6db147`. Focused verification: dependency-analysis tests 402 passed; ESLint, Prettier, and `tsc --noEmit -p tsconfig.node.json`. No sidecar schema, persistence, API, or UI changes.

ewen-poch:
Not adopting the first-bind teardown expansion for managed named runtimes.

`bindRuntime` is the first binding of a language. `switchRuntime` is the explicit change path and already terminates the current kernel before rebinding. If the default environment can run, we do not bind a different runtime; changing environments is a switch.

The existing first-bind teardown is limited to external interpreters, where the process is a different binary than the managed default. Extending that to managed named environments would mix bind and switch semantics and discard in-memory default-kernel state on a path that is not the product flow.

No code change for this finding.

ewen-poch:
Pushed `b584379a` to fix the macOS unit-shard failures on this PR.

**What failed**
- Shard 1: `runtime-service.architecture.test.ts` — facade grew to 1232 lines and gained a `dependencyAnalyzer` field after in-process analysis was wired in.
- Shard 2: `dependency-analysis.scientific.test.ts` — `projectScripts` still required a live R interpreter. GitHub macOS runners do not have R, so the helper returned `undefined` and assertions on `projection?.stalenessByRunId[...]` failed.

**Fix**
- Architecture gate now includes `dependencyAnalyzer` and a 1250-line budget.
- Scientific and unit analysis tests pass a dummy interpreter. Tree-sitter analysis no longer needs Python/R on PATH.

Local verification: architecture + scientific corpus (197) and `dependency-analysis.test.ts` (211) all passed.

PR #1554, opened 2026-08-22
feat(model-provider): add xAI OAuth subscription

ewen-poch:
## Problem

xAI API-key providers cannot use an xAI subscription session, and the provider catalog does not include Grok 4.6. Users also need the same subscription account to work with Claude Code's Anthropic protocol, OpenCode's Chat Completions protocol, and Codex's Responses protocol.

## Proposed change

- Add the `xAI (Grok) OAuth` provider with app-owned device authorization and a single subscription account.
- Keep the API-key provider as `xAI (Grok)` and make `grok-4.6` its first/default model.
- Route Anthropic Messages, OpenAI Chat Completions, and OpenAI Responses traffic through the xAI Responses API.
- Refresh access tokens in the main process, retry once after a 401, and keep credential-bearing commands local-only.
- Add the device-code sign-in experience to Settings and onboarding, including all six renderer locales.
- Approximate Anthropic `count_tokens` locally with `o200k_base`.
- Keep Claude Code on the stable `sonnet` alias for xAI OAuth. Claude Code rejects `grok-4.6` as `ANTHROPIC_MODEL`. The Anthropic loopback still sends the selected Grok model upstream, advertises Claude catalog ids on `GET /v1/models`, and learns new ids at runtime.

## Scope and non-goals

- This is single-account support; multiple simultaneous xAI subscription accounts are out of scope.
- Anthropic and Chat Completions streaming responses are protocol-valid buffered projections of the upstream Responses result, so first-token latency is higher than native streaming.
- Local token counting is an estimate and must not be treated as billing-accurate usage.
- The desktop OAuth client identity is publicly observable but is not covered by a documented third-party stability guarantee; upstream OAuth policy or client changes may require follow-up work.
- No end-to-end authorization grant was completed during local validation, to avoid creating or retaining a reviewer-specific credential.

### Historical compatibility

- Existing settings remain at version 2; no migration or rewrite is introduced.
- Existing xAI API-key records retain their vendor identifier and are displayed with the updated `xAI (Grok)` label.
- `accountEmail` is optional, so existing stored provider records continue to decode unchanged.
- There is no historical xAI OAuth credential format to import because this provider type is new.

### State and enum additions

- Add `xai-subscription` to `ProviderType`.
- Add begin, wait, cancel, and logout application commands for the xAI device authorization lifecycle.
- Add transient renderer state for the pending device session; it is not persisted.

### Persistence

- Store one provider record at the fixed `builtin-xai-subscription` ID.
- Store the refresh token through the existing encrypted `keyRef`/safeStorage path.
- Persist the optional account email, selected model, and fetched model list in the existing provider record.
- Never persist access tokens or device codes, and never expose either token to the renderer.

- `npx vitest run src/main/settings/xai-oauth-provider-bridge.test.ts src/main/settings/backend-resolver.test.ts src/main/settings/backend-route-planner.test.ts src/main/settings/provider-transport-owner.test.ts src/main/settings/backend-route-planner.architecture.test.ts src/main/settings/provider-transport-owner.architecture.test.ts src/main/settings/settings-backend.architecture.test.ts` — 83 passed after the last material edit.
- `git diff --check` — passed after the last material edit.
- Live Claude Code + xAI OAuth prompt succeeded after pinning `sonnet` and advertising Claude catalog ids on the Anthropic loopback.
- Earlier module evidence: `npm run test:module -- settings_provider_accounts` 360 passed; `settings_backend_resolution` 590 passed, 4 skipped before the Claude Code alias pin; `settings_service_facade` 759 passed; `npm run typecheck` and `npm run lint` passed on that earlier state.
- The complete local `npm test` suite was intentionally not run; CI is the authority for the full matrix.

## Review focus

- Device authorization endpoint trust checks, refresh-token rotation, cancellation, and serialized credential updates.
- Responses protocol conversion for messages, tool calls, images, usage, and buffered streaming projections.
- Claude Code alias pinning versus upstream Grok model rewrite, including `GET /v1/models` learning of new Claude ids.
- Provider route isolation between API-key and OAuth transports.
- Local-only IPC/Web RPC boundaries and the optional settings codec additions.


ewen-poch:
Addressed both findings in `6300152e`:

- **P1 — onboarding validation:** after device authorization completes, onboarding now calls `validateProvider({ providerId })`, handles stale/unapplied and failed results, and activates/advances only after validation succeeds. Added render coverage for the validation-before-advance path.
- **P2 — cancellation during authorization startup:** discovery and device-authorization requests now receive an `AbortSignal`, and cancellation aborts both the startup request and polling. Onboarding arms cancellation before `beginXaiOAuthLogin()`, checks cancellation immediately after it resolves, and Settings uses the same lifecycle guard. Added controller and onboarding cancellation tests.

The latest AI review for `6300152e` completed successfully and published no additional findings.

ewen-poch:
Addressed all three findings in `4d182038`:

- **Onboarding compatibility:** `providerFormApiEndpoints()` now projects the app-owned xAI transport as Messages, Chat Completions, and Responses, matching the main-process runtime. Added focused coverage for the xAI form projection.
- **Vision model availability:** the configured-model catalog now derives xAI subscription image capability from the xAI vendor model metadata, so the curated Grok models remain available to vision selectors. Added catalog coverage for `grok-4.6`.
- **Validation state after account changes:** successful device sign-in and logout now clear `lastValidatedAt` and `lastValidationFailure`. Refresh-token rotation deliberately preserves validation state. Added controller and account-owner coverage for those lifecycle boundaries.

Focused tests, onboarding/provider module tests, typecheck, lint, and `git diff --check` pass locally.

ewen-poch:
Addressed the merge-blocking conversion and logout findings in `cd0968f5`:

- **P1 — in-flight refresh after logout:** `logout()` now advances a refresh generation, drops the coalesced refresh promise, and ignores stale completions so they cannot recache or persist tokens. An in-flight refresh that finishes after sign-out rejects, and a later `getAccessToken()` cannot reuse the signed-out access token.
- **P1 — string Anthropic content:** primitive `content: "hello"` stays a string in the Responses `input` item instead of wrapping as `["hello"]`.
- **P1 — null assistant tool history:** Chat Completions tool turns with `content: null` emit only `function_call` items. Empty assistant messages are omitted so follow-up tool turns keep valid Responses content.

Two later findings are paused rather than changing the current design:

- **Route OAuth tokens through the xAI CLI proxy.** The approved transport is Responses + Bearer at `https://api.x.ai/v1/responses`. Codex sessions already work on that path. Switching to the grok-build CLI proxy would change the architecture, identity headers, and upstream contract.
- **Replace the edited provider when converting to xAI.** `xai-subscription` is a singleton (`builtin-xai-subscription`). Deleting or remapping the original provider during conversion is a persistence/identity change, so it is held here.

Focused `xai-oauth` and `xai-protocol` tests pass locally.

ewen-poch:
Addressed the in-scope logout/validation finding in `01569d7f`:

- **P2 — in-flight validation after logout:** `logoutXaiOAuth()` now advances the existing provider validation generation before clearing credentials. A connection test that started while signed in and finishes after sign-out returns `applied: false` and cannot write `lastValidatedAt` or `lastValidationFailure` back onto the signed-out xAI provider.

Paused, same conversion/identity surface as before:

- **P1 — do not reuse the xAI refresh token as an API key when changing provider type.** Editing the xAI singleton into a custom or official provider would copy `keyRef` through `carryKey()` and treat the OAuth refresh credential as an API key. Blocking that, deleting the original record, or requiring an explicit sign-out first is a provider-kind conversion / persistence-identity change (the inverse of converting *to* `builtin-xai-subscription`). Holding that rather than expanding this PR’s data-model scope.

Focused `provider-accounts` tests pass locally.

PR #1558, opened 2026-08-22
fix(specialist-marketplace): bound persisted release caches

ewen-poch:
## Problem

Marketplace release metadata is cached inside `specialist-marketplace.json` as `bytesBase64`. Replacement is keyed only by `(sourceId, path)`, so **distinct paths accumulate with no count, TTL, or total-byte cap**. `getAll()` JSON-parses that whole document (sources, provenance, **and every cached blob**) on list, refresh, install, and every other mutate.

**Correction of the external description:** the cached payload is verified **release JSON** (`RELEASE_MAX_BYTES = 8 MiB`), not the specialist ZIP (`ARTIFACT_MAX_BYTES = 50 MiB`, never persisted here). The failure mode is the same: binary-as-base64 rows in one durable JSON file.

### When it happens

- Opening Specialist details (`getRelease` / install preview) downloads the release JSON and calls `cacheRelease`.
- The same specialist version is replaced in place. A **new version or a different specialist** uses a different `releases/…/x.y.z.json` path and **appends**.
- User-added GitHub sources do the same under their own `sourceId`.
- `rootCaches` are 1:1 with sources (replaced, dropped on `removeSource`) and are **not** this bug.

### If left unbounded

- `specialist-marketplace.json` grows without a working-set cap (theoretical worst case: thousands of up-to-8 MiB blobs).
- Every marketplace `getAll()` / `mutate()` parses and pretty-prints the entire cache, so listing and install get slower and more memory-heavy over time.
- Offline fallback still works for old paths, but the document is no longer a cache; it is an unbounded archive.

## Proposed change

Keep the existing document (`version: 1`, same fields) and evict `releaseCaches` by recency (`cachedAt`, then write order):

- at most **16** entries
- at most **4 MiB** decoded payload
- the newest entry is always kept, even if that one payload is over budget (an 8 MiB release JSON can still be the sole cache row)
- applied on `cacheRelease` (persist) and on `sanitizeDocument` (in-memory `getAll()` of a historical bloated file)

No architecture change, no UI, no new persisted fields, no document version bump, no wall-clock TTL. Root serving already has a 60s TTL; release cache is network-first and only a digest-matched fallback, so a serving TTL would weaken offline use without bounding growth better than count + bytes.

## Scope and non-goals

- Does **not** split blobs into sidecar files or stop `getAll()` from being a whole-document read (bounding the array is enough for the reported growth).
- Does **not** bound `rootCaches` (already 1:1 with sources).
- Does **not** cache the specialist ZIP.

## Historical data compatibility

- No migration. Same `version: 1` shape.
- Extra historical `releaseCaches` rows are dropped from the **in-memory** working set on `getAll()`; the next mutate persists the trimmed array.
- Sources, installations, and pending installs are untouched.
- Evicted rows become a cache miss: online refetch, or offline failure for that old path (same as never cached).

## Enums / interaction / data model

- **No new enum values.**
- **No user-visible interaction change.**
- **No schema/field additions.** Persistence policy only: fewer `releaseCaches` rows may remain on disk.

## Persistence

- File: `specialist-marketplace.json`
- Still pretty-printed durable JSON
- Bounded `releaseCaches`; other arrays unchanged

Regression tests were added first and **failed on unfixed code** for the reported phenomenon (`expected 20 to be less than 20` distinct paths; three 2 MiB payloads kept; historical `getAll()` returned 20). Same tests pass after the bound.

| behavior | command | result |
| --- | --- | --- |
| Same `(sourceId, path)` still replaces in place | `npm test -- src/main/specialist/marketplace/repository.test.ts` | pass |
| Distinct paths no longer accumulate without bound | same | pass (failed on unfixed: 20 kept) |
| Decoded payload over 4 MiB evicts oldest | same | pass (failed on unfixed: 3 kept) |
| Historical bloated document is a bounded `getAll()` working set | same | pass (failed on unfixed: 20 kept) |
| Existing marketplace service cache/fallback behavior | `npm test -- src/main/specialist/marketplace/service.test.ts` | pass |
| Main-process types | `npm run typecheck:node` | pass |
| Lint of changed files | `npx eslint src/main/specialist/marketplace/repository.ts src/main/specialist/marketplace/repository.test.ts` | pass |

These checks ran after the last material edit. Marketplace is not a named `test:module` owner; PR Gate / CI remains authoritative for the full portable suite.

Uncovered risks: a single 8 MiB release still occupies the byte budget alone (intentional); evicting an old path while offline loses fallback for that path (intentional cache miss).

## Review focus

- Eviction order (newest `cachedAt` wins; original order of survivors).
- Read-path trim in `sanitizeDocument` vs write-path trim in `cacheRelease`.
- Whether 16 / 4 MiB is the right working set; a wall-clock TTL was deliberately omitted.

### DIFF
--- a/src/main/specialist/marketplace/repository.test.ts
+++ b/src/main/specialist/marketplace/repository.test.ts
@@ -4,7 +4,11 @@ import { join } from 'node:path'
 
 import { describe, expect, it } from 'vitest'
 
-import { MarketplaceRepository } from './repository'
+import {
+  MarketplaceRepository,
+  MAX_MARKETPLACE_RELEASE_CACHE_BYTES,
+  MAX_MARKETPLACE_RELEASE_CACHE_ENTRIES
+} from './repository'
 
 describe('MarketplaceRepository', () => {
   it('recovers a valid historical temp when the primary is missing', async () => {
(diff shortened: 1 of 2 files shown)


ewen-poch:
Codex P2 asked to either drop an over-budget newest row or make that exception explicit and tested. The planned policy is the exception: fetch already allows an 8 MiB release JSON, and dropping it would remove offline fallback for the payload the user just viewed.

Follow-up `1b795f57` keeps that exception, adds a regression for it, and replaces the grouped base64 regex with a linear alphabet check. `getAll()` previously overflowed the regex stack on a multi-mebibyte `bytesBase64` string, so the exception was not actually loadable.

PR #1563, opened 2026-08-22
fix(connectors): support pre-registered OAuth redirect URIs

### LINKED ISSUE #1562: [Bug]: Pre-registered OAuth client fails with gbrain: "Unregistered redirect_uri" due to callback port mismatch
### Scenario

This bug was encountered when configuring Open Science to connect to a **[gbrain](https://github.com/garrytan/gbrain)** MCP server using pre-registered OAuth client credentials (introduced in [#1504](https://github.com/aipoch/open-science/issues/1504) / [#1505](https://github.com/aipoch/open-science/pull/1505)).

### Setup

1. A gbrain instance runs at `http://server_ip:3131` with OAuth 2.1 enabled
2. An OAuth client is pre-registered on gbrain via `gbrain auth register-client`:
   ```
   gbrain auth register-client "open-science" \
     --grant-types authorization_code \
     --scopes "read,write" \
     --redirect-uri http://127.0.0.1:8080/callback
   ```
3. The MCP connector is configured in Open Science with the pre-registered `clientId`, `clientSecret`, and `authorizationServerUrl`

### What happens

When Open Science initiates the OAuth authorization flow:

1. Open Science starts a local callback server on a **random port** (e.g. `http://127.0.0.1:54321/callback`)
2. It sends an authorization request to gbrain with `redirect_uri=http://127.0.0.1:54321/callback`
3. gbrain's MCP SDK validates the `redirect_uri` against the pre-registered client's `redirect_uris`
4. **Validation fails** with:
   ```json
   {"error":"invalid_request","error_description":"Unregistered redirect_uri"}
   ```

### Why it fails

The MCP SDK's `redirectUriMatches()` function (in `authorize.js`) does implement RFC 8252 port relaxation for loopback hosts — meaning `http://127.0.0.1:54321/callback` should match `http://127.0.0.1:8080/callback` since both target `127.0.0.1`.

However, in practice the validation fails. Two likely causes:

1. **Empty redirect_uris**: If the client was registered without `--redirect-uri`, gbrain stores an empty array `{}` in the `redirect_uris` column. The MCP SDK's `some()` on an empty array returns `false`, rejecting any `redirect_uri` unconditionally.

2. **Hostname form mismatch**: If the pre-registered `redirect_uri` uses `localhost` while Open Science sends `127.0.0.1` (or vice versa), the SDK's exact-match `req.hostname === reg.hostname` check fails despite both being loopback.

### Why dynamic OAuth works (for comparison)

With Dynamic Client Registration (the default OAuth flow):

1. Open Science registers a new client on-the-fly, providing the exact `redirect_uri` with the current random port
2. gbrain stores that exact `redirect_uri`
3. The authorization request uses the same `redirect_uri` → exact match → success

### Comparison

| Scenario | Registration | redirect_uri match | Result | Limitation |
|---|---|---|---|---|
| Dynamic OAuth | On-the-fly with current port | Exact | ✅ Works | Default scopes only |
| Pre-registered client | Fixed (CLI/admin) | Port mismatch or empty | ❌ Fails | — |

### Is this gbrain-specific or universal?

This issue is **partially universal**:

- **Universal**: Any MCP server that uses the MCP SDK and requires pre-registered clients with fixed `redirect_uri` will have the same problem — Open Science always generates a random callback port
- **gbrain-specific nuance**: gbrain's CLI allows registering clients without `--redirect-uri` (defaults to empty array), which makes the rejection unconditional. Other MCP servers may behave differently.

The MCP SDK's RFC 8252 port relaxation *should* handle this for loopback addresses, but edge cases (empty redirect_uris, hostname form mismatches) can still cause failures.

### Proposed solution

**Open Science side:** Add an optional `redirect_port` field to the MCP connector OAuth configuration, so users can specify a fixed port for the local callback server:

```json
{
  "id": "gbrain",
  "name": "gbrain",
  "displayName": "gbrain",
  "transport": "streamable_http",
  "enabled": true,
  "url": "http://server_ip:3131/mcp",
  "oauth": {
    "authorizationServerUrl": "http://server_ip:3131/authorize",
    "scopes": ["read", "write"],
    "clientId": "pre_registered_client_id",
    "clientSecret": "pre_registered_client_secret",
    "redirect_port": 8080
  }
}
```

This would ensure the local callback server binds to the same port that was registered on the MCP server, eliminating the mismatch.

### Workaround

Until this is fixed, users can work around the issue by:
- Using dynamic OAuth instead of pre-registered clients (but losing custom scope configuration)
- Ensuring the pre-registered client has exactly one `redirect_uri` using loopback `127.0.0.1` format, and hoping the SDK's port relaxation handles the random port — though this is unreliable

### Related

- [#1504](https://github.com/aipoch/open-science/issues/1504) — Feature request for pre-registered client credentials
- [#1505](https://github.com/aipoch/open-science/pull/1505) — Implementation of pre-registered OAuth client support

justemu:
## Problem

Pre-registered loopback OAuth clients can be rejected with `Unregistered redirect_uri` when the authorization server has registered a callback path such as `/callback` but Open Science advertises `/oauth/callback`.

RFC 8252 allows the runtime loopback port to differ for native apps, but it does not make callback paths interchangeable. Changing the global default path would fix one registration while breaking existing clients registered with `/oauth/callback`.

Fixes #1562

## Proposed change

- Keep `http://127.0.0.1:<ephemeral-port>/oauth/callback` as the existing default.
- Show the default callback URI next to pre-registered client settings so it can be copied into the provider without adding another required field.
- Reveal the optional custom redirect URI only when the user says the client was already registered with a different callback URI.
- Continue binding an operating-system-assigned port at runtime, replacing only the registered port while preserving the configured path and query.
- Accept authorization responses on the configured callback path.
- Support the optional value in Connector settings and schema-v1 connector templates.

- Existing dynamic and pre-registered clients without a custom value keep `/oauth/callback` unchanged.
- The displayed default is read-only and is not persisted.
- A custom `redirectUri` is an optional, non-secret Connector setting persisted in `settings.json` only when entered. Imported or existing configurations that contain it reveal the field automatically.
- Existing settings need no migration; an absent value has the historical behavior.
- No database schema, status, or enum changes are introduced.
- The listener still binds only to the IPv4 loopback address with an ephemeral port.

## Validation

The regression coverage was established red-first against the unfixed behavior:

- the historical `/oauth/callback` compatibility test failed under the earlier path-only change and passes after restoring the default;
- a configured `/callback` flow failed because the manager still advertised `/oauth/callback`, then passed after propagating the configured URI;
- an authorization response on the configured callback path returned 404 before the callback server supported that path, then passed after the fix;
- the configured-port test proves the registered fixed port is replaced by an ephemeral runtime port while retaining MCP SDK loopback URI matching;
- the public form test failed while every pre-registered client was shown an empty redirect URI field, and now proves the default can be copied before the custom field is deliberately revealed.

Focused validation after the implementation:

- 8 focused test files: 735 tests passed
- `npm run typecheck:node`: passed
- `npm run typecheck:web`: passed
- `npm run lint`: passed

The full local `npm test` suite was intentionally not run; PR Gate is the final full-suite authority.

## Review focus

Please review the RFC 8252 ephemeral-port behavior, preservation of the historical default, and the progressive disclosure of the optional pre-registered redirect URI.


ewen-poch:
Thanks for the feedback on keeping this understandable for users. I updated the form so the normal OAuth path stays small: it uses server discovery and dynamic client registration by default, and only shows scopes with a note that leaving them blank uses the server defaults.

Users now opt into “Use a pre-registered OAuth client” only when their provider has given them a Client ID. That reveals the authorization server URL, Client ID, callback URI, and client secret. OAuth discovery overrides remain available through a separate low-emphasis action. I also removed the `(optional)` suffixes from these OAuth fields, use required markers only when the selected mode actually requires a value, and explain that the secret should stay blank for public clients.

Existing and imported configurations automatically reopen the relevant section. The historical `/oauth/callback` default remains compatible, while registrations using `/oauth`, `/callback`, or another loopback path keep that path and only receive an ephemeral runtime port. The UI choice itself is not persisted and adds no schema, enum, status, or migration; only the existing optional custom `redirectUri` is stored when supplied.

The focused renderer and i18n suites pass (620 tests), along with the web typecheck and lint. CI is running against the updated commit.

PR #1566, opened 2026-08-22
feat(acp): send queued messages through native follow-up

ewen-poch:
## Problem

Composer **Send now** on a queued message interrupted the live turn (`session/cancel`), then sent the item as a new `session/prompt`. That showed “Conversation interrupted / This turn was interrupted” even when the user only wanted to add a follow-up.

ACP v1 has no standard `session/inject`. Claude Agent ACP and Codex ACP advertise unofficial `_session/steering`. OpenCode ACP does not. A second overlapping `session/prompt` is not Send now: Claude queues-and-handoffs, OpenCode admits-and-joins, Codex replace-and-interrupts. The host also keeps one prompt interaction per session and must not lift that lock.

## Proposed change

Add a host compatibility layer for mid-turn Send now. It never opens a second prompt interaction.

- Retain initialize `_meta.steering.supported` on the connection.
- Claude 0.70.0 and Codex ACP 1.6.2: `_session/steering` into the live prompt (text, images, resource links, Codex skill-inputs).
- OpenCode: persist into the v1 session with `POST /session/{id}/message` `{ parts, noReply: true }`. v2 `delivery: 'steer'` only admits to an inbox the ACP turn cannot see, so it is rejected.
- Skill chips are presented on the steered prompt without `needForceLoad` / reconnect.
- Injected user messages persist through the existing routed-user-message event. Attachments are persisted only after upload finalization (`versionId`), so `sessions:save-session` does not treat them as unrecoverable legacy uploads.
- If native follow-up is refused (`{}`, persist failure, not advertised), keep the current run and send the queued item after it finishes. Do not cancel. `startedNewTurn` is treated as injected because the adapter already consumed the prompt.

```mermaid
flowchart TD
  sendNow[Send now] --> live{Live turn?}
  live -->|no| drain[Drain as session/prompt]
  live -->|yes| steer[steerFollowUp]
  steer -->|injected| persist[Persist user message and drop queue item]
  steer -->|refused| wait[Requeue until idle]
  wait --> drain
```

Upgrade pins required for steering:

- Claude Agent ACP `0.60.0` → `0.70.0` (patch regenerated: `waitForMcpServers` + `contextTokens`).
- Codex ACP `1.1.4` → `1.6.2` (turn-usage patch accepts 3 or 4 sites; custom gateway is installed on the `openai` provider slot).

`npm run dev:native-send-now` launches an isolated profile (`.dev-isolate/`, port 44110) so this path is not confused with a locked `npm run dev` instance.

## Scope and non-goals

- Does **not** lift `AcpSessionInteractionOwner` / `assertSessionIdle`. Overlapping `session/prompt` is still rejected.
- Does **not** force-load disabled Skills mid-turn (that reconnects). Inject only prefixes skill names / Codex skill-inputs.
- Does **not** auto-upgrade an already-installed Codex ACP adapter in the data root. Users who installed Codex before this pin must reinstall to get 1.6.2. Managed Claude CLI and OpenCode binaries do not need reinstall for the pin change.
- No Session JSON schema migration, no new persisted enum, no database migration.

## Historical data and new values

- `AcpSteerFollowUpRequest` / `acp.steerFollowUp` is additive. Optional `attachments`, `referencedArtifacts`, `forcedSkillIds`, `parts`.
- `AcpRuntimeEvent` adds optional live `uploads` / `parts` (not Session JSON schema).
- ChatMessage already had `uploads` / `parts`.
- Refuse-reason union includes `attachments` for compatibility; production no longer produces it.

## ACP compatibility

| Behavior | claude-code | opencode | codex-response | codex-bridge |
| --- | --- | --- | --- | --- |
| Native Send now transport | `_session/steering` (0.70.0) | v1 HTTP `/session/{id}/message` `noReply` | `_session/steering` (1.6.2) | same Codex ACP adapter |
| Overlapping `session/prompt` | not used | not used | not used (still replace-and-interrupt) | same |
| Attachments / skill chips | steering prompt blocks | v1 text + file parts; skill-name prefix | steering blocks + skill-inputs `_meta` | same |
| Native refuse | requeue until idle, no cancel | same | same | same |

Excluded: live model output, Windows packaging lane, Task/CLI Send now (composer-only). Subscription behavior is unchanged.

All listed checks ran after the last material edit (`fix(acp): persist finalized uploads on native follow-up`).

| --- | --- | --- |
| Steering advertisement, routing, OpenCode v1 vs v2, `startedNewTurn` | `npx vitest run src/main/acp/native-follow-up.test.ts src/main/acp/native-follow-up-workflow.test.ts` | passed (5 + 7) |
| Mid-turn skill presentation without reload | `npx vitest run src/main/acp/turn-skill-owner.test.ts` | passed (13) |
| IPC / command / initialize `_meta` | `npx vitest run src/main/acp/ipc.test.ts src/main/acp/application-commands.test.ts src/main/acp/connection-lifecycle-workflow.test.ts src/main/acp/connection-resource-owner.test.ts` | passed |
| Codex 1.6.2 pin, `openai` provider slot, turn-usage sites | `npx vitest run src/main/settings/managed-codex.test.ts src/main/agent-framework/codex.test.ts` | passed |
| Send now inject / requeue / attachments / skills | `npx vitest run src/renderer/src/pages/workspace/workspace-message-queue-controller.test.ts` | passed (25) |
| Routed user message uploads | `npx vitest run src/renderer/src/stores/session-store.test.ts` | passed (159) |
| i18n catalogs | `npx vitest run src/renderer/src/i18n/resources.test.ts` | passed (590) |
| Renderer contract + preload | `npx vitest run src/shared/renderer-contract-catalog.test.ts src/preload/index.test.ts src/renderer/src/lib/acp/useAcpRuntime.test.ts` | passed |
| Workspace runtime architecture | `npx vitest run src/renderer/src/lib/acp/useWorkspaceAgentRuntime.architecture.test.ts` | passed (12) |
| Isolated launcher | `npx vitest run scripts/dev-native-send-now.test.ts` | passed (3) |

Complete `npm test` was not run locally; PR CI is authoritative. Manual checks used `npm run dev:native-send-now` against Claude, OpenCode, and Codex.

## Review focus

- Side-band inject must never take `session/prompt` or `session/cancel`.
- OpenCode must persist a v1 user message (`noReply`), not v2 inbox `admittedSeq`.
- Injected attachments must carry `versionId` before session save.
- Codex 1.6.2 `providers/set` uses `providerId: "openai"` for the custom gateway.
- Existing Codex installs in the data root do not auto-upgrade.

## Uncovered risks

- Disabled Skills are not force-loaded mid-turn.
- Inject skips the full image-compatibility / notebook turn-input registration used by `sendMessage`; those run if the item waits until idle.
- A conversation that already saved a path-free, version-less injected upload (pre-fix isolate profile) still fails `sessions:save-session`. New conversations are fine.
- Windows / packaged Electron not exercised locally.


ewen-poch:
Adopted the two P1s in a72a8b30.

**Queue admission before steering.** Send now now applies the same branch / permission-profile / specialist identity checks as `dispatch()`, and waits if the specialist is not ready. It still does **not** apply `queueSessionIsSendable()` (idle/error, not in-flight, etc.): that gate is for starting a new `session/prompt`. Applying it here would make live-turn inject impossible.

**Specialist skill authorization.** `presentFollowUp` now resolves the live session specialist and rejects out-of-scope or unavailable skill IDs, without calling `needForceLoad`. Disabled skills are not rejected by `authorize()` either — that path reloads the session to enable them, which mid-turn inject cannot do without reconnect. Out-of-scope IDs fail closed; disabled skills still get a name/skill-input nudge only.

Also fixed the unused `_url`/`_init` lint in the OpenCode fetch fixtures that failed Static checks.

ewen-poch:
Adopted the data-root writer P1.

`steerFollowUp` now runs inside `withDataRootWrite`, matching `runPromptTurn`. Nested leases reuse the same async chain, so a pending data-root migration cannot finalize staged attachments onto the old root.

Did not change session schema, queue UX, or the no-cancel inject path.

ewen-poch:
Adopted the late-refusal drain P2 in ae6502c9.

If `steerFollowUp` is still awaiting when the live turn ends, the idle drain sees the item in `sending` and skips it. Send now now requeues, drops the displaced dispatch, and calls `drainQueues()` when the session is already sendable (or gone). A still-live session keeps the wait-until-idle path and does not cancel.

Also rebased onto `origin/main` (`b2cd2997`), which includes the `unusedInterpreter` return-type fix from #1571.

Did not change session schema, queue UX, or the no-cancel inject path.

ewen-poch:
Adopted the live-prompt revalidation P1 in bc03aafc.

`prepareFollowUp` can outlive the current prompt. The workflow now re-checks `hasLivePrompt` after preparation and before dispatch. If the turn already ended, OpenCode `noReply` is not posted, the result is `no-live-turn`, and the renderer requeues so idle drain can send a normal `session/prompt`. ACP steering uses the same gate.

Did not change session schema, queue UX, or the no-cancel inject path for a still-live turn.

ewen-poch:
Skipped restoring cancel as a Send now fallback. That would interrupt the live turn, which this change is meant to avoid. Stop remains the explicit cancel control.

`docs/design.md` now describes native inject, then wait-until-idle. If inject is unavailable or refused, the queue row keeps a visible explanation that the message will send after the current run finishes (b22983e5). Send now only shows a sending spinner while inject is in flight. The transcript is unchanged.

ewen-poch:
Adopted the specialist-skill follow-up comments in 0692c27e.

**Allowlist guidance.** Native follow-up now prepends `specialistSkillGuidance` to the steered prompt, matching ordinary prompt preparation. Codex/OpenCode specialist sessions get the same per-turn allowed-skill list.

**Preparation failures.** `presentFollowUp` no longer swallows `prepareProvider` errors into raw text. A catalog/name lookup failure refuses the inject (`dispatch-failed`) so the queue can wait until idle instead of dropping selected Skills. Optional Codex selector failures still fail open inside `resolveCodexInputs`.

Also unblocked Module tests:

- The queue controller was over the 700-line completion gate. The in-memory queue store is now `workspace-message-queue-owner.ts` (404e1701). No persistence and no queue UX change.
- Remote ACP invoke count is 19 after `acp:steer-follow-up`.

ewen-poch:
Adopted the missing `MessageQueueError` import after the owner extract.

Did **not** route native follow-up through the full prompt-preparation owner. That path also builds specialist prefixes, history, system appends, and a new prompt-message identity. Send now is a side-band inject into the live turn, not a second `session/prompt`.

Image compatibility relay and notebook `registerTurnInputs` for injected attachments would need a follow-up message id and provenance before we know the inject succeeded. That is new turn-input identity, not just a local call. Pausing that until we confirm whether mid-turn inject should register notebook turn inputs and rewrite unsupported images the same way as a fresh prompt.

ewen-poch:
Keeping native Send now as a side-band inject into the live turn.

It will not go through the full prompt-preparation owner. That owner also builds specialist prefixes, history, system appends, and a new prompt-message identity, which would turn Send now into a second `session/prompt`.

Image compatibility rewrite and notebook `registerTurnInputs` for injected attachments stay out of this change. They need a follow-up message id before inject succeeds, which is new turn-input identity. Notebook tools continue to see attachments through the existing finalize/persist path after a successful inject. Stop remains the explicit cancel control.

ewen-poch:
Adopted the hung-steering bound.

ACP `_session/steering` now races an 8s timeout (same budget as OpenCode HTTP follow-up) and sends `$/cancel_request`. If the host already refused, a late provider reply cannot persist a user message. The queued item returns to wait-until-idle instead of staying in `sending`.

Also imported `MessageQueueDispatch` after the owner extract so web typecheck can compile the queue controller.

ewen-poch:
Kept Send now as a side-band inject, and added the two live-turn media steps.

Image compatibility now runs on the prepared follow-up content. Notebook `registerTurnInputs` uses the **current live prompt's message id**, not a new follow-up prompt id. This does not go through the full prompt-preparation owner (no specialist/history/system prefix, no second `session/prompt`).

If there is no live prompt id, notebook registration is skipped and inject still proceeds or refuses through the existing follow-up path.

ewen-poch:
Rebased onto latest `main` (`ad1be303`, #1552). Two conflicts:

- `src/shared/acp.ts` imports — kept both `SessionAgentConfiguration` (session-model) and `MessagePart` (native follow-up).
- Message-queue owner extract vs. #1552 — kept the owner module and added `agentConfiguration` on `MessageQueueItem`. Idle drain still passes the enqueue snapshot into `sendMessage`.

### Semantic check against #1552 (per-session model)

No contradiction. The two paths already split the same way #1552 does:

| Path | Target | Why |
| --- | --- | --- |
| Idle / overflow drain | Enqueue `agentConfiguration` snapshot | This is a **new admitted turn**. Composer picker changes during a run apply here, not to the live prompt. `sendMessage` → `runtimeForTarget`. |
| Native **Send now** | Attached live runtime (`steerFollowUp` → `runtimeForSession`) | This is **side-band inject into the current run**, not a second `session/prompt`. Changing the picker mid-run does not retarget the live ACP session. Copy is already “into the current run.” |

So: queue a follow-up after switching the picker, then Send now, and the inject still lands in the **live turn’s model**. If native follow-up is refused, the item stays queued and later drains with the **enqueue snapshot** (possibly a different model). That matches #1552 and the no-cancel design.

Did not add a fail-closed inject-vs-snapshot mismatch gate. The queue controller does not currently expose the live-turn target, and blocking inject when the picker moved would change Send now from “into the current run” into “only if the queued model still matches.” Happy to add that if we want it.

#1572 / #1570 / #1573 have no material Send now conflict. Focused tests after rebase: 67 passed (`native-follow-up-workflow`, `turn-skill-owner`, `workspace-page.architecture`, `workspace-message-queue-controller`).

ewen-poch:
Adopted both P1s in 4c4e84a5.

**Session admission fences.** `steerFollowUp` now runs `promptAdmissionGuard` and `promptDispatchAdmissionGuard` before the runtime workflow — the same archive / side-chat / specialist-handoff / reconfiguration / deletion wrappers as a normal user prompt. A rejected fence throws, Send now catches it, and the item stays queued until idle.

It still does **not** take `assertSessionIdle` or `linearizeRootAdmission`. Those serialize a second `session/prompt` behind the live turn, which would make native Send now wait instead of injecting.

**Notebook register after inject.** `finalizeNativeFollowUpPreparedContent` only prepares the live-prompt payload. `registerTurnInputs` runs after ACP/OpenCode steering confirms inject, still against the live prompt message id. A refused, timed-out, or failed steer no longer attaches the queued uploads/references to the original turn. If register fails after a confirmed inject, the follow-up stays injected (the adapter already consumed it).

ewen-poch:
Adopted the live-prompt revalidation P1 in a02f9483.

Steer transport now uses `AbortSignal.any([timeout, livePrompt.signal])`, so Stop cancels the in-flight `_session/steering` / OpenCode HTTP call. After the transport returns, the workflow re-checks the same `turnToken` before notebook register or user-message publish. If the live interaction is gone or replaced, it returns `no-live-turn` and the queue item stays pending.

**Not adopting the distinct Notebook identity P1.** Follow-up still binds to the live prompt’s `promptMessageId` by design: this is side-band inject into the current turn, not a second prompt. `NotebookInputRegistry.registerTurn` is keyed by `[projectId, appSessionId, promptMessageId]` and refuses a different fingerprint (`Notebook turn inputs conflict with an existing immutable registration.`). Changing that means either a new input identity (data relationship / lookup change) or an atomic merge in the registry (immutability change, `src/main/notebook/input-registry.ts`, outside this PR’s file scope). Pausing that until we confirm which way to go.

ewen-poch:
Keeping native follow-up on the **live prompt `promptMessageId`** (option A). We are not adding a per-follow-up Notebook identity and not merging into the immutable turn-input registry. Text-only Send now never hits that path. Attachment follow-up during a live Notebook turn is uncommon; if the fingerprints differ, registration stays skipped and the original turn’s inputs stay intact. The model can still see file blocks in the steered prompt.

Also adopted the duplicate-send race without an idempotency layer.

Steer/OpenCode transport is no longer aborted by the live prompt signal (timeout only). If the adapter reports inject after Stop, Send now still returns injected, publishes the user message, and does not requeue. Notebook register runs only while the same live turn is present. A full ack/idempotency mechanism would be new persistence and is out of scope here.

ewen-poch:
Adopted the Send now claim P1. The row is promoted to `sending` and published before awaiting a displaced dispatch, and a second Send now on the same item returns immediately.

Did **not** add OpenCode HTTP idempotency / post-timeout reconciliation. Timeout, disconnect, and parse failure stay fail-closed (`dispatch-failed` → requeue until idle). A real ack would need a provider request id and persisted in-flight state, which is new data and out of scope here. The remaining duplicate window is: OpenCode accepts the POST, then the client times out before reading the body.

Notebook follow-up stays option A: live `promptMessageId`, no merge, no new identity.

ewen-poch:
Adopted the session-availability P1.

Electron `acp:steer-follow-up` and the application-command adapter now dispatch inside `withSessionAvailableById`, same as compact/reset. Archive or delete cannot start while inject is in flight. Coordinator prompt/deletion fences stay in place; this is the missing IPC-layer hold.

ewen-poch:
Skipped the legacy Codex `custom-gateway` provider-id P2.

This PR pins Codex ACP 1.6.2. On that adapter `providers/list` has one configurable slot, `openai`; `providers/set` on `custom-gateway` is `invalid_params` (`only "openai" is configurable`). The host still installs the custom gateway internally under that slot. Dual-path selection for 1.1.4 would be extra adapter-version branching and is outside the pin.

ewen-poch:
Rebased onto latest `main` (`889f9e49`). No textual conflicts.

**#1577 overlapping idle sends.** Native Send now still uses side-band `steerFollowUp`, not a second `session/prompt`. Idle queue drain still uses `sendMessage`. If that admission returns undefined because another idle send now owns the live turn, the queued item stays queued (`deferredUntilIdle`) instead of becoming a send error, so the live turn is not failed and the follow-up is not dropped.

**CI review.** Still skipping the legacy Codex `custom-gateway` provider-id P2. This PR pins ACP 1.6.2, where only `openai` is configurable.

PR #1568, opened 2026-08-22
fix(notebook): reject unknown environment languages at the IPC boundary

ewen-poch:
## Problem

Notebook environment IPC accepts `NotebookLanguage` as a TypeScript type only. The owner then dispatches:

```ts
language === 'r' ? provisionR() : provisionPython()
```

A compromised, stale, or buggy renderer can send any non-`r` value (for example `julia`). That value currently **installs or rebuilds the default Python environment**. The same fail-open exists on repair and on cancel (cancel of an unknown language does not abort a running Python/R install).

Reproduced on unfixed `main`: `notebook-env:provision` with `'julia'` resolved successfully and completed a `provision` operation whose language was `'julia'`. After this change the same call is rejected and neither `provisionPython` nor `provisionR` runs.

## Proposed change

Use the existing Zod / application-command contract helper. No new validation framework.

- Add `notebookLanguageSchema` and `parseNotebookLanguage()` in `src/shared/notebook.ts`.
- Fail-closed parse at `createNotebookEnvironmentLifecycle` **before** the Python/R dispatch (covers Electron `env-ipc`, local Web application commands, and any other caller).
- Attach application-command contracts on `notebook-env:provision`, `repair`, and `cancel` so Web rejects with `invalid-command-arguments` before the owner runs.
- Keep Electron on `env-ipc.ts`. Invalid language still fails, via the lifecycle.

### What this does **not** change

- **Data fields:** none
- **Data model:** none
- **Persistence / migrations:** none
- **New enum values:** none (still `'python' | 'r'`)
- **Historical compatibility:** none required. Valid renderer traffic is unchanged. Invalid language that previously installed Python now fails; that is the bugfix.
- **User interaction (valid path):** none
- **Architecture / Electron envelope:** not migrated onto catalog `RUNTIME_VALIDATED`. That flag would double-register the same `ipcMainHandle` channels `env-ipc.ts` already owns, or require retiring those handlers and switching Electron to `{ ok, result }` envelopes. Called out as follow-up.

## Scope and non-goals

**In this PR:** the reproduced Notebook install/repair/cancel language fail-open.

**Not in this PR (over-engineering as one change):**

- Marking `notebookEnv.*` as `RUNTIME_VALIDATED` and retiring overlapping `env-ipc` invoke handlers (Electron envelope cutover).
- Present-but-invalid `notebook:execute` language (omitted language still means Python by design; garbage should fail closed in a follow-up).
- File, credential, Provider, Connector, Skill, Compute, and remote-access IPC. Those do not share this ternary. Several already confine paths or require a desktop caller. Each needs its own reproduction.

- `'julia'` (and any non-`python`/`r` value) must not call `provisionPython` or `provisionR`.
- `'python'` and `'r'` still provision/repair as before.
- Local Web application commands reject unknown language with `invalid-command-arguments`.
- Remote Web mutations remain rejected before the lifecycle (unchanged).

### Test Impact Set (ran after the last material edit)

| --- | --- | --- |
| Unknown language no longer installs Python (IPC + owner) | `npm test -- src/main/notebook/env-ipc.test.ts src/main/notebook/environment-lifecycle-workflows.test.ts src/main/notebook/application-commands.test.ts src/shared/notebook.test.ts` | pass (52 tests). On unfixed `main` the three new owner/IPC cases failed because `'julia'` resolved and completed a provision/repair. |
| Application-command inventory still certifies | `npm test -- src/main/application-command-composition.test.ts src/main/application-command-router.test.ts` | pass (25 tests) |
| Lint of touched files | `npx eslint --max-warnings 0` on the 7 changed files | pass |

Not run: full `npm test` (PR Gate / CI is authoritative). No renderer UI change, so no Web typecheck or browser pass. No persistence/migration. No ACP protocol change.

**Consumers / platform lanes:** Electron and local Web share the lifecycle parse. Remote Web cannot call these mutations today. Windows path behavior is not involved.

**Uncovered risks:** `kernel-executor` still maps a present-but-invalid execute language to Python (`language === 'r' ? 'r' : 'python'`). Agent/MCP install already uses `z.enum(['python', 'r'])`. Internal `=== 'r' ? r : python` helpers remain behind typed callers.

## Review focus

- Fail-closed at the owner vs. a full catalog `RUNTIME_VALIDATED` Electron cutover.
- Whether execute-language validation should be a stacked follow-up PR rather than this one.

### DIFF
--- a/src/main/notebook/application-commands.test.ts
+++ b/src/main/notebook/application-commands.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from 'vitest'
 
+import { ApplicationCommandError } from '../../shared/application-command-contract'
 import {
   createApplicationCommandRouter,
   type ApplicationInvocation
@@ -331,6 +332,35 @@ describe('Notebook application commands', () => {
       )
     ).rejects.toThrow('Channel only available from the local app: notebook-env:cancel')
     expect(status).toHaveBeenCalledOnce()
+    expect(provision).not.toHaveBeenCalled()
+    expect(repair).not.toHaveBeenCalled()
+    expect(cancel).not.toHaveBeenCalled()
+  })
+
+  it('rejects an unknown Environment language before the lifecycle runs', async () => {
+    const provision = vi.fn()
+    const repair = vi.fn()
+    const cancel = vi.fn()
+    const router = createApplicationCommandRouter()
+    installNotebookEnvironmentApplicationCommands(router.registrar, {
+      status: vi.fn(),
+      provision,
+      repair,
+      cancel,
+      startup: vi.fn()
+    })
+
+    const error = await router.dispatcher
+      .invoke(notebookEnvironmentProvisionCommand, invocation(['julia'] as never))
+      .catch((caught: unknown) => caught)
+    expect(error).toBeInstanceOf(ApplicationCommandError)
+    expect(error).toMatchObject({ code: 'invalid-command-arguments' })
+    await expect(
+      router.dispatcher.invoke(notebookEnvironmentRepairCommand, invocation(['julia'] as never))
+    ).rejects.toMatchObject({ code: 'invalid-command-arguments' })
+    await expect(
+      router.dispatcher.invoke(notebookEnvironmentCancelCommand, invocation(['julia'] as never))
+    ).rejects.toMatchObject({ code: 'invalid-command-arguments' })
     expect(provision).not.toHaveBeenCalled()
     expect(repair).not.toHaveBeenCalled()
     expect(cancel).not.toHaveBeenCalled()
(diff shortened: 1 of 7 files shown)


ewen-poch:
**JSON-null optional arguments:** agreed. Web RPC `JSON.stringify` turns omitted optional slots into `null` (`cancel(undefined)` → `[null]`, `provision(lang, undefined)` → `[lang, null]`). The first contract treated those as invalid languages/ids.

`bd883a69` now drops trailing JSON `null` before the Environment command schemas and treats `null` as omitted on optional cancel. Required language remains fail-closed: `[null]` still cannot provision/repair, and `'julia'` still cannot install Python.

ewen-poch:
**Latest Codex review (mergeable):** no further code changes. The JSON-null optional-argument case is covered in `bd883a69`; remaining Environment language validation stays fail-closed for unknown non-null values.

PR #1570, opened 2026-08-22
fix(notebook): keep the R namespace after cancellation

ewen-poch:
## Problem

Cancelling a running R cell returns `cancelled`, then the next cell fails. The existing real-R test `cancels a run without clearing the persistent R namespace` reproduces this.

The external report guessed a late SIGINT into the next request. Instrumentation on unfixed `main` showed a different, stable failure:

1. Save `preserved_after_cancel <- 41`.
2. Dispatch `Sys.sleep(30)` and cancel as soon as the framed request is written.
3. Cancel returns `cancelled`.
4. The R process exits with code 1 and stderr `Execution halted`.
5. The next `execute()` respawns a new process.
6. `cat(preserved_after_cancel + 1)` fails with `object 'preserved_after_cancel' not found`.

SIGINT is sent as soon as the request is written. The loop is often still blocked in `read_request()` / `readLines("stdin")`, **outside** the eval `tryCatch(..., interrupt=)`. A top-level interrupt in non-interactive Rscript is fatal. The #1275 probe never runs because the process is already gone.

When SIGINT instead lands inside `Sys.sleep`, the existing eval handler and probe usually work. Waiting ~50ms for `pending` therefore made the original test flake; aborting in the same turn as the stdin write makes the read-path crash stable.

```mermaid
sequenceDiagram
  participant Driver
  participant Loop as R loop
  Driver->>Loop: framed Sys.sleep request
  Driver->>Loop: SIGINT (same turn)
  Note over Loop: still in read_request()
  Loop-->>Driver: exit 1, Execution halted
  Driver->>Loop: respawn on next execute
  Loop-->>Driver: object not found
```

## Proposed change

Keep the persistent R process alive across cancellation SIGINT. Internal loop change only.

- Retry empty / interrupted stdin reads instead of treating them as EOF.
- Catch interrupts around the whole request cycle, including `run()` setup.
- If SIGINT arrived while reading the next framed request, acknowledge that request as `interrupted` with the existing `interrupt_ack` field and do not drop the process.
- Keep the existing executor probe for the late-SIGINT-after-response race.

No executor public API change. No new protocol field.

### What this does **not** change

- **Data fields:** none
- **Data model:** none
- **Persistence / migrations:** none
- **New enum values:** none (`cancelled` stays `cancelled`)
- **Historical compatibility:** none required. Valid cells and successful cancellations keep the same results. The previously fatal read-path interrupt now preserves the kernel instead of respawning empty.
- **User interaction:** none. Cancel still returns `cancelled`.
- **Architecture:** no larger executor state machine. #1275 already holds the queue until interrupt ack or probe; that path is unreachable when R dies first.

## Scope and non-goals

**In this PR:** the reproduced R cancel-then-next-cell namespace loss, plus a dispatch-time regression and a 50-iteration cancel stress test on the public `execute()` seam.

**Not in this PR (over-engineering as one change):**

- A new executor state machine for “original response / SIGINT consumed / probe complete”. #1275 already owns that late-SIGINT path.
- New wire fields or execution statuses.
- Windows recoverable cancel. Node maps SIGINT to `TerminateProcess`; the existing POSIX skip and Windows drop path stay as-is.

- Fast cancel of `Sys.sleep(30)` returns `cancelled` and leaves the same R process alive.
- The next cell `cat(preserved_after_cancel + 1)` returns `42`.
- An in-eval cancel and 50 subsequent dispatch cancels still preserve the namespace.
- Existing fake-loop interrupt / probe tests still pass.

On unfixed `main`, the tightened real-R test failed 5/5 with `object 'preserved_after_cancel' not found`. After this change it passed 5/5.

### Test Impact Set (ran after the last material edit)

| --- | --- | --- |
| Cancel then next R cell keeps the namespace | `npx vitest run src/main/notebook/kernel-executor.test.ts -t 'cancels a run without clearing the persistent R namespace'` | pass 5/5. On unfixed code the same command failed 5/5 with `object 'preserved_after_cancel' not found`. |
| In-eval cancel + 50 dispatch cancels | `npx vitest run src/main/notebook/kernel-executor.test.ts -t 'preserves the R namespace across an in-eval cancel'` | pass |
| Owner kernel-executor file | `npx vitest run src/main/notebook/kernel-executor.test.ts` | 73 passed / 1 skipped |
| Real `r_loop.R` protocol, including immediate SIGINT | `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=... npx vitest run src/main/notebook/r-loop.integration.test.ts` | 43 passed |
| Lint of touched TypeScript tests | `npx eslint --no-error-on-unmatched-pattern src/main/notebook/kernel-executor.test.ts src/main/notebook/r-loop.integration.test.ts` | pass |

Not run: full `npm test` (PR Gate / CI is authoritative). No renderer UI change, so no Web typecheck or browser pass. No persistence/migration. No ACP protocol change.

**Consumers / platform lanes:** the R loop script is shared by Electron and headless Web notebook execution. macOS `sandbox-exec` spawn was checked with a direct SIGINT against the wrapped Rscript; the process survived and the follow-up cell printed `42`. Linux uses the same `r_loop.R` without sandbox-exec. Windows cancellation still cannot preserve this namespace.

**Uncovered risks:** Linux CI skips the real-R tests when R is not installed. User code that catches interrupts still depends on the existing #1275 probe. The fake loop cannot exercise `Execution halted`; that path is covered by the real-R tests and `r_loop.R` integration.

## Review focus

- Treating interrupted / empty stdin reads as retryable rather than EOF, without hanging shutdown when stdin is actually closed.
- Skipping eval for a request that was interrupted during read, versus waiting for `Sys.sleep` to see a leftover interrupt flag.
- Whether a larger executor state machine is still warranted after this loop-level fix.

### DIFF
--- a/resources/notebook/r_loop.R
+++ b/resources/notebook/r_loop.R
@@ -629,15 +629,54 @@ capture_environment <- function() {
   )
 }
 
+# SIGINT during a blocked stdin read looks like an empty read and, if left uncaught at top level,
+# prints "Execution halted" and exits the persistent kernel. Remember it so the request we were
+# about to read can be acknowledged without dropping the namespace.
+interrupt_during_read <- FALSE
+
 # Reads one request off the length-prefixed protocol; returns list(req_id, code) or NULL at EOF.
 read_request <- function() {
-  header <- readLines(con, n = 1L, warn = FALSE)
-  if (length(header) == 0L) return(NULL)
-  parts <- strsplit(header, " ", fixed = TRUE)[[1]]
-  req_id <- parts[1]
-  n <- as.integer(parts[2])
-  code <- if (n > 0L) readChar(con, n, useBytes = TRUE) else ""
-  list(req_id = req_id, code = code)
+  empty_streak <- 0L
+  repeat {
+    header <- tryCatch(
+      readLines(con, n = 1L, warn = FALSE),
+      interrupt = function(cnd) {
+        interrupt_during_read <<- TRUE
+        character()
+      }
+    )
+    if (length(header) == 0L) {
+      empty_streak <- empty_streak + 1L
+      if (!isOpen(con, rw = "read") || empty_streak >= 2L) return(NULL)
+      interrupt_during_read <<- TRUE
+      next
+    }
+    parts <- strsplit(header, " ", fixed = TRUE)[[1]]
+    req_id <- parts[1]
+    n <- as.integer(parts[2])
+    if (is.na(n) || n < 0L) n <- 0L
+    acc <- raw()
+    body_empty <- 0L
+    while (length(acc) < n) {
+      chunk <- tryCatch(
+        readBin(con, what = "raw", n = n - length(acc)),
+        interrupt = function(cnd) {
+          interrupt_during_read <<- TRUE
+          raw()
+        }
+      )
+      if (length(chunk) == 0L) {
+        body_empty <- body_empty + 1L
+        if (!isOpen(con, rw = "read") || body_empty >= 2L) return(NULL)
+        interrupt_during_read <<- TRUE
+        next
+      }
+      body_empty <- 0L
+      acc <- c(acc, chunk)
+    }
+    code <- if (n > 0L) rawToChar(acc, multiple = FALSE) else ""
+    return(list(req_id = req_id, code = code))
+  }
 }
 
 # User code may manage its own output sinks, but it must never pop the kernel-owned capture sink and
@@ -1294,10 +1333,59 @@ run <- base::local({
 ))
 lockEnvironment(environment(run), bindings = TRUE)
 
+interrupted_response <- function(req_id) {
+  environment <- tryCatch(
+    capture_environment(),
+    interrupt = function(cnd) {
+      list(
+        runtime_version = paste(R.version$major, R.version$minor, sep = "."),
+        packages = list()
+      )
+    },
+    error = function(cnd) {
+      list(
+        runtime_version = paste(R.version$major, R.version$minor, sep = "."),
+        packages = list()
+      )
+    }
+  )
+  list(
+    stdout = "",
+    stderr = "",
+    error = "interrupted",
+    interrupt_ack = TRUE,
+    error_line = NULL,
+    result = NA,
+    cwd = getwd(),
+    figures = list(),
+    output_truncated = FALSE,
+    environment = environment,
+    req_id = req_id
+  )
+}
+
 repeat {
-  req <- read_request()
+  req <- tryCatch(
+    read_request(),
+    interrupt = function(cnd) {
+      interrupt_during_read <<- TRUE
+      "retry"
+    }
+  )
   if (is.null(req)) break
-  resp <- run(req)
-  resp$req_id <- req$req_id
+  if (identical(req, "retry")) next
+  resp <- tryCatch(
+    {
+      if (isTRUE(interrupt_during_read)) {
+        interrupt_during_read <<- FALSE
+        interrupted_response(req$req_id)
+      } else {
+        result <- run(req)
+        result$req_id <- req$req_id
+        result
+      }
+    },
+    interrupt = function(cnd) interrupted_response(req$req_id)
+  )
   emit(resp)
 }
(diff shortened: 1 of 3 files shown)


ewen-poch:
**Keep interrupt state out of the user namespace** — adopted in 89a23576.

The read-path interrupt flag and request reader now live in a `local({ ... })` environment, not `.GlobalEnv`. A cell that assigns `interrupt_during_read <- TRUE` (or `during_read <- TRUE`) is just another user binding and no longer skips the next request.

Regression coverage:
- `does not let user R code poison cancellation interrupt state` in `kernel-executor.test.ts`
- `does not skip the next request when user code assigns interrupt_during_read` in `r-loop.integration.test.ts`

Cancel still preserves the namespace after this move (fast dispatch cancel, in-eval cancel, and the 50-cancel stress test).

ewen-poch:
**Do not discard bytes consumed before an interrupted body read** — adopted in 800b906b (test in 103698c4).

An interrupted `readBin`/`readLines` is no longer counted as EOF. The reader keeps draining the declared length-prefixed frame while `state$during_read` is set, then acknowledges the request. Two empty reads still mean a closed stdin.

R does not return the partial vector from an interrupted `readBin`, so those in-flight bytes cannot be reconstructed. The next `readBin` continues the same frame so leftover bytes stay in this request instead of being parsed as the next one.

Coverage: `keeps the protocol aligned when SIGINT arrives in the middle of a framed body`. Fast cancel, in-eval cancel, and the 50-cancel stress test still preserve the namespace.

ewen-poch:
**Do not turn EOF into a permanent read loop** and **protect the cancellation response from user-global bindings** — adopted in 56b12605.

Empty stdin reads are EOF only. The interrupt handler uses a distinct sentinel (`NA_character_` / `attr(..., "interrupt")`) and is the only path that sets `during_read`. Two empty reads still return NULL.

The request loop now runs in `list2env(..., parent = baseenv())` with `con`, `emit`, `run`, and `capture_environment` injected. Cancellation helpers such as `getwd()` resolve from base, so a cell that assigns `getwd <- function(...) 1` cannot crash the ack path.

Coverage: `does not let user R code poison cancellation interrupt state` now also shadows `getwd` and then cancels. Fast cancel, split-body SIGINT, and the 50-cancel stress test still pass.

ewen-poch:
**Use a valid connection mode in the empty-read check** and **mark transient empty reads as interrupted** — adopted in 325b83b0.

`isOpen(con)` is used with the default `rw`. Empty header/body reads increment the EOF streak first (two empties still return NULL, so a closed pipe cannot spin) and then set `during_read` because SIGINT on a blocked read often returns empty without raising `interrupt`. The next completed frame is acknowledged as interrupted instead of being evaluated.

Fast cancel, getwd-shadow cancel, split-body SIGINT, and the 50-cancel stress test still pass.

ewen-poch:
**Do not restart frame parsing after an interrupt** — not adopted on 325b83b0.

`readLines`/`readBin` already catch interrupts while a frame is in progress and keep draining that frame. The outer `"retry"` only runs when an interrupt escapes *before* `read_request()` has consumed header bytes (the idle wait between cells). Removing that retry made the original regression fail again: fast cancel of `Sys.sleep(30)` returned `cancelled`, then `cat(preserved_after_cancel + 1)` failed with `object 'preserved_after_cancel' not found` because the loop died instead of recovering.

Keeping the idle-wait retry preserves the namespace. A parser-state machine that resumes a half-consumed frame across an *uncaught* interrupt would be a larger change than this loop patch; inner I/O handlers are the path that actually sees in-frame SIGINT.

Head remains 325b83b0.

ewen-poch:
**Do not restart frame parsing after an interrupt** — still not adopted on 325b83b0.

Tried the recommended resumable parser (frame header/body in `state`, resume after SIGINT). Single-cell fast cancel and the split-body test passed, but the 50-cancel stress test started failing again with `object 'preserved_after_cancel' not found` (kernel respawned). Consecutive empty reads from cancel SIGINT were treated as EOF or otherwise tore down the loop. Reverted that approach.

In the current reader, `readLines`/`readBin` already catch interrupts while a frame is in progress and keep the same header/`acc`. The outer `"retry"` is the idle-wait path (no header consumed yet). Removing it reproduced the original namespace loss.

Happy to take a different design if we should keep iterating; I am not going to land a parser change that reintroduces the original P1.

PR #1579, opened 2026-08-23
fix(session): keep an unsaved local title across a newer remote upsert

ewen-poch:
## Problem

Client A can rename a Session in the renderer store before the ordered saver reaches disk. Client B's higher-revision `session:updated` still arrived through `upsertPersistedSession`, and `mergeNewerPersistedSessionByIdentity` started from `structuredClone(incoming)` for scalars.

A's unsaved title was replaced in the UI even though durable Session JSON already uses `expectedRevision` CAS. After the upsert, the session was marked externally hydrated, so the saver also skipped writing A's title.

```mermaid
sequenceDiagram
  participant A as Client A store
  participant S as Ordered saver
  participant M as Main JSON CAS
  participant B as Client B
  A->>A: renameSession Local draft
  B->>A: session:updated revision N+1 title Remote
  Note over A: upsert used to take Remote title
  A->>S: skipped as externally hydrated
  Note over M: disk still at B's revision, A's title never saved
```

This is renderer projection, not disk last-write-wins. Same product shape as `fix(settings): keep in-flight optimistic preferences across onChanged`: keep the local in-flight field, apply the rest of the snapshot.

## Proposed change

Keep the unsaved local title until the local save is acknowledged.

- `renameSession` sets a transient `unsavedTitle` flag that `toPersistedSession` strips.
- `upsertPersistedSession` still identity-merges the incoming snapshot (messages, graph, `pinned`, `permissionProfile`, …), then overlays only `title` when that flag is set.
- The ordered saver still acknowledges the remote revision, but if store title differs from the externally hydrated authority it enqueues a write with `conflictRebaseFields: ['title']`.
- A successful durable save acknowledgement clears the flag. A remote title still replaces local state when A did not rename.

Main `expectedRevision` CAS is unchanged.

## Scope and non-goals

**In this PR:** fence `upsertPersistedSession` for an unsaved local title only, and still persist that title after a newer remote projection.

**Not in this PR:**

- `pinned`, `permissionProfile`, `autoReviewEnabled`, and `agentConfiguration` remain incoming-wins
- Rename-dialog draft before confirm
- Overlapping `sendWorkspaceMessage`
- Settings `onChanged` (already shipped)
- R kernel cancel
- Main CAS / rebase policy
- Schema, migrations, or i18n

- After `renameSession` to `Local draft`, a higher-revision upsert with title `Remote title` keeps `Local draft`, still merges remote messages, and still applies remote `pinned` / `permissionProfile`.
- `unsavedTitle` is not present on `toPersistedSession`.
- Without a local rename, a higher-revision remote title still replaces the store title.
- After rename then remote upsert, the ordered saver writes `{ title: 'Local draft', revision: 2 }` with `conflictRebaseFields: ['title']`, then clears `unsavedTitle`.
- Existing revision-conflict / CAS tests still pass.

These checks ran after the last material edit:

|---|---|---|
| Unsaved title vs newer upsert | `npx vitest run src/renderer/src/stores/session-store.test.ts -t 'unsaved local title\|newer remote title'` | pass |
| Saver still writes title | `npx vitest run src/renderer/src/lib/session-persistence/session-persistence.test.ts -t 'unsaved local title'` | pass |
| Session store + architecture | `npx vitest run src/renderer/src/stores/session-store.test.ts src/renderer/src/stores/session-store.architecture.test.ts src/renderer/src/lib/session-persistence/session-persistence.test.ts` | pass (215) |
| Module | `npm run test:module -- session_renderer` | pass (466) |
| Durable CAS / lifecycle | `npx vitest run src/main/session-persistence/coordinator.test.ts src/main/session-persistence/coordinator-contract.test.ts src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/hooks/useLifecycleSync.test.tsx` | pass |
| Web typecheck | `npm run typecheck:web` | pass |
| Lint | `npm run lint` | pass |

**ACP:** renderer projection and saver only. `claude-code` / `opencode` / `codex-response` / `codex-bridge` excluded as Main mutex and Session JSON CAS already shared; no adapter change.

**Excluded:** live model output, E2E, packaging, database migrations, Web API map. No renderer contract catalog change. No visual UI component change, so no localhost Web walkthrough.

**Uncovered risk:** Main-owned `applyDurableSessionProjection` paths that replace persisted fields (for example continuation echo) still apply their snapshot title. That is not the other-client `upsertPersistedSession` seam.

## Review focus

- Whether overlaying title after identity merge in `upsertPersistedSession` is the right owner, versus teaching the merge helper itself.
- That `unsavedTitle` stays transient and never reaches Session JSON.
- That the saver still acknowledges the remote revision and only extra-writes when the store title differs from that authority.

### DIFF
--- a/src/renderer/src/lib/session-persistence/session-persistence.test.ts
+++ b/src/renderer/src/lib/session-persistence/session-persistence.test.ts
@@ -1564,6 +1564,53 @@ describe('renderer session persistence bridge', () => {
     expect(saveSession.mock.calls[1][0].messages).toEqual(latest.messages)
   })
 
+  it('still saves an unsaved local title after a newer remote Session projection', async () => {
+    const persisted = materializeSessionConversationGraph(
+      createPersistedSession({
+        projectId: 'project-a',
+        revision: 1,
+        title: 'Original'
+      })
+    )
+    const remote = materializeSessionConversationGraph({
+      ...persisted,
+      revision: 2,
+      title: 'Remote title',
+      messages: [
+        {
+          id: 'remote-message',
+          role: 'agent',
+          content: 'Saved in another window',
+          status: 'complete',
+          eventIds: [],
+          createdAt: persisted.updatedAt + 1,
+          updatedAt: persisted.updatedAt + 1
+        }
+      ],
+      updatedAt: persisted.updatedAt + 10
+    })
+    const saveSession = vi
+      .fn<SessionPersistenceApi['saveSession']>()
+      .mockImplementation(async (submitted) => ({
+        ...submitted,
+        revision: (submitted.revision ?? 0) + 1
+      }))
+    const api = createApi({ saveSession })
+    useSessionStore.getState().hydrateSessions([persisted])
+    const save = createStoreSaver(api, useSessionStore.getState())
+
+    useSessionStore.getState().renameSession('session-1', 'Local draft')
+    useSessionStore.getState().upsertPersistedSession(remote)
+    await save(useSessionStore.getState())
+
+    expect(useSessionStore.getState().sessions[0].title).toBe('Local draft')
+    expect(useSessionStore.getState().sessions[0].unsavedTitle).toBeUndefined()
+    expect(saveSession).toHaveBeenCalledWith(
+      expect.objectContaining({ title: 'Local draft', revision: 2 }),
+      { conflictRebaseFields: ['title'] }
+    )
+  })
+
   it('reports an earlier failed write even when a later queued write succeeds', async () => {
     const api = createApi({
       saveSession: vi.fn().mockRejectedValue(new Error('disk full')),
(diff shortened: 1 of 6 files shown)


ewen-poch:
**P1 — Clear the title-dirty marker after an acknowledged save**

Agreed. `replace-persisted-if-current` only dropped `unsavedTitle` when `current === source`. If another store update replaced the Session object during the write, the advanced-session merge kept the flag, so a later remote rename could still be fenced.

Addressed in 09d435c9:

- `applyDurableSessionProjection` now clears `unsavedTitle` when the durable title matches the live title, including `current !== source`.
- A later `renameSession` during the write still keeps the flag, so that newer local title is not discarded.
- Regression: save ack after `togglePinned` advances the object, then a later remote title applies; a second local rename during the write is retained.

ewen-poch:
**P2 — No-op renames leave the title marked unsaved**

Agreed. Confirming the rename dialog with the same trimmed title set `unsavedTitle` without a real edit, so a later remote title could be fenced.

Addressed in a7815830:

- `renameSession` is a no-op unless the trimmed title actually changes.
- `upsertPersistedSession` keeps `unsavedTitle` only when the live title still differs from the incoming snapshot.
- Regression: whitespace-only confirm does not mark dirty; a matching remote title clears the marker so a later remote rename applies.

PR #1588, opened 2026-08-23
feat(settings): add feedback entry

ewen-poch:
## Summary

- add a single-word `Feedback` entry above `Archived` in the Settings navigation
- open the GitHub feedback issue form directly, keeping the app free of an additional form or dialog
- broaden the existing feature-request form to cover ideas, improvements, and general feedback
- localize the new entry in all supported renderer locales and test its placement and destination

## Testing

- `npx vitest run src/renderer/src/pages/settings/SettingsPage.render.test.tsx src/renderer/src/i18n/resources.test.ts` (675 tests passed)

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,47 +1,34 @@
-name: Feature request
-description: Suggest an idea or improvement for Open Science
-title: '[Feature]: '
+name: Feedback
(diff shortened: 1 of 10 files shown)


ewen-poch:
Addressed the mobile settings drawer overflow. The navigation now uses vertical scrolling with overscroll containment on narrow screens, while restoring the existing non-scrolling layout at the md breakpoint and above. Added layout assertions for both responsive classes; the targeted Settings and i18n suites pass (675 tests).

PR #1589, opened 2026-08-23
fix(workspace): surface resume failures on queued Send now

ewen-poch:
## Problem

Send now on a queued message with an attachment could leave the queue row in a send error whose detail was `The queued message was not admitted.` even when the Session had already failed with `Agent session resume failed: reply was never sent`. A second Send now retried the same drain and looked like it stopped immediately.

This happens after Send now: native follow-up is refused or never replies, the live turn ends, drain calls `sendMessage`, resume `failRun`s the Session, and `sendMessage` still returns `undefined`. Error Sessions are sendable, so the queue mapped that `undefined` to a generic admission miss.

## Proposed change

When queued admission returns `undefined` against a Session that is already in `error` with an `error` string, use that Session error as the queue send-error detail. Idle admission misses without a Session error still use `The queued message was not admitted.`

No new fields, enums, or persistence. Queue remains in-memory. The only interaction change is the queue error copy matching the resume banner.

## Scope and non-goals

- Does not fix the underlying `acp:resume-session` invoke that never replies.
- Does not add Codex native follow-up / steering for attachments.

- After Send now of a queued attachment, when the live turn ends during inject and resume failRuns the Session, the queue row shows the resume error instead of a generic admission miss.
- A second Send now keeps that resume error and does not cancel the run.
- The same mapping holds when native follow-up never replies and the Session is already errored.
- Generic idle admission misses are unchanged.

### Validation (after the last material edit)

- queued Send now + resume failure copy -> `npm test -- src/renderer/src/pages/workspace/workspace-message-queue-controller.test.ts` -> pass (32)
- queue UI + conversation consumers -> `npm test -- src/renderer/src/pages/workspace/workspace-message-queue-controller.test.ts src/renderer/src/pages/workspace/ComposerMessageQueue.interaction.test.tsx src/renderer/src/pages/workspace/workspace-conversation-controller.test.ts` -> pass (69)
- renderer types -> `npm run typecheck:web` -> pass
- changed files -> `npx eslint src/renderer/src/pages/workspace/workspace-message-queue-controller.ts src/renderer/src/pages/workspace/workspace-message-queue-controller.test.ts` -> pass

Excluded `test:module -- workspace_page` and the full portable suite locally; PR Gate remains authoritative. No Electron-only or persistence/migration risk in this diff.

## Review focus

- The `undefined` + errored-Session branch in `workspace-message-queue-controller.ts`.
- The two Send now attachment regressions; they failed on unfixed code with `The queued message was not admitted.`

### DIFF
--- a/src/renderer/src/pages/workspace/workspace-message-queue-controller.test.ts
+++ b/src/renderer/src/pages/workspace/workspace-message-queue-controller.test.ts
@@ -541,6 +541,179 @@ describe('workspace message queue controller', () => {
     )
   })
 
+  it('surfaces a resume failure after Send now of a queued attachment when the live turn ends during inject', async () => {
+    const resumeError = 'Agent session resume failed: reply was never sent'
+    const attachment = {
+      id: 'upload-1',
+      sessionId: 'session-a',
+      name: 'data.csv',
+      originalName: 'data.csv',
+      path: '/tmp/data.csv',
+      mimeType: 'text/csv',
+      size: 12
+    }
+    let currentSession = session()
+    let notifySessionChanged: (() => void) | undefined
+    let finishSteer!: (result: { injected: false; reason: 'not-advertised' }) => void
+    const sendMessage = vi.fn(async () => {
+      currentSession = { ...session('error'), error: resumeError, errorReportable: true }
+      return undefined
+    })
+    const steerFollowUp = vi.fn(
+      () =>
+        new Promise<{ injected: false; reason: 'not-advertised' }>((resolve) => {
+          finishSteer = resolve
+        })
+    )
+    const input = options(currentSession, {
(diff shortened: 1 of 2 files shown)


ewen-poch:
Adopted the P2 on stale Session errors.

Queued Send now now snapshots `status`, `error`, and `updatedAt` before `sendMessage`. The Session error is copied onto the queue row only when this admission produced a new error transition (status became `error`, the error string changed, or `updatedAt` advanced, which `failRun` always does). If `sendMessage` returns `undefined` without changing that snapshot — for example a later admission miss while an older resume error is still on the Session — the queue keeps `The queued message was not admitted.`

`26098a81` adds that comparison and a regression for the stale-error case.

PR #1598, opened 2026-08-23
fix(startup): separate database checks from composition delays

ewen-poch:
## Problem

Windows launches that sat on **Checking database…** for 1–4 minutes had already verified SQLite (typically 0.2–0.4s). The remaining time was runtime composition (`createApplicationModules` / `registerIpcHandlers`). The startup shell kept the database-checking copy until `complete()`, so the hang looked like a database failure. Cold vs warm starts on the same machine also matched slow disk or antivirus first-touch more than a stuck ledger.

## Proposed change

- After database verification, publish `starting` so the shell shows **Starting Open Science…** until runtime composition finishes.
- Record CPU vs wall time on `application-startup` and `application-composition` (`phaseWaitMs`, `delayKind`: `cpu` / `io-or-wait` / `mixed`).
- Run a bounded, path-free storage probe at process start (`typical` / `likely-slow-disk` / `slow-disk-or-scanner`).
- Split composition into greppable phases (`specialist-catalog`, `builtin-specialists`, `marketplace-recover`, `acp-runtime`, `load-application-modules`, and others).

## Scope and non-goals

- Diagnostics and honest startup copy only.
- Does not shorten composition, defer Notebook/Agent work, or change database migrations.
- Does not log filesystem paths.

- After SQLite verification, the gate is no longer stuck on the database-checking copy.
- Support logs can tell a long `io-or-wait` phase from CPU-bound work, and can attribute time to a named composition phase.
- Storage probe results contain no absolute paths.

### Test impact set

Ran after the last material edit in `.worktree/startup-starting-phase`:

| --- | --- | --- |
| Starting phase after verification | `vitest run src/main/database/database-startup-owner.test.ts src/renderer/src/components/database-startup-gate.test.tsx` | pass |
| Locale catalogs for the new copy | `vitest run src/renderer/src/i18n/resources.test.ts` | pass |
| CPU vs wait classification | `vitest run src/main/diagnostics/startup-delay.test.ts src/main/diagnostics/operation.test.ts` | pass |
| Storage probe timings, cleanup, timeout | `vitest run src/main/diagnostics/startup-storage-probe.test.ts` | pass |
| Diagnostic privacy canary | `vitest run src/main/diagnostics/privacy-canary.test.ts` | pass |
| Startup module load order | `vitest run src/main/index-startup-order.test.ts` | pass |
| Main-process types | `tsc --noEmit -p tsconfig.node.json --composite false` | pass |

ACP frameworks, session branches, and renderer contract catalogs are unchanged. No Windows packaged cold-start lane was run locally.

## Review focus

- `starting` must not mark the app `ready` before IPC composition finishes.
- `delayKind` / `phaseWaitMs` should stay off default diagnostics unless `cpuUsage` is passed.
- Storage probe must never block startup beyond the 1.5s cap or leak paths into `main.log`.

## Uncovered risks

- Packaged Windows cold start (HDD + Defender) is the motivating environment and was not re-run with this build.
- Probe thresholds are coarse heuristics, not a hardware classifier.

### DIFF
--- a/src/main/database/database-startup-owner.test.ts
+++ b/src/main/database/database-startup-owner.test.ts
@@ -22,10 +22,7 @@ describe('database startup owner', () => {
     finish?.()
     await attempt
 
-    expect(owner.getState()).toEqual({
-      phase: 'migrating',
-      migrationId: '0001_runtime_schema_baseline'
-    })
+    expect(owner.getState()).toEqual({ phase: 'starting' })
     expect(owner.isMigrating()).toBe(false)
   })
 
@@ -58,15 +55,9 @@ describe('database startup owner', () => {
       migrationId: '0001_runtime_schema_baseline'
     })
 
-    await expect(owner.retry()).resolves.toEqual({
-      phase: 'migrating',
-      migrationId: '0001_runtime_schema_baseline'
-    })
+    await expect(owner.retry()).resolves.toEqual({ phase: 'starting' })
     await expect(owner.whenVerified()).resolves.toBeUndefined()
-    expect(owner.getState()).toEqual({
-      phase: 'migrating',
-      migrationId: '0001_runtime_schema_baseline'
-    })
+    expect(owner.getState()).toEqual({ phase: 'starting' })
 
     owner.complete()
     expect(owner.getState()).toEqual({ phase: 'ready' })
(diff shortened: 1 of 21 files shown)


ewen-poch:
Thanks for the review.

**Timeout does not bound the storage probe.** Adopted. `Promise.race` only stopped waiting; a stalled `fsync` could keep a handle open. The probe now runs in an isolated `worker_threads` worker (self-contained CJS, not the packaged main bundle). On timeout we `terminate()` the worker and unlink the probe file. The in-process `probeStartupStorage` helper remains for unit tests of the timing logic.

**Composition diagnostics report success too early.** Adopted. `application-composition` now starts in `registerIpcHandlers`, continues through module construction, records `ipc-adapters` after Electron adapter installation, and only then `complete()`. Construction or adapter failures call `fail()` on the same operation.

Pushed on `feat/database-startup-starting-phase`.

ewen-poch:
Thanks for the review.

**Startup probe overwrites and deletes any existing file at its fixed probe path.** Adopted. Each probe now creates an exclusive unique file (`wx` plus a UUID suffix), passes that exact path into the worker, and unlinks it only after this invocation successfully created it. A leftover `.open-science-startup-probe` in the logs directory is left untouched.

Pushed on `feat/database-startup-starting-phase`.

ewen-poch:
Thanks for the review.

**Do not gate app startup on the storage probe.** Adopted. The probe is diagnostic only, so it no longer runs on the path to failure capture, bootstrap assets, or the first window. It starts in the background after the log sink is ready and writes `startup storage probe` whenever it settles (still capped at 1.5s, still isolated). Completing it as a later diagnostic phase would have rewritten the sequential startup timeline, so the result stays on that log line.

Pushed on `feat/database-startup-starting-phase`.

ewen-poch:
Thanks for the review.

**Clean up partially created probe files.** Adopted. Exclusive creation now uses `open(..., 'wx')` and sets `created` as soon as that open succeeds, so a later write/close failure still unlinks this invocation’s file.

**Keep timeout coverage through worker cleanup.** Adopted. The worker posts its result only after unlink, and the parent always `terminate()`s after the race so a stuck cleanup cannot outlive the 1.5s bound. Generated unique paths are removed on terminate; a caller-supplied existing path is left untouched.

Pushed on `feat/database-startup-starting-phase`.

ewen-poch:
Thanks for the review.

**Diagnostic worker can delay fatal startup exit.** Adopted. The isolated probe worker is `unref()`d after create so a background fsync cannot keep Electron alive after a fatal startup failure. Generated probe files are still removed on `terminate()` when the process remains up long enough to join the worker.

Pushed on `feat/database-startup-starting-phase`.

PR #1600, opened 2026-08-23
feat(specialists): govern Marketplace-installed packages

ewen-poch:
## Problem

Marketplace-installed Specialists were represented like ordinary custom/imported profiles. Their publisher-owned content could be edited or overwritten by a manual ZIP, and Installed did not distinguish Custom, Marketplace, and Built-in ownership. Update, disable, copy, and uninstall actions also lacked one coherent lifecycle.

## Proposed change

- Persist an additive `marketplace` Specialist origin and make publisher-owned content read-only while retaining icon, color, and enabled controls.
- Route Marketplace package transactions through privileged provenance, preserve local presentation on Marketplace updates, and block ordinary ZIP overwrite.
- Require changed package content to use a higher SemVer when an exact package-content baseline exists; package version remains excluded from content identity.
- Extend Installed grouping to All / Custom / Marketplace / Built-in, with a mobile-safe category selector.
- Add a readable managed-package detail with quiet update availability, Create editable copy, separate enable/disable, and tailored uninstall confirmation.
- Remove completed/pending Marketplace provenance after uninstall and backfill historical `imported` records only when archive provenance matches exactly and imported content is still unmodified.

The interaction follows the separation used by [VS Code extension management](https://code.visualstudio.com/docs/configure/extensions/extension-marketplace) (update, disable, and uninstall are distinct), the explicit removal confirmation used by [Chrome extensions](https://support.google.com/chrome_webstore/answer/2664769), and [Apple alert guidance](https://developer.apple.com/design/human-interface-guidelines/alerts) by keeping update status contextual while reserving the modal for destructive uninstall.

## Scope and non-goals

- `specialists.json` remains document version `2`; there is no schema-version upgrade.
- No persisted grouping field or grouping enum is added; groups are derived from `kind` plus `origin`.
- No automatic-update policy or broader package release workflow is introduced.
- Legacy records without the newer package-content digest fall back to their historical payload-hash format, so same/lower-version changes to name, display name, description, or instructions are still blocked. The older hash never covered capability references or bundled Skill contents, so those historical records cannot prove changes in those fields until the next successful install writes the complete digest.
- Older app versions do not understand the new `marketplace` origin and may open the Specialist document in degraded/read-only mode; downgrade compatibility is not claimed.

All listed checks ran after the final material source edit.

- Marketplace authorization, package/version identity, transaction presentation preservation, exact-provenance recovery, uninstall cleanup, IPC contracts, stores, Settings interactions, and translations → targeted Vitest impact set (15 files) → **903 tests passed**.
- Node and renderer interface compatibility → `npm run typecheck` → **passed**.
- Repository lint contract → `npm run lint` → **passed**.
- Formatting/patch integrity → Prettier on changed files and `git diff --check` → **passed**.

Per maintainer direction, the full local `npm test` suite was not run. PR CI remains authoritative for the complete portable and platform lanes. Live Marketplace network/update journeys and packaged cross-platform rendering were not exercised locally.

## Review focus

- The main-process boundary that permits only icon/color updates for Marketplace-managed profiles while keeping enable/disable, duplicate, and uninstall available.
- Exact-digest historical backfill and provenance cleanup after uninstall.
- Version monotonicity behavior for changed content, including the format-matched legacy payload-hash fallback.
- The managed detail hierarchy and narrow-width behavior of the four-way grouping, update reminder, and destructive uninstall flow.

## Data model notes

- **Historical compatibility:** exact Marketplace archive provenance upgrades `origin: imported` to `origin: marketplace` only when imported content remains unmodified; ambiguous or locally edited history is left editable. Legacy import baselines retain their historical payload-hash comparison for core profile content; their old format cannot detect capability/bundled-Skill-only changes until refreshed. `specialists.json` stays at version `2`.
- **New enum value:** `SpecialistOrigin` adds `marketplace`. No other state or grouping enum is added.
- **Persistence:** `origin: marketplace` is stored in `specialists.json`; Marketplace provenance continues in `specialist-marketplace.json` and is removed on uninstall. Publisher/source/version enrichment remains derived for renderer use.


ewen-poch:
Addressed the actionable AI review points:

- Historical Marketplace recovery now promotes an imported Specialist only when the installed archive digest matches exactly and `modifiedSinceImport === false`; locally edited or ambiguous historical installs remain editable.
- Full-access capability summaries label `excludedSkillIds` and `excludedConnectorIds` as exclusions instead of presenting them as granted capabilities.
- Marketplace-managed capability content is explicitly protected across Skill and Connector attach/detach operations, while appearance, enablement, duplication, and uninstall remain available.
- Marketplace install/update and Specialist uninstall now share one operation coordinator, preventing an in-flight installation from recreating the profile or provenance after uninstall.
- Legacy import baselines without `packageContentDigest` now use the historical payload-hash format for same-version and downgrade governance. This safely covers the fields represented by that format; capability-only and bundled-Skill-only changes become fully detectable after the next successful install records the complete package content digest.

Validation on the current head: 15 affected test files / 906 tests passed locally, along with typecheck and lint. The current AI review completed successfully without publishing additional feedback, and the PR gate is green.

PR #1602, opened 2026-08-23
fix(network): refresh stale registry reachability on window focus

ewen-poch:
## Problem

Network reachability is probed only on startup, the browser `online` event, and manual Retry. There is no background polling. When the local link stays up (`navigator.onLine === true`) but proxy, DNS, or egress recovers, the header can remain **Unreachable** until the user clicks Check again.

Two related gaps:

- Settings → Network loads local interface details with `getInfo()` and ignores rejection, so a bridge failure looks like “no IP / no type” instead of an error.
- The probe is an HTTPS HEAD to **npmjs** and **npmmirror**, not generic internet. If both registries are blocked, the UI still said “Internet unreachable” even when other services work.

## Proposed change

- Silently re-probe when the window is focused or the document becomes visible, **only** while the last result is `unreachable` or `probe-failed`. Same-turn `focus` + `visibilitychange` are coalesced. No timer / no polling.
- Catch `getInfo()` rejection in the Network panel and show “Could not load local network details.”
- Relabel the failing probe as package-registry reachability (zh: 软件源) instead of adding a second generic connectivity check.

## Scope and non-goals

- No new IPC methods, store fields, or `NetworkConnectivity` enum values.
- No persisted network state and no database/settings migration.
- No periodic TTL timer while the window stays focused (that would reintroduce polling). Users who never leave the window can still click Check again.
- No independent “is the internet up?” probe (e.g. generate_204). The measured object remains npmjs / npmmirror.

## Compatibility / persistence / enums

- **Historical data:** none. Nothing is stored.
- **New enum values:** none. `NetworkConnectivity` is unchanged (`unknown` | `reachable` | `unreachable` | `probe-failed`).
- **Persistence:** none. `networkInfoError` is local React state in the Network panel only.
- **Interaction:** the Unreachable pill can clear without a click when the user returns to the window; the Network panel shows an explicit local-info error instead of empty detail.

- After an unreachable/probe-failed result, focusing the window or making the document visible silently re-probes and updates the header/store.
- Focus does not re-probe while already reachable or while the link is down.
- `getInfo()` rejection shows “Could not load local network details.”
- Unreachable copy refers to package registries, not generic internet.
- Probe targets remain npmjs and npmmirror.

Regression tests were written against unfixed `origin/main` first. Failures matched the report (connectivity stayed `unreachable` after `focus`; Unreachable pill remained; `getInfo()` rejection had no error copy; UI still said “Internet unreachable”). The same tests pass after the fix.

## Review focus

- Is event-driven silent re-probe (focus / visibility) enough, vs a TTL timer while focused?
- Copy-only honesty vs a future generic internet probe.

## Test impact

Checks below ran after the last material edit.

| --- | --- | --- |
| Silent re-probe on focus / visibility; no probe while reachable or offline | `npx vitest run src/renderer/src/stores/network-store.test.ts` | pass |
| Header Unreachable pill clears after a successful focus re-probe; new copy | `npx vitest run src/renderer/src/components/NetworkStatusIndicator.render.test.tsx` | pass |
| `getInfo()` rejection shows an explicit error; Retry still restores last known connectivity | `npx vitest run src/renderer/src/pages/settings/NetworkPanel.render.test.tsx` | pass |
| Probe is npmjs + npmmirror only | `npx vitest run src/main/net/network-info.test.ts` | pass |
| i18n catalog keys, placeholders, script purity | `npx vitest run src/renderer/src/i18n/resources.test.ts` | pass |
| Renderer / main types | `npm run typecheck:web`, `npm run typecheck:node` | pass |
| Lint on changed files | `npx eslint --no-cache` on the edited TS/TSX files | pass |

Not run locally: full `npm test`. These files are not a dedicated `test:module` owner; PR Gate remains the complete portable/platform authority.

Uncovered risks: a window that stays focused the entire time a proxy recovers still needs Check again; Web has no `checkConnectivity` bridge and still treats `navigator.onLine` as reachable.


ewen-poch:
Codex P2 is addressed in 3898b391: a silent focus/visibility re-probe that is already in flight is coalesced, so a later generation cannot discard an earlier success. The previous Static checks failure was `npm ci` / node-gyp (`assert(!this.paused)` in undici) while rebuilding `safe-file-publisher-native`, not a lint or typecheck finding on this change.

ewen-poch:
Addressed the follow-up Codex P2s in the latest commit:
- `getInfo()` now applies success/error only for the latest request, so a stale rejection cannot clear newer interface details.
- Passed, probe-failed, and section copy now say package registries, matching the npmjs/npmmirror probe.

PR #1605, opened 2026-08-23
feat(workspace): preview notebook figures in tool groups

ewen-poch:
## Problem

Notebook Python/R executions already persist captured display images, but message tool groups only exposed those figures inside the tool detail panel. Collapsing an individual tool therefore hid its visual result, and the tool row did not summarize figure and output-line counts.

## Proposed change

- Render captured notebook figures as siblings of their tool row under the owning tool group, so row collapse keeps the result visible while group collapse still hides it.
- Show localized metadata such as `1 figure · 2 lines of output` and `1 figure · done`.
- Preserve failed, timed-out, interrupted, and cancelled terminal status ahead of successful-run output-line metadata.
- Count legacy visible output fields once, excluding the `text.plain` projection that mirrors stdout/stderr.
- Add a compact, left-aligned, content-sized transcript figure card. Images keep their aspect ratio with a 300px height cap and a container-width cap instead of using the full message width.
- Show a restrained bottom hover gradient with the saved visual file's basename when the persisted run gives an unambiguous one-to-one association, otherwise a localized `Figure N.ext` fallback, plus a diagonal expand-arrow affordance.
- Gate each card's base64 `img` and decoder work by the existing viewport-overscan observer; offscreen cards retain only a lightweight placeholder and unload the data URL after leaving the overscan region.
- Open figures in a dedicated single-image Radix dialog with Escape and close-button support, separate from artifact provenance previews.
- Restore keyboard focus to the exact figure card that opened the dialog after either Escape or close-button dismissal.
- Anchor a tool group to its nearest current viewport position before toggling its height, so expanding grows downward without re-entering bottom-follow mode and collapsing does not snap the transcript to the bottom.
- Hydrate historical figure-bearing Notebook runs while their collapsed tool rows are near the viewport. Targeted IPC reads stay batched at 20 runs; active near-viewport runs are protected from the 20-record inactive historical LRU until they leave the viewport.
- Add all renderer locale entries and focused hierarchy, metadata, dialog, Escape, and i18n coverage.

## Scope and non-goals

- Notebook capture and kernel execution are unchanged.
- Artifact writing, artifact provenance, and working-file discovery are unchanged.
- No dependency, domain enum, persistence schema, migration, or new persisted state is introduced.
- Historical captured `outputs[].display` images render through the existing stored payload. Targeted IPC batches retain the existing 20-run limit, while the inactive historical LRU remains bounded at 20 records; the renderer may temporarily retain more than 20 records when more than 20 figure rows are simultaneously near the viewport.
- Existing persisted `workingFiles[].relativePath` supplies the display filename only when a run has exactly one captured figure and one saved visual file with a MIME-compatible extension. Multiple or incompatible candidates are not associated safely, so those runs retain the stable localized fallback such as `Figure 1.png`.

- Figure hierarchy and size constraints, metadata, MIME-compatible saved-filename association and fallback, normalized output-line deduplication, dedicated preview, Escape/close-button focus restoration, group-collapse ownership and viewport anchoring, terminal-status precedence, collapsed historical-run hydration (including 21 simultaneously near historical runs), hook-order stability, and offscreen data-URL mount/unmount behavior -> focused Vitest run covering `WorkspaceMessageScroller`, `WorkspaceToolDetailsRow`, `WorkspaceActivityGroup`, `useNotebookRunsById`, `NotebookRunOutputs`, and i18n resources -> 7 files / 741 tests passed.
- Locale catalog completeness and plural contracts -> `npm run test:module -- i18n_catalog` -> 1 file / 590 tests passed.
- Renderer type safety -> `tsc --noEmit -p tsconfig.web.json --composite false` -> passed.
- Source and test linting -> `npm run lint` -> passed.
- Reference sizing contract -> render regression asserts a content-sized/max-width card, `max-h-[300px] w-auto max-w-full` image treatment, no fixed 52rem desktop width, compact hover gradient, and diagonal expand icon.

The exact base/head impact explanation fails closed to the full Vitest plan because the changed locale path has no recognized module owner. Per request, the full local `npm test` was not run; exact-head PR Gate is authoritative for the complete plan.

## Review focus

- Confirm the result hierarchy matches the intended collapse behavior.
- Confirm group expansion/collapse preserves the group header's viewport context rather than snapping to the transcript bottom.
- Confirm batched near-viewport hydration keeps all currently visible historical figures stable without coupling them to detail-row expansion.
- Confirm per-card viewport gating keeps offscreen base64 payloads out of the DOM and image decoder.
- Confirm both dialog dismissal paths return focus to the activating figure card.
- Confirm output-line counting reflects visible structured notebook output and excludes image payloads.
- Confirm non-completed terminal status remains visible when a figure also has text output.
- Confirm the persisted saved filename is used only for an unambiguous one-figure/one-file run and that multi-candidate historical runs retain the synthetic fallback.


ewen-poch:
Addressed both review points in `827e6d01`:

- Figure filename inference now requires an unambiguous one-capture/one-file match **and** a MIME-compatible extension (including `jpg`/`jpeg` and `tif`/`tiff` aliases). A captured PNG with an unrelated saved PDF now keeps the synthetic filename.
- Output-line metadata now counts the same normalized projection shown by tool details: canonical stdout/stderr/traceback streams, duplicate traceback suppression, and echoed display text removal when that text is already present in a stream.

Added regression coverage for incompatible single-file associations, duplicate stream/display text, and duplicate stderr/traceback content. Focused validation: 7 files / 741 tests passed; renderer typecheck, lint, and diff checks passed.

PR #1613, opened 2026-08-23
fix(workspace): keep live Send now without hung resume or interrupt

ewen-poch:
## Problem

On latest `main`, **Send now** during a live Codex native Responses turn does not inject the queued follow-up. Four stacked failures show up as one user-visible break:

1. `#1552` resumes an attached Session whenever `agentConfiguration` is present, even when the live target is unchanged. That resume hangs while the cancelled native Responses stream is still aborting, and the queue surfaces `Agent session resume failed: reply was never sent`.
2. After the message appears, cancelled-turn finalization still awaits automatic compaction. Compaction can block behind the aborting stream, so `acp:send-prompt` never replies.
3. `#1593` changed native follow-up from fail-closed to interrupt fallback. When `_session/steering` is refused or times out, Send now cancels the live turn (`This turn was interrupted`). The Sending/Stopping wait is `cancelRun` waiting for the in-flight Responses stream to abort.
4. Even after a natural `end_turn`, queued drain still invoked `acp:send-prompt` while the previous turn held root admission for compact and after-release work. That IPC never replied. A later composer send, including after switching the agent framework, produced no ACP logs because admission and `promptInFlight` never cleared.

`#1566` already specified inject-or-wait: keep the current run and send after it finishes. `#1593` replaced that with inject-or-cancel.

## Proposed change

- Resume an attached Session only when the explicit target differs (provider, model, or effort), the runtime is detached, or the framework/backend changed. Same-target Send now goes to `sendPrompt` instead of `session/resume`.
- Skip automatic context compaction while finalizing a live prompt. Mark the turn idle and release root admission at stop, then compact after the prompt interaction is gone. A new user prompt preempts an in-flight automatic compact.
- Restore fail-closed Send now: try native follow-up first; if inject is unavailable, refused, or throws while the turn is still live, keep the run and send the queued item after it becomes idle. Stop remains the explicit cancel control.

Provider/model/effort adoption from `#1552` is unchanged.

## Scope and non-goals

- Renderer prompt preparation, message-queue drain, cancelled-prompt finalization, and post-stop admission.
- Does not make Codex native Responses `_session/steering` inject during an in-flight stream. If inject still refuses, the follow-up waits until the live run idles instead of cancelling it.
- Does not change persisted Session JSON, enums, or migrations.
- Does not add a live Codex native-responses steering harness.

- **Historical data:** none. Existing Sessions without `agentConfiguration` still resume when an explicit target is first supplied, or when detached.
- **New enum values:** none.
- **Persistence:** none. No Session file, database, or migration changes.

- Same-target attached send must not resume → `npx vitest run src/renderer/src/lib/acp/useWorkspaceAgentRuntime.test.ts` → 182 passed
- Effort-only and model-only target changes still resume → same file → passed
- Cancelled and `end_turn` stops must not wait on hanging compaction → `npx vitest run src/main/acp/prompt-outcome-finalizer.test.ts` → 15 passed
- Follow-up `startPrompt` after `end_turn` must not wait for post-stop work (claude-code, opencode, codex) → `npx vitest run src/main/acp/runtime-coordinator.test.ts` → 111 passed
- Idle automatic compact and no-threshold Codex skip → `npx vitest run src/main/acp/context-compaction-workflow.test.ts` → 11 passed
- Send now must not `cancelRun` when native follow-up is unavailable or refused → `npx vitest run src/renderer/src/pages/workspace/workspace-message-queue-controller.test.ts` → 35 passed
- Prompt-turn after-release compactIfIdle wiring → `npx vitest run src/main/acp/prompt-turn-workflow.test.ts` → 17 passed

Exact-head CI remains authoritative for the complete portable suite.

## Test impact (ACP)

- Affected behavior: when Send now injects, resumes, or drains after a live turn; when the next `startPrompt` is admitted after stop.
- Included: shared renderer send path for `claude-code`, `opencode`, and `codex` (covers Codex Responses and Codex Bridge at this seam). Queue drain is framework-agnostic. Coordinator admission release is shared.
- Excluded: live `_session/steering` against a real native Responses stream.

| behavior | framework/path | command | result |
| --- | --- | --- | --- |
| same-target attached send does not resume | claude-code / opencode / codex renderer send | vitest `useWorkspaceAgentRuntime.test.ts` | passed |
| effort/model change still resumes | codex renderer send | vitest effort-only and model-only cases | passed |
| stop does not await compact | shared ACP finalizer | vitest `prompt-outcome-finalizer.test.ts` | passed |
| follow-up startPrompt after end_turn | claude-code / opencode / codex coordinator | vitest `runtime-coordinator.test.ts` | passed |
| Send now does not interrupt when inject fails | shared queue controller | vitest `workspace-message-queue-controller.test.ts` | passed |

## Review focus

- Same-target resume skip vs effort/model adoption from `#1552`.
- Fail-closed requeue after refused native follow-up, including the race where the live turn ends during inject.
- Admission release at `onPromptEnded` must not allow overlapping user prompts; compaction is preempted, not concurrent with a new user turn.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -601,10 +601,10 @@ colors communicate a successful or failed probe/migration result.
   Up/Arrow Down keyboard reordering, Edit, Remove, and Send now. Edit moves an item back into an
   unchanged empty composer. Send now promotes the item and first tries to inject it into the current
   run through the agent framework's native follow-up, without cancelling that run. While inject is in
-  flight, the row shows a sending state. If inject is unavailable or refused, Send now stops the live
-  turn and dispatches the promoted item as soon as the Session becomes sendable; the row shows a
-  stopping state while cancellation is in flight. Stop remains the explicit control for cancelling a
-  live turn without sending a queued message. Branch, admission, cancellation, or edit failures keep
+  flight, the row shows a sending state. If inject is unavailable or refused, Send now keeps the live
+  turn and sends the promoted item after that run finishes; the row returns to queued. Stop remains
+  the explicit control for cancelling a live turn without sending a queued message. Branch, admission,
+  cancellation, or edit failures keep
   the row and show a recoverable inline error.
 - Queued messages are transient and Session-scoped. They are not persisted across renderer restart.
   Bind each item to its admission Message Branch and block branch switching or inline message edits
--- a/src/main/acp/application-commands.test.ts
+++ b/src/main/acp/application-commands.test.ts
@@ -494,17 +494,60 @@ describe('ACP application commands', () => {
         invocation([{ sessionId: 'session-1', reason: 'manual' }])
       )
     ).rejects.toThrow('Restore this archived Session before continuing.')
-    await expect(
-      router.dispatcher.invoke(
-        acpCommands.steerFollowUp,
-        invocation([{ sessionId: 'session-1', text: 'focus on tests' }])
-      )
-    ).rejects.toThrow('Restore this archived Session before continuing.')
-
-    expect(admittedById).toHaveBeenCalledTimes(3)
+    expect(admittedById).toHaveBeenCalledTimes(2)
     expect(admittedById).toHaveBeenCalledWith(request.sessionId)
     expect(dependencies.runtime.resetSessionContext).not.toHaveBeenCalled()
     expect(dependencies.runtime.compactSession).not.toHaveBeenCalled()
+  })
+
+  it('does not nest Session admission around the coordinator follow-up guard', async () => {
+    let archiveQueue: Promise<void> = Promise.resolve()
+    const enqueueArchive = <Result>(operation: () => Promise<Result>): Promise<Result> => {
+      const result = archiveQueue.then(operation, operation)
+      archiveQueue = result.then(
+        () => undefined,
+        () => undefined
+      )
+      return result
+    }
+    const base = createDependencies()
+    const withSessionAvailableById = <Result>(
+      _sessionId: string,
+      operation: () => Promise<Result>
+    ): Promise<Result> => enqueueArchive(operation)
+    const dependencies: AcpApplicationCommandDependencies = {
+      ...base,
+      runtime: {
+        ...base.runtime,
+        steerFollowUp: vi.fn(() =>
+          withSessionAvailableById('session-1', async () => ({
+            injected: false as const,
+            reason: 'prompt-required' as const
+          }))
+        )
+      },
+      archiveAvailability: {
+        withSessionAvailable: async <Result>(
+          _projectId: string,
+          _sessionId: string,
+          operation: () => Promise<Result>
+        ): Promise<Result> => operation(),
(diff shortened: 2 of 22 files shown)


ewen-poch:
Follow-up from the new logs: Send now now appends the queued message, then a few seconds later Electron reports `Error invoking remote method 'acp:send-prompt': reply was never sent`. There is no second `prompt start`.

Cause: after the live turn is cancelled (steering/interrupt aborting native Responses), finalization still waited on automatic native compaction. That compact prompt can block on the aborting stream, so the follow-up `acp:send-prompt` admission never completes.

`fix(acp): skip compact after cancelled Send now turns` skips auto-compact on `stopReason: cancelled`. Completed turns still compact.

ewen-poch:
The new transcript shows Send now taking the **interrupt** path (`This turn was interrupted`) instead of injecting into the live PubMed turn.

That came from #1593 (today): when native follow-up is refused or times out, Send now started cancelling the live run. Before that, inject was fail-closed — keep the current run and send after it finishes.

On Codex native Responses, `_session/steering` often does not inject. The 8s wait is that inject attempt (row loading). #1593 then cancelled the turn, which aborted the second Responses stream and left `acp:send-prompt` without a reply.

`fix(message-queue): keep live turn when Send now cannot inject` restores fail-closed. Send now still tries native follow-up first; if inject is refused, the PubMed turn continues and the queued follow-up sends when that run becomes idle.

ewen-poch:
Live reproduction and lifecycle follow-up completed in `2d837f64`.

Root cause of the unresolved `acp:send-prompt` IPC was a non-reentrant archive admission cycle. Both the Electron `acp:steer-follow-up` adapter and the application-command adapter wrapped `runtime.steerFollowUp()` in session availability, while the coordinator already acquires that same global archive queue through its prompt admission guard. The outer queue entry waited for the coordinator, and the coordinator's nested entry waited behind the outer entry, so the IPC invoke never settled.

I reproduced this against the production Electron bundle with a real Codex native Responses stream: start a long Codex turn, queue a follow-up, click **Send now**, and wait for the first `end_turn`. Before the fix, the queued message failed with `Agent session resume failed: reply was never sent` and no second prompt started. After removing the duplicate adapter admission, the same flow logged native follow-up injection and the second turn settled.

This commit also closes the two lifecycle findings from the prior review:

- Provider-native `/compact` is now cancelled and fully drained before a prompt reservation activates; automatic compaction also respects pending prompt reservations.
- Coordinator snapshots now identify visible Sessions that are still owned by a retiring runtime, so an unchanged selected target resumes instead of dispatching into a retired owner.

Validation:

- 412 focused ACP/renderer tests pass.
- The three affected real-runtime lifecycle tests pass.
- `npm run typecheck` passes.
- `npm run build:e2e` passes.
- The unsandboxed full suite completed with 18,640 passing tests. Its only three failures are unrelated existing environment/schema checks: Prisma runtime DDL/catalog drift and installed-runtime fallback selection.

PR #1618, opened 2026-08-23
perf(session-persistence): index Session metadata in SQLite

ewen-poch:
## Problem

Startup currently loads every `session.json` to build Session navigation, Project counts, title search, and Usage analytics. Transcript growth therefore increases startup I/O and JSON parsing even when the user opens only one Session.

## Proposed change

Keep live Session JSON as transcript/runtime authority and duplicate query metadata plus per-turn Usage into SQLite. Healthy startup reads SQLite summaries and loads a Session JSON only when that Session is opened or exported.

The design follows [Microsoft's Materialized View pattern](https://learn.microsoft.com/en-us/azure/architecture/patterns/materialized-view), [CQRS guidance](https://learn.microsoft.com/en-us/azure/architecture/patterns/cqrs), [SQLite transaction/FK guidance](https://sqlite.org/foreignkeys.html), and the concrete [OpenAI Codex Thread Store](https://github.com/openai/codex/blob/c9b19deb09c1841ce7acc33ddb96276030936a29/codex-rs/thread-store/README.md) precedent. The principal entity remains `Session`; no speculative `Catalog`-suffixed domain is introduced.

```mermaid
erDiagram
  Project ||--o{ Session : "owns; hard delete restricted"
  Session ||--o{ SessionTurnUsage : contains
  Session ||--o{ SessionRun : contains
  Session ||--o{ SessionArtifactRef : contains

  Project {
    String id PK
    String name
    String description
    String agentContext
    Boolean isExample
    Boolean pinned
    DateTime archivedAt "nullable"
    DateTime deletedAt "nullable; soft delete"
    DateTime createdAt
    DateTime updatedAt
  }

  SessionNumberSequence {
    String id PK "fixed: global"
    Int nextNumber "next unallocated number"
  }

  Session {
    String id PK "existing Session identity"
    Int number UK "global sequence; starts at 1"
    String projectId FK
    String title
    String status
    String presentedStatus
    Boolean pinned
    BigInt archivedAtMs "nullable"
    BigInt revision
    Int activeMessageCount
    Int artifactCount
    Int filesRevision
    BigInt createdAtMs
    BigInt updatedAtMs
    BigInt presentedActivityAtMs "nullable"
    Boolean needsStartupRecovery
    BigInt sourceByteLength "nullable; reserved"
    BigInt sourceMtimeMs "nullable; reserved"
    BigInt deletedAtMs "nullable; individual deletion"
  }

  SessionProjectionState {
    String id PK "fixed: session-projection"
    Int projectionVersion
    DateTime completedAt
  }

  PendingSessionReconciliation {
    String sessionId PK "intentionally no FK"
    String projectId
    String operation "save or delete"
    DateTime markedAt
  }

  SessionTurnUsage {
    String sessionId PK,FK
    String messageId PK
    BigInt completedAtMs
    BigInt inputTokens
    BigInt cacheTokens
    BigInt outputTokens
    Boolean isRootFrame
  }

  SessionRun {
    String sessionId PK,FK
    String messageId PK
    BigInt createdAtMs
  }

  SessionArtifactRef {
    String sessionId PK,FK
    String artifactId PK
    BigInt artifactCreatedAtMs "nullable"
  }
```

## Read, write, and deletion semantics

1. Save commits `PendingSessionReconciliation(operation = save)` and atomically allocates/reuses the global Session `number`; individual deletion first commits `operation = delete`.
2. Save atomically replaces authoritative Session JSON, including duplicated metadata, `turnUsage`, and `number`.
3. Save transactionally replaces the SQLite summary/facts and clears the pending row.
4. Startup repairs pending saves from live JSON or, during Project-deletion recovery, from the committed Project Session tombstone before that temporary JSON authority is removed. Pending deletes resume individual JSON removal. Replay runs behind the global write barrier so a stale repair cannot overwrite a concurrent newer save. Initial/re-versioned projection publication is single-flight.
5. Individually deleting a Session removes its JSON and Usage/run/artifact facts, while retaining the `Session` metadata row and first `deletedAtMs`. Ordinary saves, reconciliation, and projection rebuilds cannot reactivate that tombstone.
6. Deleting a Project removes its live JSON/runtime data but retains the `Project` row, its Session metadata rows, and all facts for Sessions not previously deleted individually. Active lists require `Project.deletedAt IS NULL`; Usage intentionally does not.
7. `Session.projectId -> Project.id` uses `ON DELETE RESTRICT`, preventing an accidental hard Project delete from cascading into retained history.

## Scope and non-goals

- Adds title/name metadata needed by global Session search without transcript hydration.
- Existing Project list/get/update paths hide soft-deleted Projects. Usage reads `createdAt` from every retained Project row, including soft-deleted rows, so total/new Project counts remain historical. This PR preserves data needed for future per-Project/global Usage; it does not add a new statistics UI or pre-aggregated rollup table.
- The Usage panel refresh key includes both Session revisions and the active Project identity/creation catalog, so creating or deleting a Project refreshes retained Project analytics even when no Session changed.
- Projection-only `sessions:list`, `sessions:load-one`, and `sessions:load-usage` are explicitly Electron-only. Web keeps its existing `sessions:load-all` authority path, including the few delegated-work/conflict/export flows that request one Session.
- Opening a lazy Session reconciles its durable Compute Host access through the existing host owner before runtime admission; removed provider IDs are filtered without hydrating every Session JSON.
- No full-text message index, daily Usage rollup, transcript storage, new database dependency, or background service is added.
- Healthy app-managed writes are crash-recoverable. Out-of-band JSON edits are not inventoried on every healthy startup because that would defeat the startup objective.

### Historical compatibility

- Existing Session JSON remains readable; `number` is optional in the historical contract.
- The first upgraded startup performs one full JSON scan and writes missing numbers in deterministic `createdAt ASC, id ASC` order. This startup may remain slow by explicit product decision.
- Existing valid unique numbers are retained. Duplicate/missing values are reassigned; the global sequence continues above every allocated/deleted number, allows gaps, and never reuses a number.
- Existing inactive-branch and delegated-frame Usage semantics remain unchanged.
- Project or Session deletions completed by older versions cannot be reconstructed because those versions did not retain their metadata/facts.

### Enum/state additions

- No Session, Message, run, artifact, deletion-kind, or UI lifecycle enum value is added.
- `PendingSessionReconciliation.operation` adds two persistence-only state values: `save` and `delete`. They distinguish cross-store recovery intent and are not exposed as Session/UI lifecycle states.
- Other new infrastructure state is limited to projection version/readiness, the global number allocator, `Project.deletedAt`, and `Session.deletedAtMs`.

### Persistence changes

- Migration `0013_session_projection` adds seven tables: `SessionNumberSequence`, `Session`, `SessionProjectionState`, `PendingSessionReconciliation`, `SessionTurnUsage`, `SessionRun`, and `SessionArtifactRef`.
- The existing `Project` table gains nullable `deletedAt`.
- `Session.id` is the PK and remains the only Session key used by FKs, paths, IPC, routes, and schedulers. `Session.number` is an unreferenced unique alternate key.
- Live Session JSON and SQLite intentionally duplicate metadata and per-turn Usage; JSON wins live reconciliation after number allocation.
- `projectCreatedAt` is a read DTO value sourced from retained `Project.createdAt`; it adds no database column.
- SQLite is not completely disposable after this change: deleting/resetting it loses the global allocator, deleted Project/Session tombstones, and Project-deleted Usage history. Remaining live Session data can still be rebuilt from JSON.
- `PendingSessionReconciliation` intentionally has no FK so parent cleanup cannot erase unresolved cross-store repair evidence.
- Its non-null `operation` column defaults to `save` and is constrained to `save | delete`; this makes single-Session deletion crash-recoverable without adding another table.

## Validation

- `npm run db:schema:check` — passed
- `npm run check:web-api-map` — passed
- `npm run lint` — passed with no errors (two pre-existing Prettier warnings on unchanged `main` files)
- targeted main/renderer Vitest modules — 24 files / 643 tests passed
- focused command-composition and Web/tombstone regression modules — 10 files / 134 tests passed
- focused macOS Electron Project create/relaunch E2E — passed
- latest Usage/Compute/session-store recovery regression modules — 4 files / 207 tests passed
- focused macOS visual recovery and accessibility recovery E2E — 2 tests passed
- focused Prettier checks and `git diff --check` — passed

Per maintainer direction, the complete local `npm test` suite was not run; exact-head PR CI is authoritative for full and platform lanes.

## Review focus

- Cross-store crash points and pending-row repair ordering.
- Project versus individual Session deletion semantics.
- Deterministic global Session number backfill and monotonic allocation.
- Active-list filtering versus retained historical Usage.
- Healthy startup remaining SQLite-only apart from the selected Session JSON.


ewen-poch:
The latest P0 is a false positive. `sessions:list` and `sessions:load-usage` are direct Electron IPC catalog entries: neither is marked `applicationCommand: runtime-validated`, and neither is declared in a data-content application-command group. Therefore they do not enter `nonWebNames`. Adding them to `ELECTRON_NATIVE_COMMAND_NAMES` would instead make `nativeCommandNames` contain entries absent from `nonWebNames` and cause the inventory mismatch described by the review. This boundary is already exercised explicitly by `src/shared/renderer-contract-catalog.test.ts` (the exact runtime-validated Electron channel list excludes both direct IPC channels) and by `src/main/application-command-composition.test.ts`; the latter passes 10/10 on head `c012911f`.

PR #1624, opened 2026-08-23
feat(side-chat): inject advisories into running main turns

ewen-poch:
## Problem

Side Chat currently receives only the Main transcript captured when it starts. Its `send_message` tool always waits for the next real Main user turn, even when Main is already running and the selected ACP backend can accept a native follow-up.

## Proposed change

- Refresh the bounded active-Branch Main transcript before every Side Chat follow-up.
- Persist each Side Chat relay first, then attempt context-only injection when Main is running.
- Reuse existing native follow-up transports:
  - Claude Code and Codex framework sessions: ACP `_session/steering` with prompt-required idle behavior.
  - OpenCode: authenticated HTTP message admission with `noReply: true`.
  - Codex Responses and Codex bridge backend variants share the Codex framework steering path.
- Suppress ordinary Main user-message publication for Side Chat advisories.
- Commit delivery against the captured live Main prompt only after native admission succeeds.
- Restore the durable relay for the next real Main user turn on refusal, transport failure, permission boundary, missing prompt identity, or turn-change race.

## Scope and non-goals

- No idle Main wake-up and no second prompt interaction.
- No renderer surface or user-visible localization changes.
- No new ACP transport or provider-specific protocol.
- No continuous transcript co-streaming; Side Chat refreshes at its own turn boundaries.

### Historical compatibility

Existing persisted Side Chat relays and sessions remain readable and use the same delivery path. No migration or backfill is required.

### State additions

`SideChatSendMessageResult` adds `status: "injected"` and `delivery: "current-turn"`. Existing `"queued"` / `"next-user-turn"` values remain unchanged.

### Persistence impact

No new tables, fields, record versions, or persisted enum values are introduced. Relays continue to use `PersistedSideChatRelay`; live injection is persist-first and only marks the existing relay delivered after a durable commit against the live prompt message id.

All listed checks ran after the final material implementation edit.

- Live injection, framework routing, refusal/failure durability, prompt-identity and turn-race guards, Side Chat runtime/IPC contracts, and representative ACP consumers:
  - `npm test -- --run src/main/acp/native-follow-up.test.ts src/main/acp/native-follow-up-workflow.test.ts src/main/acp/runtime.test.ts src/main/acp/runtime-coordinator.test.ts src/main/acp/side-chat-relay-owner.test.ts src/main/acp/application-commands.test.ts src/main/acp/ipc.test.ts src/main/side-chat/host-message-mcp-server.test.ts src/main/side-chat/main-prompt-relay.test.ts src/main/side-chat/ipc.test.ts src/main/side-chat/runtime-owner.test.ts`
  - 11 files passed, 827 tests passed.
- Main/shared TypeScript contracts:
  - `npm run typecheck:node`
  - passed.
- Changed-file lint:
  - `npx eslint <14 changed TypeScript files>`
  - passed with no issues.
- Repository lint:
  - `npm run lint`
  - 0 errors; two existing Prettier warnings in unchanged `src/main/acp/application-commands.ts` and `src/main/acp/ipc.ts`.

Per task scope, the complete local `npm test` suite was not run; PR Gate is authoritative for complete portable and platform coverage. No E2E coverage was added because this changes main-process relay admission without renderer behavior.

## Review focus

- The narrow coordinator bypass: Side Chat advisory injection skips normal user-prompt admission while retaining initialization, quit, deletion, live-turn, and permission fences.
- Persist-first/restore-on-failure ordering in `MainPromptSideChatRelay.tryInject`.
- The distinction between context-only advisory delivery and ordinary Main user-message publication.

### DIFF
--- a/src/main/acp/native-follow-up-workflow.test.ts
+++ b/src/main/acp/native-follow-up-workflow.test.ts
@@ -178,6 +178,37 @@ describe('AcpNativeFollowUpWorkflow', () => {
     expect(published).toEqual([])
   })
 
+  it('refuses an advisory when steering starts a new Main turn', async () => {
+    const { workflow } = createWorkflow({
+      request: vi.fn(async () => ({ outcome: 'startedNewTurn' }))
+    })
+
+    await expect(
+      workflow.steerSideChatAdvisory({ sessionId: 'app-1', text: 'Context-only advisory' })
+    ).resolves.toEqual({ injected: false })
+    expect(published).toEqual([])
+  })
+
+  it('refuses an advisory when Main changes turns during steering', async () => {
+    let liveTurnToken = 'turn-1'
+    const { workflow } = createWorkflow({
+      request: vi.fn(async () => {
+        liveTurnToken = 'turn-2'
+        return { outcome: 'injected' }
+      }),
+      livePromptTurn: () => ({
+        turnToken: liveTurnToken,
+        signal: new AbortController().signal,
+        promptMessageId: liveTurnToken === 'turn-1' ? 'prompt-1' : 'prompt-2'
+      })
+    })
+
+    await expect(
+      workflow.steerSideChatAdvisory({ sessionId: 'app-1', text: 'Context-only advisory' })
+    ).resolves.toEqual({ injected: false })
+    expect(published).toEqual([])
+  })
+
   it('treats startedNewTurn as injected because the adapter consumed the prompt', async () => {
     const { workflow } = createWorkflow({
       request: vi.fn(async () => ({ outcome: 'startedNewTurn' }))
--- a/src/main/acp/native-follow-up-workflow.test.ts
+++ b/src/main/acp/native-follow-up-workflow.test.ts
@@ -16,7 +16,8 @@ const createWorkflow = (
     advertised?: boolean
     livePrompt?: boolean | (() => boolean)
     pendingPermission?: boolean | (() => boolean)
-    livePromptTurn?: () => { turnToken: string; signal: AbortSignal } | undefined
+    livePromptTurn?: () =>
+      { turnToken: string; signal: AbortSignal; promptMessageId?: string } | undefined
     frameworkId?: 'claude-code' | 'opencode' | 'codex'
     openCodeHttp?: boolean
     providerSessionId?: string | null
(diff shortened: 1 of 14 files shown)


ewen-poch:
CI follow-up for head `55a1d76`:

- Module tests and coverage, Static checks, Windows core, Windows E2E, CodeQL, CI Integrity, and the second Codex review all pass. The second review verdict is mergeable with no actionable findings.
- PR Gate is blocked only by `macOS build and E2E` → visual regression → `workspace-long-dark.png` (9,327 pixels / 1%). Build, functional journeys, workspace journeys, and accessibility all pass.
- This is reproducible outside this change: PR #1625 (`test/full-suite-stability`), which has no renderer, E2E, or snapshot changes and uses the same base `0eb04fc0`, failed the same snapshot in run 32670052114 with 9,327 pixels on the first attempt.
- The artifact shows fixture/baseline drift: actual includes an earlier deterministic turn and the `e2e-model` selector, while expected omits both. This PR changes no renderer or visual fixture code, so I am not updating the snapshot in this feature branch.

Current PR Gate run: https://github.com/aipoch/open-science/actions/runs/32670501164
Independent matching failure: https://github.com/aipoch/open-science/actions/runs/32670052114

ewen-poch:
Addressed the bounded-claim delivery finding in `55a1d76`.

`tryInject` now reports `current-turn` only when the live bounded claim actually contains the current relay's `messageId`. Older relays may still be injected and committed first, while the current relay correctly remains `queued` for the next Main turn.

Added a regression test covering two oversized relays where the first fills the live preamble and the second must not be reported as injected.

ewen-poch:
Rebased onto `origin/main` at `55796686`, including the SQLite Session metadata projection.

No Side Chat code adjustment was required: the DB projection contains Session summary and usage metadata, while Side Chat still needs the single durable Session transcript loaded on demand. Live delivery state continues to come from the ACP runtime snapshot rather than persisted status.

Validation after the rebase: 827 focused ACP/Side Chat tests passed, and `typecheck:node` passed after regenerating the Prisma Client for the updated schema.

ewen-poch:
Addressed the advisory turn-race finding in `c58468b6`.

Side Chat advisory dispatch now fails closed when ACP steering reports `startedNewTurn`, so that outcome remains accepted only for ordinary user-triggered Send now messages whose prompt was consumed by the provider. After either ACP steering or OpenCode HTTP delivery returns, advisory dispatch also revalidates the captured live turn token before reporting success; a completed or replaced Main turn leaves the relay queued instead of committing it against the old prompt.

Added regression coverage for both `startedNewTurn` and a turn-token change during transport. The focused ACP/Side Chat suite passes 829 tests, along with `typecheck:node` and changed-file ESLint.

ewen-poch:
Confirmed and fixed in fef3c67b. The regression now drives the existing relay and native-follow-up public boundaries: Codex ACP consumes the advisory with startedNewTurn after the captured turn finishes, and the test verifies that the relay is committed once and cannot be claimed again. The workflow still revalidates the captured turn before transport admission, but treats injected/startedNewTurn as authoritative consumption after the provider accepts the prompt; genuine refusal, timeout, and malformed outcomes still restore the relay. Relevant validation: 758/758 focused ACP, Side Chat, and runtime tests passed; node typecheck, changed-file ESLint, and Prettier passed.

PR #1625, opened 2026-08-23
test(vitest): stabilize the full-suite worker pool

ewen-poch:
## Problem

A full `npm test` / `npx vitest run` is not a stable functional gate. After regenerating Prisma Client, a local full run reported 1189 files, 18,880 tests, 6 failures, ~499s. Every failure passed when rerun alone (ACP, Notebook R cancel, Subagent rendering, architecture scan, completion gate).

This is not six independent product-assertion bugs. Under the full concurrent pool, real R/Python loops, TCP servers, huge payloads, and whole-tree AST walks starve ordinary tests past the fixed 15s timeout. Several ACP tests also `Promise.race` success against `setTimeout(..., 50)`, so scheduler delay is treated as deadlock. A separate class is a stale generated Prisma Client: `prisma generate` runs only in `postinstall`, and a shared `node_modules` (or another worktree) can make `node_modules/.prisma/client/schema.prisma` disagree with `prisma/schema.prisma`, producing a mass of Prisma errors instead of the six flakes.

Vitest already defaults `maxWorkers` to `availableParallelism() - 1`. The missing contract is an explicit cap plus a serial group for heavy files so they cannot overlap the parallel unit pool.

## Proposed change

Test-harness only:

- Pin `maxWorkers` to Vitest’s CPU-minus-one bound.
- Run architecture scans in a later one-worker project (`isolate: false`) with a shared in-memory source index.
- Run `*.integration.test.ts`, `*.certification.test.ts`, real kernel, and TCP files in Vitest’s sequential group (`maxWorkers: 1`, `isolate: true`).
- Replace ACP 50ms deadlock races with a direct await (true hangs still fail via `testTimeout`).
- Fail `npm test` pretest when the generated Prisma Client does not match `prisma/schema.prisma`.

## Scope and non-goals

No product architecture, data model, interaction, enum, or persistence change.

| Concern | Change? |
| --- | --- |
| Historical compatibility | None |
| New enum / status values | None |
| Data persistence / Session JSON / SQLite | None |

Not done, as over-design for the reproduced failure:

- Virtual clocks for real R cancel (`abortOnNextStdinWrite` is already a handshake; fake timers cannot replace SIGINT).
- Global `restoreAllMocks` / Zustand reset (high blast radius; file isolate already separates files).
- Auto `prisma generate` in pretest (would mutate a shared install).
- Raising the 15s timeout as the primary fix.

- Harness tests failed on unfixed `vitest.config.ts` (no `maxWorkers`, no serial projects) and on a missing Prisma checker; they pass after the fix.
- A fixture Prisma mismatch throws a fingerprint error, not a Prisma query error. `node scripts/check-prisma-client.mjs` currently fails closed on this machine because the generated client does not match source schema.
- ACP tests no longer contain `setTimeout(..., 50)` deadlock races and still pass.
- Converted full-repo architecture scans use the shared source index.

Checks after the last material edit:

| --- | --- | --- |
| Worker cap, serial projects, ACP 50ms contract | `npx vitest run vitest.config.test.ts` | pass |
| Prisma fingerprint fixtures + pretest wiring | `npx vitest run scripts/check-prisma-client.test.ts` | pass |
| Shared architecture source index | `npx vitest run test/architecture-source-index.test.ts` | pass |
| Converted architecture scans + ACP files | `npx vitest run` of the touched ACP and architecture test files | pass (180) |
| `scripts.test` remains the complete portable suite | `npx vitest run scripts/ci/module-test-impact.test.ts` | pass |
| Lint of changed files | `npx eslint --cache` on the changed paths | pass |

`vitest.config.ts` and `package.json` are global PR Gate inputs, so exact-head CI (full unit shards) is authoritative for the complete portable suite. Full local `npm test` was not run.

Uncovered risks: CI uses `npx vitest run` after `npm ci`, so pretest does not run there; `npm ci` already generates a matching client. A live `npm test` on a machine whose shared `node_modules` was generated from another worktree’s schema will fail closed at pretest until `npx prisma generate` is run for that tree.

## Review focus

- Vitest `projects` grouping: architecture (`groupOrder: 1`, `isolate: false`) and process (`maxWorkers: 1`, isolate true → sequential group after the parallel pool).
- Whether the process glob list (`*.integration.test.ts`, kernel, TCP) is the right serial set.
- Prisma fingerprint compares the generated `schema.prisma` copy rather than hashing models only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "scripts": {
     "format": "prettier --write .",
     "lint": "eslint --cache .",
+    "pretest": "node scripts/check-prisma-client.mjs",
     "test": "vitest run",
     "test:module": "node scripts/ci/module-test-impact.mjs module",
     "test:affected": "node scripts/ci/module-test-impact.mjs affected",
--- a/scripts/check-prisma-client.mjs
+++ b/scripts/check-prisma-client.mjs
@@ -18,6 +18,48 @@ export function prismaClientFingerprintMismatchMessage(schemaPath, clientSchemaP
   ].join('\n')
 }
 
+// Prisma generate copies the schema into the Client package after realigning field columns.
+// Compare tokens, not bytes: comments and alignment spaces are not part of the datamodel.
+export function prismaSchemaFingerprint(schema) {
+  let out = ''
+  let index = 0
+  const source = String(schema).replaceAll('\r\n', '\n').replaceAll('\r', '\n')
+  while (index < source.length) {
+    const char = source[index]
+    if (char === '/' && source[index + 1] === '/') {
+      while (index < source.length && source[index] !== '\n') index += 1
+      continue
+    }
+    if (char === '"' || char === "'") {
+      const quote = char
+      out += char
+      index += 1
+      while (index < source.length) {
+        out += source[index]
+        if (source[index] === '\\' && index + 1 < source.length) {
+          out += source[index + 1]
+          index += 2
+          continue
+        }
+        if (source[index] === quote) {
+          index += 1
+          break
+        }
+        index += 1
+      }
+      continue
+    }
+    if (/\s/.test(char)) {
+      if (out.length > 0 && out.at(-1) !== ' ') out += ' '
+      index += 1
+      continue
+    }
+    out += char
+    index += 1
+  }
+  return out.trim()
+}
+
 export function checkPrismaClient({
   root = repositoryRoot,
   schemaPath = resolve(root, PRISMA_SOURCE_SCHEMA_RELATIVE_PATH),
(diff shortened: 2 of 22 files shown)


ewen-poch:
## Review follow-up

### Prisma fingerprint (P1)

Byte equality was wrong. Prisma 6.19.3 `generate` realigns field columns, so a fresh Client never matches the repo schema bytes. The checker now fingerprints the datamodel: strip `//` comments, preserve string literals, collapse remaining whitespace. Isolated 6.19.3 generate output (same models, different alignment) is a passing fixture; an extra field still fails.

No product schema, Session JSON, enum, or UI change.

### Verification

| Check | Result |
| --- | --- |
| `npm run typecheck` | pass |
| `npm run lint` | pass (0 errors; 2 pre-existing prettier warnings in untouched ACP files) |
| Prisma fingerprint tests | pass (including unformatted source + generate-formatted client) |
| Harness / ACP / converted architecture tests | pass |
| `npx vitest run` on this machine | 18,570 pass, 210 skip, **114 fail** — all `deletedAt does not exist` because the shared `node_modules` Client was generated from another worktree’s newer schema. `pretest` fails closed on that mismatch (intentional). CI `npm ci` generates from this branch. |

`npm test` locally cannot go green here without `npx prisma generate` against the shared install, which would desync other worktrees. PR Gate unit shards after `npm ci` are the clean full-suite path.

PR #1631, opened 2026-08-24
perf(session-persistence): complete summary-first Session loading

ewen-poch:
## Problem

Session projection backfill and startup recovery performed an unconditional second full Session JSON scan. Web did not expose the existing summary, single-Session, or Usage projection reads, so healthy Web startup still used sessions:load-all. A rejected Usage projection request was also swallowed and replaced with totals from the currently hydrated Zustand subset.

## Proposed change

- Reuse a complete authority scan during projection build and recovery. Record overlapping Session saves and deletes and reread only those identities; retain a fail-closed full rescan only for overlapping catalog-wide or unreadable mutations.
- Publish sessions:list, sessions:load-one, and sessions:load-usage through the local and remote Web application-command adapter and generated renderer map.
- Show a translated, retryable Usage error when the complete projection cannot be loaded instead of displaying partial totals.

## Scope and non-goals

- Historical compatibility is retained. The first upgrade or projection-version rebuild still scans all Session JSON once and may persist the existing Session number backfill.
- needsStartupRecovery still performs one complete authority scan because it covers cross-session recovery owners. Splitting recovery by persisted reason is not part of this PR.
- A concurrent catalog-wide mutation can still require a second fail-closed scan; ordinary startup and per-Session overlap do not.
- Persistence uses the existing JSON authority and SQLite projection. No table, column, field, migration, or data-model change is introduced.
- No persisted status or enum is added. Usage has component-local request/result state only.
- Interaction changes are limited to summary-first Web startup, on-demand Session content loading, and an explicit Usage retry action.

The behavior checks below ran against the final material state.

- Projection build, recovery, concurrent save publication, and missing-delete publication -> focused 9-file Vitest command -> 682 tests passed.
- Web application commands, renderer contract inventory, generated API shape, IPC owner contract, Usage failure UI, and translations -> same focused command -> 682 tests passed.
- Node process types -> npm run typecheck:node -> passed after the review follow-up.
- Web renderer types -> npm run typecheck:web -> passed.
- Generated Web mapping -> npm run check:web-api-map -> passed.
- Changed-file lint after the review follow-up -> npx eslint for the projection repository and test -> passed.
- Full repository lint before that two-file follow-up -> npm run lint -> 0 errors and 2 pre-existing Prettier warnings in src/main/acp/application-commands.ts and src/main/acp/ipc.ts.

The final impact classifier selects the full PR Gate because docs/design.md has no module owner. Per maintainer direction, the complete npm test suite was not run locally; PR CI is authoritative for the full and platform matrix.

## Review focus

- Correctness of suspended projection-write reconciliation, including already-missing deletes, and its catalog-wide fallback.
- Web exposure of Session projection reads across local and remote caller contexts.
- Failure behavior that prevents incomplete Usage totals from being presented as complete.


ewen-poch:
CI note: the final PR Gate is blocked by two pre-existing architecture completion-gate failures on the branch base, not by files changed in this PR:

- `src/renderer/src/stores/session-store-persistence-owner.ts`: 785 lines vs 780 limit.
- `src/renderer/src/pages/workspace/workspace-session-controller.ts`: 701 lines vs 700 limit.

Both counts and limits are already present at base `c7fb61f5`, and neither file is modified here. Static checks, type checks, Windows core/E2E, macOS build/E2E, CodeQL, CI Integrity, PR policy, and the independent Codex review passed. Concurrent PR Gate runs are failing on the same full-module lane, so I have not mixed an unrelated architecture-threshold change into this PR.

PR #1637, opened 2026-08-24
perf(startup): defer transcript and runtime probes

ewen-poch:
## Problem

App startup waited for work that existing users do not need before entering Home:

- Session bootstrap read the complete lightweight SQLite summary catalog and then opened the manifest-selected Session JSON, even though Home only consumes summary metadata.
- Settings bootstrap waited for runtime preflight, encryption, and npm probes before publishing an already-valid persisted Settings snapshot.

This kept the global `Loading saved conversations` and `Loading settings` gates visible longer than the actual startup data dependencies required.

## Proposed change

- Keep the complete lightweight SQLite Session summary catalog at startup so Home counts, recent Sessions, archived Sessions, search titles, notifications, and deep links retain their current complete view.
- Stop opening the manifest-selected Session transcript during ordinary startup. Transcript JSON remains lazy and opens when Workspace selects it.
- Continue eagerly opening Sessions marked for startup recovery, and preserve explicit selected-Session loading when retrying persistence recovery.
- Publish the persisted Settings snapshot as soon as it resolves, while runtime probes continue in the existing deduplicated load pass.
- Let existing users enter Home during those probes; keep first-run onboarding gated until runtime initialization completes.
- Treat a runtime-probe failure as a runtime availability problem instead of misreporting valid Settings as unreadable.

## Scope / non-goals

- No Session summary pagination or new aggregate/search/by-id backend API in this PR.
- No SQLite schema, migration, projection shape, JSON format, write ordering, or IPC contract changes.
- No new persisted state, enum value, or user-visible copy.
- Legacy/web `loadAll` fallback behavior is unchanged.

Historical compatibility: no new compatibility path is required. Existing SQLite projection rebuild/recovery behavior remains in place, and explicit recovery retry still reopens the selected Session.

## Acceptance / validation

- Ordinary SQLite startup hydrates all summaries without opening the manifest-selected Session JSON.
- Startup-recovery Sessions and explicit recovery retry still hydrate full JSON.
- Existing users render Home after the Settings snapshot while probes continue.
- First-run onboarding remains behind runtime initialization.
- Settings authority failures remain blocking; runtime-probe failures do not discard valid Settings.

Local evidence:

- `npm test -- src/renderer/src/lib/session-persistence/session-persistence.test.ts src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/stores/settings-store.test.ts src/renderer/src/App.test.tsx` — 4 files, 222 tests passed.
- ESLint on the 7 changed renderer files — passed.
- Prettier check on the 7 changed renderer files — passed.
- `npm run test:affected:explain -- --base origin/main --head HEAD` escalates to the complete suite solely because `App.test.tsx` has no module owner. Per the agreed module-focused validation strategy, the complete suite is left to PR CI.

## Review focus

- The distinction between ordinary startup (`preferredSelection` absent) and explicit persistence recovery retry (`preferredSelection` present).
- Whether any existing-user route still requires runtime probe results before Home renders.
- The Settings load generation/deduplication behavior when a forced retry supersedes an in-flight startup load.

### DIFF
--- a/src/renderer/src/App.test.tsx
+++ b/src/renderer/src/App.test.tsx
@@ -830,6 +830,28 @@ describe('App startup routing', () => {
     expect(mocks.syncWindowFindAppearance).toHaveBeenCalledTimes(1)
   })
 
+  it('renders Home while existing-user runtime probes continue', async () => {
+    mocks.settings.isLoaded = true
+    mocks.settings.isLoading = true
+    mocks.startupView = 'app'
+
+    await render()
+
+    expect(container.querySelector('[data-testid="home-page"]')).not.toBeNull()
+    expect(container.querySelector('[data-testid="settings-startup-loading"]')).toBeNull()
+  })
+
+  it('keeps first-run onboarding behind runtime initialization', async () => {
+    mocks.settings.isLoaded = true
+    mocks.settings.isLoading = true
+    mocks.startupView = 'onboarding'
+
+    await render()
+
+    expect(container.querySelector('[data-testid="settings-startup-loading"]')).not.toBeNull()
+    expect(container.querySelector('[data-testid="onboarding-page"]')).toBeNull()
+  })
+
   it('shows a settings load error and retries the complete initialization', async () => {
     mocks.settings.loadError = 'settings IPC unavailable'
     mocks.settings.load.mockReset().mockResolvedValueOnce(false).mockResolvedValueOnce(true)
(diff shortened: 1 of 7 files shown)


ewen-poch:
Thanks for flagging this path. The initial navigation view is always Home, so the manifest-selected Session is not rendered immediately. Every existing route into Workspace goes through `openProject` or `openSession`, and both call `selectSession`. The persistence listener subscribes to the whole Zustand store rather than a selected-id selector, so `set({ selectedSessionId: sameId })` still notifies it and starts transcript hydration even when the id is unchanged.

Loading the selected transcript during bootstrap, or firing the listener immediately after installation, would undo the intended lazy-on-Workspace behavior and perform the JSON read while the user remains on Home. I kept the production behavior and added regression coverage in d8bc3c79 that asserts both sides of the contract: startup performs no `loadOne`, then selecting the already-selected summary triggers `loadOne` with the correct locator.

PR #1650, opened 2026-08-24
feat(notebook): show package installation progress

ewen-poch:
## Summary

- render a dedicated `manage_packages` progress row inside the existing tool activity group
- show requested package count, elapsed time, a several-minute expectation, and an indeterminate progress track
- derive the presentation entirely from existing Renderer activity state so session switches preserve elapsed progress without adding Agent context or installer logs
- keep completed and failed calls on the existing expandable result details
- translate the new UI copy across all six Renderer locales

## Compatibility and data impact

- Historical compatibility: existing persisted activities continue to hydrate unchanged; active rows derive from existing `status`, `rawInput`, and `createdAt` fields
- Enum additions: none
- Persistence changes: none; no schema, database, IPC, ACP event, or tool-result fields were added
- Platform behavior: Renderer-only and installer-agnostic, so micromamba, pip, and R flows behave consistently on Windows, macOS, and Linux

## Testing

- `npx vitest run src/renderer/src/pages/workspace/WorkspaceActivityGroup.i18n.render.test.tsx src/renderer/src/pages/workspace/workspace-components.test.tsx src/renderer/src/pages/workspace/workspace-tool-activity-details.test.ts src/renderer/src/i18n/resources.test.ts` (661 passed)
- `npm run typecheck:web`
- `npx eslint --no-warn-ignored src/renderer/src/pages/workspace/WorkspaceManagePackagesActivityRow.tsx src/renderer/src/pages/workspace/WorkspaceActivityGroup.tsx src/renderer/src/pages/workspace/WorkspaceActivityGroup.i18n.render.test.tsx`
- `git diff --check`

Full cross-platform coverage is delegated to PR CI as requested.


ewen-poch:
Addressed the structured package failure finding in 5f749f56. The package activity now treats result.ok === false as a failure even when the ACP tool phase is completed, uses the failure icon and semantic color, suppresses active/success progress treatment, and shows the compact actionable error. Added a regression test for a completed tool response carrying ok: false.

ewen-poch:
Addressed both uninstall-copy findings in ee786941. Active uninstall package rows and the trailing progress status now use the localized ‘Removing…’ copy, while a completed R uninstall that requires a restart now explains that the removed packages need a kernel restart to unload. Added regression coverage for both active and completed uninstall states across the localized renderer.

ewen-poch:
Addressed the plain-text failure diagnostic finding in 08bc648a. When no structured manage_packages result can be parsed, the special package row now falls back to the existing ACP tool-content or raw text and presents it as the compact failure detail. Structured results continue to prefer their explicit error field, so successful output and installer logs are not added to the card. Added regression coverage for a failed plain-text result.

ewen-poch:
Added the confirmed R source-installer scope in fdd141d4 and 77dccbe5. manage_packages now supports explicit BiocManager and GitHub installers for managed R runtimes, verifies requested versions from the post-install inventory, records Bioconductor release or GitHub repository/ref/SHA metadata, and retains non-requested inventory changes. The existing tool-group card keeps requested packages in the primary table and exposes passive dependency/update changes in a default-collapsed, localized Related changes disclosure. Existing requests and historical sessions remain compatible; no database migration is introduced.

ewen-poch:
Addressed the three actionable review findings in 84910618:

- GitHub package verification now requires the installed DESCRIPTION source ref to match an explicitly requested `@ref`, while repository-only requests continue to accept any ref.
- `BiocManager::install` and `remotes::install_github` now fail closed before spawning R when the managed environment’s protected `r-base` version/build identity cannot be verified.
- Artifact provenance now preserves optional package source evidence for both the captured inventory and operation-level package changes (GitHub repository/ref/commit or Bioconductor release). The fields are additive and remain compatible with historical evidence that omits them.

Added focused regression coverage for ref mismatch, the protected-runtime guard, and provenance projection. Targeted suites, ESLint, diff checks, and the repository-standard node/web typecheck pass. The known host-dependent fail-closed child-process test remains unrelated.

ewen-poch:
Addressed three of the four findings in 892a4994:

- GitHub source matching is now restricted to R requests, and Python local path specs use their path basename for normal installed-package verification.
- Package delta classification now includes normalized source metadata, so same-version GitHub ref/commit changes are reported as updates.
- `host.lineage` now validates and preserves source metadata for installed packages and operation package changes across both the service projection and the REPL trust boundary; GitHub installer attempts are also accepted by the lineage validator.

I intentionally did not expose source metadata through the compact `inspect_packages` result. This feature keeps detailed package-source evidence in renderer/provenance projections rather than adding it to agent context; changing that boundary would conflict with the agreed renderer-only context constraint.

Focused tracker/lineage tests, repository-standard node/web typecheck, ESLint, diff checks, and the socket-backed host.lineage REPL integration test pass.

ewen-poch:
Addressed the Bioconductor source persistence finding in 80f2ca69. The mutation owner now passes the installer source into the environment tracker before publication. The tracker applies that source to requested package changes in the immediate operation manifest and stores the same optional field in the pending operation journal when inventory refresh fails, so recovery republishes it into the completed operation log.

This is additive and historical records remain compatible: no schema-version change, database migration, or new state enum is introduced. Added regression coverage for immediate Bioconductor source publication, caller propagation, and failed-refresh recovery. The focused suites, repository-standard node/web typecheck, ESLint, and diff checks pass.

ewen-poch:
No code change is planned for this repeated `inspect_packages` recommendation. The omission is intentional: detailed package-source metadata is available to the renderer and persisted provenance/host.lineage projections, while the compact MCP inspection result remains bounded and does not add these details to agent context. This preserves the agreed renderer-only context boundary. The PR is mergeable and the current head remains unchanged.

ewen-poch:
Addressed the actionable review findings in 859251f2:

- Exact Python `==` requests are now verified against the post-install inventory version, so a name match with the wrong version is reported as unsatisfied.
- The package ledger now starts from every requested package and overlays available change records, so partially failed requests remain visible.
- The package activity now follows the approved compact interaction: it stays expanded while running, defaults collapsed at terminal state, and expands to the package/version/dependency ledger. Successful calls intentionally omit duplicate Command/Result blocks; failed calls retain the existing error summary.
- `environmentName` remains in the compact tool result by product decision: it is useful final result metadata returned to the Agent, unlike renderer-only progress and intermediate logs.

No enum or persistence changes were introduced. Targeted validation passed: 4 test files / 668 tests, plus `tsc --noEmit`.

ewen-poch:
Addressed the package-spec matching finding in 46aa5e75. The renderer now normalizes exact Python specs and R package prefixes to canonical package names, and matches GitHub requests by both canonical package name and verified source repository. This prevents placeholder/verified duplicate rows while retaining the backend-reported version and source evidence. Added regression coverage for `numpy==2.0.0` and `tidyverse/ggplot2@main`; the focused renderer/i18n suites (603 tests), `tsc --noEmit`, and diff checks pass.

ewen-poch:
Addressed both latest review findings in cfb9a657. Exact Python pins are now compared through normalized release/epoch/pre-release/local-version components rather than raw strings, including equivalent release forms such as `2.0` and `2.0.0`. R inventory capture now records `RemoteType` and `RemoteHost` and emits GitHub provenance only when the package explicitly identifies a GitHub remote on `github.com` or `api.github.com`; lookalike metadata from other hosts remains unattributed. Added regression coverage for normalized exact-version success and rejection of non-GitHub remote attribution. The focused tracker suite (27 tests), `tsc --noEmit`, and diff checks pass.

ewen-poch:
Addressed the failed-installer provenance finding in d233f797. Bioconductor/GitHub installer source metadata is now returned only when the installer transaction succeeds; failed results retain their error and attempt evidence without propagating an unverified source into mutation verification or persisted provenance. Inventory-derived source metadata remains independent. Added failure coverage for both source installers; the focused tests, `tsc --noEmit`, and diff checks pass.

ewen-poch:
Addressed the compact-result prioritization finding in 4fc93d10. `compactManagePackagesResult` now orders verified requested changes before related/dependency changes, then applies the existing count and serialized-size budgets. Related changes preserve their original order, and `omittedPackageChangeCount` still reports every excluded entry. Added a regression case with the requested package after 60 oversized dependency records; the requested version/status remains present under truncation. Focused MCP tests, `tsc --noEmit`, and diff checks pass.

ewen-poch:
Addressed the GitHub inspection identity finding in 8fffeafc. R GitHub specs are now resolved by normalized repository/ref evidence and use the repository basename as the package identity, so `tidyverse/ggplot2@main` reports the installed `ggplot2` package rather than the owner. Added focused inspection coverage; the tracker suite (28 tests), `tsc --noEmit`, and diff checks pass.

No change is planned for the repeated request to expose source metadata through `inspect_packages` compaction. That boundary remains intentional: source details are retained for renderer/provenance/lineage use, while the read-only Agent inspection result stays bounded. Source evidence from `manage_packages` remains available for the mutation the Agent initiated.

ewen-poch:
Addressed the crash-recovery source gap in 35e69151. `refreshAfterPackageMutation` now atomically persists the terminal installer result, attempts, fallback state, and optional source before reading the baseline inventory or starting the terminal inventory probe. A successful refresh removes the pending record as before; a failed refresh or process exit recovers from the same already-terminal record, preserving Bioconductor/GitHub source evidence. This reuses the existing schema-v1 optional fields: no migration, enum, or interaction change. Added an assertion that the source-bearing terminal record exists before the inventory probe runs; the tracker suite (28 tests), `tsc --noEmit`, and diff checks pass.

PR #1653, opened 2026-08-24
fix(acp): preserve permission denial boundary

ewen-poch:
## Problem

Claude Code cancels its provider prompt after a rejected permission. Open Science resumes the same logical turn with a hidden continuation, but the old continuation encouraged the Agent to use an alternative. That could turn an explicit user denial into repeated commands or another permission request for the same operation.

## Proposed change

Keep the logical run alive, but carry an explicit authorization boundary into both live Claude continuation and restored durable permission continuation. The Agent is told that the user denied the operation, it has no authorization for it, and it must not retry or approximate that operation through another command, tool, route, or permission request in the current turn. Independent already-permitted work may continue.

## Scope and non-goals

- Preserve the ACP reject response and existing provider wire behavior.
- Preserve Claude provider-turn recovery and the user-visible logical run.
- Do not terminate the whole message or run.
- Do not add permission states, persisted fields, migrations, or compatibility codecs.

- explicit live Claude denial boundary -> `npm test -- src/main/acp/runtime.test.ts` -> 564 passed
- Claude Code, OpenCode, Codex Responses, and Codex Bridge permission projection -> `npm test -- src/main/acp/runtime.test.ts -t projects session/request_permission` -> passed as part of 13 focused permission cases
- Node runtime types -> `npm run typecheck:node` -> passed
- source lint -> `npm run lint` -> passed with 0 errors and 2 pre-existing formatting warnings in unrelated `application-commands.ts` and `ipc.ts`
- diff formatting -> `git diff --check` -> passed before commit

A sandboxed runtime test attempt first reported three loopback-bind failures with `listen EPERM: operation not permitted 127.0.0.1`; the same focused runtime suite passed outside that network sandbox.

Excluded: renderer, Web typecheck, database, migration, and platform-specific lanes because the final diff changes only main-process continuation text and its protocol-level regression assertion. PR CI remains authoritative for complete portable and platform coverage.

## Review focus

Confirm that the denial boundary forbids semantic retries of the rejected operation while still allowing independent permitted work, and that sharing the text between live and restored continuations keeps both lifecycle paths aligned.

### DIFF
--- a/src/main/acp/runtime.test.ts
+++ b/src/main/acp/runtime.test.ts
@@ -21247,7 +21247,7 @@ describe('ACP runtime session management', () => {
     expect(permissionResponses).toEqual([{ outcome: { outcome: 'selected', optionId: 'reject' } }])
   })
 
-  it('continues a Claude turn after a rejected permission cancels the provider prompt', async () => {
+  it('continues a Claude turn with an explicit authorization boundary after permission denial', async () => {
     const process = new FakeAgentProcess()
     const continuationGate = createDeferred<PromptResponse>()
     const prompts: string[] = []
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the P2 cancellation/denial distinction in 311f555c. A restored `cancelled` response now uses cancellation-specific continuation context: authorization was not granted and the parked call must not execute or retry, without claiming that the user explicitly rejected the operation. A selected reject option still uses the stronger explicit-denial boundary. Added a black-box restored-permission regression test; the focused ACP runtime suite passes 565/565.

PR #1658, opened 2026-08-24
perf(settings): cache usage projection for ten minutes

ewen-poch:
## Problem

Opening the Usage settings panel always reloads the SQLite usage projection. Switching away from the panel and back therefore repeats an expensive load and makes navigation feel sluggish.

## Proposed change

- Cache each renderer usage projection for 10 minutes.
- Reuse a fresh projection when the Usage panel remounts.
- Reload automatically when the panel is reopened after the cache expires.
- Keep the existing manual refresh action as an unconditional reload.

## Scope and non-goals

- Renderer-memory cache only; no persisted data or migration.
- No historical-data compatibility changes.
- No new enum or durable state values.
- No polling while the panel remains open.

- Fresh projection reuse and expiry reload -> `npm test -- src/renderer/src/pages/settings/TokenUsagePanel.render.test.tsx` -> 10 tests passed.
- Changed-file lint -> `npx eslint --no-cache src/renderer/src/pages/settings/TokenUsagePanel.tsx src/renderer/src/pages/settings/TokenUsagePanel.render.test.tsx` -> passed.

These checks ran after the last material source edit. The dependency-aware planner selects the full suite because `docs/design.md` has no module owner; per maintainer direction, the full local suite was not run and PR CI is authoritative.

## Review focus

- Cache lifetime and remount behavior.
- Manual refresh continuing to bypass the cache.
- Renderer-only cache ownership and absence of persistence.


ewen-poch:
Addressed the expired-cache retention finding in bf7fa2d1. The cache reader now deletes the WeakMap entry whenever it is missing a valid refresh timestamp or exceeds the 10-minute TTL, so the projection and its usage events can be released after expiry. The focused renderer test suite, Web typecheck, and changed-file lint pass.

ewen-poch:
Added the Usage loading skeleton in 55bd570a. The initial projection load now preserves the final page structure with placeholders for the header, range control, refresh row, eight summary groups, and both charts. The loading surface retains an accessible status, aria-busy, and reduced-motion behavior. Focused renderer tests, Web typecheck, and changed-file lint pass.

PR #1670, opened 2026-08-24
fix(notebook): keep REPL errors concise for agents

ewen-poch:
## Problem

`repl_execute` returned the complete Node VM stack to the Agent for ordinary JavaScript failures. The response exposed internal `<repl>`, `node:vm`, and `resources/notebook/repl_loop.js` frames without helping the Agent correct the request.

## Proposed change

Use a REPL-specific agent-facing projection that keeps the actionable first error line and removes internal stack frames before the existing execution-result compactor runs. The durable Notebook run remains unchanged, so the full diagnostic is still available outside the immediate Agent reply.

## Scope and non-goals

- Applies only to the immediate `repl_execute` MCP result shared by all Agent frameworks.
- Keeps `notebook_execute` Python/R tracebacks and `bash_execute` diagnostics unchanged.
- Adds no dependencies, enum/status values, persistence changes, migrations, or historical-data compatibility behavior.

- REPL error projection (including SyntaxError preambles and multiline messages) and Python traceback preservation -> `npm test -- src/main/notebook/mcp-server.test.ts` -> 75 passed.
- Node contract -> `npm run typecheck:node` -> passed.
- Changed source and tests -> `npx eslint src/main/notebook/mcp-server.ts src/main/notebook/mcp-server.test.ts` -> passed.
- Formatting and whitespace -> `prettier` on both changed files plus `git diff --check` -> passed.

Local full `npm test` was intentionally not run. `mcp-server.ts` is not currently owned by a Module in `scripts/ci/module-impact.json`, so the local affected-test explainer selected no Module; exact-head PR Gate CI is the final authority and may fail closed to broader coverage. No local platform lane was included because this is a platform-independent string projection.

## Review focus

- The projection must not mutate the raw result that is persisted for Notebook history.
- Only the REPL tool should use the concise projection; Python/R and shell consumers must retain existing diagnostics.

### DIFF
--- a/src/main/notebook/mcp-server.test.ts
+++ b/src/main/notebook/mcp-server.test.ts
@@ -1139,11 +1139,14 @@ describe('compactNotebookExecutionResult', () => {
   })
 
   it('applies the compact projection and global budget to every execution tool', () => {
-    for (const name of ['notebook_execute', 'repl_execute', 'bash_execute']) {
+    for (const name of ['notebook_execute', 'bash_execute']) {
       const tool = NOTEBOOK_RPC_TOOLS.find((entry) => entry.name === name)
       expect(tool?.mapResult).toBe(compactNotebookExecutionResult)
       expect(tool?.resultLimitChars).toBe(NOTEBOOK_MCP_EXECUTION_RESULT_LIMIT)
     }
+    const replTool = NOTEBOOK_RPC_TOOLS.find((entry) => entry.name === 'repl_execute')
+    expect(replTool?.mapResult).not.toBe(compactNotebookExecutionResult)
+    expect(replTool?.resultLimitChars).toBe(NOTEBOOK_MCP_EXECUTION_RESULT_LIMIT)
   })
 
   it('keeps diagnostic streams once and removes duplicated structured stream outputs', () => {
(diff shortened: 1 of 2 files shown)


ewen-poch:
CI investigation: the two failed macOS shards contain three renderer architecture/source-shape assertion failures, all unrelated to this PR's `src/main/notebook/mcp-server*` diff:

- `session-store.test.ts` — `keeps production consumers on the public store facade`
- `workspace-components.test.tsx` — `starts session persistence from the app shell and passes readiness into the workspace`
- `workspace-page.architecture.test.ts` — `keeps private controller consumers explicit and bounded`

I reproduced the same failures at the latest `origin/main` (`1f4032a5`) in a clean detached worktree:

```text
Test Files  3 failed (3)
Tests       3 failed | 205 passed (208)
```

The received values reflect the app-shell move to `ApplicationPresentationHost.tsx` / `useApplicationEventBindings.ts`, while the assertions still expect the old `App.tsx` ownership. PR #1669, which introduced that move, skipped the full macOS module shards under its affected-test plan, so these baseline failures were first exposed by this PR's fail-closed full plan.

I am leaving the notebook PR scoped and not mixing an unrelated renderer baseline repair into it. All checks covering this diff pass, including the 75 focused notebook MCP tests, node typecheck, targeted ESLint, static CI checks, builds/E2E, CodeQL, and the latest Codex review (mergeable, no findings).

PR #1678, opened 2026-08-24
feat(composer): attach long pasted text

ewen-poch:
## Problem

Large plain-text pastes occupy most of the composer and are difficult to review, while converting them to attachments without preserving their insertion point risks losing prompt structure or user data.

## Proposed change

- Convert plain-text pastes over 10,000 characters or 300 lines into a managed `Pasted text.txt` attachment.
- Keep a transient `pasted-text` anchor in the live draft so **Show in text field** restores the exact text and caret position.
- Keep the empty text field focused with its placeholder and a drawable caret while the attachment card is present.
- Delete the pending upload when the card is closed or restored; `Cmd+Z` / `Ctrl+Z` after either action re-stages it at the same logical position without showing an Undo notice.
- Restore the original text inline on upload failure or attachment-limit rejection.
- Filter transient paste anchors at every message-parts boundary so only the attachment reaches the runtime.
- Reuse the existing hybrid text-upload delivery path: bounded text resources when budget permits, otherwise preview plus resource link.

```mermaid
flowchart LR
  A[Long plain-text paste] --> B[Transient draft anchor]
  A --> C[Managed pending upload]
  C --> D[Attachment card]
  D -->|Show in text field| E[Restore exact text and caret]
  E -->|Cmd/Ctrl+Z| C
  D -->|Close| F[Delete pending upload]
  F -->|Cmd/Ctrl+Z| C
  D -->|Send| G[Immutable Upload Version]
  B -. filtered from message parts .-> G
```

## Scope and non-goals

- Uses the existing upload staging, publication, cleanup, and Agent delivery contracts.
- Adds no database schema, migration, IPC contract, dependency, settings surface, or cross-restart draft persistence.
- The new `pasted-text` discriminant, `pastedTextId` transfer metadata, and one-step paste undo receipt are renderer-only transient state and are never persisted in historical session JSON.
- Existing messages, uploads, and storage layouts need no compatibility adapter.
- The compact card follows existing composer tokens and interaction conventions; no visible Undo snackbar or success toast is added.

- Long-paste conversion, placeholder/caret hosting, metadata binding without selection loss, Show then Undo, close Undo, exact restoration, send filtering, and card interaction -> focused Vitest set -> **5 files / 244 tests passed**.
- Workspace ownership, representative consumers, and architecture limits -> `npm run test:module -- workspace_page` -> **25 files / 519 tests passed**.
- Translation catalog integrity -> `npx vitest run src/renderer/src/i18n/resources.test.ts` -> **590 tests passed**.
- Renderer interface/type safety -> `npm run typecheck:web` -> **passed**.
- Source lint policy -> `npm run lint` -> **passed**.
- Production renderer bundle -> `npm run build:web` -> **passed** (existing dynamic-import and bundle-size warnings only).
- Browser interaction/responsive behavior -> isolated real-app QA -> **passed** for exact restore, Show then `Cmd+Z`, placeholder/card coexistence, focused text-node caret hosting, absence of an Undo notice, and the previously verified no-overflow layouts at 320, 375, 414, and 768 CSS pixels.
- The first PR AI review identified an attachment-cap replacement race; `5cd7e355` fixes it. Commit `a336e639` adds Show Undo and visible empty-editor coverage requested during manual QA. The later caret review is fixed by `6b817dde`, which updates upload metadata in place so the contenteditable DOM and selection are preserved.

Per maintainer direction, the complete local `npm test` suite was not run. PR CI remains authoritative for the complete portable and platform lanes.

## Review focus

- Lossless transitions among inline text, pending transfer, uploaded attachment, Show/close Undo, and send.
- Cleanup races when an anchor is removed, a draft/session changes, or an upload completes late.
- `docToMessageParts` filtering at immediate, queued, and edited-message send boundaries.
- The renderer-only state boundary: no transient paste node should enter persisted message data.


ewen-poch:
Addressed the actionable review points:

- **Attachment-cap replacement:** `5cd7e355` excludes the pasted-text attachment/transfer being replaced from the effective occupancy count, so replacement still succeeds at the cap. Regression coverage exercises that boundary.
- **Caret preservation during upload metadata binding:** `6b817dde` synchronizes `transferId`/`attachmentId` onto the existing pasted-text anchor without rebuilding the contenteditable DOM. The regression test verifies both the unchanged middle-of-text selection and the updated anchor metadata.

The latest head review reports no actionable findings and marks the PR mergeable.

ewen-poch:
Updated in 2e4df510:\n\n- Aligned the pasted-text attachment card with standard attachment cards at 36px.\n- Replaced the generic upload filename in the card with a normalized 20-character content preview plus an ellipsis; the underlying upload remains Pasted text.txt.\n- Replaced the zero-length trailing text node with a non-serializing word-joiner caret host, so Chromium has a drawable caret line box while the placeholder remains visible.\n\nValidated with the three focused renderer suites (195 tests), ESLint, Prettier, web typecheck, a production web build, and browser QA (focused editor, 17.5px caret range, 36px card).

ewen-poch:
Follow-up in e9425d87: the invisible pasted-text anchor now has zero layout height. It no longer enlarges the composer body while the drawable caret host remains intact. Browser measurements now match the intended formula exactly: 94.375px base composer + 45px attachment row + 8px existing gap = 147.375px, with the editor staying at 36.375px and the caret range at 17.5px.

ewen-poch:
Aligned in-progress attachment cards with the shared 36px attachment height by removing the temporary h-11 override. This prevents small uploads from expanding the composer to 155.375px and then snapping back to 147.375px on completion. Added a regression assertion for the shared height contract; the full ConversationPanel interaction suite passes (123 tests), and browser measurements confirm queued, uploading, and completed states remain 36px throughout.

ewen-poch:
Confirmed. Pasted-text anchors are intentionally live-draft-only today, while the inline edit flow reconstructs only durable message parts and keeps original uploads read-only. That leaves a pasted-text-only message with an empty edit document and the current resend gate rejects it. The narrow fix is to permit an attachment-only edited resend while preserving the original uploads; restoring exact positions after send would instead require a durable ordered paste-to-upload association and historical fallback behavior. I am pausing implementation pending the product/data-model decision.

ewen-poch:
Implemented the approved inline pasted-text anchor behavior:

- A compact inline marker now preserves each pasted-text insertion position and can locate its matching attachment by click, Enter, or Space. The attachment scrolls into view and uses a one-shot reduced-motion-aware emphasis animation.
- Copying a composer selection that contains one or more pasted-text markers writes a private clipboard fragment plus a full text/plain fallback. Pasting that fragment into the composer creates fresh pastedTextIds and independent uploads; it never reuses the source attachment relationship.
- Multiple copied markers are staged atomically so all draft bindings are applied together.
- The inline placeholder remains visible only while the marker is the sole content. Any user-entered character, including whitespace, hides it.
- Uploading and completed pasted-text chips share the existing fixed attachment height and content preview.

No persistent data, enum values, shared data model, dependencies, or historical compatibility path were added.

Validation: 218 focused tests passed across composer doc/editor, ConversationPanel, and workspace composer controller; npm run typecheck:web passed; npm run lint passed.

ewen-poch:
Pasted-text uploads now use short content-derived filenames instead of the generic Pasted text.txt name.

Naming format: Pastedtext-<safe excerpt>.txt. The excerpt is normalized, limited to Unicode letters and numbers plus separators, and capped at 20 characters, making the total filename at most 35 Unicode characters. If no safe excerpt remains, the fallback is Pastedtext.txt.

The generated name is used by the real File and upload transfer, so the composer, sent attachment, and Files library remain consistent. Existing uploads keep their stored names; there is no migration, schema change, enum change, or new persistence field.

Validation: 218 focused tests passed; npm run typecheck:web passed; npm run lint passed.

ewen-poch:
Adjusted the naming presentation boundary: while pasted text is still in the composer, its attachment chip continues to show the original content preview. The short content-derived Pastedtext-<excerpt>.txt name is used by the real upload and therefore appears after send and in the Files library.

This is a presentation-only correction on top of the existing filename change; no persistence shape, enum, or compatibility behavior changed.

Validation: ConversationPanel interaction tests passed (123 tests); npm run typecheck:web passed; npm run lint passed.

ewen-poch:
Shortened the composer-only pasted-text content preview from 20 to 17 source characters before the ellipsis, bringing its visual width closer to the Show in text field row. The persisted upload filename limit remains unchanged.

Validation: composer document and ConversationPanel interaction tests passed (167 tests); npm run typecheck:web passed; npm run lint passed.

ewen-poch:
Addressed both actionable caret/undo findings in 1847d0a0:

- Pending caret restoration is discarded when the active draft changes, so a blocked composer cannot later apply an old node index/offset to another draft. This uses the existing caret request state and adds no draft-key state shape.
- An editor change that exactly removes one atomic pasted-text marker now reuses the existing attachment-removal receipt path. Backspace/Delete therefore remains Cmd/Ctrl-Z reversible, while subsequent ordinary typing still clears the custom undo boundary.

No architecture, persistence, enum, or interaction-model changes were introduced.

Validation: all four focused composer suites passed (219 tests); npm run typecheck:web passed; npm run lint passed.

ewen-poch:
The latest review repeats the known post-send editing boundary for draft-only pasted-text anchors. Fixing it is not a local serialization correction: it requires one of two product decisions—either preserve attachment-backed uploads through edited resend, or disable editing for attachment-backed messages. Rehydrating exact inline positions would additionally require a durable ordered paste-to-upload association and historical fallback behavior.

No code change is being made for this finding until the interaction/data-model choice is confirmed.

ewen-poch:
Addressed P1 in 2480e726. domToDoc now strips the reserved U+2060 caret marker from every text node, so a browser-split caret host cannot leak an invisible character into the serialized message. The new regression test pastes a second long-text attachment at the trailing host and verifies that the resulting document contains only the two pasted-text anchors. Focused composer tests pass (79 tests), along with targeted ESLint, Prettier, and web typecheck.\n\nI am not changing P2. Hiding the placeholder after any user-entered character, including whitespace, is the explicitly approved interaction. The sendability rule remains unchanged: a whitespace-only draft is still rejected after trimming. This is an intentional distinction between visual input feedback and submission validity.

ewen-poch:
Addressed the cut-path finding in d061ec35. Cut now reuses the same private composer fragment and full text/plain serialization as Copy before removing the selected marker through the normal DOM-to-doc change path. A focused-marker cut therefore places the complete pasted payload on the clipboard, removes the matching anchor atomically, and continues through the existing upload cleanup/undo receipt behavior instead of copying the visible ellipsis.\n\nValidation: focused composer suites pass (80 tests); targeted ESLint and Prettier pass; web typecheck passes.

ewen-poch:
Addressed the word-joiner ownership finding in 59bdb5c4. domToDoc again strips U+2060 only from renderer-owned caret-host Text nodes tracked by the existing WeakSet. Before plain or private-fragment long-paste insertion, the editor records whether the active range is inside an owned host; if the browser splits that host, the split marker is replaced with a newly tracked host. A user-entered U+2060 therefore remains ordinary message text.\n\nValidation: focused composer suites pass (81 tests); targeted ESLint and Prettier pass; web typecheck passes. This concludes the configured five-round review follow-up window, so I am not starting another review-response loop automatically.

ewen-poch:
Addressed the paste-selection ownership finding in 0d2ca80b. All composer paste insertion helpers now require the active Selection range to belong to the current editor root and report whether insertion actually occurred. Long plain-text and private-fragment uploads are staged only after a successful insertion; ordinary text paste uses the same boundary and cannot modify an unrelated DOM selection.\n\nValidation: focused composer suites pass (82 tests); targeted ESLint and Prettier pass; web typecheck passes.

PR #1691, opened 2026-08-25
feat(global-search): add session number lookup

ewen-poch:
## Problem

Session numbers are available in persisted Session summaries and in Composer references, but global search neither displays them nor accepts them as a lookup key.

## Proposed change

- display each valid Session number as stable trailing metadata in global search
- align Session rows with the existing Artifact result skeleton through a matching leading visual slot and two-line text hierarchy
- match pure numeric queries against Session-number prefixes and rank an exact number first
- retain title-only matching for nonnumeric Session queries

## Scope and non-goals

- no database, migration, persistence, IPC, or renderer contract changes
- no new state or enum values
- historical Sessions without a valid positive number remain searchable by title and omit number metadata
- no number presentation changes outside global search

- Session number filtering and exact-match ordering -> `npx vitest run src/renderer/src/components/global-search/global-search-catalog.test.ts src/renderer/src/components/global-search/GlobalSearchDialog.test.tsx` -> 23 passed
- localized global-search rendering and renderer i18n catalog guard -> `npx vitest run src/renderer/src/components/global-search/GlobalSearchDialog.i18n.render.test.tsx src/renderer/src/i18n/resources.test.ts` -> 597 passed
- renderer type safety -> `npm run typecheck:web` -> passed
- source lint -> `npm run lint` -> passed
- changed-file formatting -> `npx prettier --check ...` -> passed

`npm run test:affected -- --base origin/main --head HEAD --explain` conservatively selected the full suite because `docs/design.md` has no module owner. Per the requested validation scope, the local full suite was not run; PR Gate remains authoritative for complete and platform-specific coverage.

## Review focus

- whether the Session and Artifact result rows now form one coherent hierarchy
- whether numeric lookup remains narrowly scoped to Session identity without expanding text search into other metadata
- whether missing historical numbers degrade without visual or search regressions

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -723,6 +723,12 @@ colors communicate a successful or failed probe/migration result.
   list is empty, each Project row also shows its artifact count from a complete Project Files index;
   partial index counts are omitted. While those counts are visible, Project Files change events
   refresh the affected Project so index repair and file changes do not leave stale totals.
+- Global search Session rows share the same result skeleton as Artifact rows: a `size-10` leading
+  visual anchor, two text lines, and stable trailing metadata. Artifact rows use their thumbnail and
+  contextual actions; Session rows use a neutral icon tile and show their SQLite-backed `#number` as
+  quiet monospaced tabular text. Do not render a redundant Session type badge. A pure numeric query
+  matches positive Session-number prefixes and ranks an exact number first; nonnumeric queries remain
+  title-only so global search does not expand into message-body or metadata search.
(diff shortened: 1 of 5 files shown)


ewen-poch:
Addressed the cross-Project exact-number ranking issue in b9774828. A numeric exact match from another Project is now promoted ahead of current-Project prefix matches, retained with its source Project metadata, and removed from the capped Other projects results to avoid duplication. Added catalog and dialog regressions covering display order, default Enter selection, and cross-Project navigation. Focused tests pass (25/25), along with affected ESLint, Web typecheck, and formatting checks.

PR #1699, opened 2026-08-25
feat(composer): add draft redo history

ewen-poch:
## Problem

PR #1694 moved Composer undo into one application-managed history so text, long pasted text, and
attachments undo consistently. That history had no forward operation, so `Cmd+Shift+Z` could not
reapply the most recently undone draft state.

## Proposed change

- Route `Cmd+Shift+Z` on macOS and `Ctrl+Shift+Z` on Windows/Linux through a Composer redo action.
- Add a bounded, per-draft in-memory redo stack and restore the post-edit caret on redo.
- Preserve completed staged uploads while their state remains reachable from redo, then delete them
  when a new edit or lifecycle boundary discards that redo branch.
- Retain transient source `File` objects for undone pending transfers, restart them with a new
  transfer ID on redo, and rebind long-paste anchors to the replacement transfer.
- Reconcile upload completion and failure across both history directions, while leaving transfers
  shared by the before/after states running.
- Release staged files and transfer sources owned only by history when the Composer controller
  unmounts, while preserving resources still owned by active or stored drafts.

## Scope and non-goals

- Builds on #1694 and preserves its per-draft, process-local history model.
- No database, schema, persisted-data, IPC, renderer API, or enum changes.
- No historical-data compatibility requirement.
- Undo/redo history and retained `File` references are not restored after restart.
- `Ctrl+Y` is not added; the forward shortcut is `Cmd/Ctrl+Shift+Z`.

- Shortcut dispatch, exhausted redo isolation, Composer document/IME/selection/mention consumers ->
  direct Vitest run including `ComposerEditor.interaction.test.tsx` and
  `ConversationPanel.interaction.test.tsx` -> passed.
- Text, caret, long-paste, completed attachment, pending-transfer restart, upload race, invalidation,
  draft switching, send/clear, customize prefill, prompt history, and Session deletion -> direct
  Vitest run of the `workspace_page` module impact plan -> passed.
- Final combined Test Impact Set -> 27 files, 702 tests passed.
- Changed-file lint -> `npx eslint` over all seven changed TypeScript files -> passed with zero
  warnings.

`npm run test:module -- workspace_page` produced the expected 25-file plan, but its pretest stopped
because this linked worktree has no generated Prisma Client. The exact plan was executed directly
with Vitest, together with the two direct interaction consumers. Full `npm test` was intentionally
not run per task scope; exact-head PR CI remains the final authority.

## Review focus

- Confirm completed staged uploads remain owned while redo can restore them and are deleted when the
  redo branch is invalidated.
- Confirm an undone pending transfer is restarted only when it is absent, with a new transfer ID and
  correctly rebound pasted-text anchor.
- Confirm upload completion/failure reconciliation cannot resurrect an aborted transfer or leave a
  dangling attachment path in either history direction.
- Confirm every existing undo invalidation boundary also invalidates redo.
- Confirm teardown deletes resources retained only by history without deleting active-draft files.

Remaining risk: local tests use mocked upload staging and synthetic macOS/Windows keyboard events;
exact-head CI remains responsible for packaged Web/Electron interaction coverage.


ewen-poch:
Addressed the P2 upload lifecycle issue in 53e496c3. Applying an undo/redo snapshot now releases resources from both the popped snapshot and the replaced current snapshot after active/history ownership has been updated; resources still referenced by active drafts, saved drafts, deletion cleanup, or either history stack remain protected. Added a regression test covering pasted-anchor replacement followed by undo, redo, and undo, verifying that the superseded re-staged upload is deleted. Validation: focused controller suite (37 tests), Web typecheck, ESLint on the changed files, and git diff --check all pass.

ewen-poch:
Addressed both P2 findings in f560cd45:\n\n- Restoring a pasted attachment into the text field now clears the active draft's redo branch while keeping the dedicated removal receipt, so the restoration itself remains undoable.\n- Redo now restarts only queued/uploading transfers. Failed ordinary uploads are restored with their existing error state and are not silently retried; their transient source File reference is released when staging fails.\n\nRegression coverage verifies redo invalidation after "Show in text field" and verifies that undo/redo restores a failed upload without another staging call. Validation: focused controller suite (38 tests), Web typecheck, ESLint on the changed files, and git diff --check all pass.

PR #1700, opened 2026-08-25
perf(skills): defer user skill startup scanning

ewen-poch:
## Problem

Application startup scanned every Personal and Imported Skill twice: once while validating the
bundled Specialist registry and once to establish the writable Skill observer fingerprint. Local
profiles show 32–125 ms for the first scan and 14–61 ms for the second, with cost growing with user
package count and size.

## Proposed change

- Restrict bundled Specialist startup validation to Featured Skills and stop reading Specialist
  package sidecars from Personal/Imported directories on that path.
- Install the writable Skill watcher before returning from startup, then reconcile its initial
  fingerprint in the background through the existing serialized queue.
- Coalesce concurrent user-catalog reads while one authoritative repository scan is in flight. If a
  user sends a Session immediately, observer reconciliation and agent Skill materialization share
  that scan; the Session still waits for the authoritative projection before the agent starts.

## Scope and non-goals

- Connector discovery remains a background refresh with its existing bounded first-ACP-session
  barrier.
- Custom Specialist loading and Marketplace/package transaction recovery are unchanged.
- No catalog cache, worker, dependency, user-visible copy, schema, migration, or persisted format is
  added.
- Historical Skill directories and the rebuildable `user-skill-compatibility-v1.json` format are
  unchanged.
- No persisted/domain state or enum value is added. Coordination consists only of process-local
  in-flight Promise/fingerprint state plus a boolean cleared after the first successful retry of a
  failed initial reconciliation.

The checks below ran after the last material edit.

| Behavior | Check | Result |
| --- | --- | --- |
| Observer startup is non-blocking, retains concurrent events, and publishes after a failed baseline recovers | `npm test -- --run src/main/skills/user-skill-catalog-observer.test.ts` | 11/11 passed |
| Bundled Specialist startup excludes user Skills; immediate agent materialization shares the observer scan | `npm run test:module -- user_skills_repository` | passed |
| User Skill owner, contracts, Settings/renderer consumers, Specialist packages, preload, and local RPC integration | same module command | 19 files, 777/777 passed |
| Node and renderer TypeScript contracts | `npm run typecheck` | passed |
| Repository lint rules | `npm run lint` | passed |
| Formatting and whitespace | Prettier check on changed files; `git diff --check` | passed |

`npm run test:affected:explain -- --base origin/main --head HEAD` fails closed to the complete
Vitest plan because `src/main/ipc.ts` is a centralized file with no module owner. Per the requested
workflow, the complete local suite was not run; PR Gate is authoritative for the full portable and
platform lanes. Filesystem watcher/platform behavior is therefore intentionally left to CI beyond
the module's local coverage. No packaged-app startup benchmark was run; the diff removes the two
user-catalog startup dependencies directly and pins those boundaries with tests.

## Review focus

- The watcher is installed before the background baseline starts, and its existing one-running /
  one-pending queue retains concurrent filesystem and explicit mutation notifications.
- The in-flight catalog Promise is cleared immediately on settlement and does not become a cache.
- Runtime Specialist catalogs still include Personal and Imported Skills; only bundled registry
  startup uses the Featured-only mode.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -1058,22 +1058,19 @@ const createApplicationModules = async (
   // the same repository while keeping their dynamic Connector/custom-Skill catalog separate.
   const specialistRepository = new SpecialistRepository(resolveStorageRoot())
   const appVersion = app.getVersion()
-  const specialistSkills = await settingsService.listSpecialistSkillCatalog()
-  const packageSkills = await specialistPackageSkillAdapter.snapshot()
+  const specialistSkills = await settingsService.listSpecialistSkillCatalog({ bundledOnly: true })
   composition.phase('specialist-catalog')
   const builtinRegistry = new BuiltinSpecialistRegistry({
     appVersion,
     builtinSkills: composeBuiltinSkillCatalog(appVersion, specialistSkills),
     skills: specialistSkills.map((skill) => {
-      const packageSkill = packageSkills.find((candidate) => candidate.id === skill.id)
       return {
         id: skill.id,
         name: skill.frameworkName,
         builtin: skill.source === 'featured',
         displayName: skill.displayName,
         source: skill.source,
-        mainEnabled: skill.mainEnabled,
-        ...(packageSkill ?? {})
+        mainEnabled: skill.mainEnabled
       }
     }),
     connectorIds: ALL_CONNECTOR_IDS,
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed the Codex P1 in 2e81577: a failed initial reconciliation is tracked in process memory, and the first successful retry now publishes one catalog change before clearing that marker. Added a focused regression test; the final user Skills module run passes 777/777 tests.

PR #1726, opened 2026-08-25
fix(session-persistence): validate projection before authority write

ewen-poch:
## Problem

`sessions:save-session` can receive runtime values that TypeScript cannot protect against. A Session update with `updatedAt: Number.MAX_VALUE` reproduces a cross-store split: the authoritative JSON is atomically replaced, then Prisma rejects the SQLite projection value and leaves a pending reconciliation row. The caller sees a failed command even though the Session authority changed.

## Proposed change

Validate the exact Session projection storage shape in `prepareSave`, before the transaction creates or updates any pending reconciliation state. The validation covers SQLite `Int` bounds, safe integer-backed `BigInt` values, required and nullable text values, and the uniqueness constraints used by turn, model-call, and run projections.

Add an integration regression test at the public `SessionRepository.saveSession` boundary. It creates a real migrated SQLite database and Session JSON, submits the malformed update, and verifies that the save rejects while both the original JSON and an empty pending queue remain unchanged.

## Scope and non-goals

- No persisted fields, enum values, database schema, migration, or data-model changes.
- No Electron/Web API shape or valid user interaction changes.
- No change to the JSON-first crash-recovery protocol; genuine post-authority I/O failures remain recoverable through the existing pending mechanism.
- No whole-Session byte or array-count budget. That needs a separately chosen product limit and capacity test; this PR does not invent one while fixing the reproduced consistency defect.
- No historical-data migration or read-path change. Existing Session files continue through the current durable decoder and reconciliation behavior.

- malformed projection value is rejected before JSON/pending mutation -> `npm test -- src/main/session-persistence/projection.test.ts src/main/session-persistence/repository.test.ts src/main/session-persistence/ipc.test.ts` -> 3 files, 115 tests passed
- Node contracts and Prisma projection types remain valid -> `npm run typecheck:node` -> passed
- changed files satisfy lint rules -> `npx eslint src/main/session-persistence/projection.ts src/main/session-persistence/projection.test.ts` -> passed

The regression was also run against the unfixed final base: Prisma rejected the save, the JSON diff showed the rejected title/timestamp/revision had already replaced authority, and `pending()` returned the Session save intent. With the fix restored, the same test passes.

Per maintainer request, the complete local `npm test` suite was not run. The impact selector falls back to the full plan because `projection.ts` is not assigned to a module owner; PR CI remains authoritative for the complete portable and platform suites.

## Review focus

- Confirm `prepareSave` remains the last side-effect-free boundary before pending reconciliation is written.
- Confirm the validator mirrors every scalar and uniqueness constraint currently written by `sessionData` and `replaceChildren`, including model-call usage introduced in projection version 2.
- Confirm the narrower storage-shape guard is preferable here to duplicating the full evolving `PersistedChatSession` model in a second runtime schema.

### DIFF
--- a/src/main/session-persistence/projection.test.ts
+++ b/src/main/session-persistence/projection.test.ts
@@ -374,6 +374,60 @@ describe('Session projection', () => {
     await expect(projection.pending()).resolves.toEqual([])
   })
 
+  it('rejects whitespace projection identifiers before replacing Session authority', async () => {
+    storageRoot = await mkdtemp(join(tmpdir(), 'open-science-session-text-validation-'))
+    client = createProjectDbClient(storageRoot)
+    await migrateApplicationDatabase(client)
+    await client.project.create({ data: { id: 'project-1', name: 'Project' } })
+    const projection = new SessionProjectionRepository(async () => client!)
+    const repository = new SessionRepository(storageRoot, {}, projection)
+    const saved = await repository.saveSession(session('blank-source'))
+    const authorityPath = join(storageRoot, 'sessions', 'project-1', 'blank-source.json')
+    const originalAuthority = await readFile(authorityPath, 'utf8')
+    const invalid = { ...session('blank-source'), number: saved.number, revision: saved.revision }
+    invalid.messages[1].modelCallUsage![0].sourceInvocationId = ' '
+
+    const saveError = await repository
+      .saveSession(invalid)
+      .then(() => undefined)
+      .catch((error: unknown) => error)
+
+    expect.soft(await readFile(authorityPath, 'utf8')).toBe(originalAuthority)
+    expect(saveError).toEqual(
+      expect.objectContaining({
+        message: 'Session projection modelCall.sourceInvocationId must be a non-empty string.'
+      })
+    )
+    await expect(projection.pending()).resolves.toEqual([])
+  })
+
+  it('rejects invalid Session statuses before replacing authority', async () => {
+    storageRoot = await mkdtemp(join(tmpdir(), 'open-science-session-status-validation-'))
+    client = createProjectDbClient(storageRoot)
+    await migrateApplicationDatabase(client)
+    await client.project.create({ data: { id: 'project-1', name: 'Project' } })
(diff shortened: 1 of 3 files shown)


ewen-poch:
Addressed in 3b7a249e. Projection storage-shape validation now runs before a suspended save can write authority JSON, before a rebuild can call clearForRebuild, and inside commitSave, commitReconciliation, and replaceAll before their projection mutations. Regression coverage reproduces both the suspended-save bypass and the destructive rebuild ordering, and verifies that invalid input leaves authority and the existing projection unchanged. The reachable fixture uses an oversized updatedAt; malformed model-call usage is removed by the existing persistence normalization before reload, while the oversized timestamp exercises the same projection-storage failure.

ewen-poch:
Addressed both findings in f8f021d6.\n\n- Revision overflow: a shared revision increment helper now rejects a revision that cannot be incremented safely before prepareSave can create a pending marker or authority JSON can be written. The committed-project save path uses the same guard.\n- Zero context-window size: projection validation now requires a non-null contextWindowSize to be a positive safe integer, matching the existing SQLite CHECK constraint.\n\nThe regression tests first reproduced authority replacement plus pending/SQLite failure on the previous code, and now verify that authority remains unchanged and no pending reconciliation is created. The targeted session-persistence suite passes all 119 tests.

ewen-poch:
Addressed the reachable text/status constraint gap in a146f795. Preflight now mirrors the existing SQLite constraints for non-empty Session/model-call identifiers, nullable attribution text, and the persisted Session status set. Regression tests reproduce the authority replacement followed by SQLite CHECK failure on the previous code and now verify unchanged authority with no pending reconciliation.\n\nI did not add the proposed cache-token pair rejection. A focused integration reproduction showed that a one-sided cachedReadTokens/cachedWriteTokens pair is accepted by SQLite and the save completes; SQLite CHECK expressions that evaluate to NULL pass, so the reported post-authority failure does not occur. Enforcing a stricter pair invariant would change accepted persisted data and requires an explicit schema/history-compatibility decision outside this fix.

PR #1731, opened 2026-08-25
fix(resources): enforce provider and file operation limits

ewen-poch:
## Problem

Provider settings accepted unbounded names, URLs, model IDs, API keys, provider counts, and fetched
model catalogs. Model refresh and provider validation also fully buffered response bodies, so a fast
malicious or broken gateway could consume substantial main-process memory before the existing timeout.
Local directory listing capped results at 5,000 entries but awaited every `stat` serially, amplifying
network-volume latency.

The regression tests were run three times before each corresponding production fix. They consistently
showed the reported behavior: oversized provider values and provider 65 were persisted, oversized
model responses/catalogs were accepted, validation fully consumed oversized direct/bridge responses,
and `LocalFsService.listDir()` started exactly one `stat` before the first completed.

## Proposed change

- Enforce main-process provider limits: 64 providers, 128-character names, 2,048-character Base URLs,
  512-character model IDs, and 16 KiB UTF-8 API keys.
- Stream provider response bodies through bounded readers: 2 MiB for model catalogs and 1 MiB for
  validation responses.
- Reject model catalogs atomically when they contain more than 2,000 models or an oversized model ID;
  keep the previous/bundled catalog instead of silently truncating persisted data.
- Process local directory metadata in batches of 16 while preserving the existing sort, 5,000-entry
  cap, fallback metadata, IPC contract, and response shape.

## Scope and non-goals

- No database/schema migration, persisted field, relationship, enum value, or new state.
- No renderer interaction or form change.
- Existing oversized `settings.json` records remain readable and are not rewritten. New or edited
  draft values must satisfy the limits; omitted historical values remain grandfathered.
- This does not add directory pagination or attempt to cancel a filesystem syscall already stuck in
  the operating system.

All passing checks below ran after the last material edit:

- Provider persistence/model refresh/validation behavior -> direct Vitest execution of the complete
  `settings_provider_accounts` module plan -> 21 files, 390 tests passed.
- Local directory batching, ordering, cap, symlink-candidate, and invalid-path behavior ->
  `npx --no-install vitest run src/main/local-fs/service.test.ts -t "lists entries|entry cap|directory symlink candidates|bounded concurrency|rejects relative paths|control characters"`
  -> 6 tests passed.
- Module ownership/classification ->
  `npx --no-install vitest run scripts/ci/validate-module-impact.test.ts scripts/ci/module-test-impact.test.ts scripts/ci/module-impact-shadow.test.ts`
  -> 3 files, 60 tests passed.
- Whitespace/patch integrity -> `git diff --cached --check` -> passed before commit.

`npm run typecheck:node` was attempted after the last material edit and is blocked by the existing
shared dependency tree: `@agentclientprotocol/claude-agent-acp` does not export `waitForMcpServers`
required by `src/main/acp/claude-agent-acp-patch.test.ts`. The worktree lockfile differs from the main
checkout, and its Prisma schema also differs from the main checkout's generated client, so project
guidance prohibits sharing that dependency tree. No private dependency installation was performed.
PR CI is authoritative for the dependency-consistent Node typecheck and complete suite. The full
local `npm test` was intentionally not run.

## Review focus

- Whether the selected limits are conservative enough for supported gateways without permitting
  material memory/disk amplification.
- Whether preserving existing oversized settings without automatic migration is the desired
  compatibility policy.
- Whether 16-entry batches strike the right balance for local and network filesystems.


ewen-poch:
Addressed the resource-limit follow-up findings in bc2ed33bc: Native Responses compatibility validation now bounds the upstream body before the proxy can buffer it; isolated Claude credential writes enforce the same 16 KiB UTF-8 limit while existing stored credentials remain readable; and application-generated provider limit errors are localized at renderer display boundaries while gateway messages remain verbatim. Added red/green regression coverage for early upstream cancellation, credential byte limits, and localized validation copy. No schema, enum, migration, persisted-field, architecture, or user-flow changes.

ewen-poch:
Addressed both P2 findings in 58888caab. Browser-captured Claude setup tokens now pass through the same 16 KiB UTF-8 controller admission check before any credential-store write, while existing stored tokens remain readable. Onboarding now localizes recognized application-generated provider resource errors through the shared renderer helper and continues to leave gateway text verbatim. Added public-boundary red/green tests for browser sign-in persistence rejection and the zh-Hans onboarding alert. No schema, migration, enum, persisted-field, architecture, or navigation-flow changes.

ewen-poch:
Addressed all three P2 findings in fb414fed4. The bounded validation adapter now keeps non-2xx upstream statuses, and the validator recovers the proxy upstream status from its existing response header so oversized 401/403 and server failures retain their classification; oversized successful responses remain a local 413. The onboarding pasted-token catch and the Settings string-error presentation now both use the recognized resource-error localization helper, while unrecognized gateway text remains unchanged. Added stable public-boundary red/green tests for oversized 401 classification plus zh-Hans onboarding and Settings error rendering. No new protocol field, schema, migration, enum, persisted state, architecture layer, or navigation-flow change.

PR #1734, opened 2026-08-25
feat(context-window): redesign calls view as per-call chart with pinned details

ewen-poch:
## Problem

The Context window dialog's Calls granularity was a text-heavy accordion: a group-by select
(turn/model/framework/none), collapsible groups, and rows of bare `In/Cache/Out` numbers. It was
hard to scan how each model call actually consumed tokens, and the header metrics surfaced
internal data-quality diagnostics (Reported/Detailed/Coverage) rather than usage facts.

## Proposed change

Rebuild the Calls view around a per-call stacked-bar chart, in the spirit of dsh-context's
context trend:

- **Session call summary** — reduced to three core metrics: total calls, total tokens
  (with an `In · Cache · Out` sub-line), and peak window (`used / size · occupancy%`).
- **Call history chart** — one stacked bar per model call (Input / Cache read / Cache write /
  Output, falling back to a single Cache segment when the read/write split is unreported),
  banded by turn with `T{n}` labels and band separators. Hover/focus previews a call, click pins
  it — same interaction model as the existing Turns history chart.
- **Pinned call details** — turn/call title with the opening prompt, agent/framework/model
  attribution, window occupancy meter, a token-mix list with proportional mini bars, and
  cache-read share. Missing values render `—`.
- **Removals** — the group-by select, the accordion list, and the Coverage/Reported-vs-Detailed
  diagnostics; the now-dead selectors (`selectContextWindowCallCoverage`,
  `groupContextWindowCallPoints`) are deleted and replaced by
  `summarizeContextWindowCallPoints` + `groupContextWindowCallsByTurn` in
  `context-window-trend.ts`.
- **Turns view alignment** — the shared `CategoryLegend` gains proportional mini bars
  (Current composition + run details), and both charts now scale the y-axis to the data instead
  of the context-window capacity, so a 1M-token window no longer squashes every bar onto the
  baseline. Capacity dashes/chips render only when capacity lands inside the data range.

Compatibility notes:

- No persistence changes, no new enum values, no schema/migration work.
- Sessions that predate call tracking keep the existing empty state; no historical-data
  compatibility concerns.

## Scope and non-goals

- Renderer + selector layer only (`ContextWindowDialog.tsx`, `context-window-trend.ts`, their
  tests, and the six locale catalogs). No main-process, IPC, or persisted-format changes.
- Turns view keeps its data model and interactions; only visual alignment (legend mini bars,
  chart scaling).
- Not in scope: per-call timestamps or per-call category composition (would require new
  persisted fields — deliberately excluded).

- Calls view renders the summary strip, banded chart, and pinned detail panel from real session
  data; hover previews, click pins, and empty/missing-data states render `—` instead of
  `undefined`/`NaN`.
- i18n guard passes for all six locales (new keys translated, orphaned keys removed).

Evidence (all run after the last material edit, in the worktree):

- `npx vitest run src/renderer/src/pages/workspace/context-window-trend.test.ts src/renderer/src/pages/workspace/ContextWindowDialog.render.test.tsx` → 18/18 passed
- `npx vitest run src/renderer/src/i18n/resources.test.ts` → 694/694 passed
- `npm run typecheck:web` → clean
- `npx eslint` on the four changed source/test files → clean
- Visual verification: `npm run dev:headless` (isolated `OPEN_SCIENCE_USER_DATA`, port 44123) +
  Web UI screenshots of both tabs against a real session with call data — chart banding, hover
  preview, pin, token-mix rows, and the data-scaled axes confirmed

Uncovered risks: full portable suite and platform lanes are left to PR CI per the verification
policy; no Electron-specific behavior is touched (dialog renders identically on Electron/Web).

## Review focus

- `context-window-trend.ts`: the two new pure selectors (summary math, turn-band grouping).
- `ContextWindowDialog.tsx`: `ContextCallChart` band layout and the capacity-visibility rule
  (charts scale to data; capacity line/chip only when within range) applied to both charts.


ewen-poch:
Addressed both review findings in b33879c:

**Capacity marker removed from the calls chart.** The bars stack billing-style totals (input + cache + output), while `contextWindowSize` limits context usage — different metrics, so the overlay was both mispositioned and usually hidden. The chart no longer plots capacity; window occupancy remains visible in the per-call details panel, where the meter compares `contextUsedTokens` against `contextWindowSize` on the same metric. The Turns view keeps its capacity marker, since those bars already plot context usage.

**Legend is now derived from the data.** The legend chips are built from the union of segments the calls actually report: providers that only return aggregate `cacheTokens` get a single `Cache` chip, and `Cache read` / `Cache write` chips appear only when the split is present. Added render-test coverage for both the aggregate and the split cases.

PR #1736, opened 2026-08-25
fix(acp): log Codex MCP failure reasons

ewen-poch:
## Problem

Codex MCP failures can carry their model-facing error text under `rawOutput.result.content` without populating ACP `content`. The ACP failure audit only inspected `content`, so logs reduced a concrete validation error to a tool call id with no actionable reason.

The reported Artifact calls were not failing on file paths: the provider emitted an empty `{}` argument object repeatedly, and the MCP server returned `-32602` because `filename` was missing. The runtime Agent received that validation result, but the application log discarded it.

## Proposed change

- preserve the existing ACP `content` diagnostic as the first choice;
- fall back to text blocks in the Codex MCP `rawOutput.result.content` shape;
- reuse the existing 300-character bound and continue excluding raw arguments, diffs, terminal output, and arbitrary raw payloads;
- add a regression fixture matching the observed empty-argument Artifact failure.

## Scope and non-goals

- No automatic filename inference or retry state. Provider/model argument generation is outside this logging fix.
- No protocol fields, architecture, data model, persisted format, enum state, or user interaction changes.
- No historical data compatibility work is required.

- Codex MCP raw-result failure reason -> `vitest run src/main/acp/session-update-projector.test.ts -t 'uses the Codex MCP raw result'` -> passed three consecutive runs
- ACP failure projection and bounded text behavior -> `vitest run src/main/acp/runtime-events.test.ts src/main/acp/session-update-projector.test.ts` -> 47 passed
- Namespace argument preservation -> `vitest run src/main/settings/native-responses-compatibility.test.ts` -> 27 passed
- Artifact bare-filename contract -> `vitest run src/main/artifacts/mcp-server.test.ts` -> 22 passed
- Node process types -> `npm run typecheck:node` -> passed
- Lint -> `npm run lint` -> passed

The complete `npm test` suite was intentionally not run locally; PR CI remains authoritative for the complete portable and platform suites.

Compatibility impact:

- `claude-code` and `opencode`: unchanged ACP `content` path remains first and is covered by the existing projection test.
- `codex-response` and `codex-bridge`: share the Codex ACP raw-result event shape covered by the new regression test.
- No launcher-, model-, subscription-, lifecycle-, cancellation-, or platform-specific behavior changes.

## Review focus

Confirm that the fallback reads only MCP result text blocks and remains bounded, while preserving the existing diagnostic path for other frameworks.

### DIFF
--- a/src/main/acp/runtime-events.test.ts
+++ b/src/main/acp/runtime-events.test.ts
@@ -624,4 +624,12 @@ describe('extractToolFailureText', () => {
       extractToolFailureText([{ type: 'terminal', terminalId: 't' } as unknown as ToolCallContent])
     ).toBeUndefined()
   })
+
+  it('does not treat arbitrary raw MCP result text as a failure reason', () => {
+    expect(
+      extractToolFailureText(undefined, {
+        result: { content: [{ type: 'text', text: 'artifact contents: api_key=do-not-log' }] }
+      })
+    ).toBeUndefined()
+  })
 })
--- a/src/main/acp/runtime-events.ts
+++ b/src/main/acp/runtime-events.ts
@@ -10,6 +10,7 @@ import {
 // Bounds how much of a failed tool's result text reaches the log, so large or sensitive tool output
 // cannot flood it. Tuned to fit a typical error message (e.g. WebFetch's domain-safety preflight).
 const TOOL_FAILURE_TEXT_LIMIT = 300
+const MCP_INPUT_VALIDATION_ERROR_PREFIX = 'MCP error -32602: Input validation error:'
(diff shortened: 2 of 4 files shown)


ewen-poch:
Addressed in de6af74b. Raw MCP result text is no longer forwarded to the warning log. The fallback now recognizes only the provider's MCP `-32602` input-validation prefix and emits the fixed diagnostic `MCP input validation failed.`, so any arbitrary or sensitive suffix remains omitted. Regression coverage now verifies both the original Codex MCP failure shape and that sensitive raw result content does not reach the diagnostic.

PR #1737, opened 2026-08-25
fix(compute): cancel poller work during shutdown

ewen-poch:
## Problem

The Compute Job runtime cleared its polling interval during application disposal but returned before already-started polling, notification, and harvest work settled. The runtime wrapper also discarded an asynchronous Poller stop result. Application shutdown could therefore report the module disposed while tracked Compute work was still using SSH, database, or event-publication resources.

Simply awaiting that work was not bounded: the default module disposal budget is one second, while polling may run for 30 seconds and SSH cancellation itself has a three-second termination boundary. Shutdown could time out and continue tearing down dependencies underneath those operations.

This affects the shared main-process runtime used by the Electron UI and headless/Web modes. MCP helper processes, preload, and renderer lifecycles do not construct this Poller.

## Proposed change

- Make `JobPoller.stop()` prevent future ticks, abort active polling and harvest operations, then drain all tracked work.
- Keep deletion `pause()` non-cancelling so the same runtime can safely resume after deletion coordination.
- Thread one lifecycle `AbortSignal` through the existing connection broker, SSH/SCP runners, and harvest dependencies; suppress cancellation-derived job status writes, including free-space-query failure paths.
- Propagate and await the Poller stop result from `createComputeJobRuntime()`.
- Give the production compute module the existing five-second normal-quit budget, which covers the transport cancellation boundary without introducing a new timeout policy.
- Add regression coverage for early disposal, active cancellation, cancelled harvest retry eligibility, and production disposal registration.

## Scope and non-goals

- No schema, data-model, persistence-format, historical-data compatibility, enum, IPC payload, or user-interaction changes.
- No new generic lifecycle framework or resource registry.
- No new timeout constant; the compute module reuses `QUIT_SHUTDOWN_BUDGET_MS`, already used by other long-running shutdown owners.

All listed successful checks ran after the last material edit.

- Early-disposal regression: failed twice before its fix at `expected true to be false`.
- Poll/harvest cancellation regression: failed twice before its fix at `expected undefined to be true`.
- Free-space/abort retry regression: failed twice before its fix because `harvestJob()` resolved and finalized.
- Production registration regression: failed twice before its fix because `compute-job-runtime` had no `disposeTimeoutMs`.
- All four regressions pass after their fixes.
- `compute_service` module-impact file list plus `harvest-engine.test.ts`, invoked directly through the canonical Vitest runtime because the worktree intentionally has no private dependency install: 32 files passed, 1 skipped; 810 tests passed, 25 skipped.
- Compute architecture gate: passed with `job-poller.ts` at 656 lines against the 660-line limit.
- Targeted ESLint and `git diff --check`: passed.
- `npm run typecheck:node` did not provide valid branch evidence because the worktree resolves the main checkout's mismatched generated Prisma Client and ACP patch. Failures remain confined to unchanged `sessionModelCallUsage` and `waitForMcpServers` consumers; there are no new Compute diagnostics. Exact-head PR CI is authoritative for typechecking.
- Full `npm test` was not run; the PR Gate runs the repository-selected exact-head plan.

The `compute_service` manifest includes the changed owners, runtime registration architecture check, and Windows-sensitive capability overlay. No renderer or Web consumer lane is needed because no interface crosses into those processes.

## Review focus

- Confirm final stop aborts the active lifecycle generation before draining tracked work.
- Confirm cancellation reaches poll SSH calls and harvest SSH/SCP transfers without persisting a false failure or consuming restart eligibility.
- Confirm the five-second compute disposal budget covers the three-second SSH cancellation boundary while retaining deletion pause/resume semantics.

### DIFF
--- a/src/main/compute/compute-service.architecture.test.ts
+++ b/src/main/compute/compute-service.architecture.test.ts
@@ -340,6 +340,17 @@ describe('Compute service architecture', () => {
     expect(backgroundProjectRecovery).toBeGreaterThan(backgroundSessionRecovery)
   })
 
+  it('gives Compute Job shutdown the transport cancellation budget', () => {
+    const source = readSource(computePaths.mainIpc)
+    const runtimeStart = source.indexOf('const jobPoller = createComputeJobRuntime')
+    const runtimeEnd = source.indexOf('const agentComputeService', runtimeStart)
+    const runtimeRegistration = source.slice(runtimeStart, runtimeEnd)
+
+    expect(runtimeStart).toBeGreaterThan(-1)
+    expect(runtimeEnd).toBeGreaterThan(runtimeStart)
+    expect(runtimeRegistration).toContain('disposeTimeoutMs: QUIT_SHUTDOWN_BUDGET_MS')
+  })
+
   it('awaits Compute Job barrier rollback when a new Project deletion aborts', () => {
     const source = readSource(computePaths.mainIpc)
     const abortStart = source.indexOf('abortProjectDeletion: async (projectId) => {')
--- a/src/main/compute/harvest-engine.test.ts
+++ b/src/main/compute/harvest-engine.test.ts
@@ -277,16 +277,19 @@ describe('harvestJob — clean harvest', () => {
     const scp = makeScpRunner()
     const { repo: jobRepo, updates } = makeJobRepo(job)
     const connectionBroker = brokerFromRunners(ssh, scp)
+    const signal = new AbortController().signal
 
     await harvestJob(job, {
       connectionBroker,
       hostRepository: makeHostRepo(host),
       jobRepository: jobRepo,
-      storageRoot
+      storageRoot,
+      signal
     })
 
     expect(connectionBroker.acquire).toHaveBeenCalledWith(job.provider_id, {
-      intent: 'job_harvest'
+      intent: 'job_harvest',
+      signal
     })
     expect(vi.mocked(ssh.run).mock.calls[0]?.[1]).toContain(
       "find ~/'.openscience/jobs/job-1' -type f"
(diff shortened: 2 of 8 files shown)


ewen-poch:
Addressed the P1 shutdown-budget finding in 3c3924219. Final stop now aborts the active poller lifecycle generation before draining; the signal reaches job_poll acquisition/run and job_harvest SSH/SCP operations, and cancellation exits without persisting false poll/harvest failures. Deletion pause remains a non-cancelling drain. The new public runtime regression failed twice before the change because both signals were undefined, then passed after signal-responsive poll and harvest promises settled via stop(). Final local compute_service coverage: 32 files passed, 1 skipped; 808 tests passed, 25 skipped.

ewen-poch:
Addressed the P2 free-space/abort race in 53739ffab. The initial free-space failure handler now checks the lifecycle signal before finalizing, matching the host lookup and enumeration paths; a shutdown cancellation therefore leaves harvestedAt/notifiedAt unset for restart recovery. The regression failed twice before the fix because harvestJob resolved and finalized, then passed after the guard with zero repository updates. Final related suite: 78 passed; final compute_service manifest plus harvest-engine: 32 files passed, 1 skipped, 809 tests passed, 25 skipped.

ewen-poch:
Addressed the P2 production disposal-budget finding in 5cbf1e078. The compute-job-runtime module now uses the existing QUIT_SHUTDOWN_BUDGET_MS (5s), which exceeds SSH's documented 2s SIGTERM grace + 1s final event boundary while preserving the repository's normal quit policy. The production-registration regression failed twice before the fix because the module had no disposeTimeoutMs, then passed after the change. Final compute_service manifest plus harvest-engine: 32 files passed, 1 skipped, 810 tests passed, 25 skipped; targeted lint and diff checks passed.

ewen-poch:
Addressed the P1 fallback-kill cancellation race in c5d451f3e. After the best-effort kill settles or throws, the poller now exits immediately when the lifecycle signal is aborted, before finishPolled can persist timeout. The job remains submitted/running and restart recovery can poll it again. The regression failed twice before the fix with one timeout updateIfStatus call, then passed with zero status writes. Final compute_service manifest plus harvest-engine: 32 files passed, 1 skipped, 811 tests passed, 25 skipped; architecture line gate is 657/660 and targeted lint/diff checks passed.

PR #1742, opened 2026-08-26
fix(elicitation): enforce form schema invariants

ewen-poch:
## Problem

Live ACP form normalization copied defaults after only shallow type checks, while the Renderer used
a looser submit predicate than Main. Invalid enum, integer, range, or formatted defaults could make
a form appear submittable before Main rejected the response.

Persisted elicitation projections also accepted duplicate field identities, contradictory
constraints, and kind-incompatible defaults. In addition,
`AgentTurnProvenanceContext.originMessageId` was declared and consumed downstream but dropped by the
durable projection sanitizer.

## Proposed change

- Share one field/value validation boundary across live request normalization, Main response
  validation, Renderer submit enablement, and persisted request restoration.
- Retain only schema-valid defaults; discard an invalid default without rejecting the otherwise
  usable form.
- Reject persisted elicitation payloads with duplicate identities, duplicate options,
  contradictory constraints, or incompatible field shapes.
- Reject oversized persisted option arrays before reading or sanitizing their entries.
- Preserve `originMessageId` when sanitizing durable elicitation provenance.
- Validate local `datetime-local` values before converting them to canonical ISO strings for Main,
  while displaying already-canonical RFC3339 values in the browser-compatible local input shape.

## Scope and non-goals

- No Prisma or database schema change, migration, IPC addition, or wire-format version.
- No new status or enum value.
- No new user-visible text, layout, or interaction flow.
- No provider-specific translation or subscription change. Claude Code, OpenCode, Codex Responses,
  and Codex bridge paths converge on the shared owner/persistence/Renderer boundaries changed here.

## Historical compatibility and persistence

- Historical invalid defaults are omitted while retaining the remaining form.
- Ambiguous or impossible elicitation payloads are omitted; their containing tool activity and
  Session remain readable.
- A pending durable response cannot be resumed from a duplicate or contradictory field model.
- Missing historical `originMessageId` remains valid because the field is optional. When present,
  it now survives Session JSON sanitization.

Every regression was first run against the unfixed implementation and failed with the reported
symptom:

- invalid live defaults were present in all five normalized pending fields;
- wrong-kind, out-of-enum, and invalid-format presets all enabled Continue;
- an RFC3339 date-time default accepted by Main left Continue disabled and could not submit;
- an oversized persisted option array was traversed before its configured bound was enforced;
- duplicate IDs, contradictory constraints, and incompatible persisted defaults were accepted;
- `originMessageId` was absent after sanitization.

- live schema normalization, persisted sanitization/restore, provenance round trip, and Renderer
  submit state -> `vitest run` on the three focused files -> 174 passed;
- framework routing, online ACP form, app-owned choice, persisted choice restoration, and Agent Frame
  provenance -> focused consumer `vitest run` -> 7 passed, 569 skipped;
- linted source -> full `eslint .` without cache -> passed;
- Node/Web typechecks -> no branch-specific errors, but both remain locally blocked by the shared
  root dependency tree. The same commands on unchanged `main` produce the identical missing
  `waitForMcpServers` export and stale Prisma `sessionModelCallUsage` errors. PR CI is authoritative
  with a clean dependency install.

Per maintainer direction, the complete `npm test` suite was not run locally.

## Review focus

- Whether invalid defaults should remain repairable annotations while duplicate/contradictory field
  identity is rejected.
- Whether the shared value predicate remains aligned with Main's canonical response validation and
  the Renderer's local date-time conversion.
- Whether preserving the existing optional `originMessageId` closes the persistence drift without
  introducing a migration requirement.

### DIFF
--- a/src/main/acp/elicitation-owner.test.ts
+++ b/src/main/acp/elicitation-owner.test.ts
@@ -210,6 +210,25 @@ describe('AcpElicitationOwner', () => {
     expect(onProjection).not.toHaveBeenCalled()
   })
 
+  it('does not rehydrate a detached choice with duplicate field identities', () => {
+    const owner = new AcpElicitationOwner({ onProjection: vi.fn() })
+
+    expect(
+      owner.restoreDetached({
+        requestId: 'choice-1',
+        sessionId: 'app-session-1',
+        toolCallId: 'tool-choice-1',
+        message: 'Choose an approach',
+        fields: [
+          { id: 'answer', label: 'First answer', kind: 'text' },
+          { id: 'answer', label: 'Second answer', kind: 'text' }
+        ],
+        durable: { kind: 'agent-user-choice', requestId: 'choice-1' }
+      })
+    ).toBeUndefined()
+    expect(owner.getPendingRequests()).toEqual([])
+  })
+
   it('parks a form request and resumes it with validated typed content', async () => {
     const onProjection = vi.fn()
     const owner = new AcpElicitationOwner({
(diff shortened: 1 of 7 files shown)


ewen-poch:
Follow-up on the sub-millisecond precision P2:

`valueForTextInput` only derives the controlled input's browser-compatible display value. The
canonical RFC3339 default remains in React state, and an unchanged submission passes that original
state value to `valueForSubmission`.

Commit 4545ef32a strengthens the public Renderer interaction regression to use
`2026-08-02T12:00:00.123456Z` and asserts that the exact six-digit value reaches `onRespond`. The
focused test passes on the PR head, so no production-code change was needed for this follow-up.

PR #1744, opened 2026-08-26
fix(net): validate downloads and link classification

ewen-poch:
## Problem

`resilientDownload` used manifest `expectedSize` for progress and short-read detection but did not enforce it as a hard upper bound. It also resumed from the local `.part` length without binding the partial to ETag/Last-Modified or validating `Content-Range`, allowing an overlong or changed response to consume unnecessary bandwidth/disk and allowing generic no-hash callers to adopt a spliced representation.

Network link classification also treated any active macOS IPv4 interface missing from the hardware-port map, and any known non-Wi-Fi port, as Ethernet. Windows/Linux intentionally remain unclassified, but the settings row rendered `unknown` with a Wi-Fi icon.

## Proposed change

- Enforce `expectedSize` against `Content-Length`, `Content-Range`, and every body chunk before writing excess bytes.
- Require valid 206 range metadata, the requested start offset, consistent range/content lengths, and a bounded representation total.
- Bind resumable partials to a strong ETag or Last-Modified through `If-Range`; restart from zero when no usable validator exists or when the server returns 200 for a changed representation.
- Persist a versioned `.part.meta` sidecar containing a SHA-256 URL fingerprint, validator, and optional expected size so a later retry call can resume safely without storing signed URLs in plaintext.
- Treat missing/malformed/mismatched historical metadata as unbound: discard the old partial and download from zero.
- Classify only explicit macOS Wi-Fi/Ethernet hardware ports; keep VPN, virtual, lookup-failure, and other port types `unknown`.
- Render `unknown` with a neutral Network icon.

## Scope and non-goals

- No manifest, IPC, database, or shared network-model fields change.
- No new enum/status values; `NetworkConnectionType` remains `wifi | ethernet | unknown`.
- The sidecar is temporary persisted data with the same lifecycle as `.part`; it is removed on success and invalidated on integrity or terminal local failures.
- Existing historical `.part` files are not resumed once because they lack trustworthy validator metadata. Completed downloads and database data are unaffected.
- This does not add route-aware Windows/Linux/macOS discovery. Cross-platform default-route and physical-link diagnostics remain a separate feature.
- No new UI copy or interaction flow.

The regression tests were written against the existing public boundaries and observed failing before the production fixes. Repeated red runs reproduced: oversized responses being adopted, excess chunks being written, missing `If-Range`, lost validators across calls, blind weak/no-validator Range retries, invalid 206 ranges being accepted, changed ETags being spliced, macOS virtual/bridge interfaces becoming Ethernet, and `unknown` rendering as Wi-Fi.

- Download bounds, validator sidecar, range validation -> `vitest run src/main/net/resilient-download.test.ts` within the final targeted set -> passed.
- macOS conservative classification -> `vitest run src/main/net/network-info.test.ts` within the final targeted set -> 10 passed; 4 pre-existing macOS-only conditional cases skipped on Windows and left for CI.
- Updater/language-pack consumers and network IPC -> targeted final set -> passed.
- Unknown settings presentation -> `NetworkPanel.render.test.tsx` within the final targeted set -> passed.
- Combined final targeted set (6 files) -> 60 passed, 4 skipped.
- `npm run lint` -> passed with no warnings after formatting.
- `npm run typecheck:node` -> blocked by existing shared dependency state (`waitForMcpServers` export and stale Prisma `sessionModelCallUsage` client); the unmodified `main` checkout produces the identical errors.
- `npm run typecheck:web` -> blocked by the same stale Prisma client; the unmodified `main` checkout produces the identical errors.

Per maintainer instruction, the full local `npm test` was not run. These changed files are not mapped to a single owner in `module-impact.json`, so PR Gate is authoritative for the expanded clean-environment and platform lanes.

## Review focus

- Hard-bound checks occur before the offending chunk write.
- Sidecar validation and cleanup cannot make an unbound partial resumable.
- 200/206 handling never appends bytes from a changed representation.
- Conservative link classification does not claim Wi-Fi/Ethernet without evidence.

### DIFF
--- a/src/main/net/network-info.test.ts
+++ b/src/main/net/network-info.test.ts
@@ -73,6 +73,36 @@ describe('parseHardwarePorts', () => {
 // The resolver shells out to networksetup, which only exists on macOS; the injected execFile
 // seam is exercised on darwin only.
 describe('createConnectionTypeResolver', () => {
+  it('keeps an active interface missing from the hardware-port table unknown', async () => {
+    const originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform')
+    Object.defineProperty(process, 'platform', { configurable: true, value: 'darwin' })
+    const execFile = vi.fn((_cmd, _args, _options, callback) => {
+      callback(null, HARDWARE_PORTS_OUTPUT, '')
+    })
+
+    try {
+      const resolve = createConnectionTypeResolver(execFile as never)
+      await expect(resolve({ utun3: [ipv4('10.8.0.2')] })).resolves.toBe('unknown')
+    } finally {
+      Object.defineProperty(process, 'platform', originalPlatform!)
+    }
+  })
+
+  it('keeps a known non-Wi-Fi and non-Ethernet hardware port unknown', async () => {
+    const originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform')
+    Object.defineProperty(process, 'platform', { configurable: true, value: 'darwin' })
+    const execFile = vi.fn((_cmd, _args, _options, callback) => {
+      callback(null, HARDWARE_PORTS_OUTPUT, '')
+    })
+
+    try {
+      const resolve = createConnectionTypeResolver(execFile as never)
+      await expect(resolve({ bridge0: [ipv4('10.0.0.3')] })).resolves.toBe('unknown')
+    } finally {
+      Object.defineProperty(process, 'platform', originalPlatform!)
+    }
+  })
+
   it.runIf(process.platform === 'darwin')('classifies a Wi-Fi hardware port as wifi', async () => {
     const execFile = vi.fn((_cmd, _args, _options, callback) => {
       callback(null, HARDWARE_PORTS_OUTPUT, '')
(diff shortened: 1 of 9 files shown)


ewen-poch:
The validator-less restart is intentional and covered by the regression test 'restarts from zero rather than resuming without a usable validator'. A same-process retry does not prove that the remote representation stayed unchanged, and without a strong ETag or Last-Modified there is no valid If-Range precondition; retaining the partial would restore the blind cross-version splice reported by this issue. Update packages still have SHA-256, but relying on the final hash would only detect the splice after wasting bandwidth and disk, and the shared helper must be safe without a caller hash. Commit 4254e27 updates the stale UpdateService/downloader comments that incorrectly described all same-session retries as resumable.

ewen-poch:
Fixed the orphaned-sidecar finding in fe6003c42. Reproduction first failed in two places: the shared helper sent Range: bytes=8- using an orphaned old sidecar, and UpdateService removed only the .part file. The shared helper now deletes metadata whenever no corresponding partial exists, and UpdateService removes both .part and .part.meta before marking a target fresh. The new regression test and all seven affected modules pass locally: 92 passed, 4 skipped.

ewen-poch:
Fixed the cleanup retry finding in 99efb5afd. The new regression test failed before the fix because deleting the unresumable .part was attempted twice after one network request. Safety-critical cleanup now marks filesystem failures terminal, so permission and lock errors surface immediately instead of entering network backoff. Final affected-module run: 93 passed, 4 skipped; lint passed.

ewen-poch:
Fixed the sidecar symlink finding in 8735ef60d. The regression test first demonstrated validator JSON overwriting a modeled symlink target. Metadata is now written to an unpredictable same-directory temp path using exclusive wx creation and atomically renamed over the sidecar path, so a pre-existing destination symlink is replaced or the operation fails without following it. Final affected-module run: 94 passed, 4 skipped; lint passed.

ewen-poch:
Fixed the mapped wired-port finding in 451465919. The new regression test first confirmed that 'USB 10/100/1000 LAN' resolved to unknown. Explicit hardware-port names containing the standalone word Ethernet or LAN now classify as ethernet; WLAN does not match, while unmapped interfaces and Thunderbolt Bridge remain unknown. Final affected-module run: 95 passed, 4 skipped; lint passed.

PR #1746, opened 2026-08-26
fix(conversation-graph): enforce persisted graph ownership

ewen-poch:
## Problem

Persisted conversation graphs accepted several internally inconsistent shapes:

- Activity Group ids and Activity/Frame/Branch/Prompt membership were not validated.
- Messages introduced on one sibling Branch could be wired into another sibling Branch head.
- Unreachable Messages and unreachable Message cycles were accepted.
- Root Frame invariants and Activity fork ownership/path constraints were incomplete.
- Validation repeatedly rebuilt Message maps and retraced paths for each Branch, producing at least Branches × Messages work plus quadratic Frame/Branch ancestry walks.

These states can cross-wire edited histories, surface Activity or Artifact ownership on the wrong path, or preserve corrupt persisted graphs until a later consumer fails closed.

## Proposed change

- Index Frame, Branch, Message, Activity, Activity Group, and Runtime Segment ids once.
- Build iterative ancestry intervals for Frame, Branch, and Message trees.
- Enforce Message introduction monotonicity, global Message reachability, root/child Frame invariants, reciprocal Activity Group membership, and fork ownership on both the child and parent Branch paths.
- Reject fork markers on root Branches.
- Reuse indexed ancestry and membership sets so validation remains linear in graph nodes and declared memberships.
- Batch existing delegated-work Activity and Activity Group events through one internal persistence mutation so reciprocal validation never observes a dangling intermediate reference.

## Scope and non-goals

- No user interaction or UI changes.
- No new persisted data fields, schema version, data model shape, event kind, or enum value.
- The internal `applyAgentEvent` command input now accepts either one existing event or an array of existing events; the coordinator facade and architecture boundaries are unchanged.
- No automatic repair or migration.
- The per-sync structured clone remains unchanged. A local 2,000-Message/500-Branch measurement put clone median at 2.272 ms, validator median at 0.731 ms, and the full synchronization boundary at 4.257 ms, so clone removal did not have a stable performance regression to justify changing mutation semantics.

- Historical compatibility: valid schema-v1 graphs, including the existing legacy Activity placement behavior, remain accepted.
- Historical data risk: already-corrupt graphs matching the newly rejected sibling-path, orphan, root, fork, or Activity Group inconsistencies will now fail persisted-graph validation. They are not rewritten or repaired.
- Staged terminal compatibility: grouped terminal Activities and their Group now commit atomically, preserving the existing workflow under strict reciprocal validation.
- New enum states: none.
- Persistence impact: the serialized shape is unchanged; decode/import/migration boundaries now reject additional invalid references and ownership relationships.

## Validation

Red/green regression coverage was added through the exported validator, existing public graph mutation APIs, and the delegated-work terminal staging boundary. Each reported case was observed failing on the unfixed implementation before the production change. The AI review findings were also reproduced before their fixes:

- Grouped terminal staging failed with `Activity Group membership is invalid` before atomic batching.
- Child-introduced Message/Activity fork markers and root-Branch fork markers passed validation before parent-path enforcement; all three regression tests failed with `expected [Function] to throw`.

- Focused/module Vitest: 15 files, 620 tests passed.
  - Conversation graph, session decode/import, Artifact visibility, migration, and Message snapshot validation.
  - Delegated-work adapter, runtime transcript, coordinator owner/contract/architecture tests.
  - Session persistence artifact-finalization and deletion integrations.
  - Renderer session-store graph projection and editing behavior.
- Node TypeScript check passed.
- Web TypeScript check passed.
- Full-repository ESLint passed with cache disabled; later test-only fixture changes passed targeted lint/static CI checks.
- Prettier check and `git diff --check` passed.
- Deterministic performance regression: id reads were 302,695 before the rewrite and are now below the 15,980 linear budget.

PR CI exposed existing fixtures that encoded newly forbidden ownership shapes. They were reproduced locally and changed to model valid graph ancestry while preserving the behavior each test originally exercised. The latest AI PR Review verdict is mergeable with no actionable findings.

The module-impact classifier conservatively requested the complete suite because `src/shared/artifact-visibility.test.ts` is not registered to a module owner. Per task scope, the complete local `npm run test` was not run; PR CI is authoritative.

## Review focus

- Whether the newly rejected persisted shapes need a separate repair/migration policy.
- Whether the accepted legacy Activity placement semantics are preserved while sibling Branch ownership is rejected.
- Atomic grouped terminal staging through the existing internal event command.
- Linear ancestry/reachability behavior for deep and highly branched conversations.

### DIFF
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -413,7 +413,7 @@ jobs:
       - name: Test complete suite shard
         id: unit_macos_shard
         continue-on-error: true
-        run: npx vitest run --coverage --coverage.reporter=text-summary --passWithNoTests --shard=${{ matrix.shard }}/2 --reporter=blob --outputFile=vitest-reports/blob-${{ matrix.shard }}.json
+        run: npx vitest run --coverage --coverage.reporter=text-summary --shard=${{ matrix.shard }}/2 --reporter=blob --outputFile=vitest-reports/blob-${{ matrix.shard }}.json
       - name: Upload full-suite blob report
         if: ${{ always() }}
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -413,7 +413,7 @@ jobs:
       - name: Test complete suite shard
         id: unit_macos_shard
         continue-on-error: true
-        run: npx vitest run --coverage --coverage.reporter=text-summary --shard=${{ matrix.shard }}/2 --reporter=blob --outputFile=vitest-reports/blob-${{ matrix.shard }}.json
+        run: npx vitest run --coverage --coverage.reporter=text-summary --passWithNoTests --shard=${{ matrix.shard }}/2 --reporter=blob --outputFile=vitest-reports/blob-${{ matrix.shard }}.json
       - name: Upload full-suite blob report
         if: ${{ always() }}
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
--- a/src/main/artifacts/provenance-message-snapshot.test.ts
+++ b/src/main/artifacts/provenance-message-snapshot.test.ts
@@ -170,7 +170,24 @@ describe('Provenance Message snapshots', () => {
     if (!staleSession.conversationGraph) {
       throw new Error('Expected the Session conversation graph.')
     }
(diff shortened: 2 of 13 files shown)


ewen-poch:
Addressed the fork-boundary finding in de85ea8aa. Fork markers are now rejected on root branches, and forkMessageId/forkActivityId must resolve on the parent Branch's actual Message path in addition to the child path. Added red/green regressions for child-introduced Message and Activity markers plus a root-Branch marker; 15 related test files pass (620 tests). This does not add fields, enum states, persistence shape, or interaction changes.

PR #1748, opened 2026-08-26
feat(notebook): add live namespace browser

ewen-poch:
## Problem

Notebook users can execute Python and R code in persistent in-memory kernels, but cannot inspect the live namespace the way RStudio/Posit exposes an Environment. Persisted notebook history also outlives the interpreter, so a variable surface must not imply that values still exist after app or kernel restart.

## Proposed change

- Add a read-only `notebook:inspect-namespace` path across the Python/R loops, kernel executor, runtime workflow, Electron IPC, preload API, and local/remote Web contract.
- Inspect only an already-live kernel; opening Variables never starts a kernel or restores variables from history.
- Return bounded metadata: name, type, optional shallow size/shape, and a safe preview. Responses are capped at 500 variables, 512 preview bytes, and 256 KiB total.
- Avoid arbitrary user display code: unknown objects receive a type summary, Python containers receive shallow count summaries, and R active bindings are reported without evaluation.
- Add a second-level Variables view in Notebook with filtering, an opt-in private-name toggle, manual refresh, stale/refreshing/unavailable states, and all six renderer translations.

### Snapshot and visibility semantics

- Refresh on opening Variables, manual refresh, changing the private-name toggle, and once after notebook execution settles while the view is open.
- Runtime change events mark the visible snapshot stale; repeated events are coalesced and the read waits for the selected kernel to return to idle. There is no polling.
- Kernel/app restart, termination, inactive historical environments, target changes, epoch changes, and superseded requests cannot publish an old snapshot.
- User-bound values such as `fig`, `ax`, and `plt` are shown. Python `_name` and R `.name` bindings are hidden by default and available through the toggle. Library-internal registries are not scanned.

## Scope and non-goals

- No recursive object explorer, dataframe pagination, editable values, polling, or library-global discovery.
- No dependency additions.
- No historical compatibility migration is needed. Existing notebook history remains readable, but Variables is unavailable until a live kernel has executed in the current app process.
- State additions are limited to the new non-persistent response union (`available` / `unavailable`, with `kernel-not-live` / `kernel-restarted`) and renderer-local view states (`empty`, `loading`, `ready`, `stale`, `refreshing`, `unavailable`, `error`).
- No namespace snapshot or variable data is persisted.

- Python live namespace, mutation, private filtering, large values, UTF-8, and hostile `repr` safety -> `RUN_KERNEL=1 OPEN_SCIENCE_TEST_PY_ENV=/opt/homebrew/bin/python3 npm test -- src/main/notebook/python-loop.integration.test.ts` -> 7 passed.
- R live namespace, mutation, private filtering, active binding safety, large values, and C-locale UTF-8 transport -> `RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local npm test -- src/main/notebook/r-loop.integration.test.ts` -> 46 passed.
- Runtime/executor/protocol behavior -> focused Notebook Vitest set -> new tests passed. One pre-existing R cancellation timing case completed before cancellation in the combined run; its isolated rerun passed.
- Application command, IPC, and workflow routing -> focused Vitest set -> 16 passed.
- Variables UI, filtering/private reload, and translations -> focused Vitest set -> 734 passed.
- Electron/Web renderer contract catalogs and generated API map -> focused Vitest set -> 34 passed; `npm run check:web-api-map` passed.
- Static validation -> `npm run lint`, `npm run typecheck:node`, and `npm run typecheck:web` -> passed.

Full `npm test` was intentionally not run; PR Gate remains authoritative for the complete portable and platform lanes.

## Review focus

- The inspector must never lazily spawn a kernel or invoke arbitrary object preview behavior.
- Epoch/request/target invalidation must prevent stale data after restart or target changes.
- Snapshot bounds and R base64 text transport must remain enforced at the loop/protocol boundary.


ewen-poch:
Addressed the latest Codex review findings in 10bd26477:

- R character previews now slice to a bounded character prefix before charToRaw, preventing a full raw copy of very large strings while retaining the 512-byte UTF-8 cap.
- Python namespace inspection now ignores non-string globals keys before private-name filtering and sorting.
- Python protocol responses now use the same compact JSON separators used by namespace entry budgeting, so the 256 KiB cap is enforced against matching serialization.

Regression coverage was added for a one-million-character R value and a non-string Python globals key. Focused validation: Python loop 8/8, R loop 46/46, lint, and Node typecheck.

ewen-poch:
Addressed the untracked R promise finding in ec1dcce5e with a conservative read policy:

- Namespace inspection only calls get() for bindings confirmed eager by a completed direct top-level assignment.
- Active bindings, tracked delayedAssign bindings, and bindings created through indirect or otherwise unprovable paths are never evaluated; unknown bindings render as binding / <not evaluated>.
- A later completed direct assignment upgrades that name back to normal type and bounded-preview inspection.

This keeps compatibility with base R plus jsonlite and avoids a new runtime dependency or native/internal R inspection. Regression coverage includes delayedAssign through an alias, an indirect eager assign(), and direct reassignment of both names. Focused validation: R loop 46/46, lint, and Node typecheck.

ewen-poch:
Addressed both bootstrap-shadow findings in 31c0741c5:

- Python now stores the original bootstrap binding objects and hides a name only while its current value is still that original object. User assignments such as sys = 1 and json = "user json" are visible.
- R now reveals bootstrap names after a confirmed direct assignment, tracked delayed assignment, or active-binding replacement. The request loop already holds private references to trusted run/con functions and connections, so user shadows do not affect protocol execution.

Regression coverage now assigns the reported Python and R bootstrap names and verifies their values in namespace snapshots. Validation: Python loop 8/8, focused R namespace test passed, lint passed, and Node typecheck passed. A full local R run reached 44/46; the two failures were pre-existing SIGINT read-framing timing tests that exited before replying and do not call namespace inspection.

ewen-poch:
Addressed the spoofed Python property finding in 70ad1b9ff:

- Namespace inspection no longer dereferences library-specific shape or axes properties based on spoofable type metadata.
- Built-in containers retain safe exact-type length summaries; NumPy, Pandas, Matplotlib, and custom objects still show their type and bounded generic preview without touching user properties.
- Type metadata reads now bypass custom metaclass attribute overrides through type.__getattribute__.

Regression coverage uses a numpy.ndarray-spoofing class whose shape property records access and verifies that inspection leaves the access count at zero. Validation: Python loop 9/9, lint, and Node typecheck.

ewen-poch:
Addressed both findings in 471fe3937:

- Variable names are now UTF-8 bounded to 1,024 bytes before namespace-entry serialization, with an ellipsis when truncated. The integration test creates a 2,000,000-character global name and verifies both the per-name and whole-response budgets.
- Namespace snapshots are now hidden immediately and cleared when the selected kernel is lost or the view becomes historical. In-flight inspections are invalidated, queued refreshes are discarded, and an open variables view can reload after the kernel becomes live again. A renderer test covers the live-to-terminated transition.

Focused validation: Python loop + NotebookPreview tests (50/50), ESLint, Node typecheck, and Web typecheck.

ewen-poch:
Addressed both findings in 6418b66ab:

- R namespace inspection is now strictly non-forcing. It never calls `get()` or reads value attributes, shapes, sizes, or previews. The Type column reports only binding metadata (`active binding`, tracked `lazy binding`, or `binding`), and Preview is empty. The regression test overwrites a previously direct-written name through an alias of `delayedAssign` and verifies inspection does not evaluate the promise.
- Restart now invalidates `namespaceLoadKey` and queues a refresh for an open Variables view. The queued refresh runs after the restarted kernel reaches idle, while the existing request-id cancellation still rejects old responses. Renderer coverage verifies the automatic second inspection.

Focused validation: R non-forcing namespace test (1/1), NotebookPreview tests (42/42), ESLint, Node typecheck, and Web typecheck.

ewen-poch:
Addressed the truncated-name row-key finding in 1a836cbc4. Namespace rows now use their bounded sorted position together with the display name, so distinct variables that truncate to the same visible prefix cannot produce duplicate React keys. The table rows have no local row state, so this avoids expanding the namespace contract with an identifier or hash.

Focused validation: NotebookPreview tests (42/42), ESLint, and Web typecheck.

ewen-poch:
Addressed the `assign()` bootstrap-shadow finding in e3609bb11. The existing R namespace syntax tracker now records literal-name writes from top-level `assign()` and `base::assign()` calls when their target is explicitly `.GlobalEnv`/`globalenv()`/position 1, or when the top-level default target is used. This keeps user-created shadows such as `assign("run", 1, envir = .GlobalEnv)` visible without reading or forcing any binding.

The namespace regression now creates the `run` shadow through `assign()` and verifies it remains listed. Focused validation: R non-forcing namespace test (1/1), ESLint, and Node typecheck.

ewen-poch:
Addressed both findings in 4e63ce45c:

- `inspectNamespace` now passes through the same workflow trust-boundary sanitizer as renderer execution requests. Forged `provenanceContext`, `executionInvocationId`, registered inputs, and input lease fields are removed before runtime lane selection. The workflow regression verifies the runtime receives only public namespace fields.
- R bootstrap-shadow tracking now recognizes `->` and `->>` direct assignments. The non-forcing namespace regression creates `run` and `con` shadows through both right-assignment forms and verifies they remain listed.

Focused validation: notebook workflows (4/4), R non-forcing namespace test (1/1), ESLint, and Node typecheck.

PR #1779, opened 2026-08-26
fix(session-persistence): bound live Session save cadence

ewen-poch:
## Problem

The attached Windows log covers long-running Sessions whose persisted Session directory grew from about 356 MB to 380 MB. During a live run, renderer presentation updates can arrive at roughly 30 fps. The existing latest-wins persistence queue coalesced only while an IPC or disk write was still backpressured, so a fast write path could still serialize, clone, send, stringify, and atomically rewrite the entire Session once per frame.

The deterministic regression reproduced that amplification at the existing ordered-persistence boundary: 30 updates in one second produced 30 whole-Session writes on the pre-fix code. It failed three consecutive runs with the report-consistent reason (expected 3 writes, received 30).

The log does not contain an OOM, disk-corruption, or kernel-level cause, so this PR does not claim that Session persistence alone caused the full Windows freeze or the unrecoverable result folder. It removes one evidenced, avoidable source of sustained CPU, memory-allocation, IPC, and disk pressure.

## Proposed change

- Bound automatic latest-Session snapshots to a 500 ms cadence while retaining latest-wins coalescing.
- Keep one replaceable waiting snapshot per Session target, so alternating active Sessions cannot build a frame-by-frame timer backlog.
- Treat explicit saves, retry/recovery, manifest writes, and flush as barriers that release every waiting target immediately and preserve existing ordering.
- Invalidate delayed snapshots when a newly hydrated store generation starts, preventing pre-reload local state from writing after authoritative hydration.

At a 30 fps update rate this reduces steady-state whole-Session rewrites from up to 30 per second to at most 2 per second. The leading and trailing boundaries make the one-second regression observe 3 writes.

## Scope and non-goals

- No data fields, schema, migration, enum values, architecture boundary, data model, or user interaction changes.
- Internal transient state only: a per-target pending-save map and hydration generation counter inside the existing ordered persistence module. Neither is exposed or persisted.
- Persistence impact: the existing Session JSON format and atomic temp-file rename remain unchanged. Automatic intermediate snapshots may be delayed by at most 500 ms; explicit barriers and quit flush remain immediate and still wait for the newest queued snapshot to finish.
- Historical compatibility: none required; existing Session data is read and written unchanged.
- No aggregate Session-size cap, truncation, compaction, or deletion behavior.
- No OS process quota or notebook/agent compute throttling. The supplied log has no process telemetry proving that such a broader mechanism is warranted.
- No claim to repair an already corrupted external result directory.

This is intentionally a narrow fix rather than a persistence redesign: the expensive behavior is reproducible at a public ordering boundary, and the change substantially reduces pressure without changing durable data contracts.

- Fast 30 fps whole-Session persistence is bounded, alternating Session targets flush without draining cadence timers, and a new hydration generation invalidates stale delayed writes -> `npm test -- --run src/renderer/src/lib/session-persistence/session-persistence.test.ts src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/hooks/useQuitPersistenceFlush.test.ts` -> 3 files passed, 78 tests passed.
- Retry, revision-conflict, manifest, hydration, and quit-flush consumers retain their behavior -> same targeted command -> passed.
- Renderer types remain valid -> `npm run typecheck:web` -> passed.
- Source style remains valid -> `npm run lint` -> passed with no findings.
- Patch whitespace remains valid -> `git diff --check` -> passed.

Per the requested workflow, the full portable suite was not run locally; exact-head PR CI is authoritative. The remaining uncovered risk is an end-to-end Windows resource-exhaustion reproduction: the log lacks per-process CPU/memory/I/O telemetry, so CI and future telemetry may identify an additional independent compute-heavy cause.

## Review focus

- Whether 500 ms is the appropriate bounded persistence cadence for live presentation updates.
- Whether per-target coalescing bounds alternating Session streams and every explicit ordering barrier releases all cadence waits.
- Whether hydration-generation invalidation prevents stale pre-reload snapshots without surfacing a false write failure.
- Whether the regression adequately represents the observed amplification without adding a production-only test seam.

Independent review is still required before marking the final Test Impact Set verified under CONTRIBUTING.md.

### DIFF
--- a/src/renderer/src/lib/session-persistence/session-persistence.test.ts
+++ b/src/renderer/src/lib/session-persistence/session-persistence.test.ts
@@ -1144,6 +1144,60 @@ describe('renderer session persistence bridge', () => {
     }
   })
 
+  it('flushes a fast alternating multi-Session stream without draining cadence timers', async () => {
+    vi.useFakeTimers()
+    vi.setSystemTime(0)
+    let flushing: Promise<void> | undefined
+    try {
+      const sessions = [
+        createPersistedSession({ id: 'session-1' }),
+        createPersistedSession({ id: 'session-2' })
+      ]
+      const writeLatest = vi.fn(async (session: PersistedChatSession) => session)
+      const persistence = createOrderedSessionPersistence(createApi())
+
+      for (let frame = 0; frame < 30; frame += 1) {
+        const session = sessions[frame % sessions.length]
+        void persistence.saveLatestSession(`session:${session.id}`, () => writeLatest(session))
+        await vi.advanceTimersByTimeAsync(33)
+      }
+
+      let flushed = false
+      flushing = persistence.flush().then(() => {
+        flushed = true
+      })
+      await vi.advanceTimersByTimeAsync(0)
+
+      expect(flushed).toBe(true)
+      expect(writeLatest.mock.calls.length).toBeLessThanOrEqual(5)
+    } finally {
+      await vi.runAllTimersAsync()
+      await flushing
+      vi.useRealTimers()
+    }
+  })
+
+  it('invalidates a delayed save when a new hydration generation starts', async () => {
+    const session = createPersistedSession({ revision: 1, title: 'Old local title' })
+    const authority = createPersistedSession({ revision: 2, title: 'Hydrated title' })
+    const persistence = createOrderedSessionPersistence(createApi())
+    persistence.seedAcknowledgedSessions([session])
+    await persistence.saveLatestSession('session:session-1', async () => session)
+
+    const staleWrite = vi.fn(async () => session)
+    const staleSave = persistence.saveLatestSession('session:session-1', staleWrite)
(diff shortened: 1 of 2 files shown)


ewen-poch:

- Replaced the single pending-latest slot with per-target pending entries. Alternating Session streams now coalesce independently, and explicit barriers plus flush release every pending cadence wait. Added a two-Session 30 fps regression that verifies flush completes without advancing cadence timers and keeps writes bounded.
- Added hydration-generation invalidation. Delayed automatic snapshots from a previous store generation now exit before invoking their write task, and the internal superseded signal is ignored by the retired store saver instead of surfacing a false write failure. Added a regression that proves stale pre-hydration state is not written.

Focused persistence, render-consumer, and quit-flush tests pass (78/78), along with renderer typecheck and lint. This remains internal transient scheduler state only; there are no schema, persisted-format, data-model, architecture-boundary, or interaction changes.

PR #1789, opened 2026-08-26
perf(markdown): lazy-load rendering plugins

ewen-poch:
## Problem

`AgentMarkdown` and workspace code surfaces statically imported the Streamdown Mermaid and code plugins. As a result, ordinary prose conversations pulled Mermaid, Shiki, and theme runtimes into the eagerly loaded renderer graph.

## Proposed change

- Detect fenced code and Mermaid fences before requesting optional rendering plugins.
- Dynamically load Mermaid only when a Mermaid fence is present.
- Dynamically load the code highlighter and themes only when fenced code is present.
- Keep Streamdown mounted while plugins load so Markdown and code source remain readable.
- Reuse the same lazy code-highlighter hook in workspace code surfaces to remove their static imports.
- Preserve the existing Mermaid error panel and source-code fallback on load or render failure.
- Keep Mermaid and code highlighting in separate optional runtime chunks.

## Scope and non-goals

- No additional Markdown rendering system.
- Math and CJK plugin behavior is unchanged.
- No user-visible copy or translation changes.
- No historical-data compatibility handling is required.
- No enum or durable state is added.
- No persistence or data migration is involved.

- `npm test -- src/renderer/src/components/streamdown src/renderer/src/pages/workspace/WorkspaceToolCodeBlock.test.tsx src/renderer/src/pages/workspace/previews/PreviewFileContent.test.tsx src/renderer/src/pages/workspace/workspace-components.test.tsx` — 15 files / 160 tests passed.
- `npm run lint` — passed.
- `npm run build:web` — passed.
- The production build entry has no static Mermaid dependency and dynamically imports distinct code-highlighter and Mermaid runtime entries.

The full cross-platform suite is left to PR CI as requested.

## Review focus

- Fence detection during streaming transitions.
- Source fallback while plugins load or fail.
- Production chunk boundaries for Mermaid versus Shiki/themes.

### DIFF
--- a/src/renderer/src/components/streamdown/AgentMarkdown.lazy-failure.test.tsx
+++ b/src/renderer/src/components/streamdown/AgentMarkdown.lazy-failure.test.tsx
@@ -1,0 +1,55 @@
+// @vitest-environment jsdom
+import { act, type PropsWithChildren } from 'react'
+import { createRoot } from 'react-dom/client'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+;(globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = true
+
+const harness = vi.hoisted(() => ({
+  plugins: undefined as Record<string, unknown> | undefined
+}))
+
+vi.mock('./code-highlighter-runtime', () => ({ code: { name: 'shiki' } }))
+vi.mock('@streamdown/cjk', () => ({ cjk: {} }))
+vi.mock('@streamdown/math', () => ({ createMathPlugin: () => ({}) }))
+vi.mock('./mermaid-runtime', () => {
+  throw new Error('Mermaid chunk unavailable')
+})
+vi.mock('streamdown', () => ({
+  Streamdown: ({
+    children,
+    plugins
+  }: PropsWithChildren<{ plugins?: Record<string, unknown> }>): React.JSX.Element => {
+    harness.plugins = plugins
+    return <div data-testid="markdown-source">{children}</div>
+  }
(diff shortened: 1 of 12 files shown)


ewen-poch:
Addressed the nested-fence detection issue in cf4a0f1.

The lazy-plugin scanner now removes blockquote prefixes at the fence's recorded depth and removes a list marker only while detecting an opener. This lets list-nested code and blockquoted Mermaid fences request their plugins without treating list-like source lines inside an open fence as closing markers.

Added regression coverage for a list-nested TypeScript fence and a blockquoted Mermaid fence. The focused test passes (16/16), ESLint reports no issues, and `typecheck:web` passes.

ewen-poch:
Addressed the indented list-continuation case in ef5a04b.

The scanner now records the active list item's content indentation and removes only that container indentation before tracking a nested fence. The same recorded indentation is applied while the fence is open, so its closing marker is recognized without reinterpreting root-level four-space-indented text as a Mermaid fence.

Added regression coverage for code and Mermaid fences beginning on indented continuation lines below list items. The focused test passes (17/17), ESLint reports no issues, and `typecheck:web` passes.

ewen-poch:
Addressed deeply nested list containers in 0309439.

The scanner now keeps a stack of active list content indents. It strips the deepest applicable continuation indent, re-evaluates nested list markers, and records the resulting indent for the fence lifetime. This covers `outer list -> inner list -> fence` while preserving dedent behavior for sibling and root content.

The Mermaid regression case now uses the deeply nested structure from the review. The focused test passes (17/17), ESLint reports no issues, `typecheck:web` passes, and the diff check is clean.

PR #1805, opened 2026-08-27
feat(compute): enable Windows SSH password authentication

ewen-poch:
## Problem

Compute SSH password authentication was hard-disabled on Windows in two places even though Windows OpenSSH supports `SSH_ASKPASS` and Electron `safeStorage` provides the existing DPAPI-backed credential boundary. As a result, a Windows user could save neither a Compute password nor use it for SSH connections.

## Proposed change

- Allow the Compute credential vault to use an available Windows `safeStorage` backend.
- Add a Windows `SSH_ASKPASS` bridge that runs the app executable in Electron Node mode and exchanges one password through a random named pipe.
- Preserve the existing one-shot capability, expected-account prompt matching, and fail-closed handling for passphrases, proxy prompts, repeated prompts, keyboard-interactive, OTP, and MFA flows.
- Keep unrelated environment variables and the password itself out of the askpass child environment and argv.

## Scope and non-goals

- Scope: Windows Compute SSH account-password authentication.
- Unix socket-based askpass behavior remains unchanged.
- No keyboard-interactive, OTP/MFA, proxy-password, key-passphrase, or arbitrary-prompt support is added.
- This intentionally uses the same Windows-account trust boundary as existing provider credentials stored with Electron `safeStorage`; it does not attempt to isolate secrets from arbitrary code already running as the same Windows user.

- **Historical compatibility:** no migration is required. Existing Windows records remain readable where DPAPI can decrypt them. Ciphertext copied from macOS/Linux, another Windows user, or another machine remains unavailable and must be reset, matching current `safeStorage` behavior.
- **State/enums:** no enum values or persisted states are added.
- **Persistence:** no schema or serialized-format changes. `ComputeCredential.ciphertext` continues to hold the bytes returned by Electron `safeStorage`; Windows now uses its DPAPI backend instead of rejecting the platform.

## Acceptance criteria

- Windows reports password storage available when Electron `safeStorage` encryption is available.
- A matching target-password prompt receives the password once through the named pipe.
- The password does not appear in child argv or environment variables.
- Unsupported or repeated interactive prompts fail closed.
- Existing non-Windows behavior remains covered by the same tests.

## Validation

- `npx vitest run` over the Compute module test plan's 32 files: **31 files passed, 1 skipped; 797 tests passed, 15 skipped**.
- Password-auth architecture/release guards: **1 passed, 1 skipped** (real-host SSH gate skipped because its opt-in environment was not configured).
- Focused Windows credential-vault, connection-broker, and architecture tests: **3 files / 63 tests passed**.
- Windows helper executed successfully through the repository Electron 39 binary in Node mode: **exit 0**.
- `npm run lint`: **passed**.
- `node --check resources/compute-askpass-win.cjs`: **passed**.
- `npm run typecheck:node`: blocked by the shared checkout dependency state (stale Prisma client and an unrelated `@zed-industries/claude-agent-acp` export mismatch); all reported diagnostics are outside this diff. PR CI is the authoritative clean-environment check.
- Full `npm test` was intentionally not run; PR CI covers the complete suite.

## Review focus

- Windows OpenSSH/Electron `SSH_ASKPASS` process behavior.
- Named-pipe request framing and one-shot capability enforcement.
- Preservation of the existing credential-storage trust boundary and Unix behavior.

### DIFF
--- a/resources/compute-askpass-win.cjs
+++ b/resources/compute-askpass-win.cjs
@@ -1,0 +1,56 @@
+'use strict'
+
+// Windows OpenSSH starts SSH_ASKPASS as an executable and passes the prompt as argv[1]. Running the
+// app executable in Electron Node mode makes Node interpret that prompt as its script path, so this
+// preload completes the one-shot named-pipe exchange synchronously and exits before script loading.
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const fs = require('node:fs')
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const path = require('node:path')
+
+let descriptor
+let response = ''
+let password
+const buffer = Buffer.allocUnsafe(4096)
+
+try {
+  const socketPath = process.env.OPEN_SCIENCE_ASKPASS_SOCKET
+  const capability = process.env.OPEN_SCIENCE_ASKPASS_CAPABILITY
+  const promptArgument = process.argv[1] || ''
+  const prompt = path.isAbsolute(promptArgument)
+    ? path.relative(process.cwd(), promptArgument)
+    : promptArgument
+
+  if (!socketPath || !capability || !prompt) throw new Error('Missing askpass capability.')
+
+  descriptor = fs.openSync(socketPath, 'r+')
+  const request = Buffer.from(JSON.stringify({ capability, prompt }), 'utf8')
+  let written = 0
+  while (written < request.length) {
+    written += fs.writeSync(descriptor, request, written, request.length - written)
+  }
+  request.fill(0)
+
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed the UTF-8 pipe-boundary finding in 081194b55.

- Added a regression case that places the first byte of a multibyte character at the end of the helper's 4096-byte read buffer; it failed against the original implementation with replacement characters.
- Switched the helper to Node's streaming `StringDecoder` and verified the same case returns the exact password.
- Re-ran the affected Compute test set: 3 files, 63 tests passed; Prettier and ESLint also passed.

PR #1814, opened 2026-08-27
test(workspace): honor OpenAI Responses protocol in picker tests

ewen-poch:
## Problem

Recently merged PRs that ran the full macOS Vitest suite (`#1807`, `#1810`) failed CI on `ComposerModelPicker.render.test.tsx`. Targeted PRs that skipped the full suite (`#1811`) looked green, so the breakage landed on `main`.

`#1807` started resolving official vendor models against their documented protocols. Official OpenAI catalog entries are Responses-only. The picker tests still used those ids under the store's initial Claude Code fixture, which only speaks Anthropic, so the trigger rendered as "No compatible model" instead of the selectable catalog those tests assert.

## Proposed change

- Drive the OpenAI catalog UI tests under a Responses-capable OpenCode fixture (Anthropic remains included so mixed OpenAI + custom Gateway cases stay selectable).
- Add a regression that official OpenAI Responses models warn under Claude Code instead of appearing selectable.

## Scope and non-goals

- Tests only. No product, persistence, or catalog changes.
- Windows E2E failures on recent PRs are a separate pattern (see review notes) and are not addressed here.

- Full-suite classification no longer fails these five picker tests for the `#1807` protocol change.
- Claude Code + official OpenAI remains a warning, matching the documented protocol.

### Evidence (after last material edit)

| --- | --- | --- |
| Picker OpenAI protocol fixtures | `npm test -- src/renderer/src/pages/workspace/ComposerModelPicker.render.test.tsx` | 22 passed |
| Catalog protocol contract | `npm test -- src/shared/configured-model-catalog.test.ts` | 6 passed |
| Web types | `npm run typecheck:web` | passed |
| Lint of the changed test | `npx eslint src/renderer/src/pages/workspace/ComposerModelPicker.render.test.tsx` | no issues |

Excluded: full `npm test` (PR Gate remains authoritative), Windows/macOS E2E (no product/UI change), `typecheck:node` (renderer test only).

## Review focus

- Confirm the Claude Code incompatibility warning is the intended `#1807` product behavior, not a test-only workaround.
- Historical compatibility of stored official OpenAI `apiEndpoints` is unchanged here and called out in the discussion for a product decision if needed.

### DIFF
--- a/src/renderer/src/pages/workspace/ComposerModelPicker.render.test.tsx
+++ b/src/renderer/src/pages/workspace/ComposerModelPicker.render.test.tsx
@@ -88,19 +88,20 @@ const provider = (overrides: Partial<ProviderView>): ProviderView => ({
 })
 
 // Official OpenAI models are documented as Responses-only. Tests that pick those catalog ids must
-// run under a Responses-capable framework; the store's initial Claude Code state only speaks
-// Anthropic, so the picker would otherwise hide the trigger as "No compatible model".
-// Include Anthropic as well so mixed-catalog cases (OpenAI + a custom Gateway) stay selectable.
-const responsesCapableFramework: {
-  agentFrameworkId: 'opencode'
+// run under Codex, the framework that actually speaks Responses. The store's initial Claude Code
+// state only speaks Anthropic, so the picker would otherwise hide the trigger as
+// "No compatible model". OpenCode is not used here: its adapter only drives Anthropic and Chat
+// Completions, even if a fixture claimed otherwise.
+const codexFramework: {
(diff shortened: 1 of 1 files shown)


ewen-poch:
Addressed the Codex review in a837c23e.

The picker fixture now uses the real Codex descriptor (`supportedApiTypes: ['responses']`) instead of advertising Responses on OpenCode. The mixed OpenAI + Gateway case gives the custom Gateway `apiEndpoints: ['openai']` so it stays selectable through the Codex Chat Completions bridge, matching the existing DeepSeek Codex test in the same file.

PR #1819, opened 2026-08-27
test(architecture): relax implementation-detail assertions

ewen-poch:
## Problem

Architecture tests rejected behavior-preserving internal refactors by fixing physical line counts, private field and method inventories, constructor locations, complete public method inventories, and single-return facade bodies. A one-line formatting-only change reproduced the failure deterministically: the Notebook completion gate failed with `expected 1251 to be <= 1250`.

## Proposed change

- Remove fixed line-count gates and exact private implementation inventories.
- Replace constructor-location assertions with one-owner-construction-site boundaries.
- Remove single-return facade delegation requirements.
- Preserve public export contracts, write/read separation, forbidden reverse dependencies, private-owner import boundaries, shared transaction policy, and cross-surface capability contracts.
- Replace the arrow-property shape check for `handleJobUpdated` with an observable detached-callback behavior test.
- Add fixture coverage proving field-initializer owner composition remains valid while duplicate owners and mutable module state remain detectable.

## Scope and non-goals

Only architecture and public-behavior tests change. Production code, user-visible behavior, persisted formats, historical compatibility, data relationships, and state/enum inventories are unchanged.

Checks below ran against the final committed state:

- Architecture/refactor tolerance -> targeted five-file Vitest run -> 5 files, 30 tests passed.
- Compute owner/contract/consumer set -> module-impact manifest file set -> 31 files passed, 1 skipped; 799 tests passed, 15 skipped.
- Project Files owner/contract/consumer set -> module-impact manifest file set -> 6 files passed; 233 tests passed, 1 skipped.
- Uploads owner/contract/consumer set -> module-impact manifest file set -> 8 files and 129 tests passed; two existing Windows symlink fixtures were blocked during setup by `EPERM` before entering product logic.
- Module-impact manifest -> `scripts/ci/validate-module-impact.test.ts` -> 9 tests passed.
- Changed files -> ESLint -> passed.
- Full repository ESLint -> passed.

`npm test` was intentionally not run locally; exact-head PR Gate is authoritative for the complete portable and platform suites. Local Node typecheck was blocked by the installed shared dependency tree: the available `@agentclientprotocol/claude-agent-acp` package does not export `waitForMcpServers`, which is imported by an unchanged ACP test.

## Review focus

- Confirm the remaining assertions describe ownership, dependency, export, persistence-safety, or observable capability boundaries.
- Confirm removed assertions constrained implementation shape without protecting external behavior.
- Confirm the detached callback test protects the runtime reason `handleJobUpdated` remains bound without prescribing arrow-field syntax.

### DIFF
--- a/src/main/compute/compute-service.architecture.test.ts
+++ b/src/main/compute/compute-service.architecture.test.ts
@@ -2,21 +2,15 @@ import { readFileSync, readdirSync } from 'node:fs'
 import { basename, dirname, extname, relative, resolve } from 'node:path'
 
 import {
-  canHaveModifiers,
   createSourceFile,
   forEachChild,
-  getModifiers,
-  isArrowFunction,
   isCallExpression,
-  isClassDeclaration,
   isExportDeclaration,
   isIdentifier,
   isImportDeclaration,
   isImportTypeNode,
   isLiteralTypeNode,
-  isMethodDeclaration,
   isPropertyAccessExpression,
-  isPropertyDeclaration,
   isStringLiteralLike,
   ScriptKind,
   ScriptTarget,
@@ -44,9 +38,7 @@ const computePaths = {
   concurrencyManager: resolve(mainRoot, 'compute/concurrency-manager.ts'),
   jobDispatcher: resolve(mainRoot, 'compute/job-dispatcher.ts'),
   jobPoller: resolve(mainRoot, 'compute/job-poller.ts'),
-  sessionEnabledHostsOwner: resolve(mainRoot, 'compute/session-enabled-hosts-owner.ts'),
   ipc: resolve(mainRoot, 'compute/ipc.ts'),
-  electronIpcAdapter: resolve(mainRoot, 'compute/electron-ipc-adapter.ts'),
   mainIpc: resolve(mainRoot, 'ipc.ts'),
   applicationCommands: resolve(mainRoot, 'compute/application-commands.ts'),
   jobRuntime: resolve(mainRoot, 'compute/job-runtime.ts'),
@@ -54,7 +46,6 @@ const computePaths = {
 } as const
 
 const readSource = (path: string): string => readFileSync(path, 'utf8')
-const rawLineCount = (source: string): number => source.trimEnd().split(/\r?\n/).length
 const modulePath = (path: string): string => path.replace(/\.[cm]?[jt]sx?$/, '')
 const portableProjectPath = (path: string): string =>
   relative(projectRoot, path).replaceAll('\\', '/')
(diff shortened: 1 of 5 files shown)


ewen-poch:
Addressed the five AI review findings in 97d1cd0d4: owner assertions now distinguish stable class-instance lifetime (constructor or field initializer) from transient method/nested-function construction; notebook module-state detection covers array/object literals while allowing function-only policy objects; and compute lifecycle repository writes use a closed allowed set. Targeted architecture tests, module-impact validation, compute/project-files module tests, and full lint pass locally. Upload module's only local failure remains the existing Windows symlink fixture EPERM before repository logic.

ewen-poch:
Review follow-up:

- **Mutable module-state detection misses factory-created state:** addressed in 736c273c0. Top-level call and 
ew initializers are now rejected, with regression fixtures for createSessionCache() and 
ew SessionCache(). Function declarations and the existing function-only runtime-settings policy object remain allowed.
- **Notebook facade API inventory:** not restored. Exact public/private method inventories couple the test to implementation shape and are the behavior this PR intentionally removes; compatibility remains covered at public call sites and observable behavior boundaries.
- **Upload facade delegation and method contract:** not restored. Exact constructor syntax, method inventory, and one-hop delegation mappings would reintroduce the implementation-detail lock. The retained checks enforce exports, stable owner lifetime, ownership boundaries, and behavior tests.
- **Project Files repository interface checks:** not restored for the same reason. Public consumers and behavior tests protect the contract, while architecture tests retain the single stable mutation/query owners and read/write separation.
- **Compute facade operation inventory:** not restored. Selected cross-surface capability boundaries and the closed lifecycle write set remain enforced; the detached callback is covered behaviorally without fixing the complete facade method inventory.

Validation: all four affected architecture test files pass (25 tests), and the changed file passes ESLint and git diff --check.

ewen-poch:
Review follow-up:

- **Restore Notebook facade contract checks:** not restored. Exact public/private method and mutable-field inventories are implementation-shape assertions that this PR intentionally removes. The retained checks enforce the observable module-state and owner-lifetime boundaries without freezing the facade layout.
- **Retain UploadRepository delegation checks:** not restored. The requested exact constructor, field, method, and direct-delegation assertions would recreate the shallow-facade coupling under removal. Export, stable-owner, ownership-separation, and behavior boundaries remain covered.
- **Retain Project Files facade checks:** not restored for the same reason. The architecture suite retains public exports, one stable mutation/query owner, and read/write separation without fixing constructor parameters or private field names.
- **Detect wrapped module-state initializers:** addressed in e30a6cb60. Initializer expressions are now inspected recursively, with a regression fixture for nabled ? new Map() : undefined. Traversal intentionally stops at function definitions and function-only policy objects so allocations that occur only when a function executes are not treated as module initialization.

Validation: all four affected architecture test files pass (25 tests), and the changed file passes ESLint and git diff --check.

ewen-poch:
Review follow-up:

- **Do not treat every property initializer as facade-lifetime construction:** addressed in 0e6f20c9a. Stable owner construction is now limited to each named top-level facade's constructor or non-static instance fields. Static fields, unrelated classes, nested classes, methods, and nested functions are classified as transient. Regression fixtures cover static fields and method-local classes; the same correction was applied to Notebook, Upload, and Project Files owner checks.
- **Restore the Notebook facade surface contract:** not restored. Exact method/private-field inventories are the implementation-detail coupling this PR removes; module state and stable owner placement remain enforced directly.
- **Restore the upload facade delegation contract:** not restored. Exact method shape and direct-delegation mappings would freeze the shallow facade implementation rather than an observable boundary.
- **Restore the project-files facade contract:** not restored. Public exports, one stable mutation/query owner, and read/write separation remain enforced without fixing private composition syntax.
- **Restore the ComputeService operation inventory check:** not restored. Cross-surface capability boundaries, the closed lifecycle write set, and detached callback behavior remain covered without locking the complete class operation inventory or callback field syntax.

Validation: all four affected architecture test files pass (25 tests); the three changed files pass ESLint and git diff --check.

ewen-poch:
Review follow-up:

- **Verify the actual shared transaction policy:** addressed in b287b73bb. The architecture check now requires the sole Prisma interactive transaction call to be staged-publication-owner.ts:UPLOAD_TRANSACTION_OPTIONS, whose initializer remains locked to { maxWait: 10_000 } as const. A regression fixture proves that a shadowed same-name local policy is rejected.
- **Do not miss aliased mutable module state:** not adopted. A static identifier/property-access initializer does not reveal whether the referenced value is mutable; rejecting all such aliases would also reject immutable imported constants and reintroduce an implementation allowlist. The retained check conservatively blocks allocations, factories, mutable literals, wrapped allocations, and non-const bindings while leaving ambiguous aliases to behavioral and ownership tests.

Validation: all four affected architecture test files pass (25 tests); the changed Upload architecture test passes ESLint and git diff --check.

ewen-poch:
**Resolve the transaction policy identifier instead of matching its text:** addressed in f91750388. The policy helper source is now checked for value bindings named UPLOAD_TRANSACTION_OPTIONS: exactly one binding must exist, and it must be the top-level variable declaration. Local variables, parameters, destructuring bindings, and named function/class/enum shadows are rejected before the $transaction call-site assertion is evaluated. A same-file local-shadow regression fixture fails under the previous text-only check and now passes by being rejected.

Validation: the Upload architecture suite passes (5 tests); all four affected architecture files pass together (25 tests); ESLint and git diff --check pass.

PR #1826, opened 2026-08-27
fix(annotations): keep trigger when selection geometry is empty

ewen-poch:
## Problem

After #1821, merged PRs that ran the full macOS unit shards failed on `WorkspaceMessageItem.actions.test.tsx` (`Annotate entry not found`). The new viewport check treated jsdom's zero-size `DOMRect` as a real selection box, so the portalled Annotate trigger stayed hidden. Nightly on current `main` would hit the same failure.

## Proposed change

Treat empty selection geometry (missing methods or zero-size rects) as unavailable and keep the previous jsdom/detached-range fallback. Live browser text selections still provide a box, so the scroll-clipping behavior from #1821 is unchanged.

## Scope and non-goals

- Restores the Annotate trigger in jsdom and for empty geometry.
- Does not change when a real in-viewport or clipped selection is shown or hidden.
- Does not change annotation persistence or the two-step confirm flow.

- Empty or missing range geometry keeps the trigger visible.
- A selection outside the window or clipped by an overflow ancestor still hides it.
- A selection that intersects the window and its clipping ancestor stays visible.

### Test impact set

- `isRangeTriggerVisible` empty-geometry fallback → `npx vitest run src/renderer/src/pages/workspace/annotations/annotation-trigger-anchor.test.ts src/renderer/src/pages/workspace/WorkspaceMessageItem.actions.test.tsx src/renderer/src/pages/workspace/annotations/TextAnnotationSurface.test.tsx` → pass (64)
- renderer types → `npm run typecheck:web` → pass
- changed files → `npx eslint src/renderer/src/pages/workspace/annotations/annotation-trigger-anchor.ts src/renderer/src/pages/workspace/annotations/annotation-trigger-anchor.test.ts` → pass

Excluded: full `npm test` (PR Gate remains authoritative), Windows lanes (no path or process change), consumer E2E (viewport hiding already covered by #1821's surface test).

## Review focus

- Whether zero-area rects are the right "geometry unavailable" signal versus only missing methods.
- That #1821's clip-outside-viewport behavior still fails closed for real boxes.

### DIFF
--- a/src/renderer/src/pages/workspace/annotations/AnnotationTrigger.tsx
+++ b/src/renderer/src/pages/workspace/annotations/AnnotationTrigger.tsx
@@ -20,7 +20,12 @@ const observeSelectionMutations = (range: Range, onMutate: () => void): (() => v
     observers.push(observer)
   }
   observe(element, { childList: true, characterData: true, subtree: true })
-  if (element.parentElement) observe(element.parentElement, { childList: true })
+  let ancestor = element.parentElement
+  while (ancestor) {
+    observe(ancestor, { childList: true })
+    if (ancestor === document.body) break
+    ancestor = ancestor.parentElement
+  }
   return () => {
     for (const observer of observers) observer.disconnect()
   }
--- a/src/renderer/src/pages/workspace/annotations/AnnotationTrigger.tsx
+++ b/src/renderer/src/pages/workspace/annotations/AnnotationTrigger.tsx
@@ -8,6 +8,24 @@ import { anchorRangeTrigger, isRangeTriggerVisible } from './annotation-trigger-
 const FALLBACK_TRIGGER_WIDTH = 82
 const FALLBACK_TRIGGER_HEIGHT = 24
 
+const observeSelectionMutations = (range: Range, onMutate: () => void): (() => void) => {
+  if (typeof MutationObserver === 'undefined') return () => {}
+  const node = range.commonAncestorContainer
+  const element = node instanceof Element ? node : node.parentElement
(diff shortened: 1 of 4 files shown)


ewen-poch:
**P2 Do not show triggers for detached ranges** — adopted.

`isRangeTriggerVisible` now returns false when the cloned range’s common ancestor is disconnected, or when removing the quoted nodes collapses the live Range onto a connected parent. That keeps a remounted message from parking the Annotate button at the fallback margin. Connected jsdom ranges with empty geometry still use the previous fallback.

ewen-poch:
**P1 Reevaluate visibility after selected content mutations** — adopted.

`AnnotationTrigger` now observes the selected node and its parent for `childList`/`characterData` mutations and reruns the same position/visibility update used on scroll and resize. Removing or streaming-replacing the quoted content hides the portalled Annotate button immediately, instead of waiting for a later scroll or resize.

ewen-poch:
**P1 Observe the full selection ancestor chain** — adopted.

The mutation observers now walk from the selected node through `document.body`. Removing a nested containing block (for example a `p` around an inline `span`) is a `childList` mutation on a grandparent, which now reruns visibility immediately. Added a nested-inline removal regression test.

ewen-poch:
**P1 Hide stale trigger when character data changes** — adopted.

`AnnotationTrigger` snapshots the captured range text and hides the button when a `characterData` mutation changes `range.toString()` while the range stays connected and non-collapsed. That prevents committing the original quote against streamed-in replacement copy.

PR #1815, opened 2026-08-27
feat(annotations): send selected context to agents

wen2zhou:
## Summary

- Add text-range annotations across transcript, activity, elicitation, and file-preview surfaces, plus point annotations for images.
- Persist source metadata, restore and reveal annotation targets, preserve context through edit/resend flows, and serialize annotations into agent and side-chat messages.
- Validate count, size, range, and source constraints and localize the new UI for es, fr, ja, ko, ru, zh-Hans, and zh-Hant.

## Integration with current main

- Preserve the dedicated web-source preview while passing annotation ports to file previews.
- Preserve the new per-question elicitation summary/accordion and reveal the exact annotated question without mutating recorded answers.
- Update React reveal effects for the current lint rules and Strict Mode cleanup behavior.

## Reviewability follow-up

- Replace repeated annotation callback/state prop bundles with a typed `AnnotationPort`.
- Share note markers/editors across workspace and preview surfaces while keeping their separate DOM Range and CSS Highlight adapters.
- Share sent-annotation card presentation across transcript and side chat.
- Centralize repeated CSS Highlight test setup.
- The follow-up changes 22 files with `+639/-734` (net -95 lines).

## Test impact and results

All listed passing checks ran after the final material edit.

- Direct annotation surfaces, cards, propagation, interaction, and reveal flows -> 11 files, 339 tests passed.
- Annotation capture, reveal, validation, persistence, localization, and renderer architecture contracts -> `npm test -- src/renderer/src/pages/workspace/WorkspaceMessageScroller.annotations.test.tsx src/renderer/src/pages/workspace/WorkspaceActivityGroup.annotations.test.tsx src/renderer/src/pages/workspace/PreviewPanel.test.tsx src/renderer/src/i18n/resources.test.ts src/shared/annotations.test.ts src/shared/session-persistence.test.ts src/renderer/src/lib/acp/useWorkspaceAgentRuntime.architecture.test.ts` -> 7 files, 963 tests passed.
- Node and renderer TypeScript surfaces -> `npm run typecheck` -> passed.
- Source and localization lint rules -> `npm run lint` -> passed.
[...]

PR #1828, opened 2026-08-27
fix(context-window): disclose incomplete call history

ewen-poch:
## Problem

After a user sends a prompt, switches ACP or model, and sends another prompt, the Context Window dialog can retain every aggregate Turn while the Calls view shows only the latest Turn. This happens when an earlier adapter reports a reliable aggregate model-call count but cannot provide trustworthy per-call details, while the later adapter does provide exact details. The current UI silently filters the earlier Turn out, which looks like switching runtimes overwrote call history.

## Proposed change

Derive reported call coverage from the existing Turn usage and compare it with the exact call points already selected for the Calls view. When exact details cover only part of the reported calls, show an inline localized notice with both counts.

The chart and token totals continue to include only trustworthy per-call records. The change does not synthesize missing calls.

## Scope and non-goals

- Renderer-only interaction change.
- No persisted field, database or JSON schema change, migration, new enum value, or state-model change.
- No historical backfill. Existing sessions with aggregate counts but no exact call details become transparently partial; unavailable historical details remain unavailable.
- No adapter behavior change and no attempt to infer per-call token breakdowns from aggregate Turn usage.

- Cross-runtime session with 3 reported calls and 1 exact call shows the exact T2 call plus: "Showing 1 of 3 reported calls because some turns have no exact call details."
- Sessions whose exact call details cover all reported calls do not show the partial-coverage notice.

The regression test ran twice against the unfixed implementation and failed both times with the reported behavior: only one T2 call was rendered and no coverage explanation was present.

Final checks below ran after the last material edit and after rebasing onto the latest origin/main:

- Calls dialog behavior -> ContextWindowDialog.render.test.tsx -> 15 passed.
- Renderer translation contract -> resources.test.ts -> 743 passed.
- Renderer types -> TypeScript with tsconfig.web.json -> passed.
- Changed renderer sources -> ESLint on ContextWindowDialog.tsx and its render test -> passed.
- Changed files -> Prettier check and git diff --check -> passed.

The full npm test suite was not run locally; exact-head PR Gate and platform CI remain authoritative. No persistence, migration, adapter, or platform lane is required by this renderer-only change. The only uncovered behavior is provider-side availability of exact call details, which this change intentionally reports rather than reconstructs.

## Review focus

- Whether comparing aggregate Turn call counts with trusted per-call points is the correct boundary for partial coverage.
- Whether the notice makes the distinction between reported calls and available exact details clear without implying data recovery.

### DIFF
--- a/src/renderer/src/pages/workspace/ContextWindowDialog.render.test.tsx
+++ b/src/renderer/src/pages/workspace/ContextWindowDialog.render.test.tsx
@@ -317,6 +317,105 @@ describe('ContextWindowDialog', () => {
     return withCalls
   }
 
+  const sessionWithCallsAfterRuntimeSwitch = (): ChatSession => {
+    const switched = session()
+    switched.messages.push(
+      {
+        id: 'answer-claude',
+        role: 'agent',
+        responseToMessageId: 'prompt-1',
+        content: 'Claude result',
+        eventIds: [],
+        status: 'complete',
+        turnUsage: { inputTokens: 30, cacheTokens: 6, outputTokens: 8, turnCount: 2 },
+        createdAt: 2,
+        updatedAt: 3,
+        completedAt: 3
+      },
+      {
+        id: 'prompt-2',
+        role: 'user',
+        content: 'Continue with Codex',
+        eventIds: [],
+        status: 'complete',
+        createdAt: 4,
+        updatedAt: 4
+      },
+      {
+        id: 'answer-codex',
+        role: 'agent',
+        responseToMessageId: 'prompt-2',
+        content: 'Codex result',
+        eventIds: [],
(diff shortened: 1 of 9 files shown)


ewen-poch:
CI note: macOS full-suite shard 2/2 failed twice on src/renderer/src/pages/workspace/WorkspaceMessageItem.actions.test.tsx > creates a text annotation only after the two-step confirmation, with Annotate entry not found. Neither that test nor WorkspaceMessageItem is changed by this PR; the same focused test also fails on the current rebased branch locally. The introducing PR #1815 previously passed this shard. I stopped after one failed-job rerun to avoid masking or mixing an unrelated baseline failure. The regression test and all directly relevant focused/static checks for this change pass.

PR #1831, opened 2026-08-27
feat(agent-framework): add CodeBuddy ACP framework

ewen-poch:
## Problem

Open Science supports three ACP agent frameworks, but CodeBuddy could not be installed, configured, or used through the same framework lifecycle without a CodeBuddy login. Its ACP surface also differs in session routing, cancellation, steering, model effort, usage reporting, Skill discovery, and upstream model protocol.

## Proposed change

- Add an app-managed, version-pinned CodeBuddy 2.138.0 runtime with detection, installation, uninstall, and Settings integration.
- Run CodeBuddy login-free over stdio ACP with an Open Science-owned Chat Completions provider bridge.
- Adapt session activation, cancellation, private live steering, model/effort changes, host-controlled compaction, image input, reasoning output, tool ordering, and per-call usage.
- Reuse the app-owned Skill pre-router, exact-name loader, Notebook MCP, Connector routing, and process-per-Attempt delegation instead of enabling CodeBuddy-native Skill and Agent discovery.
- Remove native Web tools and common direct-network shell commands from the CodeBuddy tool surface while preserving legitimate HTTP/SSE MCP proxy access.

```mermaid
flowchart LR
  UI[Open Science session] --> ACP[ACP runtime]
  ACP --> CB[CodeBuddy 2.138.0 stdio ACP]
  CB --> Bridge[Chat Completions loopback bridge]
  Bridge --> Provider[Configured model provider]
  ACP --> Router[App-owned Skill pre-router]
  Router --> Loader[Exact-name Skill loader]
  Loader --> Notebook[Notebook MCP / host.mcp]
  Notebook --> Connector[Configured Connector]
```

## Scope and non-goals

- CodeBuddy is pinned to 2.138.0; upgrades require ACP and private steering regression spikes.
- CodeBuddy talks to upstream models through `/v1/chat/completions`. Messages and Responses providers require the existing Open Science compatibility bridge.
- Native CodeBuddy Skills, Agents, workflows, background tasks, WebFetch, and WebSearch remain disabled.
- The adapter does not claim lossless session fork, persistent-session delete, live native main-agent switching, or Agent Team restoration.
- Keeping arbitrary Bash means command deny rules are not an absolute network sandbox. Windows exposes no CodeBuddy shell tool because the pinned runtime has no equivalent shell sandbox there.

- CodeBuddy installs through the app-managed runtime without system npm or a CodeBuddy login.
- ACP prompt, image, cancellation, live injection, MCP, Connector Skill routing, delegation, compaction, model effort, and usage paths are covered by deterministic tests and repository spikes.
- Skill selector failures and unsafe selections fail closed; an initial valid empty selection receives a second independent verification pass.
- Existing frameworks retain their protocol-specific paths and shared behavior.

Checks run after the final material implementation change:

- `npx vitest run src/main/agent-framework/codebuddy.test.ts src/main/acp/turn-skill-owner.test.ts src/main/acp/context-compaction-workflow.test.ts` — 49 passed.
- `npm run typecheck:node` — passed.
- `npx vitest run src/renderer/src/i18n/resources.test.ts` — passed as part of the final focused run.
- `npx vitest run src/main/acp/runtime.test.ts -t "fails the CodeBuddy turn closed when Skill discovery is unavailable"` — passed.

The complete `npm test` suite was intentionally not run locally; required CI remains the final authority. A broad runtime test attempt was constrained by the workspace sandbox: localhost `listen` returned `EPERM` in three cases and one oversized MCP case timed out. The changed fail-closed assertion passes in isolation.

Persistence and compatibility notes:

- Adds persisted framework id `codebuddy` and model route `codebuddy-openai`.
- Adds no database migration or CodeBuddy-specific usage table; per-call usage reuses the existing optional projection.
- Historical turns without upstream usage updates cannot be backfilled losslessly.

## Review focus

- CodeBuddy-specific behavior should remain behind the framework and output/request adapters.
- Dynamic MCP configuration must be replayed on every pre-turn session resume.
- Skill routing and usage correlation must remain fail closed and session-scoped.
- Tool restrictions must not break the app-owned Connector path or other ACP frameworks.


ewen-poch:
Applied the requested fixes:

- CodeBuddy now receives the isolated Skill runtime whenever Skill materialization is enabled, including generations with no enabled Connector Skills.
- Restricted CodeBuddy inference now copies configuration into the per-job profile, removes the Skill runtime and prior system-prompt file, and passes an empty native tool allowlist. This shared restricted profile covers Side Chat and artifact reconstruction.
- Delegated CodeBuddy attempts now copy configuration into the per-attempt runtime directory; the existing attempt cleanup removes that isolated directory.
- Artifact reconstruction cache validation now accepts CodeBuddy entries, with a durable round-trip regression test.
- Managed and fallback CodeBuddy installs now use the shared exclusive-install guard.
- The local-only renderer surface inventory includes CodeBuddy install/uninstall, and the spike file is formatted.

Validation: targeted suites passed (CodeBuddy resolver/framework, restricted inference, delegation, Side Chat, artifact reconstruction, runtime manager, renderer inventory), plus node typecheck, i18n guard, Web API map check, ESLint, and Prettier.

ewen-poch:
Applied the latest review findings:

- Managed CodeBuddy verification now invokes Windows .cmd/.bat shims with shell-aware execution while keeping direct execFile behavior on other platforms.
- Current-mode projection now carries the originating framework and recognizes CodeBuddy fullAccess as the native Full access mode.
- CodeBuddy pre-routing now injects Skill document content without the projected base-directory prefix or absolute placeholder expansion. The absolute path remains only in app-owned Skill activity metadata.

Added regressions for Windows shim verification, CodeBuddy fullAccess updates through the shared projector, and absence of the projected root in CodeBuddy Skill prompt guidance. The five affected test modules pass (98 tests), along with node typecheck, ESLint, and Prettier.

ewen-poch:
Addressed both review findings in aedccee6:

- Added CodeBuddy to the explicit framework override allowlist and covered the override path with a regression test.
- Certified the CodeBuddy OpenAI visual route for host.viewImage and added behavior-level coverage for both availability and image staging.

Validation: 30 targeted tests passed, along with Node typecheck, ESLint, Prettier, and git diff checks.

ewen-poch:
Applied the current high-value review findings on top of the latest origin/main:\n\n- Removed the global npm installation path. CodeBuddy installation is app-managed only, while manually installed binaries remain discoverable.\n- Enforced the 2.138.0 runtime pin for detection, cached restoration, and launch readiness.\n- Added the missing Spanish managed-install description and accepted CodeBuddy in OPEN_SCIENCE_AGENT_FRAMEWORK.\n- Kept the default CodeBuddy provider on the MiniMax Chat Completions route.\n- Preserved precise Notebook tool observation in full-access sessions and passed the isolated CodeBuddy process into delegated runtimes.\n- Added direct managed installer coverage for extraction, idempotent replacement, uninstall, traversal rejection, and wrong-version rejection.\n- Added a full CodeBuddy session/steer workflow regression.\n- Removed the model-callable Skill loader from CodeBuddy session/new; pre-routed Skill documents are injected directly, and stale loader permission requests are no longer auto-approved.\n- Restored Codex Skill selection fail-open behavior while keeping CodeBuddy selection fail-closed.\n\nValidation completed locally with targeted suites: the pre-review aggregate passed 985 tests; the latest focused runs passed 167 tests plus focused selector, transport, and runtime cases. Node and web typechecks, targeted ESLint, Prettier, and git diff checks passed. The full suite is intentionally left to CI.

ewen-poch:
Addressed the installation safety finding:\n\n- The candidate package and shim are now created and version-verified in a sibling staging runtime before the live runtime is touched.\n- After verification, the staged shim is finalized for the live path and the runtime is swapped with a backup/restore guard.\n- A failed verification leaves the existing package and shim unchanged; a failed swap restores the previous runtime, and an unrecoverable backup is retained instead of deleted.\n- Added a regression proving that a wrong-version replacement cannot destroy a working installation.\n\nValidation: managed-codebuddy tests pass (5 tests), along with Node typecheck, targeted ESLint, Prettier, and git diff checks. The branch was rebased onto the latest origin/main and remains a single commit.

ewen-poch:
Addressed the CodeBuddy transcript-correlation finding:\n\n- Tool events and their following assistant chunks now use the same projected message ID for each tool boundary.\n- Boundaries are keyed to the provider response message ID, replaced for subsequent tools, and cleared when the next provider response begins, so they cannot leak into later turns.\n- Added adapter- and projector-level coverage for two tool boundaries plus a subsequent response.\n- Routed the selector transport's ProviderRuntimeTarget type through the public ProviderAccountsModule seam and updated the locked export/import inventories.\n- Registered the three Windows-sensitive CodeBuddy runtime sources in the trusted CI change-impact manifest.\n\nValidation: 101 targeted classification, architecture, adapter, and projector tests pass, along with Node typecheck, targeted ESLint, Prettier, and git diff checks. The branch is rebased onto the latest origin/main and remains one commit.

ewen-poch:
Addressed the current CodeBuddy review findings in ae38093e:

- CodeBuddy Skill materialization now uses the existing agent-facing layout, so public catalog names map directly to `.claude/skills/<name>/SKILL.md`. Added a closed-loop regression that materializes a Skill whose local ID differs from its public name and then loads the selected document through `AcpTurnSkillOwner`.
- Live CodeBuddy steering now preserves the active Skill route when no Skill chip is supplied. Explicit follow-up Skills receive the isolated runtime root and are loaded before injection instead of clearing the route or failing with `document-error`.
- Hardened the managed runtime swap lifecycle: staged swap failure restores the working runtime; restore failure reports both errors and the retained backup path; uninstall removes CodeBuddy-owned live, staging, and backup directories. Added deterministic failure-path coverage.

Validation: 94 focused ACP/materializer/managed-runtime tests and all 63 backend-resolver tests pass, along with Node typecheck, targeted ESLint, Prettier, and git diff checks. The branch is based on current origin/main and remains a single commit.

The Messages/Responses provider compatibility bridge is intentionally not expanded here: CodeBuddy remains a Chat Completions consumer, and adding request/stream translators would change the advertised Provider compatibility and user-facing selection behavior.

ewen-poch:
The previous AI review body was withheld by GitHub output masking, but its normalized findings were recoverable from the job log. Addressed all three findings in b4252f58:

- Agent-facing Skill synchronization is now idempotent for persistent CodeBuddy runtimes and rebuilds the app-owned projection when the enabled Skill set changes.
- Injected CodeBuddy Skill documents now carry a provider-safe, environment-backed resource root. `${CLAUDE_SKILL_DIR}` references are translated to `${CODEBUDDY_CONFIG_DIR}/skill-runtime/.claude/skills/<name>`, and the route guidance defines how relative references, scripts, and assets resolve without exposing the absolute local configuration path.
- CodeBuddy uninstall now removes only directory entries with the exact UUID-shaped staging/backup names generated by the installer. Prefix look-alikes and UUID-named plain files are preserved.

Added regressions for repeated/changed agent-facing synchronization, resource-root translation without absolute-path disclosure, and exact orphan cleanup. Focused tests pass (58 direct regression tests; 157 impacted tests with 132 intentionally filtered), along with Node typecheck, targeted ESLint, Prettier, and git diff checks. The branch remains one commit on current origin/main.

ewen-poch:
Implemented the actionable review findings in 3e0b4ad5.

- Removed the CodeBuddy-specific Skill and ACP behavior from the reusable UI/UX design document and updated the PRD framework inventory.
- Kept CodeBuddy on its native Chat Completions consumer contract. When an upstream provider exposes only Responses or Messages, an app-owned loopback now translates that provider wire to Chat Completions. The same loopback drives Skill selection, and routing tests cover both provider types.
- Native compaction now runs the shared framework pre-dispatch activation before sending /compact. A multi-session regression verifies that CodeBuddy resumes the requested provider session before compacting it.
- Saved CodeBuddy executable paths are re-probed at launch and accepted only when they still report the pinned 2.138.0 version.

Validation: Node and web typechecks passed; targeted routing, protocol, provider, compaction, runtime, version, architecture, validation, and history-replay modules passed.

ewen-poch:
Addressed the latest AI review findings after rebasing onto origin/main (bd5ff177):

- CodeBuddy xAI subscription traffic now uses the authenticated xAI OAuth bridge, retains forced token refresh on 401, and applies the CodeBuddy Chat Completions request adapter before Responses translation. Skill selection is created against the authenticated loopback target.
- Native CodeBuddy compaction and normal provider prompts now share one process-level serialization owner, so session activation and /compact cannot race across sessions.
- CodeBuddy tool events preserve the provider response message identity, allowing late-arriving tool updates to be ordered before already-projected final prose while post-tool assistant chunks remain segmented.
- Rebase conflicts were resolved against the generated renderer contract architecture from current main.

Targeted validation passed: Node/Web typechecks, web API map check, ACP runtime (575 tests), CodeBuddy projection/workflow tests (64), provider/xAI bridge tests (16), preload/renderer contract tests (89), targeted ESLint, and diff checks.

ewen-poch:
Processed the completed CI results:

- Removed this feature PR change to scripts/ci/change-impact.json because that established gate control-plane file requires a separate maintainer ruleset bypass. The CodeBuddy runtime changes were already exercised by Windows core, which passed.
- Added chat-provider-compatibility.ts to the exact Provider loopback adapter architecture inventory. This was the sole failing assertion recovered from the macOS shard blob report.

Local verification now passes for the Provider loopback architecture test and the CI integrity checker against the current origin/main base. The branch remains one commit.

ewen-poch:
Fixed the remaining CodeBuddy tool-ordering case at 60c29a0a. CodeBuddy tool notifications do not guarantee a provider messageId, so exact stream correlation could be skipped and late tool activities remained below the final assistant prose. The renderer now falls back, only for CodeBuddy, to the latest agent message belonging to the same active prompt. A workspace event regression covers multiple late tool calls without provider message IDs and verifies that every tool remains before the final prose. Rebased onto origin/main 439e7194; the branch remains one commit. Validation: 265 workspace/session-store tests passed, the focused multi-tool regression passed after rebase, typecheck:web passed, and targeted ESLint/Prettier checks passed.

ewen-poch:
Corrected the CodeBuddy transcript fix at 12e6ec51 after reproducing the three-model-call sequence. The previous renderer fallback moved the most recent assistant message behind every later tool, which incorrectly collapsed an intermediate assistant segment into the following tool group. CodeBuddy tool notifications can omit messageId, so the adapter now treats each initial tool notification as a boundary for the next assistant stream and derives the segmented message ID when that assistant chunk arrives. Existing prose is never reordered in the renderer. This preserves the intended timeline: Skill/tool activity -> intermediate assistant message -> later tool activity -> final assistant message. Removed the now-unneeded renderer provider-message correlation path. Validation: 293 affected tests passed, including adapter/projector and three-segment workspace regressions; Node and Web typechecks, targeted ESLint, Prettier, and diff checks passed. The branch remains one commit on origin/main 439e7194.

PR #1832, opened 2026-08-27
fix(notebook): arbitrate queued execution races

ewen-poch:
## Problem

Four Notebook execution races produced inconsistent lifecycle outcomes:

- A run admitted while runtime repair was required could wait behind the repair's exclusive environment lease, then still reuse its stale pre-lock rejection after the repair completed. The run was persisted as `failed` with `RUNTIME_REPAIR_REQUIRED` instead of executing.
- Production repair released its exclusive environment lock before the lifecycle cleared the durable repair marker and restored bindings. A queued run could acquire a shared lease in that gap and still observe `RUNTIME_REPAIR_REQUIRED` after a successful rebuild.
- Cancellation could fire synchronously during Kernel request dispatch before the timeout was armed. Dispatch then armed the timeout anyway, so one request could be both cancelled and timed out and receive a duplicate SIGINT.
- A Session could change or disable its runtime binding while a repair-blocked run waited for the environment lock. Clearing only the stale repair rejection then dispatched the cell with the old route/interpreter admission even though the Session's current binding selected a different or unavailable runtime.

An active Kernel-crash scenario was also reproduced through the Notebook service boundary. Existing behavior already persists the run as `failed` and the Kernel as `terminated`, so no change or always-green regression test is included for that case.

## Proposed change

- Distinguish runtime-repair admission errors internally and discard only that stale rejection when the post-lock repair check is clear.
- Re-read the Session binding after the environment lock is acquired. A changed route or interpreter identity fails closed with retryable `RUNTIME_BINDING_CHANGED`; a currently unavailable binding retains the existing `RUNTIME_BINDING_UNAVAILABLE` classification.
- Add an internal repair-completion callback that runs only after rebuild verification and before the provisioner releases the per-environment lock; the lifecycle uses it to clear repair markers and restore bindings atomically with repair.
- Keep repair cancellation protection active through the awaited completion callback so a late Cancel cannot leak into the next environment operation.
- Do not arm a request timeout when cancellation has already won during synchronous dispatch.
- Add deterministic regression tests at the lifecycle, provisioner, Notebook runtime, and Kernel request-arbitration boundaries. Each regression failed on the relevant pre-fix implementation for the reported symptom and passes with this change.

## Scope and non-goals

- No database/schema, persisted-format, request/response shape, UI, or interaction changes.
- No new run status or other enum value. `RUNTIME_BINDING_CHANGED` is a transient execution-error classification, not persisted model state.
- No historical migration; previously persisted false failures are not rewritten.
- The `RuntimeProvisioner.repair` options type gains an internal callback only; there is no persisted data-model change and no change to environment contents or provisioning algorithms.
- No changes to Python/R dependency parsing or package-manager error taxonomy.

- lifecycle publishes repair completion before the environment lock releases -> regression failed twice pre-fix with `lock:enter -> lock:exit -> repair:completed`, then passed after the fix
- the real provisioner invokes repair completion after verification and before unlock -> focused provisioner regression passed
- Cancel during awaited repair finalization does not cancel the next operation -> regression failed twice on the intermediate implementation with `Runtime setup cancelled`, then passed after extending the repair protection boundary
- repaired environment releases a queued run without a stale repair failure -> focused runtime-service regression passed
- binding changed while that run is queued never dispatches the old runtime -> regression failed twice pre-fix with a completed run, then passed with `RUNTIME_BINDING_CHANGED` and zero stale dispatches
- cancellation wins synchronous dispatch without later timeout -> focused kernel-executor regression passed
- affected Notebook modules -> `npx vitest run src/main/notebook/environment-lifecycle-workflows.test.ts src/main/notebook/environment-operation-foundation.test.ts src/main/notebook/provisioner.test.ts src/main/notebook/runtime-service.test.ts src/main/notebook/kernel-executor.test.ts` -> 377 passed, 66 existing platform/interpreter-gated tests skipped; after the final error-priority refinement, the complete runtime-service module passed 221 tests with 5 skipped
- all changed source/test files -> `npx eslint` and `npx prettier --check` -> passed
- Node typecheck before rebase -> `npm run typecheck:node` -> blocked only by the unchanged `src/main/acp/claude-agent-acp-patch.test.ts` because the shared main-checkout dependency tree did not export `waitForMcpServers`; no changed Notebook file errors remained. CI's clean dependency install passed Node/Web typecheck on the prior head.

The repository currently has no Notebook module ID in `scripts/ci/module-impact.json`, so affected Notebook test files were run directly instead of claiming `test:module` coverage. Exact-head PR Gate remains authoritative for downstream and platform lanes. Full local `npm test` was intentionally not run.

## Review focus

- Repair verification, durable marker cleanup, in-memory repair-gate cleanup, and binding restoration must finish before the per-environment lock is released.
- Per-language Cancel must remain ignored through repair finalization and must not arm the next environment operation.
- The post-lock admission path must clear only a stale repair-required rejection when the current binding still resolves to the same route and interpreter identity; changed/unavailable bindings must fail closed.
- Existing repair-required error priority must remain intact for quarantined managed bindings.
- Cancellation must remain the sole terminal cause when it fires before timeout arming.

### DIFF
--- a/src/main/notebook/data-execution-admission.ts
+++ b/src/main/notebook/data-execution-admission.ts
@@ -62,6 +62,48 @@ const repairRequiredError = (language: NotebookLanguage): Error =>
       'integrity check. Run the runtime Repair workflow before executing another cell.'
   )
 
+const bindingChangedError = (language: NotebookLanguage): Error =>
+  new Error(
+    `RUNTIME_BINDING_CHANGED: the bound ${language} runtime changed while this cell was queued. ` +
+      'Retry the cell so it is admitted against the current runtime.'
+  )
+
+const bindingUnavailableError = (
+  language: NotebookLanguage,
+  binding: NotebookSessionRuntimeBinding
+): Error =>
+  new Error(
+    `RUNTIME_BINDING_UNAVAILABLE: the bound ${language} runtime is ${binding.status}` +
+      (binding.reason ? ` (${binding.reason})` : '') +
+      '. Call list_notebook_runtimes then notebook_switch_runtime to choose another runtime ' +
+      '(an unspecified choice falls back to the app-managed default). Any prior kernel memory ' +
+      '(variables, imports) for this language was lost.'
+  )
+
+const equalStringArrays = (left: string[] | undefined, right: string[] | undefined): boolean =>
+  left === right ||
+  (left !== undefined &&
+    right !== undefined &&
+    left.length === right.length &&
+    left.every((value, index) => value === right[index]))
+
+const sameRuntimeTarget = (
+  admitted: NotebookSessionRuntimeBinding | undefined,
+  current: NotebookSessionRuntimeBinding | undefined
+): boolean => {
+  if (admitted === undefined || current === undefined) return admitted === current
+  return (
+    admitted.language === current.language &&
+    admitted.runtimeId === current.runtimeId &&
+    admitted.source === current.source &&
+    admitted.interpreterPath === current.interpreterPath &&
+    admitted.envName === current.envName &&
+    admitted.resolvedInterpreter?.command === current.resolvedInterpreter?.command &&
+    equalStringArrays(admitted.resolvedInterpreter?.args, current.resolvedInterpreter?.args) &&
+    admitted.resolvedInterpreter?.condaPrefix === current.resolvedInterpreter?.condaPrefix
+  )
+}
+
 /** Owns data-cell routing plus every fail-closed gate before the executor may be dispatched. */
 class NotebookDataExecutionAdmissionOwner {
   constructor(private readonly options: NotebookDataExecutionAdmissionOwnerOptions) {}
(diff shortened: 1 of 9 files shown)


ewen-poch:
Addressed the P1 in a6b87b6e3. The lifecycle now passes repair cleanup through an internal onVerified callback, and DefaultRuntimeProvisioner awaits it after rebuild verification but before releasing the per-environment lock. Added a lifecycle regression that failed twice on the previous head with lock:enter -> lock:exit -> repair:completed, plus a real provisioner lock-order assertion. Affected Notebook modules now pass (375 passed, 66 platform/interpreter-gated skipped).

ewen-poch:
Addressed the follow-up P2 on final head 105238528. Added a serialized-provisioner regression that calls Cancel from the awaited repair finalization callback; it failed twice on the intermediate implementation because the next provision rejected with Runtime setup cancelled. The callback now remains inside runLanguage's uninterruptible boundary, so that late Cancel is ignored and cannot arm the next operation. Rebased onto current main to resolve the separate two-dot changed-format CI false positive without modifying the three unrelated files.

ewen-poch:
Addressed the latest AI Review P1 in 78a7c894d. Reproduced through the public NotebookRuntimeService boundary before the fix: while a repair-blocked run was queued, the Session bound an external Python runtime; the regression failed twice because the run still completed on the stale default target. The post-lock path now re-reads and validates the current route, runtime target identity, interpreter identity, and binding status before clearing a stale repair rejection. A changed target fails closed with RUNTIME_BINDING_CHANGED; an unavailable target keeps RUNTIME_BINDING_UNAVAILABLE; an active unchanged target repaired during the wait may proceed. Existing RUNTIME_REPAIR_REQUIRED priority is preserved. Exact-head affected-module validation: 377 passed, 66 platform/interpreter-gated tests skipped; ESLint and Prettier pass.

PR #1837, opened 2026-08-27
fix(annotations): keep the annotate trigger after code highlighting

ewen-poch:
## Problem

PRs based on `main` fail **PR Gate / Full Module tests (macOS, shard 2/2)** because

`WorkspaceMessageScroller.annotations.test.tsx` → `annotates and restores independent Tool and Notebook transcript sections`

asserts the Annotate trigger exists after selecting tool/notebook code, and the trigger is missing.

Two things hide it after a live selection:

1. `WorkspaceToolCodeBlock` runs Shiki after paint and replaces the selected text node with token spans. The cloned draft `Range` disconnects or collapses, and `isRangeTriggerVisible` then hides the portalled trigger.
2. Tool rows use `overflow-hidden` / `overflow-auto`. In jsdom those ancestors often report a zero-size box. The clip check treated that as “selection is outside the viewport” and hid a still-valid draft.

## Proposed change

- Retarget the live draft `Range` from the surviving quote when surface content mutates but the quoted text is still there (the same path saved annotations already use after highlighting).
- Drop the draft only when the quote is gone, which keeps the existing hide-on-stream / hide-on-remove behavior.
- Skip window and overflow clip checks when the measured box has no usable width or height, matching the existing jsdom missing-geometry fallback.

## Scope and non-goals

- Renderer annotation trigger only. No schema, migration, persisted format, or enum changes.
- Does not change macOS E2E product behavior. The shard 2/2 failure is this unit test; see notes below for the separate E2E finding.

- Selecting tool/notebook code still shows Annotate after async syntax highlighting.
- Streaming or removing the quoted text still hides the trigger.
- Overflow clipping still hides the trigger when the ancestor box is a real, usable rectangle.

### Evidence (after last material edit)

| --- | --- | --- |
| Tool/notebook transcript annotate + restore | `npx vitest run src/renderer/src/pages/workspace/WorkspaceMessageScroller.annotations.test.tsx` | pass |
| Draft retarget, hide-on-remove, highlight spans | `npx vitest run src/renderer/src/pages/workspace/annotations/TextAnnotationSurface.test.tsx` | pass |
| Trigger visibility / zero-size clip fallback | `npx vitest run src/renderer/src/pages/workspace/annotations/annotation-trigger-anchor.test.ts` | pass |
| Quote retarget helper | `npx vitest run src/renderer/src/pages/workspace/annotations/text-annotation-range.test.ts` | pass |
| Changed-file lint | `npx eslint` on the six edited files | pass |

Consumers outside the annotation surface were excluded: the public `TextAnnotationSurface` contract is unchanged. Platform E2E was excluded from local runs; PR Gate remains authoritative for the full portable and macOS suites.

## Review focus

- Mutation-observer retarget vs. the existing hide-when-quote-is-gone tests.
- Zero-size clip fallback still hiding real off-screen selections.

## Historical compatibility

None. No persisted annotation format or stored Range is changed; retarget is in-memory for the live draft only.

## New state / enum values

None.

## Data persistence

None in this PR.

### Related CI note (not fixed here)

**PR Gate / macOS build and E2E** is not failing on every main-based PR. The consistent red check is shard 2/2. One open PR (`#1818`) also failed `e2e/message-tool-layout-stability.spec.ts` because `session-persistence-alert` (“This conversation changed in another window…”) covered the Send button. That is a session-revision race on the persistence path. I did not change persistence, conflict rebase, or stored session shape here — confirm if that should be a follow-up.

### DIFF
--- a/src/renderer/src/pages/workspace/annotations/TextAnnotationSurface.test.tsx
+++ b/src/renderer/src/pages/workspace/annotations/TextAnnotationSurface.test.tsx
@@ -715,6 +715,38 @@ describe('TextAnnotationSurface annotate trigger', () => {
     expect(annotateTrigger()).toBeUndefined()
   })
 
+  it('keeps the trigger after syntax highlighting retargets the selected code', async () => {
+    await act(async () =>
+      root.render(
+        <TextAnnotationSurface
+          source={{
+            kind: 'session-item',
+            sessionId: 'session-1',
+            itemId: 'tool-1',
+            itemType: 'tool-activity',
+            sectionId: 'code'
+          }}
+          activeAnnotations={[]}
+          onAdd={vi.fn()}
+          onError={vi.fn()}
+        >
+          <WorkspaceToolCodeBlock code="const answer = 42" language="typescript" />
+        </TextAnnotationSurface>
+      )
+    )
+    const code = container.querySelector<HTMLElement>('[data-testid="tool-code-block"] code')!
+    await commitSelection(code)
+    expect(annotateTrigger()).toBeDefined()
+
+    await vi.waitFor(() => {
+      expect(
+        container.querySelectorAll('[data-testid="tool-code-block"] code span').length
+      ).toBeGreaterThan(0)
+    })
+    await act(async () => Promise.resolve())
+    expect(annotateTrigger()).toBeDefined()
+  })
+
   it('hides the trigger when streaming replaces the selected message content', async () => {
     const paragraph = await renderSurface()
     await commitSelection(paragraph)
--- a/src/renderer/src/pages/workspace/annotations/TextAnnotationSurface.test.tsx
+++ b/src/renderer/src/pages/workspace/annotations/TextAnnotationSurface.test.tsx
@@ -747,6 +747,66 @@ describe('TextAnnotationSurface annotate trigger', () => {
     expect(annotateTrigger()).toBeDefined()
   })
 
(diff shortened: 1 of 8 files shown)


ewen-poch:
**[P2] Preserve the selected occurrence when retargeting after DOM replacement** — adopted.

Live draft retarget now captures the quote occurrence at selection time and rebuilds that same occurrence after Shiki replaces the text node. Confirming the draft no longer jumps a duplicate quote to the first match.

This occurrence is in-memory draft metadata only. Saved annotations still store the quote string (no persisted offset/occurrence field). After remount, a single saved duplicate quote continues to reconcile as occurrence 0, which is the existing saved-annotation contract.

Also tightened `rangeForTextOccurrence` so a match that starts on a node boundary lands on the following token span instead of the exclusive end of the previous span.

ewen-poch:
**[P2] Retarget selections in file previews too** — adopted, without extracting a shared surface owner.

`PreviewTextAnnotationSurface` now captures quote occurrence at selection time and retargets the live draft when `HighlightedCodeLines` (or any child mutation) replaces the selected text node. It reuses `quoteOccurrenceForRange` / `retargetTextAnnotationRange` from the session surface. The two surfaces still own their highlight registries and source mapping; a shared helper would only wrap this wiring.

No persisted annotation fields were added. Preview drafts stay in memory, same as session drafts.

PR #1843, opened 2026-08-28
fix(workspace): align session hover preview card and add inline rename

ewen-poch:
## Problem

The session detail hover card in the workspace sidebar had three issues:

1. The card's vertical position did not line up with the hovered session row — the Radix `Tooltip` trigger was the inner open button (inset by the row's `py-1.5`) while the card added its own `p-4`, so the card visibly floated away from the row.
2. The card title (15px semibold) was larger than the row title (14px), and the description (14px) was not smaller than the title.
3. The pointer could not travel horizontally from the row onto the card: the trigger button ended before the row's right edge (action button + gradient overlay), leaving a dead zone that closed the card mid-traversal. The card also offered no interaction.

## Proposed change

- **Primitive**: switch from Radix `Tooltip` to Radix `HoverCard` (already exported by the installed `radix-ui@1.6.1` umbrella package — no new dependency), via a new `ui/hover-card.tsx` wrapper. HoverCard content is designed to be interactive and provides close-delay grace for pointer traversal.
- **Alignment**: the hover trigger is now the whole session row instead of the inner button, which removes the dead zone between the button and the card and anchors the card to the row's top edge. A fixed `alignOffset` (`SESSION_HOVER_PREVIEW_ALIGN_OFFSET_PX`) lands the card title on the same baseline as the hovered row title (and vertically centers a title-only card on the row).
- **Typography**: card title is `text-sm font-semibold` (same 14px as the row); the description drops to `text-xs`.
- **Inline rename**: the card title becomes a button when `canMutateConversations` allows; clicking swaps it for an inline input. Enter or blur commits a non-empty trimmed title, Escape cancels, and blank input cancels (mirroring `renameSession`'s blank-ignore rule). Commits go through a new `sessionController.actions.renameTitle`, which reuses the store's existing `renameSession` → `unsavedTitle` persistence path (no `editDetails` call, so the description is never touched). While editing, the card is pinned open; when the edit ends and the pointer is outside the hover region, the card closes.

## Scope and non-goals

- Desktop sidebar only; mobile mode never rendered the hover preview and is unchanged.
- The `Edit…` dialog remains the full title+description editor; only the title is editable inline.
- Data/persistence: no schema or migration change, no new persisted fields, no new state/enum values, no historical-data compatibility impact — renames use the existing session JSON persistence.
- i18n: two new renderer keys (`Rename session title`, `Session title`) added to all seven translated locales.

- Behavior -> command -> result (all run after the last material edit):
  - open/close, focus, row-removal, traversal and inline-edit behavior -> `npx vitest run src/renderer/src/pages/workspace/WorkspaceSidebar.render.test.tsx` -> 42 passed (4 new inline-rename tests; tooltip-content selector updated to hovercard-content)
  - container preview wiring -> `npx vitest run src/renderer/src/pages/workspace/WorkspaceSidebarContainer.test.tsx` -> passed
  - workspace module owner/contract/consumer tests -> `npm run test:module -- workspace_page` -> 25 files / 582 tests passed
  - i18n guard -> `npx vitest run src/renderer/src/i18n/resources.test.ts` -> passed (745 tests incl. sidebar suite)
  - renderer typecheck -> `npm run typecheck:web` -> clean
  - lint on changed files -> `npx eslint <changed files>` -> clean
- Visual verification through the authenticated localhost Web UI on an isolated dev instance (seeded sessions): measured geometry shows the card title within 0.3px of the row title's vertical position, card title 14px == row title 14px, description 12px; the pointer crossed from the row onto the card without closing it; clicking the card title opened the inline editor and Escape cancelled it (no rename committed).
- Full local `npm test` intentionally not run; PR CI is the final authority.

Uncovered risks: exact subpixel alignment can vary with display zoom; Electron-specific chrome was not re-verified (Web and Electron share the renderer).

## Review focus

- `SessionHoverPreview.tsx`: controlled `HoverCard` + manual pointer handlers interplay, and the editing pin (`editingRef`) that suppresses closes mid-edit.
- `WorkspaceSidebar.tsx`: the row element doubling as the hover trigger (desktop only).


ewen-poch:
Addressed all three findings in `e86378e2`:

- **P1 (route inline renames through the session-details owner):** confirmed — a store-level `renameSession` never marks the durable title `sessionDetailsSource: 'manual'`, so a later details generation could overwrite an inline rename. `renameTitle` now delegates to the session-details controller's new `rename`, which loads the authoritative session first when `contentLoaded === false` and calls `sessions.editDetails` with the preserved description, then upserts the returned durable session. Covered by controller tests: editDetails called with trimmed title + preserved description, blank/unchanged titles ignored, unopened sessions loaded before renaming.
- **P2 (title length contract):** the inline input now sets `maxLength={SESSION_DETAILS_TITLE_MAX_LENGTH}` (80), and the editDetails mutation enforces the same limit durably. Asserted in the render test.
- **P2 (keyboard focus transitions):** trigger blur now treats a `relatedTarget` inside the portaled card as internal, and close requests are suppressed while focus is inside the card, so keyboard users can reach the rename control without the card unmounting. Covered by a new render test for both the internal and the leaving transitions.

ewen-poch:
Addressed the remaining P2 in `b8300258`:

- **Prevent HoverCard from closing when focus enters the interactive card:** confirmed in the Radix source — `HoverCardTrigger` composes `onBlur: composeEventHandlers(props.onBlur, context.onClose)`, so the primitive close ran even when my handler returned early. The blur handler now calls `event.preventDefault()` when `relatedTarget` is inside the portaled card (Radix's `composeEventHandlers` honors `defaultPrevented`), which skips the primitive close entirely. For focus moves with a null `relatedTarget` (programmatic focus), the close decision is deferred to a zero-delay timer so `document.activeElement` reflects the settled destination, and the existing "focus inside the card" guard in `requestClose` keeps the card open. The strengthened render test advances fake timers past the 300ms close delay for both the explicit-`relatedTarget` and null-`relatedTarget` paths, and still asserts the card closes once focus leaves the hover region.

I kept the HoverCard primitive rather than moving to a Popover/focus-capture implementation: the hover card is a passive preview whose only interactive element is the rename affordance, and the guard above covers focus entering it. A Popover with focus trap would change the keyboard interaction model of a hover surface (Tab trapping, Esc hierarchy, focus return), which I'd rather not take on in this PR.

PR #1849, opened 2026-08-28
fix(agent-framework): block CodeBuddy Bash Python execution

ewen-poch:
## Problem

CodeBuddy exposed its native Bash tool on macOS and Linux. A request such as “用 python 绘制 sin” could therefore produce `Bash(python -c ...)` and bypass the app-owned persistent Notebook runtime. Command-prefix deny rules are not a complete boundary because absolute paths, versioned interpreters, and wrappers can bypass them.

## Proposed change

- Remove native `Bash` from CodeBuddy's model-callable tool list on every platform.
- Keep ordinary shell work available through the app-owned `open-science-notebook/bash_execute` tool.
- Add a regression test at the public framework launch-configuration boundary.

The final regression test was run three times before the production change and failed consistently because the Linux tool list was `Read,Write,Edit,Glob,Grep,Bash`. It passes after native Bash is removed.

## Scope and non-goals

- CodeBuddy-specific; Claude Code, OpenCode, Codex Responses, and Codex Bridge behavior is unchanged.
- Applies to every model routed through CodeBuddy. Windows already omitted Bash; macOS and Linux now match that boundary.
- No ACP wire-shape, MCP registration, subscription, lifecycle, model, UI, data-model, persisted-format, historical-compatibility, migration, or new-state changes.

- CodeBuddy does not expose native Bash and continues to mount its Notebook MCP server -> `npm test -- src/main/agent-framework/codebuddy.test.ts` -> 11/11 passed.
- Ordinary app-owned shell execution remains registered and forwarded -> `npm test -- src/main/notebook/mcp-server.test.ts -t 'bash_execute'` -> 4/4 passed.
- Node adapter types remain valid -> `npm run typecheck:node` -> passed.
- Changed files satisfy lint and formatting -> `npx eslint src/main/agent-framework/codebuddy.ts src/main/agent-framework/codebuddy.test.ts` and `npx prettier --check ...` -> passed.
- Repository lint -> `npm run lint` -> 0 errors; 256 pre-existing unrelated formatting warnings in database/Notebook tests.

Per the requested local scope, the full portable suite was not run. Exact-head PR Gate remains authoritative for selected consumer and platform coverage.

## Review focus

Please verify that CodeBuddy has no native Bash surface while ordinary shell work remains available through the app-owned Notebook tool.

### DIFF
--- a/src/main/agent-framework/codebuddy.test.ts
+++ b/src/main/agent-framework/codebuddy.test.ts
@@ -55,10 +55,8 @@ describe('codebuddy framework', () => {
       '--setting-sources',
       'user',
       '--tools',
-      expect.not.stringMatching(/Agent|Skill|Workflow|Task|WebFetch|WebSearch/),
+      expect.not.stringMatching(/Agent|Skill|Workflow|Task|WebFetch|WebSearch|Bash/),
       '--disallowedTools',
-      'Bash(python:*)',
-      'Bash(python3:*)',
       'Bash(curl:*)',
       'Bash(wget:*)',
       'Bash(aria2c:*)',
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed in 49a438b5: removed CodeBuddy's native Bash tool instead of relying on bypassable command-prefix deny rules. The regression test now asserts the Linux launch configuration has no Bash, and the existing app-owned Notebook MCP path retains bash_execute for ordinary shell work.

PR #1853, opened 2026-08-28
fix(acp): resume interrupted Claude responses

ewen-poch:
## Problem

Claude Code can terminate an upstream response stream with:

`Internal error: API Error: Connection closed mid-response. The response above may be incomplete.`

The locked `claude-agent-acp` adapter exposes this as a `RequestError` (`code: -32603`, `errorKind: server_error`). Open Science did not classify that fixed wrapper as provider-owned, so a partially streamed turn became an ordinary reportable failure. Users saw no supported way to continue the interrupted request.

External reports show this is an upstream Claude Code transport failure rather than an Open Science ACP defect: [canonical issue](https://github.com/anthropics/claude-code/issues/69336), [duplicate with the exact message](https://github.com/anthropics/claude-code/issues/80825), and [frequency report](https://github.com/anthropics/claude-code/issues/83183).

## Proposed change

- Recognize only Claude Code's fixed `API Error: Connection closed mid-response` wrapper.
- Tag the ACP failure as provider-owned.
- Project a live matching error as the existing `connection-lost` interrupted-turn state, which exposes the existing user-triggered Resume flow and preserves partial output, prompt identity, usage, and context-window evidence.
- Treat the same exact text as non-reportable for historical sessions without rewriting persisted records.

## Scope and non-goals

- No schema, field, architecture, or new state-enum change. This reuses persisted `resumeRecovery.kind = resume-required` and `cause = connection-lost`.
- No persistence migration. Newly observed matching failures persist through the existing interrupted-turn format; historical records only receive the exact text-classification fallback and do not gain a retroactive Resume marker.
- No automatic retry. Resume remains user initiated because already-completed tool side effects are not generally idempotent.
- No broad connection-error heuristic. Nearby generic text remains reportable.
- The new trigger is Claude Code-specific. OpenCode, Codex Responses, and Codex Bridge behavior is unchanged; the existing continuation implementation and parameterized continuation tests remain shared across frameworks.

The regression tests were first run against the unfixed code and failed in three ways matching the report: `providerError` was false, the text was reportable, and the renderer produced no `connection-lost` Resume recovery marker.

Final evidence below was run after the last material edit:

| --- | --- | --- |
| Claude ACP wrapper is provider-owned | `vitest run src/main/acp/prompt-error.test.ts src/main/acp/prompt-outcome-finalizer.test.ts` | passed |
| Exact historical text fallback hides Report without broad matching | `vitest run src/shared/run-error-classification.test.ts` | passed |
| Partial response becomes a resumable interrupted turn | `vitest run src/renderer/src/lib/acp/workspace-events.test.ts` | passed |
| Existing continuation contract remains valid across supported frameworks | `vitest run src/main/acp/interrupted-turn-continuation.test.ts` | passed |
| Combined affected test set | the five targeted files above | 157/157 passed |
| Main/shared types | `tsc --noEmit -p tsconfig.node.json --composite false` | passed |
| Renderer/shared types | `tsc --noEmit -p tsconfig.web.json --composite false` | passed |
| Changed source/test lint and whitespace | `eslint --no-cache <changed files>`; `git diff --check` | passed |

The full `npm test` suite was intentionally not run; exact-head PR Gate/CI is authoritative for the complete portable and platform lanes.

Uncovered risk: a real provider stream drop is nondeterministic and was not induced against the live service. The regression fixture uses the exact externally reported message and the locked adapter's observed wire shape. Resume asks the agent not to repeat completed work, but cannot guarantee idempotency for external side effects already performed before disconnection.

## Review focus

- Whether the matcher is sufficiently narrow for Claude Code's stable wrapper.
- Whether `interruptRun(..., 'connection-lost')` is the right projection for partial provider output.
- Whether preserving forward-only Resume compatibility for historical records is preferable to a persistence migration.

### DIFF
--- a/src/main/acp/prompt-error.ts
+++ b/src/main/acp/prompt-error.ts
@@ -1,5 +1,6 @@
 import {
   isClaudeApiConnectionFailure,
+  isClaudeApiResponseInterruption,
   PROVIDER_CONNECTION_FAILED_PREFIX,
   PROVIDER_RESOURCE_NOT_FOUND_PREFIX
 } from '../../shared/run-error-classification'
@@ -28,9 +29,9 @@ type UpstreamDetail = { text: string; type?: string }
 // bare "not found" substring, so a benign message like "rate limit config not found" isn't reworded.
 const NOT_FOUND_PATTERN =
   /resource[\s_-]?not[\s_-]?found|no such (?:model|resource)|not[\s_-]?found\s*:/i
-// Claude Code wraps upstream provider HTTP 4xx as `Internal error: API Error: 4xx …`. Unreachable-API
-// connection failures use the shared `API Error: Unable to connect to API` matcher. 5xx stays excluded:
-// an untagged internal 5xx may still be an adapter defect.
+// Claude Code wraps upstream provider HTTP 4xx as `Internal error: API Error: 4xx …`. Its two fixed
+// connection wrappers use shared exact matchers. 5xx stays excluded: an untagged internal 5xx may
+// still be an adapter defect.
 const CLAUDE_PROVIDER_CLIENT_ERROR_PATTERN = /^\s*internal error:\s*api error:\s*4\d{2}\b/i
 
 // Converts an unknown thrown value into its base message string.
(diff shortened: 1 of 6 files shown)


ewen-poch:
I traced the reported ordering through `dispatchPrompt` → `failPrompt` → `failRun`, but the recovery state is not erased. `projectFailedRun` spreads the current Session and does not clear `resumeRecovery`, `interrupted`, or the prompt message's `interrupted` marker; it only clears the active-run fields and updates the terminal error projection.

I also exercised the public `sendWorkspaceMessage` path with the matching error event applied before the `sendPrompt` rejection, explicitly waited until the rejected-send fallback entered `getState`, and verified twice that the `connection-lost` recovery and Prompt ID remained intact. No production guard was added because it would duplicate the existing idempotent projection rather than fix a reproducible defect.

PR #1855, opened 2026-08-28
fix(tasks): release raw runtime events after aggregation

ewen-poch:
## Problem

`TaskRunner` retained every matching `AcpRuntimeEvent` for the lifetime of an active Run and then for up to 200 retained terminal Runs. Completion also filtered, reversed, mapped, and joined the full array, increasing peak memory.

A public-boundary regression reproduced the report on the unchanged implementation through `startRun -> runtimeEvents -> waitForRun`: a unique raw event payload remained strongly reachable after completion (`expected 0, received 1`) in three consecutive runs. A mismatched-Session control released it, isolating the retention to the Run event array.

## Proposed change

- Project runtime events incrementally into the data needed for the final result: assistant output/event ids, bounded images, latest stop/error metadata, per-tool activity, and Artifact claim ids.
- Stop retaining complete provider event objects during the Run.
- Snapshot completion arrays before asynchronous Artifact finalization and release the transient accumulator after Session persistence.
- Preserve Claude Code refusal normalization and the shared event behavior used by Claude Code, OpenCode, Codex Responses, and Codex Bridge.

## Scope and non-goals

- No public Task API fields or IPC contracts change.
- No database schema, migration, persisted-format, enum/status, historical compatibility, or UI interaction change.
- This PR does not impose new truncation limits on final assistant output or the number of durable tool activities. Those require a separate product/persistence policy.

Checks below ran after the last material implementation edit:

- Raw payload lifetime during and after a Run -> public TaskRunner lifecycle regression -> passed.
- Async completion snapshot consistency -> regression failed before the review fix by appending a late event id without its text, then passed after projection cloning.
- Assistant output, stop usage, tool activities, Artifact finalization, failure, cancellation, and async snapshot projections -> complete `src/main/tasks/task-runner.test.ts` -> 59/59 passed.
- Changed-file formatting -> Prettier check on both changed files -> passed.
- Source lint -> targeted ESLint -> passed.
- Repository lint -> `npm run lint` -> exited 0; existing warnings were limited to unrelated files.
- PR impact classification -> `main_runtime` selective plan generated successfully; CI lanes include policy, format, lint, Node/Web typecheck, interface contracts, i18n, unit, build, and macOS functional/workspace E2E.

Local `npm run typecheck:node` was attempted but the shared main-checkout dependency tree does not match this worktree lockfile: `@agentclientprotocol/claude-agent-acp` lacks the unrelated `waitForMcpServers` export referenced by `claude-agent-acp-patch.test.ts`. Project guidance prevents linking or installing a second dependency tree without an explicit dependency strategy, so PR CI is authoritative for typechecks.

## Review focus

- Equivalence between the former terminal array scan and the incremental projections.
- Completion snapshot immutability across Artifact finalization and Session persistence awaits.
- The point where the accumulator is released relative to failure handling, automatic review, and terminal status publication.

### DIFF
--- a/src/main/tasks/task-runner.test.ts
+++ b/src/main/tasks/task-runner.test.ts
@@ -2293,6 +2293,111 @@ describe('TaskRunner', () => {
     ])
   })
 
+  it('freezes completion projections before asynchronous artifact finalization', async () => {
+    let emitEvent: ((event: AcpRuntimeEvent) => void) | undefined
+    let signalFinalizeStarted: (() => void) | undefined
+    let finishFinalize: (() => void) | undefined
+    const finalizeStarted = new Promise<void>((resolve) => {
+      signalFinalizeStarted = resolve
+    })
+    const finalizeGate = new Promise<void>((resolve) => {
+      finishFinalize = resolve
+    })
+    const savedSessions: PersistedChatSession[] = []
+    const runner = createRunner({
+      sessions: {
+        list: async () => [],
+        save: async (saved) => {
+          savedSessions.push(structuredClone(saved))
+        }
+      },
+      agent: {
+        withSessionAvailable: async (_projectId, _sessionId, operation) => operation(),
+        listAttachedSessionIds: async () => [],
+        createSession: async () => ({ sessionId: 'session-artifact-gate' }),
+        resumeSession: async (request) => ({ sessionId: request.sessionId }),
+        setPermissionProfile: async () => undefined,
+        cancelPrompt: async () => undefined,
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the Codex P2 in 548442b4d. Completion now snapshots assistant event IDs, images, and per-activity event IDs before asynchronous Artifact finalization. Added a public-boundary regression that failed before the fix by appending a late assistant event ID without its text, then passed after cloning the completion projections. Full TaskRunner suite: 59/59 passed.

PR #1858, opened 2026-08-28
fix(acp): bound agent stderr diagnostics

ewen-poch:
## Problem

Each Agent process stderr chunk was copied verbatim into the main-process log and an ACP system warning. Ordinary research content, local paths, command output, and tool diagnostics are not credential-shaped, so the logger's redaction policy preserved them. A noisy Agent therefore amplified the same raw payload through the logger write chain, ACP publication, TaskRunner event capture, and WebSocket delivery.

The ACP retained history is already bounded, so this is not an unbounded-history bug. The reproducible problem is raw-content exposure plus per-chunk real-time fan-out and queued log work.

## Proposed change

- Aggregate stderr per Agent process into one-second windows.
- Log only metadata by default: framework, lifecycle status, Session count, diagnostic category, chunk count, byte count, window duration, and observed chunk rate.
- Publish one bounded system-warning summary per window, preserving a Session id only while attribution remains unambiguous for the whole window.
- Keep raw stderr support-only and opt-in through `OPEN_SCIENCE_AGENT_STDERR=raw`.
- Cap an opted-in raw sample at 4096 UTF-8 bytes and mark truncation.
- Flush a pending summary when the Agent process exits.

## Scope and non-goals

- No database, persisted format, migration, or historical-compatibility change.
- No new state enum or shared ACP event field.
- No renderer setting or interaction flow is added. The existing warning becomes a metadata summary instead of verbatim stderr by default.
- This PR does not redesign the general logger Promise queue, TaskRunner retention, or WebSocket backpressure. Reducing this source to one small record per second removes the reported amplification at its common origin; other noisy log sources remain a separate concern.

The regression test was run before the fix and failed deterministically: ten stderr chunks produced ten `agent stderr` log calls containing `private-path:/lab/run-42/result.csv`, matching the report.

Source-impact checks after the implementation edit:

- Default privacy, rate limiting, raw opt-in byte cap, lifecycle behavior, and all four execution routes (Claude Code, OpenCode, Codex Responses, Codex Bridge) -> `vitest run src/main/acp/runtime.test.ts` -> 571 passed.
- Common process stderr adapter plus representative TaskRunner and WebSocket consumers -> targeted Vitest set -> 97 passed.
- Final combined owner/consumer Test Impact Set -> targeted Vitest set -> 668 passed.
- Changed TypeScript files -> targeted ESLint -> passed.

Downstream interaction check after aligning the existing status-bearing layout fixture with the metadata summary:

- `playwright test e2e/message-tool-layout-stability.spec.ts` -> 3 passed locally.

Local `npm run typecheck:node` was blocked by the shared dependency tree mismatch: `@agentclientprotocol/claude-agent-acp/dist/acp-agent.js` does not export `waitForMcpServers`. The identical error reproduced on the unmodified `main` checkout. Exact-head PR CI subsequently passed Static checks, including node/web typecheck.

Full `npm test` was intentionally not run. Exact-head PR CI passed the complete required matrix, including macOS module shards, module coverage, Windows core, macOS build and E2E, Windows E2E, CodeQL, and PR Gate.

## Review focus

- Default outputs contain no raw stderr.
- Per-process windows cannot mix Session attribution.
- The support-only raw path remains strictly byte-bounded for multibyte scientific text.
- Process replacement and exit cannot publish a stale Session warning.

### DIFF
--- a/e2e/fixtures/fake-opencode.mjs
+++ b/e2e/fixtures/fake-opencode.mjs
@@ -571,7 +571,7 @@ if (process.argv.includes('--version')) {
           if (prompt.includes(TOOL_STATUS_LAYOUT_SHIFT_PROMPT)) {
             process.stderr.write('Layout fixture status.\n')
           }
-          await delay(750)
+          await delay(prompt.includes(TOOL_STATUS_LAYOUT_SHIFT_PROMPT) ? 2_000 : 750)
 
           const finalMessageId = `e2e-message-${nextMessageId++}`
           await context.client.notify(acp.methods.client.session.update, {
--- a/e2e/message-tool-layout-stability.spec.ts
+++ b/e2e/message-tool-layout-stability.spec.ts
@@ -9,6 +9,7 @@ const TOOL_LAYOUT_SHIFT_PROMPT = 'Run the tool layout stability journey.'
 const TOOL_STATUS_LAYOUT_SHIFT_PROMPT = 'Run the status-bearing layout stability journey.'
 const BUFFERED_TEXT_TOOL_LAYOUT_SHIFT_PROMPT =
   'Run the buffered text tool layout stability journey.'
+const AGENT_STDERR_SUMMARY = 'Agent process stderr: 1 chunk, 22 bytes; raw output omitted.'
 
 test.use({ windowMode: 'normal' })
 
@@ -17,7 +18,7 @@ const cases = [
   {
     name: 'with agent status',
     prompt: TOOL_STATUS_LAYOUT_SHIFT_PROMPT,
-    agentStatus: 'Layout fixture status.'
+    agentStatus: AGENT_STDERR_SUMMARY
   }
 ] as const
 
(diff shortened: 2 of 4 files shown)


ewen-poch:
Addressed both review findings in `7a4b217ea`:

- Delayed stderr windows now retain the originating interaction sequence and revalidate the uniquely active Session and interaction at flush time. A warning from an earlier prompt can no longer inherit the next prompt's `sessionId` or `promptMessageId`.
- Windows made entirely from the two exact non-actionable Codex diagnostics still produce the bounded metadata log summary, but no runtime warning event. This preserves the existing UI suppression without retaining additional raw stderr or adding a shared ACP event field; any other content keeps the window actionable.

Both regressions were added test-first and failed against the previous head for the reported reasons. Targeted validation now passes: 670 ACP owner/consumer tests, changed-file formatting/lint, and all 3 status-layout Electron E2E tests.

PR #1865, opened 2026-08-28
feat(credentials): centralize credential management

ewen-poch:
## Summary

- add a key-icon Credentials destination with Services and Custom sections
- centralize GitHub, literature/NCBI, and OpenAlex credential entry while linking Custom MCP and model-provider credentials back to their existing authoritative editors
- send the saved OpenAlex key on every OpenAlex Connector request and validate candidates in the main process before saving
- park an exact Connector call when its required OpenAlex key is missing, show an Approval-style credential dialog, then resume the same call after a successful save
- add contextual credential links from the Literature Connector, Connector contact settings, and GitHub Skill import

## Architecture and persistence

- `SettingsService` remains the credential owner; renderer snapshots expose configured-state booleans only
- OpenAlex is persisted in `settings.json` as `connectors.openAlexApiKeyRef`, encrypted through Electron `safeStorage`; plaintext is decrypted only in main for validation/dispatch
- internal IPC and persisted fields use camelCase; `api_key` exists only at the external OpenAlex HTTP protocol boundary
- Custom MCP headers/env/OAuth, model-provider keys, GitHub tokens, and Compute SSH credentials remain in their existing owners; Credentials links to those editors instead of duplicating secret storage
- no SQLite tables, columns, migrations, or Prisma changes

## Compatibility and state impact

- historical settings without `openAlexApiKeyRef` load as not configured; no data migration is required
- existing GitHub, NCBI, Custom MCP, provider, and Compute SSH data is reused in place and is not copied
- no persisted enum value or durable status state was added; credential-request lifecycle state is transient and fail-closed on decline, timeout, abort, or application disposal

## Reference behavior

The ego-lite reference was inspected at `http://localhost:8990/`. It exposes a Credentials catalog and a separate `credential_request` card, but an already-failed MCP call is not parked. This implementation preserves the single-owner model while adding contextual links and exact-call resume to avoid the observed interaction gap.

## Verification

- `npm run typecheck:node -- --pretty false`
- `npm run typecheck:web -- --pretty false`
- `npm run lint`
- `npm run check:web-api-map`
- targeted main Connector/settings tests: 144 passed
- targeted renderer/settings/i18n tests: 889 passed
- targeted renderer-contract/application-event tests: 135 passed

Per project guidance, the full `npm test` suite was not run locally; CI is the full-suite authority for this PR.


ewen-poch:
Addressed the review findings in `8933b64d`:

- **Policy revalidation:** bundled connector enablement and tool policy are re-read after credential recovery and immediately before dispatch. A previously satisfied Once approval is retained, while a newly applied Block or Disable decision fails closed.
- **NCBI key removal:** the Literature credential view now exposes a conditional **Remove key** action that submits an empty API key while preserving the current contact email.
- **Concurrent credential requests:** one successful save now settles every parked request for the same credential, and the renderer removes the matching queued requests together so no redundant follow-up dialog appears.

Regression coverage was added for Block/Disable changes during the prompt, Once-approval preservation, grouped parked-call resumption, grouped renderer settlement, and NCBI removal. Targeted result: 97 tests passed; lint and both node/web typechecks passed.

ewen-poch:
Addressed the remaining current-head review finding in 2c2e17ad:

- OpenAlex credential validation now treats HTTP 429 from /rate-limit as a valid authenticated credential, allowing the key to be saved while its quota is exhausted.
- HTTP 401/403 remain rejected, and other non-success or network failures remain unavailable.
- Added a focused regression test for the 429 case.

Validation: 49 targeted tests passed, ESLint passed, and git diff --check passed. This reuses the existing validation result contract and adds no status, persistence, schema, or interaction changes.

ewen-poch:
Addressed both current-head findings in c380deaa:

- **Specialist credential freshness:** bundled Specialist calls now read the durable Connector snapshot immediately before their initial credential lookup. Replacing an OpenAlex key sends only the new key; revoking it fails with credential_required and never dispatches the stale projected key.
- **Credential recovery ownership:** request, settlement, replay, and response capabilities are now Electron-only. Browser surfaces no longer receive a credential dialog they cannot complete, and the shared renderer treats those Electron-only members as optional.

The generated Web API map and explicit surface omission inventory were updated accordingly. This keeps credential ownership on the local Electron host and does not add persistence, schema, enum, notification-inbox, or SQLite changes.

Validation: Connector 60/60, renderer store/App 87/87, contract/preload 90/90; Node and Web typechecks, ESLint, Web API map check, and git diff --check all passed.

ewen-poch:
Removed the stale Web application-command declarations and handlers for connector credential recovery. The credential recovery request/response path remains Electron IPC-only, matching the renderer contract and Web API exposure already established in the previous change.\n\nValidation:\n- 17/17 settings integration and application-command composition tests\n- 160/160 related settings, renderer-surface, preload, and connector-service tests\n- Node and Web typechecks\n- ESLint on touched files\n- Web API map check\n- git diff --check\n\nThis does not change the credential architecture, data model, persistence, enums, or user interaction.

ewen-poch:
Addressed the current-head P1 in f4c646f0. Connector credential recovery now checks for a live local Electron credential owner at call time. Headless processes, and desktop processes with no BrowserWindow, fail closed immediately with credential_required instead of parking the call until the broker timeout. Desktop calls with a local window keep the existing prompt-save-resume flow.\n\nAdded a composition-level regression test confirming that no credential request is broadcast and no OpenAlex request is dispatched when the local owner is unavailable.\n\nValidation: 181 targeted tests passed; Node and Web typechecks, ESLint, Web API map check, and git diff --check passed. This adds no persistence, schema, enum, architecture, or visible desktop interaction changes.

ewen-poch:
Addressed the current-head AI review finding in b27ce8b7. MaskedPasswordField now lets a caller handle keydown before its default multiline Enter behavior; the GitHub credential form prevents the default and verifies/saves on Enter, so no newline is inserted into the token. The focused regression test now submits with Enter and verifies the exact token sent. Validation: 9 targeted tests passed, Web typecheck passed, ESLint passed, and git diff --check passed. This changes no architecture, data model, persistence, enums, or visual layout.

ewen-poch:
Addressed the current-head Remote Web P2 in 8009d1a4. The existing LOCAL getGitHubTokenStatus command now acts as the capability probe: Remote Web keeps GitHub and OpenAlex visible but disables their actions as Desktop only, while stale/direct detail routes show a clear desktop-only explanation without rendering credential fields. Literature access remains available remotely, and Electron/local Web behavior is unchanged. Verified with the focused CredentialsPanel tests, the i18n resource guard (747 tests), Web typecheck, and ESLint. This does not change the architecture, data model, persistence, enums, or credential security boundary.

ewen-poch:
Addressed the current-head Specialist authorization P1 in 99b0d1d0. After a credential prompt resolves, bundled Specialist calls now re-resolve the current Specialist profile immediately before dispatch. Disabling or deleting the Specialist, or removing the Connector from its capability scope while the call is parked, fails closed and sends no external request. Added parameterized regression coverage for all three revocation paths. Validation: ConnectorService 63/63, Node typecheck, ESLint, and git diff --check passed. This reuses the existing access gate and adds no architecture, data model, persistence, enum, or user-interaction changes.

ewen-poch:
Addressed the current-head credential-revocation P1 in b05cd30d. After the Specialist access recheck succeeds, the call now reloads credentials from the current durable Connector snapshot and verifies the required credential again before dispatch. If the OpenAlex key is removed while the access recheck is pending, the call fails closed with credential_required and sends no external request. Added a focused regression test for this exact race. Validation: ConnectorService 64/64, Node typecheck, ESLint, and git diff --check passed. This adds no architecture, data model, persistence, enum, or user-interaction changes.

ewen-poch:
Rebased the branch onto the latest origin/main and addressed the current-head GitHub feedback accessibility P2 in 1b287052. The saved-token status and transient feedback now have distinct IDs, and the token field dynamically references every description element currently rendered. This preserves the saved mask during verification failures while eliminating duplicate IDs. The post-rebase Specialist tests were also aligned with the upstream SpecialistView rename in 2fb9f4b6. Validation: 816 focused module/i18n tests, 69 ConnectorService/GitHub control tests after the compatibility adjustment, Node and Web typechecks, ESLint, and git diff --check passed. Prisma Client was regenerated locally for the updated upstream schema; no schema, table, persistence, data-model, enum, architecture, or user-interaction changes were introduced.

ewen-poch:
Resolved the current-head documentation mismatch in 9423755b while preserving the intentional centralized credential workflow. Skill Import keeps its neutral Manage GitHub credential action and navigates to Credentials > GitHub; the centralized detail page remains the single place to inspect, verify, replace, or remove the token. docs/design.md now documents that approved interaction instead of the obsolete inline editor convention. This avoids duplicate secret-management surfaces and does not change runtime behavior, architecture, data model, persistence, schema, enums, or credential security boundaries. Validation: Prettier check and git diff --check passed; no code logic changed, so npm test was not required by the repository rules.

PR #1868, opened 2026-08-28
fix(updater): allow retry after failed download

ewen-poch:
## Problem

When an in-place update download fails, the strategy replaces the active offer with a minimal error status. This drops the offered version, release notes, and artifact size, which closes the update dialog. Error statuses that do retain an offered version still cannot retry because download only accepts the available state.

## Proposed change

- preserve the active offer when a download fails
- allow an error status with an offered version to transition back to downloading
- clear attempt-local error and progress fields before retrying
- extend the existing public strategy test to prove the first failure preserves metadata and the second download reaches ready

## Scope and non-goals

- no new UpdateState value
- no new UpdateStatus field
- no persistence, migration, or historical-data compatibility change
- no renderer or interaction redesign
- no generalized state-machine refactor

The regression test was run twice against the original implementation and deterministically failed because latest, notes, and totalBytes were missing, the retry remained in error, and the updater started only once.

Final checks, run after the last material edit:

- update module behavior: focused src/main/update run passed 8 files and 107 tests
- linted source: full ESLint run passed
- diff hygiene: git diff --check passed
- Node typecheck: attempted, but the shared local dependency tree does not match this worktree lockfile; an unrelated ACP test imports waitForMcpServers from an installed package version that does not export it

PR CI is authoritative for the clean-lockfile Node typecheck and packaged/platform lanes. No full local npm test was run, per maintainer request.

## Review focus

Please verify that treating error statuses carrying latest as restartable downloads is the intended recovery behavior for in-place updates, and that attempt-local error/progress fields are reset without dropping offer metadata.

### DIFF
--- a/src/main/update/electron-updater-strategy.test.ts
+++ b/src/main/update/electron-updater-strategy.test.ts
@@ -363,10 +363,13 @@ describe('ElectronUpdaterStrategy', () => {
     expect(updater.downloadUpdate).toHaveBeenCalledTimes(1)
   })
 
-  it('records a failed in-place download without the provider error message', async () => {
+  it('preserves the offer and retries after a failed in-place download', async () => {
     const updater = new FakeUpdater()
+    let starts = 0
     updater.runDownload = async () => {
-      throw new Error('private updater diagnostic detail')
+      starts += 1
+      if (starts === 1) throw new Error('private updater diagnostic detail')
+      updater.emit('update-downloaded', { version: '0.3.0' })
     }
     const log = createLogSpy()
     const strategy = new ElectronUpdaterStrategy({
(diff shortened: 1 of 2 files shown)


ewen-poch:
CI follow-up: Windows E2E failed in the workspace journey on both the original run and a job-only rerun. Both artifacts show the same three message-tool-layout-stability cases (including their Playwright retries) timing out before the first fake-agent reply because the app displays 'Conversation storage needs attention' / 'This conversation changed in another window.' The updater path is not involved. All other PR Gate jobs, including Windows core, static checks, both macOS module shards, module coverage, macOS build/E2E, CodeQL, and Codex review, passed. I did not add unrelated session-storage or E2E changes to this focused updater PR.

PR #1872, opened 2026-08-28
fix(notifications): enforce mobile message-center modality

ewen-poch:
## Problem

On mobile, the message center declares `role="dialog"` and `aria-modal="true"`, but previously only locked body scrolling. Keyboard focus could move into the obscured application, dismissal did not always restore a usable bell trigger, and responsive or navigation transitions could leave focus on `document.body` or outside the modal.

## Proposed change

- Let Radix `FocusScope` own mobile initial focus, focus trapping, and focus looping.
- Mark the application root `inert` and `aria-hidden` while the portaled mobile drawer is open, restoring its prior values on close.
- Restore the invoking bell after mobile dismissal, or fall back to another visible bell when the opener becomes inert or unmounts during navigation.
- Preserve focus restoration if a mobile center crosses to desktop before unmounting.
- Exclude inert and `aria-hidden` notification bells from fallback visibility checks.
- Move focus into an already-open center when it transitions from desktop to mobile modality.
- Add public interaction regressions for background focus, Escape and overlay dismissal, inert and unmounted openers, and breakpoint transitions.

## Scope and non-goals

- Desktop message-center behavior and positioning are unchanged.
- No dependencies or user-visible copy are added.
- No architecture, data fields, enum values, data models, migrations, or persistence behavior change.

The regression tests were first confirmed to fail on the pre-fix implementation for the reported focus-isolation and focus-restoration behavior.

| Affected behavior | Final check | Result |
| --- | --- | --- |
| Mobile modality, dismissal, fallback focus restoration, responsive transitions, and unmount lifecycle | `npx --no-install vitest run src/renderer/src/components/NotificationBell.test.tsx` | 29 passed |
| Renderer type safety | `npm run typecheck:web` | Passed |
| Lint rules | `npm run lint` | Passed |
| Final impact classification | `npm run test:affected:explain -- --base origin/main --head HEAD` | Full plan selected because this shared component has no module owner; complete coverage delegated to PR Gate per maintainer direction |

All listed passing checks ran after the last material edit.

PR Gate passed static checks, both full module-test shards, module coverage, Windows core, and macOS build/E2E. Its remaining Windows E2E failure is unrelated to the notification center: one settings-persistence test passed only on retry and the workspace message-tool layout test timed out without opening this surface.

Uncovered risk: browser-level manual verification was unavailable because the in-app browser blocked localhost navigation. The retained component tests exercise the actual DOM focus, responsive, dismissal, and unmount boundaries.

## Review focus

Please review `FocusScope` lifecycle timing, background-isolation cleanup and focus-restoration ordering, fallback selection when the invoking bell becomes inert or unmounts, and focus ownership across both breakpoint directions.

### DIFF
--- a/src/renderer/src/components/NotificationBell.test.tsx
+++ b/src/renderer/src/components/NotificationBell.test.tsx
@@ -67,6 +67,35 @@ const stubMobileViewport = (): void => {
   )
 }
 
+const stubMutableViewport = (): { setMobile: (mobile: boolean) => void } => {
+  let matches = false
+  const listeners = new Set<() => void>()
+  const media = {
+    get matches() {
+      return matches
+    },
+    media: '(max-width: 47.999rem)',
+    onchange: null,
+    addListener: vi.fn(),
+    removeListener: vi.fn(),
+    addEventListener: vi.fn((_event: string, listener: () => void) => listeners.add(listener)),
+    removeEventListener: vi.fn((_event: string, listener: () => void) =>
+      listeners.delete(listener)
+    ),
+    dispatchEvent: vi.fn()
+  }
+  vi.stubGlobal(
+    'matchMedia',
+    vi.fn(() => media)
+  )
+  return {
+    setMobile: (mobile) => {
+      matches = mobile
+      listeners.forEach((listener) => listener())
+    }
+  }
+}
+
 describe('NotificationBell', () => {
   it('renders a red-dot entry point with an accessible unread count and pending state', async () => {
     await act(async () => root.render(<NotificationBell />))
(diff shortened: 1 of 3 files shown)


ewen-poch:
Investigated the latest Codex P1 as a false positive. FocusScope renders the panel element itself through Primitive.div and forwards the supplied classes; it does not add a wrapper around another panel element here. A temporary DOM assertion against the existing mobile drawer test confirmed that the role=dialog element itself has flex and flex-col, and its min-h-0 flex-1 overflow-y-auto list is a direct child. The focused test passed, and the temporary diagnostic assertion was removed. Using asChild directly would also be invalid for the current two sibling children. No production change is warranted for this finding.

ewen-poch:
CI note for final head 3092b41f9: all static checks, both full Module-test shards, coverage, Windows core, and macOS build/E2E passed after the failed-job rerun. Windows E2E remains red for unrelated instability: settings-persistence was flaky (Project was not created on the first attempt, retry passed), and three message-tool-layout-stability cases timed out waiting for the deterministic agent reply. Those journeys do not open the message center, and this branch only changes NotificationBell behavior while its panel is open. The same workspace failures recurred across attempts, so I am not expanding this accessibility PR into the unrelated deterministic-fixture/E2E issue. Run: https://github.com/aipoch/open-science/actions/runs/33200324615/attempts/2

PR #1875, opened 2026-08-28
fix(task-api): secure and resume event streams

ewen-poch:
## Problem

The public Task WebSocket projected the global ACP application event bus directly. A Desktop or ordinary Web Session could therefore expose detailed messages, tool activity, or permission requests to every Task API subscriber even though the Task API could otherwise read only Session summaries.

The public event variants also lacked a stable common Run identity and had no replay protocol. Reusing one Session for sequential Runs made sessionId-only correlation ambiguous, and an established SDK socket ended permanently after disconnect. Remote pairing exposed Task HTTP routes but rejected the corresponding /api/v1/events WebSocket.

The regression tests first reproduced all three failures on the unmodified implementation: an unowned Session event was delivered, an established SDK socket did not reconnect, and paired authorization returned no capability for the public Task event path.

## Proposed change

- Make TaskRunner the authority for sessionId to active Run ownership, with an additional promptMessageId guard when ACP supplies it.
- Fail closed when projecting ACP events that cannot be resolved to an active Task Run.
- Give every business event additive sequence, runId, sessionId, and projectId fields.
- Add protocol version 1 with a bounded process-local replay suffix (2048 frames or 16 MiB), cursor resume, and explicit resync-required control events.
- Reconnect the Node SDK with its last cursor only after transient disconnects; complete on normal shutdown and fail on terminal protocol or policy closes.
- Make the CLI correlate by runId and preserve a historical sessionId fallback.
- Authorize paired /api/v1/events sockets through the same remote-session checks already used for the internal Web event socket.

## Scope and non-goals

- No Prisma, SQLite, Session JSON, or other persisted-format change.
- No new Session or Run status.
- New wire-only control state: stream.resync-required, with reason stream-changed or cursor-expired.
- Replay is deliberately process-local and bounded; daemon restart or eviction requires an HTTP Run/Session resync.
- Legacy clients that omit eventProtocol remain live-only. Existing business frame fields remain, with the common identity fields added at the top level.
- No ACP provider protocol, translation, or provider lifecycle changes. claude-code, opencode, codex-response, and codex-bridge adapter suites are outside the impact set because filtering occurs after the shared application event hub.

Historical compatibility: no stored data requires migration. A new SDK can still consume an older server's live frames, but cursor replay becomes available only when the server advertises protocol version 1 through its readiness frame.

- Task ownership, projection, replay owner, SDK reconnect/control frames, CLI Run correlation, remote pairing, and shared contract behavior -> targeted Vitest command -> 8 files, 150 tests passed.
- SDK terminal-versus-transient socket close lifecycle -> Vitest packages/open-science/client.test.ts -> 26 tests passed.
- Public WebSocket isolation and disconnect replay through the HTTP server boundary -> Vitest src/main/web-service/http-server.test.ts -> 34 tests passed.
- Main-process and shared TypeScript contracts -> npm run typecheck:node -> passed.
- Repository lint and formatting rules -> npm run lint -> passed.
- git diff --check -> passed.

The full npm test suite was intentionally not run, per the requested module-focused workflow. Exact-head PR Gate remains authoritative for additional portable and platform lanes. The only intentionally uncovered behavior is replay across daemon restarts, which reports resync-required instead of persisting raw event payloads.

## Review focus

- Whether active Run ownership is registered and cleared at the correct TaskRunner lifecycle boundaries.
- Whether cursor-expired and stream-changed are sufficient explicit recovery states for this versioned public protocol.
- Whether terminal WebSocket close codes and live-only behavior for clients without eventProtocol form the right compatibility boundary.


ewen-poch:
CI follow-up for exact head `40011c74`:

- Full Module tests shard 2 was cancelled on attempt 1 only because dependency installation consumed 10m08s of the 15-minute job budget. The failed-job rerun completed in 10m43s, and both the shard and merged Module coverage passed.
- Windows functional E2E passed on attempt 2. Windows workspace E2E still fails in the three existing `message-tool-layout-stability.spec.ts` scenarios before the layout assertions, because the fake Agent baseline reply is absent.
- This is not introduced by this PR: an unrelated earlier PR Gate run (`33191425968`, job `98918021458`) has the same three tests, locator, expected reply, and source lines failing; five other concurrent unrelated PRs also have a failed Windows E2E job. This PR does not modify the spec, Electron fixture, renderer conversation path, or fake Agent, and the same workspace lane passes on macOS for this head.

I did not mix an unrelated Windows E2E/fake-Agent repair into the Task API event-stream change. The PR Gate is currently blocked only by that repository-wide Windows lane; all Task API tests, both macOS full Module shards, merged coverage, static checks, macOS E2E, Windows core, CodeQL, and the independent Codex review passed.

PR #1892, opened 2026-08-29
fix(notebook): preserve R kernel across repeated cancellation

ewen-poch:
## Problem

A late POSIX SIGINT can escape between the R loop local read/evaluation handlers and response emission. Rscript then exits with Execution halted, the persistent namespace is lost, and the stress test can attach its next abort hook to the stale child so that the next request reports completed instead of cancelled.

## Proposed change

- Keep one recoverable interrupt boundary active across R request transitions and response emission.
- Track only the current in-memory request ID so a late interrupt can emit the matching interrupt acknowledgement before the loop continues.
- Restore the exact original stdin write method in the stress helper and shorten the cancelled sleep while retaining 50 repeated cancellations and per-iteration diagnostics.

## Scope and non-goals

No public API, architecture boundary, UI interaction, data model, persisted format, database schema, migration, external status enum, or historical-data compatibility change. Windows retains its existing terminate-and-respawn cancellation behavior.

The checks below ran against the final material implementation:

- repeated R cancellation and namespace preservation -> rtk npm test -- src/main/notebook/kernel-executor.test.ts -> 89 passed, 1 skipped
- R SIGINT protocol alignment and acknowledgement -> RUN_KERNEL=1 OPEN_SCIENCE_TEST_R_ENV=/usr/local rtk npm test -- src/main/notebook/r-loop.integration.test.ts -t SIGINT -> 3 passed, 43 not selected
- Node process types -> rtk npm run typecheck:node -> passed
- linted source and tests -> rtk npm run lint -> passed

A complete local npm test run was intentionally not run; PR CI is the authority for the full portable and platform suites.

## Review focus

Please focus on the lifetime of current_req_id around EOF, normal response emission, and the outer interrupt recovery handler. The state is process-local and never reaches the wire or persistence.

### DIFF
--- a/resources/notebook/r_loop.R
+++ b/resources/notebook/r_loop.R
@@ -1470,9 +1470,9 @@ rm(namespace_state)
 base::local({
   state <- new.env(parent = emptyenv())
   state$during_read <- FALSE
-  # Retain the fully read frame's identity until its response is emitted. A SIGINT can otherwise
-  # escape between the local read/eval handlers and terminate Rscript without acknowledging it.
-  state$current_req_id <- NULL
+  # Retain the fully read frame until its response is emitted. A SIGINT can otherwise unwind
+  # read_request after consuming the body and discard the request before it is acknowledged.
+  state$current_req <- NULL
 
   read_request <- function() {
     empty_streak <- 0L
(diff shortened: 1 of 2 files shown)


ewen-poch:
Addressed the completed-frame interrupt window in d4fb5507. The R loop now retains the full decoded request while finalizing it; if SIGINT unwinds that step, the read handler resumes with the retained request and acknowledges cancellation under the original request ID before processing the probe. I also added a deterministic integration test that injects SIGINT after frame decoding: it reproduced kernel replacement and namespace loss before the fix and passes afterward. Focused validation passed: kernel-executor.test.ts (90 passed, 1 skipped), typecheck:node, and lint.

ewen-poch:
I am not changing the response-emission path for this finding. In the supported protocol, every POSIX SIGINT sent to an R kernel is immediately followed by queueRInterruptProbe while the original pending request retains queue ownership. If SIGINT is deferred until after current_req is cleared, during_read applies to that private probe; building the probe's interrupted response resets during_read before any subsequent user request is accepted. The existing response-before-ack cancellation and timeout tests each execute subsequent namespace requests successfully, and the real-R 50-cancellation regression also verifies the next user request and process identity. Adding a separate emission-phase state would duplicate this probe state machine and risk consuming the acknowledgement the executor is waiting for.

PR #1894, opened 2026-08-29
fix(session-persistence): reject linked authority paths

ewen-poch:
## Problem

Active Session persistence did not validate symbolic-link boundaries consistently. A linked `sessions` root or Project directory allowed `saveSession()` to write authoritative JSON outside the configured storage root. Directory scans also ignored linked Project entries and JSON-shaped special nodes while reporting `isComplete: true`, allowing downstream reconciliation to treat hidden Session authority as absent.

## Proposed change

- validate the active `sessions` root, Project directory, and Session file with non-following metadata checks
- reject writes through linked or non-regular authority paths
- mark scans incomplete when Project links or JSON-shaped special nodes are present
- validate the same boundaries for direct Session reads
- keep anomalous nodes in place instead of deleting, quarantining, or migrating them
- add cross-platform regression coverage using POSIX directory links and Windows junctions
- complete the queue-test filesystem mock with the new `lstat` boundary metadata

## Scope and non-goals

- no database, persisted JSON, public interface, renderer, or domain-model changes
- no new persisted or domain enum values; the `missing | valid | invalid` classification is repository-internal only
- existing regular historical Session data remains compatible
- intentionally linked historical Session layouts now fail closed and require manual conversion to regular directories/files; this PR does not migrate them automatically
- does not attempt to eliminate a same-user concurrent path-swap TOCTOU race, which would require a substantially different native directory-handle design

Production behavior checks ran after the final production edit. After CI identified an incomplete hoisted filesystem mock, the latest test-only edit was validated together with its repository owner file.

- linked root/Project writes are rejected and no external JSON is created -> focused symbolic-link regression loop -> 5 passed
- linked Project and JSON-shaped nodes make scans incomplete -> focused symbolic-link regression loop -> 5 passed
- existing Session persistence/recovery behavior remains intact -> `repository.test.ts` -> 79 passed
- Session persistence owner, contract, and representative consumer behavior after the final production edit -> module impact set -> 7 files, 360 passed
- queue scheduling plus repository owner behavior after the latest test-mock edit -> `repository.test.ts` + `repository-queue.test.ts` -> 2 files, 81 passed
- changed files match repository formatting and lint rules -> Prettier, targeted ESLint, and `git diff --check` -> passed
- Node typecheck was attempted after the final production commit but the shared local dependency tree lacks the `waitForMcpServers` export expected by `src/main/acp/claude-agent-acp-patch.test.ts`; no diagnostic referenced the changed files. PR CI uses a clean install and is authoritative.

The complete portable suite was not run locally; PR Gate and its selected platform lanes are authoritative.

## Review focus

- fail-closed completeness semantics for anomalous authority nodes
- preservation of missing-directory and durable-temp recovery behavior
- whether rejecting, rather than migrating, intentional historical symlink layouts is the correct compatibility boundary
- metadata-operation cost: each scanned boundary is validated once, with repeated checks skipped after validation

### DIFF
--- a/src/main/session-persistence/repository-queue.test.ts
+++ b/src/main/session-persistence/repository-queue.test.ts
@@ -3,6 +3,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
 import type { PersistedChatSession } from '../../shared/session-persistence'
 
 const fsMock = vi.hoisted(() => ({
+  lstat: vi.fn(),
   mkdir: vi.fn(),
   open: vi.fn(),
   readFile: vi.fn(),
@@ -56,6 +57,15 @@ const createSession = (id: string, projectId = 'project-a'): PersistedChatSessio
 describe('session persistence repository save ordering', () => {
   beforeEach(() => {
     vi.resetAllMocks()
+    fsMock.lstat.mockImplementation((path: string) =>
+      path.endsWith('.json')
+        ? Promise.reject(Object.assign(new Error('not found'), { code: 'ENOENT' }))
+        : Promise.resolve({
+            isDirectory: () => true,
+            isFile: () => false,
+            isSymbolicLink: () => false
+          })
+    )
     fsMock.open.mockResolvedValue({
       close: vi.fn().mockResolvedValue(undefined),
       sync: vi.fn().mockResolvedValue(undefined)
--- a/src/main/session-persistence/repository.test.ts
+++ b/src/main/session-persistence/repository.test.ts
@@ -1,4 +1,14 @@
-import { mkdir, mkdtemp, readFile, readdir, rename, rm, utimes, writeFile } from 'node:fs/promises'
+import {
+  mkdir,
+  mkdtemp,
+  readFile,
+  readdir,
+  rename,
+  rm,
+  symlink,
+  utimes,
+  writeFile
+} from 'node:fs/promises'
 import { join } from 'node:path'
 import { tmpdir } from 'node:os'
 import { pathToFileURL } from 'node:url'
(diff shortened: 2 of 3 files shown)


ewen-poch:
AI review disposition: no code change in this PR. The findings describe a same-user process concurrently replacing filesystem names between operations, rather than the reported persistent symlink layout. A process with that access can already overwrite or delete Session authority directly, so it is not a separate privilege or trust boundary in the current local-desktop threat model. Closing only the cited read and write gaps would also be incomplete: durable recovery, directory enumeration, quarantine, backup, rename, and cleanup would all need one cross-platform descriptor-relative implementation, most likely by substantially expanding the native filesystem module. The PR body therefore explicitly lists concurrent path-swap TOCTOU elimination as a non-goal. This PR remains intentionally limited to rejecting links and special nodes that exist when an operation observes them, and to reporting scans incomplete. If same-user adversarial namespace mutation becomes an in-scope threat, it should be handled as a separate native-storage design with race-injection coverage across every persistence operation.

ewen-poch:
Addressed the linked deleted-sessions root finding. SessionRepository now validates the deletion root before committed-authority reads and writes, and applies the same fail-closed check to deletion moves, tail cleanup, state classification, and tombstone enumeration. The regression tests exercise the public repository behavior and demonstrated external JSON write, read, move, and removal before the fix; all six linked-root cases now pass, and the full repository plus queue module suite passes 87/87. This does not change persisted JSON, data models, enum values, or user interaction.

ewen-poch:
Not changing this in this PR. `SessionRepository` receives `resolveStorageRoot()` as the configured trust anchor, while this change is scoped to authority nodes below that root. Rejecting a symlink in the root or in every ancestor would change the storage-root contract and historical compatibility; the codebase explicitly accounts for legitimate platform paths behind aliases or mounts (for example macOS `/var` and some Linux `/home` layouts). Checking only the final `storageDir` component would be incomplete, while a complete race-resistant solution requires the cross-platform descriptor-relative native I/O design already declared out of scope. The regression coverage in this PR therefore continues to validate the `sessions`, project, Session JSON, and deletion-authority nodes below the configured root.

PR #1895, opened 2026-08-29
fix(platform): correct update and CLI platform handling

ewen-poch:
## Problem

- Settings always instructed users to open an installer for a ready update, even when Windows/Linux use the restart apply strategy.
- Manifest selection mapped every Windows/Linux architecture to x64 artifacts and could therefore select an incompatible download on ARM64.
- Windows CLI PATH checks were case-sensitive and did not normalize trailing separators; uninstall removed the managed shim without removing the app-managed user PATH entry.

## Proposed change

- Make the Settings ready message follow `applyKind`, reusing the existing translated `Restart to update` copy.
- Return no manifest artifact for unsupported architectures and limit the Linux x64 AppImage fallback to x64 hosts.
- Normalize Windows PATH entries case-insensitively with trailing separators removed for both process and persistent-user checks.
- Before Windows installation mutates the user PATH, write and durably flush the before/after PATH snapshots to a unique same-directory temporary journal, then atomically publish it as `pending`. Atomically rename `pending` to the owned receipt after the registry mutation; a later install or uninstall can reconcile an interrupted mutation from those snapshots.
- During Windows uninstall, reconcile a valid journal independently of whether the shim still exists, then remove the shim if present. Restore the recorded `beforePath` only when the current PATH exactly matches `afterPath`; if no equivalent entry remains, clean up only the journal and shim. Any other continuity mismatch, conflicting or unmanaged journal, or unreconcilable pending state fails closed while preserving PATH, journal, and any shim.

## Scope and non-goals

- This does not add ARM64 Windows/Linux artifacts or manifest keys; unsupported architectures fail closed until those artifacts exist.
- This does not claim existing or legacy PATH entries retroactively. A managed shim without the new receipt is removed while its PATH entries are preserved.
- Strict ownership continuity means automatic uninstall can require manual PATH resolution if the user changes PATH after installation while an equivalent app bin entry remains.
- No dependency, database, renderer contract, or update-state shape changes are included.

## Historical compatibility, state, and persistence

- Historical compatibility: strict forward-only ownership. Legacy shims without a receipt still install, report status, and uninstall normally, but uninstall preserves their PATH entries. There is no receipt adoption or orphan cleanup heuristic.
- New enum/state values: the private Windows PATH journal has `pending` and `owned` states, encoded by `.open-science-path-pending` and `.open-science-path-receipt`; no shared update or renderer enum changes.
- Persistence: the journal is stored beside the managed shim and contains a version, ownership marker, canonical bin directory, and the local user PATH snapshots immediately before and after the app-owned append. Creation uses a unique `.open-science-path-pending.<guid>.tmp` name; only a temporary file successfully created by the current process is removed on handled failure, and only a fully flushed file is atomically published as `pending`. A process or machine crash before publication can leave one small temporary file beside the shim; later runs ignore it and do not delete it because ownership cannot be proven. The committed journal is removed after successful cleanup. There is no database field, migration, or schema change.

- Settings ready copy follows `applyKind` -> `npm test -- src/renderer/src/pages/settings/AppVersionSection.render.test.tsx` (included in combined focused run) -> passed.
- Unsupported Windows/Linux architectures cannot select x64 artifacts -> `npm test -- src/shared/update.test.ts` (included in combined focused run) -> passed.
- Windows PATH normalization, atomic pending-journal publication, interrupted-write retry safety, crash recovery, strict snapshot continuity, pre-write journal validation, shim-independent journal cleanup, unmanaged-journal rejection, fail-closed behavior, cleanup ordering, and legacy preservation -> `npm test -- src/main/cli-install/launcher.test.ts` -> 51 tests passed after rebasing onto `origin/main`.
- Combined focused run -> `npm test -- src/renderer/src/pages/settings/AppVersionSection.render.test.tsx src/shared/update.test.ts src/main/cli-install/launcher.test.ts` -> 3 files / 70 tests passed after rebasing onto `origin/main`.
- Changed-file formatting -> `npx prettier --check ...` -> passed.
- Impact explanation -> `npm run test:affected:explain -- --base origin/main --head HEAD` -> full CI plan because these paths are not registered to a module owner.

The complete local `npm test` suite was intentionally not run; PR CI is the final authority for the classifier-selected full plan and Windows-specific execution.

## Review focus

- Confirm that unsupported architectures should fail closed rather than inventing not-yet-supported manifest keys.
- Confirm the Windows ownership protocol: a uniquely named temporary journal is flushed and atomically published as pending before the registry mutation, temporary-file cleanup is limited to files created by the current process, owned state is committed by same-directory rename, valid journals are reconciled independently of the shim, and automatic cleanup requires exact `afterPath` continuity while receipt-less legacy entries remain untouched.

### DIFF
--- a/src/main/cli-install/launcher.test.ts
+++ b/src/main/cli-install/launcher.test.ts
@@ -443,11 +443,23 @@ describe('buildWindowsPathCommand', () => {
     const script = args[args.length - 1]
     expect(script).toContain(`$binDir = '${binDir}'`)
     expect(script).toContain(`$pendingPath = '${join(binDir, '.open-science-path-pending')}'`)
+    expect(script).toContain(
+      `$pendingTempPath = '${join(binDir, '.open-science-path-pending.tmp')}'`
+    )
     expect(script).toContain(`$receiptPath = '${join(binDir, '.open-science-path-receipt')}'`)
     expect(script).toContain(`$receiptOwner = '${WINDOWS_PATH_RECEIPT_OWNER}'`)
     expect(script).toContain("TrimEnd([char[]]'\\/') -ieq $normalizedBinDir")
     expect(script).toContain("[Environment]::SetEnvironmentVariable('Path'")
+    expect(script).toContain(
+      '$stream = [IO.File]::Open($pendingTempPath, [IO.FileMode]::CreateNew,'
+    )
+    expect(script).toContain(
+      'if ([IO.File]::Exists($pendingTempPath)) { [IO.File]::Delete($pendingTempPath) }'
+    )
     expect(script.indexOf('$stream.Flush($true)')).toBeLessThan(
+      script.indexOf('[IO.File]::Move($pendingTempPath, $pendingPath)')
+    )
+    expect(script.indexOf('[IO.File]::Move($pendingTempPath, $pendingPath)')).toBeLessThan(
       script.lastIndexOf("[Environment]::SetEnvironmentVariable('Path', $next")
     )
     expect(script.lastIndexOf("[Environment]::SetEnvironmentVariable('Path', $next")).toBeLessThan(
--- a/src/main/cli-install/launcher.test.ts
+++ b/src/main/cli-install/launcher.test.ts
@@ -324,7 +324,7 @@ describe.each([
     await expect(getCliLauncherStatus(env)).resolves.toMatchObject({ installed: true })
     await installCliLauncher(env, () => true)
     await expect(readFile(plan.target, 'utf8')).resolves.toBe(plan.shim)
-    await uninstallCliLauncher(env)
+    await uninstallCliLauncher(env, () => true)
     await expect(readFile(plan.target, 'utf8')).rejects.toMatchObject({ code: 'ENOENT' })
   })
 })
@@ -420,6 +420,7 @@ describe('buildWindowsPathCommand', () => {
     expect(args).not.toContain('-args')
     const script = args[args.length - 1]
     expect(script).toContain("$binDir = 'C:\\Users\\me\\AppData\\Roaming\\Open Science\\bin'")
+    expect(script).toContain("TrimEnd([char[]]'\\/') -ieq $normalizedBinDir")
     expect(script).toContain("[Environment]::SetEnvironmentVariable('Path'")
   })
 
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed the PATH ownership concern in `75aa8f84`.

- Installation now writes a validated receipt only after it actually appends the canonical bin directory to the persistent user PATH. If receipt creation fails, it attempts to roll the PATH mutation back.
- Uninstall removes at most one normalized matching entry only when both the managed shim and the exact receipt are present. Pre-existing entries, remaining user duplicates, unmanaged receipts, and legacy receipt-less installations are preserved.
- PATH cleanup and receipt deletion complete before the managed shim is removed, so failures leave the operation retryable.
- Added regression coverage for pre-existing PATH entries, owned cleanup, single-entry removal, receipt cleanup, and retry safety.

Validation after the change: focused tests passed (3 files, 61 tests), along with Node/Web type checks, lint, and changed-file formatting.

ewen-poch:
Addressed both PATH ownership findings in `4500d86d`.

- Receipt creation is now crash-recoverable: installation writes and `Flush(true)`-persists a versioned `.open-science-path-pending` journal before changing the registry, then commits ownership by atomically renaming it to `.open-science-path-receipt` after the PATH mutation.
- The journal records the exact before/after PATH snapshots. A later install can complete a pending transaction only when the current PATH exactly matches one of those snapshots; otherwise it fails closed.
- Uninstall treats `pending` and owned journals explicitly. It removes a normalized PATH entry only when there is exactly one match; multiple equivalent entries are ambiguous and leave PATH, journal, and shim intact.
- Receipt-less legacy installs and pre-existing user PATH entries remain unclaimed and preserved.

Validation after the change: focused tests passed (3 files, 62 tests), along with Node/Web type checks, lint, and changed-file formatting.

ewen-poch:
Addressed both journal boundary findings in `f7f2185a`.

- Windows uninstall now discovers and reconciles a valid pending/owned journal independently of the launcher shim. An app-owned PATH entry and its journal are therefore cleaned even if the shim was manually removed; a present shim is removed afterward.
- Any existing journal that is malformed, unmanaged, symlinked, hard-linked, or unstable now propagates an ownership error. Uninstall aborts before PATH or shim deletion instead of silently skipping journal cleanup.
- Receipt-less legacy shims retain the existing forward-only behavior: the shim is removed and PATH is preserved.
- Added regression coverage for missing-shim cleanup and unmanaged-journal fail-closed behavior.

Validation after the change: focused tests passed (3 files, 64 tests), along with Node/Web type checks, lint, and changed-file formatting.

ewen-poch:
Addressed the ownership-continuity finding in `42465006`.

- An owned journal now authorizes PATH mutation only when the current user PATH exactly matches its recorded `afterPath`; cleanup then restores the exact recorded `beforePath`.
- If no normalized app-bin entry remains, uninstall treats it as already removed and cleans up only the journal and shim.
- If an equivalent entry exists but the current PATH differs from `afterPath`—including a remove-and-recreate sequence or unrelated PATH edits—uninstall fails closed and preserves PATH, journal, and shim.
- Journal validation now also rejects snapshots whose `beforePath` already contained an equivalent bin entry.

Validation after the change: focused tests passed (3 files, 64 tests), along with Node/Web type checks, lint, and changed-file formatting.

ewen-poch:
Addressed the partial-journal finding in `12e25403`.

- Pending journal creation now writes to `.open-science-path-pending.tmp`, calls `Flush(true)`, disposes the stream, and atomically renames the complete file to `.open-science-path-pending` before any registry mutation.
- Failed writes remove the temporary file. A temporary file left by process termination is not treated as a journal and is removed before the next install attempt, so it cannot block reconciliation.
- The generated-command regression coverage now verifies the temporary path, failed-write cleanup, and the required `flush -> temp-to-pending rename -> PATH mutation -> pending-to-receipt rename` ordering.

The `afterPath` ownership-continuity recommendation is already implemented in `42465006`. In the owned branch, no normalized match means the entry was already removed; otherwise `current -ceq afterPath` is required before restoring `beforePath`. A remove-and-recreate sequence or any other mismatch throws while preserving PATH, journal, and shim, so no additional change was made for that duplicate finding.

Validation after the change: focused tests passed (3 files, 64 tests; launcher module 45 tests), along with Node/Web type checks, lint, and changed-file formatting.

ewen-poch:
Addressed the temporary-file ownership finding in `aa25d22f`.

- Each write now uses a unique same-directory `.open-science-path-pending.<guid>.tmp` path instead of a predictable shared name.
- Cleanup is gated by `$pendingTempCreated`, which is set only after `File.Open(..., CreateNew)` succeeds in the current invocation. The command therefore never enumerates or deletes a pre-existing temporary file.
- A crash-left unique temporary file is ignored by journal discovery and cannot block a later install. It is intentionally not removed by later runs because they cannot prove ownership; this persistence trade-off is now explicit in the PR description.
- Regression coverage verifies unique naming, cleanup ownership gating, and the required `flush -> temp-to-pending rename -> PATH mutation -> pending-to-receipt rename` ordering.

Validation after the change: focused tests passed (3 files, 64 tests; launcher module 45 tests), along with Node/Web type checks, lint, and changed-file formatting.

ewen-poch:
Rebased onto current `origin/main` at `8e1d5d0a` and force-pushed with an explicit lease.

The only conflict was in `src/main/cli-install/launcher.ts`, where #1898's launcher hardening overlapped the PATH ownership work. The resolution preserves #1898's exact versioned/legacy ownership headers, retained-handle validation, durable same-directory temporary publication, and atomic shim replacement, while keeping this PR's independent Windows PATH journal protocol. No architecture, shared state, persistence format, or user-interaction behavior was added by the rebase.

Post-rebase validation passed: focused tests 3 files / 68 tests (launcher module 49 tests), Node/Web type checks, lint, and changed-file formatting. The full local test suite was not run.

ewen-poch:
Addressed the pre-write validation finding in `d1062be0`.

- Windows install now discovers and validates pending/owned PATH journals immediately after ensuring the bin directory exists and before any launcher creation or replacement.
- An unmanaged journal fails closed before touching the shim. Simultaneous pending and owned journals are rejected as ambiguous at the same boundary.
- The later duplicate validation block was removed; the existing PATH reconciliation protocol remains unchanged.
- Added regression coverage for both unmanaged and ambiguous journals with an absent launcher, asserting that failed installation leaves no shim behind.

Validation after the change: focused tests passed (3 files, 70 tests; launcher module 51 tests), along with Node/Web type checks, lint, and changed-file formatting.

PR #1896, opened 2026-08-29
fix(projects): report pending deletion cleanup

ewen-poch:
## Problem

Project deletion soft-deletes Project metadata before several fallible cleanup tails. When one of those tails failed, the command rejected even though the Project was already absent from ordinary queries, so Home and Archived Settings retained stale Project and Session state and offered a misleading deletion retry. Background recovery could later finish the durable intent, but only the foreground command adapter published `project:deleted`, leaving Renderer caches stale until reload.

## Proposed change

- Add an explicit Project deletion command outcome: `deleted` or `cleanup-pending`.
- Keep failures before the soft-delete commit as rejections.
- Return `cleanup-pending` after the commit while retaining the existing deletion intent for background retry.
- Make `ProjectDeletionCoordinator` the single publisher of terminal `project:deleted` events for foreground and recovery completion.
- Remove the Project and its Sessions from the initiating Renderer for both committed outcomes; show a translated non-blocking warning while cleanup remains pending.

```mermaid
flowchart LR
  A[Delete command] --> B{Soft-delete committed?}
  B -- No, operation fails --> C[Reject and keep Project visible]
  B -- Yes, tail completes --> D[deleted]
  B -- Yes, tail fails --> E[cleanup-pending]
  E --> F[Background recovery]
  F --> D
  D --> G[Coordinator publishes project:deleted]
```

## Scope and non-goals

- No Prisma schema, migration, persisted field, or persisted enum changes.
- No historical data migration or compatibility path is required; existing deletion intents and Session tombstones remain the retry authority.
- The new status values are transient application-command/Renderer contract values only.
- No change to deletion ordering, retry delay, or legacy tombstone adoption.
- Durable event delivery across a process crash remains out of scope; this keeps the existing in-process lifecycle-event model.

Red regression evidence on the unmodified implementation:

- post-commit cleanup test rejected with `side chat profile busy` instead of resolving `cleanup-pending`;
- background recovery completed with zero `project:deleted` publications;
- the command adapter returned `undefined` instead of the owner outcome;
- Renderer tests had no cleanup-pending status notice.

Final evidence after rebasing onto the latest `main`:

- affected Project deletion modules: 8 files, 126 tests passed;
- locale owner and i18n catalog guards: 2 files, 750 tests passed;
- PR Gate: both full macOS module shards, coverage aggregation, static checks, Windows core, Windows E2E, macOS build/E2E, CodeQL, PR policy, and AI review passed;
- the latest `main` introduced one unrelated French high-punctuation catalog violation; the existing guard reproduced it in two macOS CI runs, and the one-character catalog correction is isolated in `fix(i18n): keep French high punctuation attached`.

Per the requested workflow, no full local `npm test` run was performed; complete portable coverage came from PR Gate.

## Review focus

- Confirm only failures after the Project soft-delete commit become `cleanup-pending`; pre-commit failures still reject.
- Confirm pending outcomes leave recovery non-sticky so the durable intent is retried.
- Confirm `project:deleted` is emitted once from the coordinator on foreground or background terminal completion and never duplicated by the command adapter.
- Confirm the warning copy and seven locale entries accurately describe a committed deletion rather than asking the user to retry.

### DIFF
--- a/src/main/compute/compute-service.architecture.test.ts
+++ b/src/main/compute/compute-service.architecture.test.ts
@@ -259,6 +259,11 @@ describe('Compute service architecture', () => {
       'await projectDeletionCoordinator.recoverPendingDeletions()',
       backgroundSessionRecovery
     )
+    const committedDeletionWake = source.indexOf(
+      "event.payload.status === 'cleanup-pending'",
+      backgroundProjectRecovery
+    )
+    const wakeCall = source.indexOf('projectDeletionRecovery.wake()', committedDeletionWake)
 
     expect(projectBarriers).toBeGreaterThan(-1)
     expect(jobBarriers).toBeGreaterThan(projectBarriers)
@@ -268,6 +273,8 @@ describe('Compute service architecture', () => {
     expect(backgroundOrphanRecovery).toBeGreaterThan(backgroundRecovery)
     expect(backgroundSessionRecovery).toBeGreaterThan(backgroundOrphanRecovery)
     expect(backgroundProjectRecovery).toBeGreaterThan(backgroundSessionRecovery)
+    expect(committedDeletionWake).toBeGreaterThan(backgroundProjectRecovery)
+    expect(wakeCall).toBeGreaterThan(committedDeletionWake)
   })
 
   it('gives Compute Job shutdown the transport cancellation budget', () => {
--- a/src/main/data-content-application-commands.test.ts
+++ b/src/main/data-content-application-commands.test.ts
@@ -800,9 +800,7 @@ describe('Data and content application commands', () => {
     expect(deps.sessions.editDetails).toHaveBeenCalledWith(editDetailsRequest)
     expect(deps.withDataRootWrite).toHaveBeenCalledTimes(6)
     expect(deps.events.publish).toHaveBeenCalledWith('project:updated', deps.project)
-    expect(deps.events.publish).not.toHaveBeenCalledWith('project:deleted', {
-      projectId: 'project-1'
-    })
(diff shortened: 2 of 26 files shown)


ewen-poch:
Addressed the review finding in f72f4c2f6: pending cleanup notices now derive from a transient Renderer Set keyed by Project ID. A terminal project:deleted lifecycle event already projects through removeProject(), which now clears the matching pending ID; the notice disappears after background recovery completes and remains visible while any other cleanup is still pending. Added stable red/green regression coverage for the store, Home, and Archived surfaces. No persisted fields or schema changes.

ewen-poch:
Addressed both P1 review findings in b8fb55b59. Project deletion now publishes a staged project:deleted event at the metadata commit point (cleanup-pending) and at terminal cleanup (deleted), so every renderer immediately evicts the Project/Sessions and clears its transient pending marker only at terminal recovery. The main-process event projection explicitly wakes the serialized recovery loop, including after its startup scan has completed; active wakes coalesce and failed recovery retains the existing retry delay. Added red/green coordinator, cross-window lifecycle, recovery-loop, event-order, and production-wiring coverage. This adds one required transient lifecycle-event status field by reusing the existing outcome union; there are no persisted fields, schema changes, migrations, or historical-data compatibility requirements.

ewen-poch:
Addressed the event/RPC ordering race in 95bb9cfcf. The Project Store now records a transient generation for each in-flight local deletion and the newest lifecycle status observed during that request. If terminal deleted arrives before the RPC resolves, the later cleanup-pending response remains the caller's outcome but cannot overwrite the authoritative terminal Store projection. Request tracking is removed when the RPC settles. Added a stable red/green public-store regression test for the exact ordering. No persistence or schema changes.

PR #1899, opened 2026-08-29
fix(session-persistence): scope deletion recovery and expose damaged files

ewen-poch:
## Problem

Deleting one Session entered a global persistence barrier, reread every Session JSON file, and blocked saves in unrelated Projects. When damaged Session authority was quarantined, the renderer reduced Main's `projectId` and `fileName` diagnostics to a count, leaving users without a way to identify or locate the affected files.

## Proposed change

- Reconcile an ordinary Session deletion inside the owning Project scope when cached metadata and the Project scan are complete.
- Preserve the existing full-catalog/global-barrier path as a fail-closed fallback for incomplete scans, incomplete metadata, duplicate identities, or reconciliation failures.
- Scope managed-file reconciliation and cached metadata replacement to the target Project.
- Preserve damaged Session file identities in the renderer, show them in a details dialog, and provide an Electron-only action that opens the managed Project Session folder.
- Add all seven required renderer translations and pin the new Electron-only renderer contract.

## Scope and non-goals

- Issue 5 (whole-Session renderer save ownership) is intentionally unchanged.
- No export workflow is added.
- Opening the recovery folder is Electron-only; local and remote Web surfaces remain unavailable.
- No database schema, on-disk Session format, persisted field, enum value, or durable state is added.
- The renderer recovery view-model changes from a count to `{ projectId, fileName }[]`; it is derived at load time and is not persisted.
- Historical compatibility: no migration is required. Existing Session JSON and existing `.invalid-*` quarantine files remain valid and are surfaced through their existing diagnostics.

All checks below ran from the final commit state after the last material edit.

- Unrelated-Project saves remain unblocked during deletion reconciliation; incomplete metadata still uses the global fallback -> focused coordinator regression tests -> passed.
- Project-scoped file reconciliation preserves another Project's indexed files -> managed-file repository regression -> passed.
- Damaged file identities reach the recovery dialog and the Electron IPC opens only the validated managed Project folder -> renderer model/component, repository, IPC, preload, and contract tests -> passed.
- Final Test Impact Set (18 owner/adapter/consumer/architecture/i18n test files) -> 1338 passed, 1 existing skip.
- `npm run typecheck` -> passed (Node and Web).
- `npm run lint` -> passed.
- `npm run check:web-api-map` -> passed.
- `git diff --check origin/main...HEAD` -> passed.

Full `npm test` was intentionally not run; PR Gate is authoritative for the complete portable and platform lanes. Local dependencies were installed with lifecycle scripts skipped because this machine lacks the native C++ toolchain; Prisma generation, dependency patches, and Electron installation were then run separately. The focused tests exercise SQLite/Electron-backed modules successfully, but canonical Node 22 clean-install behavior remains covered by CI (local Node is v24).

## Review focus

- The completeness checks around the Project-scoped fast path and its transition to the global fallback.
- Project scoping in managed-file cleanup and metadata replacement.
- The Electron-only capability declaration and path-segment validation for opening recovery folders.


ewen-poch:
Addressed the incomplete-session key delimiter finding. Project-scoped reconciliation now matches the existing colon-delimited session key format, so a removed Session no longer leaves its Project permanently marked with an incomplete file index. Added regression coverage through the public Project Files repository boundary; the test failed twice before the fix with isIndexComplete remaining false and passes after the fix. Validation: Project Files repository tests (40 passed, 1 skipped), typecheck, and lint.

ewen-poch:
Addressed the recovery-folder open failure finding. The Renderer now awaits the existing Promise<void> callback, catches rejection, clears any prior error before retrying, and shows the existing localized 'Could not open that folder.' message inline for the affected Project. The IPC contract is unchanged. Added public component-boundary regression coverage; the refined test failed twice before the fix and passes afterward. Validation: recovery alert and i18n tests (754 passed), typecheck, and lint.

PR #1900, opened 2026-08-29
fix(project-files): persist index completeness across restarts

ewen-poch:
## Problem

Project Files index completeness was held primarily in `ProjectFilesMutationOwner` memory. A failed Session file sync did persist `ManagedFileSessionSync.filesRevision = -1`, but queries ignored that durable retry marker. Catalog-wide reconciliation failure was not persisted at all.

After an application restart, a ready Session Summary projection can avoid rescanning Session JSON. The recreated repository therefore had empty in-memory failure state and could report `isIndexComplete: true` for an incomplete index, allowing zero or partial results to be treated as authoritative by Files Overview, downloads, and global search.

## Proposed change

- Derive project completeness from active `ManagedFileSessionSync.filesRevision < 0` rows.
- Persist catalog-wide reconciliation completeness in a constrained singleton `ManagedFileIndexState` row.
- Mark reconciliation incomplete before reconciliation work and complete only after it succeeds.
- Read completeness in the query owner from SQLite instead of a synchronous mutation-owner callback.
- Preserve the original operation error when the completeness marker cannot immediately be written, and retry the conservative global marker before the next observable Project Files read.
- Track fallback marker writes by Session so a successful retry clears only that Session without hiding unrelated pending failures.
- Normalize SQLite raw booleans across `true`, `1n`, and numeric `1`, and flush pending state before Artifact group projection reads.
- Make the singleton migration idempotently seed a missing row when adopting an already-generated table, while preserving an existing persisted `false` value.
- Serialize pending-marker flushes with synchronization of the same Session so a late `-1` write cannot overwrite a successful retry.
- Preserve Session deletion tombstones when flushing a deferred incomplete marker.
- Flush deferred Session markers only for the projects involved in the current read, so one
  project's locked marker cannot block an unrelated project's Files view.
- Pin migration `0020_project_files_index_state` in packaged migration-ledger certification and replay fixtures.
- Add public Repository regression coverage that recreates the Repository over the same database after both a Session indexing failure and a reconciliation failure.

## Scope and non-goals

- Data persistence changes: adds migration `0020_project_files_index_state` and the `ManagedFileIndexState` singleton table.
- Historical compatibility: existing databases are seeded with catalog reconciliation complete. Existing negative Session revisions remain authoritative and immediately make their projects incomplete. A catalog-wide failure from a pre-fix process cannot be reconstructed historically; the next reconciliation establishes the durable state.
- No new enum values.
- No Project Files IPC contract, renderer interaction, or user-visible behavior changes.
- No Session JSON or Session Summary data-model changes.
- No attempt to force a Session JSON rescan when its Summary projection is already ready.
- Includes an independent one-character French punctuation correction inherited from current
  `main`; without it, the full Module shard fails its existing locale invariant.

Before the production fix, the two new restart tests failed repeatedly at the public `ManagedFileIndexRepository.getOverview()` boundary with `expected false, received true`. Both pass after the fix. Eight AI-review follow-up regressions also failed before their fixes: successful retry remained incomplete, numeric SQLite true parsed as false, Artifact group reads failed to persist pending state before restart, an adopted table could omit the singleton row, a concurrent pending flush could overwrite a successful retry, a pending flush could revive a deleted Session group, packaged ledger certification stopped at migration 0019, and a Project A read attempted to flush Project B's locked marker. All now pass.

Final checks after the last material edit:

- `vitest` Project Files module: 66 passed, 1 skipped.
- Project Files index-state migration suite: 3 passed.
- Packaged database migration-ledger smoke: 12 passed.
- Module-impact validation: 9 passed.
- Generated database schema consistency check: passed.
- `vitest` Session Persistence coordinator: 149 passed.
- `vitest` Session deletion integration: 16 passed.
- `vitest` artifact finalization recovery integration: 14 passed.
- Targeted ESLint for the final touched implementation and migration files: passed.
- i18n resources guard: 743 passed; locale-owner suite: 7 passed.
- `git diff --check`: passed before commit; committed worktree is clean.

Additional affected checks run during implementation:

- Other affected migration suites: 23 passed.
- Migration service suite excluding one stale generated-client case: 60 passed, 1 skipped.
- Application database integration excluding two stale generated-client cases: 20 passed, 2 skipped.
- Full repository ESLint: passed.

Uncovered/local environment risks:

- This worktree intentionally reused the main checkout dependencies because its lockfile differs. That Prisma Client was generated from the old schema, so three legacy-classification tests see `ManagedFileIndexState` as an unknown table. Their remaining assertions pass when those cases are excluded; CI's fresh generated client is authoritative.
- Node TypeScript checking reaches one unrelated dependency mismatch: the reused `@agentclientprotocol/claude-agent-acp` package lacks the branch-expected `waitForMcpServers` export. No changed file reports a type error.
- Full `npm test` was not run locally as requested; PR CI owns the full matrix.

## Review focus

- Compatibility-first `true` seed for existing databases versus conservatively marking every upgraded catalog incomplete.
- State transitions around reconciliation failure/crash and successful retry.
- Completeness SQL for active negative Session revisions, including searched overviews and cross-project artifact search.
- Keeping the query owner read-only while removing completeness authority from process memory.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -357,6 +357,13 @@ model ManagedFileSessionSync {
   @@index([projectId, deletedAt, groupSortAtMs, sessionId])
 }
 
+// Singleton completeness marker for catalog-wide Project Files reconciliation. Per-Session retry
+// state remains represented by ManagedFileSessionSync.filesRevision < 0.
+model ManagedFileIndexState {
+  id                       String  @id
+  isReconciliationComplete Boolean @default(true)
+}
+
 // Narrow origin lifecycle retained after a live Session JSON is deleted. This is not a second
 // editable Session model; it exists only to preserve managed-file ownership and deletion state.
 model FileOriginSession {
--- a/prisma/sqlite-check-constraints.json
+++ b/prisma/sqlite-check-constraints.json
@@ -196,6 +196,11 @@
       "expression": "\"source\" IN ('artifact', 'upload')"
     },
     {
+      "tableName": "ManagedFileIndexState",
+      "name": "ManagedFileIndexState_identity_check",
+      "expression": "\"id\" = 'project-files-index' AND \"isReconciliationComplete\" IN (false, true)"
+    },
+    {
       "tableName": "UploadVersion",
       "name": "UploadVersion_state_check",
       "expression": "\"state\" IN ('staging', 'ready')"
--- a/scripts/ci/module-impact.json
+++ b/scripts/ci/module-impact.json
@@ -626,6 +626,7 @@
     },
     "project_files_repository": {
       "ownerPaths": [
+        "src/main/project-files/index-state.ts",
         "src/main/project-files/mutation-owner.ts",
         "src/main/project-files/mutation-projection.ts",
         "src/main/project-files/query-owner.ts",
--- a/scripts/database-migration-ledger-smoke.mjs
+++ b/scripts/database-migration-ledger-smoke.mjs
@@ -80,6 +80,10 @@ const EXPECTED_MIGRATION_LEDGER = [
   {
     id: '0019_session_usage_attribution',
     checksum: 'c505fe7e55e8428d29e8506ad305c04fd46d8def53385a1f7839fba21047ab9d'
+  },
+  {
+    id: '0020_project_files_index_state',
+    checksum: '3db4a30baa2b4614e29205ed56b2682c0579f1a91f5a2d293d21925253ed7fd1'
   }
 ]
 const LEGACY_PROJECT_ID = 'package-smoke-legacy-project'
--- a/scripts/database-migration-ledger-smoke.test.ts
+++ b/scripts/database-migration-ledger-smoke.test.ts
@@ -20,7 +20,7 @@ import { PrismaClient } from '@prisma/client'
 describe('packaged database migration ledger smoke', () => {
   it('pins every packaged application migration identity and checksum', () => {
     expect(MIGRATION_MANIFEST.at(-1)?.checksum).toBe(
-      'c505fe7e55e8428d29e8506ad305c04fd46d8def53385a1f7839fba21047ab9d'
+      '3db4a30baa2b4614e29205ed56b2682c0579f1a91f5a2d293d21925253ed7fd1'
     )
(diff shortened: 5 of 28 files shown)


ewen-poch:
Addressed in 13896b18d. Pending fallback writes are now tracked per Session, successful retries clear only their own entry, and reconciliation pending state remains independent. Added a public Repository regression test that failed before the fix with expected true / received false and now passes.

ewen-poch:
Addressed both findings in 798495d79. Completeness parsing now shares a true/1n/1 normalizer, and listArtifactGroups flushes pending state before reading the projection. Added focused regressions for numeric SQLite true and restart persistence through the Artifact-group read path; both failed before the fix and now pass. Full Project Files suite: 59 passed, 1 skipped.

ewen-poch:
Addressed both findings in 46fdeb8ea. Migration 0020 now verifies the singleton row and idempotently seeds it when an existing generated table is adopted, without requiring or overwriting the persisted completeness value. Pending Session marker flushes now share a per-Session serialization lane with sync retries, preventing a late -1 upsert from overwriting a successful revision. Added focused regressions for both cases; each failed before the fix and now passes. Final Project Files module: 64 passed, 1 skipped; migration suite: 3 passed.

ewen-poch:
Addressed the latest-head deletion finding and the prior packaged-ledger finding in 4898b9726. Deferred Session marker updates no longer clear deletedAt/deleteOperationId, so a tombstoned Session cannot be reactivated by a later Project Files read. Packaged migration certification now pins 0020 and replay fixtures remove/reapply it with dependent suffixes. Both regressions failed before the fixes (revived group count 1; ledger checksum still 0019) and now pass. Project Files module: 65 passed, 1 skipped; packaged ledger smoke: 12 passed.

ewen-poch:
Addressed the latest cross-project flush finding in 50fc8ca7c. The read hook now receives the projects participating in the query and flushes only pending Session markers owned by those projects; the global reconciliation marker remains global. A public Repository regression creates a pending marker for Project B, makes B's marker write fail, and verifies Project A's overview still resolves. It failed before the fix with 'unrelated project is locked' and now passes. After merging current main, I also isolated and corrected its existing French pre-colon spacing failure in 8a8a4cffd; the exact locale test and the 743-case i18n guard pass. Final Project Files module: 66 passed, 1 skipped.

PR #1902, opened 2026-08-29
feat(notebook): persist immutable file generations

ewen-poch:
## Summary

- replace silent working-file observation failure with explicit, persisted evidence coverage;
- freeze created and modified Notebook files as immutable, SHA-256-addressed generations without overwriting an earlier generation;
- persist per-run created/modified/deleted relations and a checksummed evidence summary for later provenance-graph work;
- analyze strongly identified data-science output layouts as advisory logical outputs, including single files, companion file sets, and directory-backed datasets;
- keep every inferred relation and logical-output group advisory: this PR does not use `dependency-analysis.json` or filename heuristics as execution authority.

This is PR 1 of the result-reproducibility work. It establishes production-time file and output-shape evidence only; it does not add a reproducibility UI, recipe planner, format validator, or rerun/reproduction executor.

## Architecture

```mermaid
flowchart LR
  Run[Notebook Run identity] --> Observe[Working-file observation]
  Observe --> Watch[Recursive watcher candidates]
  Observe --> Snapshot[Bounded before/after fallback]
  Watch --> Relations[created / modified / deleted relations]
  Snapshot --> Relations
  Relations --> Analyze[Advisory scientific-output analysis]
  Analyze --> Single[single-file]
  Analyze --> Set[file-set]
  Analyze --> Tree[directory-tree]
  Relations --> Worker[Descriptor-bound evidence worker]
  Single --> Worker
  Set --> Worker
  Tree --> Worker
  Worker --> Freeze[Freeze final file generations]
  Freeze --> Stage[Run-owned staging directory]
  Stage --> Publish[Atomic evidence + blobs publish]
  Publish --> Sidecar[evidence.json schema v1]
  Sidecar --> Summary[Checksummed run.json summary]
  Summary --> Terminal[Run terminalization]
  Summary --> Restart[Startup reconciliation]
  Restart --> Keep[Keep referenced evidence]
  Restart --> Sweep[Remove staging + unpublished Runs]
```

`startWorkingFileObservation` remains the narrow facade. `NotebookExecutionOwner` supplies the app-owned `runId` to Python, R, REPL, and shell paths. Kernel and shell terminalization carry the observation result into the durable Run.

The observer watches only the app-owned `data/` and `handoff/` roots. Watcher events are candidate evidence; a bounded snapshot is the fallback when the watcher is unavailable or yields no usable candidates. Concurrent observations of the same root fail closed rather than cross-attribute a file to the wrong Run.

`analyzeScientificOutputs` is a pure advisory projection over observed relation paths. It never opens, parses, deserializes, checkpoints, or mutates user outputs. Strong layout signatures group related members; an unrecognized path remains an independent `single-file` output. A run-scoped hash supplies each logical output ID, so identical paths in later Runs remain distinct output generations.

### Data-science coverage

The filesystem capture itself is API-independent: local regular files are observed whether Python/R produced them through pandas, NumPy, PyArrow, xarray, base R, readr, data.table, arrow, sf, ggplot2, or another library. The analysis adds structural coverage for the output shapes that a per-file list cannot represent:

| Shape | Representative Python/R outputs | Persisted interpretation |
| --- | --- | --- |
| Tabular/text/single binary | CSV, TSV, JSON, Excel, Parquet, Feather, FST | `single-file`, format hint, immutable member generation |
| Scientific arrays/containers | NPY/NPZ, HDF5, netCDF, MAT, RDS/RData/qs, H5AD/H5MU/Loom | `single-file`; runtime-dependent serialization is disclosed where applicable |
| Partitioned datasets | PyArrow/Dask/Polars and `arrow::write_dataset`, including Hive partitions and `_metadata`/`_SUCCESS` | one `directory-tree` output containing all observed members |
| Chunked stores | Zarr/xarray-to-Zarr | one `.zarr` `directory-tree` output |
| Databases | SQLite/DuckDB plus WAL/SHM/journal companions | one `file-set`; transaction/checkpoint state is explicitly unverified |
| Geospatial outputs | Shapefile companions, GeoTIFF sidecars, GeoPackage and common geospatial files | one companion `file-set` where strongly identified, otherwise one file |
| Models/checkpoints | TensorFlow SavedModel/checkpoints, sharded or unsharded model directories, ONNX/PyTorch/Keras files | directory/file-set grouping plus runtime-dependency risk |
| Reports | HTML/htmlwidgets assets, images, PDF | HTML plus its `_files` tree is one `file-set`; ordinary reports remain one file |
| Single-cell matrices | 10x Matrix Market plus barcode/feature files | one `file-set` even when only a strongly named subset was observed |
| Statistical packages | SPSS, Stata, and SAS interchange files | `single-file` statistical-data hint |

The analyzer does not claim format validity, schema equality, complete companion membership, database transaction consistency, or runtime portability. Those remain explicit risk codes for later readiness and isolated validation phases.

### Spike findings

The filesystem spike exercised reads, creates, same-path rewrites, deletion, atomic replacement, create-then-delete, external writes, symlinks, partitioned directories, Zarr, database companions, and runtime-bound Python/R serialization. Recursive `fs.watch` can asynchronously fail on a resource-constrained host; final snapshots still find final created/modified/deleted state. Reads, transient create-then-delete files, delayed/background writes, remote sinks, and writes outside managed roots are not recoverable as complete evidence.

Those findings determine the shipped semantics:

- watcher failure adds `watcher-unavailable` and falls back instead of returning a misleading empty list;
- file reads, initial mutable-file generations, transient/delayed writes, remote/external outputs, and complete writer attribution remain explicitly unavailable;
- relations use `authority: advisory`, and scientific-output classifications use `classificationAuthority: path-heuristic`;
- PR 1 never claims authoritative lineage or exact reproducibility readiness.

This follows the Entity/Activity separation used by [W3C PROV-O](https://www.w3.org/TR/prov-o/#description-starting-point-terms) and the plan/run distinction in [Workflow Run RO-Crate](https://www.researchobject.org/workflow-run-crate/profiles/workflow_run_crate/), while keeping the implementation a narrow typed JSON contract rather than introducing RDF or a workflow engine.

## Persistence

No Prisma schema or migration changes.

New additive data under each Notebook Session:

- `run.json`
  - optional `workingFiles[].generationId`, `checksum`, and `change`;
  - optional `fileEvidence` summary with schema version, state, coverage, reason codes, relation/generation counts, `scientificOutputCount`, `scientificOutputAnalysis`, sidecar `storageKey`, and checksum.
- `file-evidence/run-<runId>/evidence.json`
  - this sidecar is introduced by this PR and remains its first `schemaVersion: 1`; there is no artificial v2 development-time upgrade;
  - immutable-generation relations for `created`, `modified`, and `deleted` paths;
  - explicit relative-path portability and advisory authority;
  - `scientificOutputs[]` with run-scoped output ID, storage shape, optional advisory `formatHint`, member relation paths, classification authority, and reproducibility risk codes;
  - generation storage keys and capture metadata.
- `file-evidence/run-<runId>/sha256-<checksum>`
  - immutable bytes for successfully frozen created/modified generations;
  - identical bytes within one Run deduplicate by checksum; different Runs retain independent ownership.

The content copy opens the source with non-blocking/no-follow flags, immediately validates the descriptor as the observed regular file, streams once into the Run staging directory, hashes while copying, caps the stream at the observed size, honors cancellation, checks source metadata before and after copying, and enforces the existing per-file/per-run limits plus disk reserve. The helper has a 10-minute hard timeout. Each staged blob is durably flushed before the sidecar and blobs are atomically published as one run-owned directory. A sidecar/publication failure removes only this invocation's staging directory and clears unpublished generation references without failing an otherwise successful scientific computation. A final-name collision fails closed and preserves existing evidence.

Evidence directory creation canonicalizes the app-owned Session root and rejects symbolic-link or non-directory roots. Persistence runs in a packaged Node helper whose `cwd` is bound by the OS to the validated evidence directory object and verified by device/inode before mutation. Staging, no-follow blob creation, fsync, publication rename, targeted cleanup, and startup reconciliation use relative paths inside that bound capability. The main process revalidates the visible root identity before publishing references to `run.json`.

On Session startup, the same bound worker removes every `staging-*` entry and every `run-*` directory not referenced by durable `run.json`; referenced evidence remains immutable.

PR 1 storage is Notebook-Session-owned. It survives process restart but is not yet pinned to an Artifact Version. A later Artifact capture/recipe phase must copy or pin referenced generations into Artifact-owned retention before it may declare a recipe ready.

## Input and output behavior

- Existing registered Upload/Artifact input Versions remain unchanged on `run.inputFiles`.
- PR 1 does not claim which mutable workspace files were read and does not freeze their initial bytes.
- Created and modified final files inside managed roots are frozen as new generations.
- Deleted files are represented as relations without fabricated bytes.
- Multi-file outputs reference observed relation members; missing/unchanged historical companions are not fabricated and keep the output consistency risk unresolved.
- Raw external absolute paths and remote sinks are not watched or rebound.
- Rewriting the same logical path in a later Run creates distinct file and logical-output generations. Earlier evidence remains readable and unchanged.

## User interaction

None in this PR. There is no renderer entry point and no user-visible string or i18n change. Users cannot yet start a reproducibility check or browse these internal evidence records.

A later provenance/readiness UI can present one logical output node with expandable member files and show why multi-file consistency, database state, or runtime portability is still unverified. Ordinary Notebook execution behavior is preserved: failure to capture evidence does not turn a successful computation into a failed Run.

## Historical compatibility

- Historical `run.json` documents without `fileEvidence` remain readable without migration or backfill.
- `evidence.json` is new and unpublished in this PR, so `scientificOutputs` is defined directly in schema v1; no unreleased sidecar variant is retained or migrated.
- Missing evidence is not fabricated as complete or inferred from historical `dependency-analysis.json`.
- Injected/direct executors without a `runId` retain the historical `workingFiles` result and receive explicit unavailable evidence.
- Existing results and paths are never overwritten by generation capture.
- Unsafe replaced evidence paths are rejected rather than adopted.

## New states and enum values

- working-file change: `created | modified` (deletion lives in the sidecar relation because no current file remains);
- evidence state/coverage: `complete | partial | unavailable`;
- scientific output storage shape: `single-file | file-set | directory-tree`;
- classification authority: `path-heuristic`;
- scientific-output risks: `format-validity-not-verified`, `multi-file-consistency-not-verified`, `database-state-not-verified`, `runtime-dependent-serialization`;
- new global gaps: `delayed-writes-not-observed`, `remote-outputs-not-observed`;
- existing reason codes continue covering unobserved reads/initial generations/external paths/transient files/writer attribution, watcher failure, observer conflict/limits/failure, generation budget/freeze failure, persistence failure, and missing Run identity.

`formatHint` is deliberately not an enum; it is advisory vocabulary that can expand without a persisted state migration. No new execution lifecycle state is introduced. PR 1 emits only `partial` or `unavailable`; `complete` is reserved for a future enforced capture backend.

## Validation

The final head is rebased onto `origin/main@00de1982`.

| Behavior / impact | Command | Result |
| --- | --- | --- |
| scientific-output classification for Python/R single files, partitioned Arrow/text datasets, Zarr, SQLite/DuckDB companions, Shapefile, GeoTIFF, 10x matrices, SavedModel/checkpoints/model directories, HTML assets, deterministic run-scoped IDs, and conservative incomplete groups | `npm test -- src/main/notebook/scientific-output-analysis.test.ts` | 20 passed |
| final file relations, immutable generations, scientific-output sidecar persistence, same-path rewrite, watcher/fallback/conflict, cancellation, storage limits, symlink/root replacement, FIFO replacement, collision preservation, and orphan reconciliation | included `working-file-observer.test.ts` in focused owner run | passed |
| final owner/persistence/Artifact contract set | `npm test -- src/main/notebook/scientific-output-analysis.test.ts src/main/notebook/working-file-observer.test.ts src/main/notebook/run-terminalization.test.ts src/main/notebook/repository.test.ts src/main/artifacts/provenance-write-contract.test.ts` | 103 passed |
| Notebook owner and consumers | focused eight-file Notebook run including kernel executor, session aggregate, and read model | 186 passed, 1 platform-skipped |
| packaged evidence worker | `npm test -- build/packaging.test.ts` | 24 passed |
| Notebook workflow smoke | focused workflow + full-stack smoke | 4 passed, 3 environment-skipped |
| Artifact provenance consumer module | `npm run test:module -- artifact_provenance` | 1,418 passed; one unrelated 30-second ACP timing test passed on isolated rerun |
| main/Web types | `npm run typecheck:node`; `npm run typecheck:web` | passed |
| lint | `npm run lint` | passed |

Per the requested workflow, the complete local `npm test` suite was not run. PR CI is the authority for the exact-head full portable and platform-sensitive plan.

## Platform and uncovered risks

- Local filesystem integration ran on macOS. Portable path normalization is unit-covered; Windows/Linux watcher and filesystem behavior remain CI/platform validation concerns.
- Watcher candidates are not a complete process-tree filesystem tracer. PR 1 cannot prove reads, transient/delayed effects, remote/external writes, or detached-child attribution.
- Path/layout classification cannot prove bytes are a valid format, that all companion members were observed, that a database is transactionally consistent, or that runtime-bound serialization will load in another environment.
- Initial mutable workspace bytes are not frozen in PR 1. Such runs cannot become exact-ready until a strict prospective-capture policy exists.
- Notebook-Session evidence is not Artifact-pinned and must not be treated as retention-safe Artifact evidence.
- No UI, recipe, DAG slicing, environment restoration, isolated execution, or result comparison is included.


ewen-poch:
Implemented the review follow-ups in 7fbecdd7. File evidence now lives in an app-private project/session root, and recovery scans only durable receipts with exact directory identities; matching run-* or staging-* names without owned receipts are preserved. Receipt publication is atomic, the internal prepared/allocated/capturing/published phases cover allocation and rename crash boundaries, terminal Run commit retires the receipt, and reconciliation failures degrade without blocking Notebook session load. The evidence v1 payload now freezes the initial managed-root view and links modified/deleted relations to immutable previous generations through previousGenerationId. I also handled worker stdin EPIPE, preserved literal POSIX backslashes, marked R save(list=...) and save(envir=...) as dynamic namespace access, included notebook-file-evidence in data-root migration, and wired Project deletion cleanup through the durable deletion tail. Focused verification: 449 tests, Node/Web typecheck, changed-file ESLint, worker syntax check, and an independent worktree review.

ewen-poch:
Addressed the control-execution evidence gap in 6360aaa7. NotebookControlResult now carries fileEvidence from the REPL executor through executeControlExclusive, so terminalization persists the same evidence returned to the caller instead of recording it as unavailable. Added a regression test that verifies both the returned result and the repository-backed session state. Verified with the focused runtime-service test, Node typecheck, ESLint on the changed files, and git diff --check.

ewen-poch:
Addressed the storage-accounting gap in 47a0062f and rebased the branch onto the latest main (f7506550). notebook-file-evidence is now a first-class usage category in the shared contract, main usage scan, and Settings presentation, so its bytes contribute to both total usage and the existing non-runtime migration estimate. Added a contract invariant that every relocatable data directory has a usage category, plus storage total, migration estimate, renderer, color-token, and seven-locale coverage. No persisted schema or historical migration is required because StorageInfo is computed on demand. Post-rebase verification: 779 focused tests, Node/Web typechecks, ESLint on the changed files, and git diff --check.

ewen-poch:
Addressed the latest ownership and repeated-baseline findings in e66c3e34 after rebasing onto origin/main 4998af34.

Project evidence creation now publishes a root-external ownership receipt and a token-matched marker before any Project-private evidence is accepted. Project deletion verifies that proof, atomically renames the directory to a tokenized tombstone, persists an internal deleting phase, and safely resumes after rename gaps or partial recursive removal. Unknown, replaced, or unowned paths fail closed. The receipt deliberately does not persist inode identity, so copied data-root migrations are revalidated by token and marker.

Run generations remain distinct, but their bytes now use a Project-scoped SHA-256 CAS blob pool. Existing blobs are opened no-follow and checksum-verified before reuse, publication is exclusive and durable, and the worker rechecks live filesystem capacity before every new blob. A 10 GiB per-Project unique-blob budget degrades new evidence to partial without overwriting or evicting historical results. CAS work is globally serialized within the app, with a signal-aware 5-second queue deadline so evidence capture cannot indefinitely delay Notebook dispatch.

Persistence impact is limited to the internal Project ownership receipt, CAS blob layout, and deleting receipt phase. evidence.json remains v1; there is no database migration, shared status enum, or UI/i18n change. Historical run-local storage keys remain readable; unmarked development-only Project directories are intentionally not adopted.

Post-rebase verification: 8 focused files / 468 tests passed, Node typecheck passed, changed-file ESLint passed, worker syntax and git diff checks passed. An independent final review found no remaining high-value issues.

ewen-poch:
Addressed the final review findings in ca492997.

Session reconciliation now establishes Project ownership before creating the session evidence root, so startup cannot create an unowned Project directory. Project deletion keeps the portable token-based receipt and avoids persisted inode identity: after the atomic tombstone rename and internal deleting phase, deletion binds to the verified tombstone directory, removes all children while retaining the ownership marker, removes the marker last, then returns to the bound root and revalidates both root and tombstone identity before removing only the same empty directory. A missing marker is accepted only for an empty tombstone, which covers the final marker-to-rmdir crash gap; non-empty replacements, symlinks, and path swaps fail closed.

Failed or aborted initial capture now immediately invokes receipt-proven cleanup for its staging directory and receipt. CAS blobs are intentionally retained when their lack of references cannot be proven safely; they remain immutable, are never overwritten or evicted, and stay bounded by the existing 10 GiB per-Project unique-blob budget. This avoids adding a refcount or a new persistence relationship.

Persistence impact: no new fields beyond the already documented internal deleting phase, no persisted inode, no evidence.json change, no database migration, no shared enum, and no UI/i18n change. Data-root copy migration therefore remains token/marker based.

Final verification after the last material edit: 8 focused files / 472 tests passed, Node typecheck passed, changed-file ESLint passed, worker syntax and diff checks passed. A fresh independent Standards/Spec review pair found no remaining high-value issues; during review, the tombstone deletion was further bound to the verified cwd to close a path-replacement window. The branch contains the latest origin/main 51ac9fcd.

ewen-poch:
Reviewed this finding. I am not deleting CAS blobs from the current cleanup path because a content-addressed blob may already be reused by another active or committed generation, while the current model intentionally has no durable reference count or project-wide reference ledger. Treating absence from one failed receipt as proof of global non-reference could delete immutable historical evidence.

The leak is bounded by the existing 10 GiB per-Project unique-blob budget and cannot overwrite or evict prior results. A complete fix requires one of two explicit architecture changes: a durable per-generation/blob reference ownership journal, or a fail-closed project-wide mark-and-sweep that serializes every reference publication and cleanup and preserves all blobs whenever any receipt/sidecar cannot be validated. Both add a new persistence/data-relationship and automatic deletion policy, so I am pausing for maintainer confirmation rather than expanding the model implicitly.

ewen-poch:
Implemented the approved ownership model in `a99df613` after rebasing onto the latest `origin/main` (`d7d9a81b`).

Each published Run now owns its immutable generation bytes through hardlinks under its receipt-proven evidence directory, and `evidence.json.contentStorageKey` points to that Run-owned path. The Project SHA-256 pool is now an internal deduplication mechanism only. New content is copied and fsynced into receipt-owned staging, linked into the CAS, and linked into the Run before the sidecar can be published.

Cleanup and startup reconciliation first remove failed receipt-owned staging/final directories, then reclaim only validated CAS entries whose link count proves that the pool link is the sole remaining owner. The sweep validates every pool entry and candidate before deleting anything; unknown entries, symlinks, identity changes, or extra links fail closed or preserve the blob. All evidence creation, capture, publication, cleanup, reconciliation, and Project deletion operations share one in-process mutation queue, preventing reference publication from racing reclamation.

Persistence impact: Run evidence directories gain an internal `blobs/sha256-*` subtree and the Project keeps its existing internal CAS. `evidence.json` remains v1, with no new field or status. There is no database migration, shared enum, UI, or i18n change. Historical main-branch runs without evidence remain readable as incomplete. The unreleased development-only CAS-direct storage-key layout is intentionally not migrated; copied data-root migrations remain safe because the sidecar points to Run-owned bytes even when hardlinks become independent copies.

Verification: 50/50 owner tests passed after the final rebase; the broader evidence/provenance suite passed 234 tests with 1 skipped; storage/migration/packaging passed 158 tests; Node typecheck, full ESLint, worker syntax, and diff checks passed. A separate final worktree review found no remaining high-value issue.

ewen-poch:
Addressed the recovery-cleanup finding in `e464bbe4` and rebased onto the latest `origin/main` (`ac6133f8`).

Receipt-driven cleanup now requires both the recorded directory identity and the receipt token's Run ownership marker before removing allocated/capturing/published staging or final directories. The marker is checked as a regular, non-symlink, zero-byte file and the directory identity is revalidated before recursive removal. A missing or mismatched marker fails closed, preserving both the target directory and receipt for inspection/recovery. Published evidence verification now checks the same marker.

Persistence impact: the existing staging ownership marker is retained as an internal file in the published Run evidence directory instead of being removed immediately before publication. There is no new receipt field, schema version, database migration, shared status enum, data relationship, UI, or i18n change.

Added regression coverage for a staging-to-final rename followed by marker removal, proving reconciliation preserves the directory contents and receipt. Also verified that the marker remains after terminal Run completion. Post-rebase verification: 51/51 owner tests passed, Node typecheck passed, changed-test ESLint, worker syntax, and diff checks passed.

ewen-poch:
Applied the independent follow-up review in `9c8d5d01`.

Recovery cleanup now distinguishes a genuinely absent receipt-owned target from an existing unsafe replacement. If a staging or final path is a symlink, regular file, or other non-directory entry, cleanup fails before retiring the receipt. This applies to immediate failure cleanup, prepared allocation recovery, allocated/capturing staging cleanup, and renamed final cleanup. Legitimately absent targets remain idempotent.

Added regression tests for a regular-file replacement at the staging path and a symlink replacement at the final path; both prove that the replacement data and ownership receipt are preserved. The owner suite now passes 53/53, with Node typecheck, changed-test ESLint, worker syntax, and diff checks also passing. A fresh independent read-only review found no remaining high-value issue.

There is no additional persistence, schema, status, data-relationship, UI, or i18n impact beyond the Run ownership-marker retention documented in the previous comment. The branch is based on current `origin/main` (`ac6133f8`).

ewen-poch:
Applied a high-value finding from a fresh independent Standards/Spec worktree audit in `9c9350c6`.

`begin()` could fail on a pre-existing staging directory before acquiring its identity, then pass an undefined identity into failure cleanup. The generic cleanup helper interpreted that as permission to recursively remove the colliding directory. Cleanup now requires a valid expected identity before removing any owned directory, so an allocation collision fails closed and retains both the pre-existing path and the durable recovery receipt.

Added a regression test that pre-creates the exact staging name with user data, triggers allocation, and verifies that the data and receipt remain intact. Verification: 54/54 owner tests passed, Node typecheck passed, changed-test ESLint, worker syntax, and diff checks passed. The independent reviewer rechecked the patch and found no remaining issue in this fix.

No persistence schema, status, data relationship, UI, or i18n behavior changed.

ewen-poch:
Addressed the independent cleanup-race audit in cce45172 after rebasing onto the latest origin/main (c90f2780).

Receipt-owned Run directories and sole-link CAS blobs are now moved to recoverable, unpredictable UUID quarantine names before destructive cleanup. Run recovery accepts only the receipt ownership token, staging/final kind, and a valid UUID; deletion stays bound to the verified quarantine directory, removes the ownership marker last, and revalidates the directory identity before removing the empty top level. CAS recovery accepts only hash-plus-UUID names and revalidates regular-file type, link count, fingerprint, and content hash. A Run source/quarantine collision, invalid or multiple Run quarantines, or multiple CAS entries for one hash now fails closed before deleting anything.

The random destination is generated immediately before the single rename and is not exposed by a preceding existence check, closing the deterministic check/use overwrite window reported by the reviewer without adding a native dependency. Crash recovery discovers the quarantine from existing ownership evidence instead of persisting a new lifecycle journal.

Persistence impact: crash-surviving quarantine names are internal filesystem lifecycle state only. There is no new receipt field or phase, evidence.json remains v1, and there is no database migration, shared status enum, or new data relationship. The previously approved Settings “Notebook evidence” category remains a storage-accounting surface only; no result-reproduction workflow or control was added.

Verification after the rebase: 11 affected test files passed (810 passed, 1 skipped), including 59/59 working-file owner tests; Node typecheck, full ESLint, worker syntax, and diff checks passed. A fresh independent read-only review specifically rechecked the former P1 and found no remaining high-value correctness or safety issue.

ewen-poch:
Investigated and reran the failed PR Gate jobs. The only underlying failure was a macOS workspace E2E flake in “keeps Memory reversible while the replacement session awaits history replay”: the first attempt timed out after Auto-review was checked in the renderer but a concurrent Session revision update produced the existing “Conversation storage needs attention” conflict, leaving the durable value false. Playwright retry passed, but --fail-on-flaky-tests correctly failed the lane.

The uploaded trace showed the first Memory persistence becoming visible immediately before the Auto-review click while Agent completion was still settling. This path is unrelated to the Notebook file-evidence changes, and no Notebook evidence/storage code was implicated. I reran only the failed workflow jobs against the same head cce45172; macOS build, web build, functional journeys, workspace journeys, accessibility, visual regression, and the final deterministic PR Gate all passed. No code or persistence/UI contract change was made for this transient failure.

PR #1908, opened 2026-08-30
fix(uploads): validate session finalization requests

ewen-poch:
## Problem

`uploads:finalize-session` accepted unvalidated renderer payloads. A remote or abnormal renderer could submit malformed, oversized, or duplicate attachment batches, which the upload repository processed concurrently. Compatibility calls that omitted `projectId` also finalized into the default Project without entering that Project and Session's mutation lane.

## Proposed change

- add a strict runtime contract for finalize requests and results;
- cap finalize batches at the existing 10-attachment composer limit;
- reject duplicate attachment IDs and exact duplicate path strings before the upload owner runs;
- route Electron and Web finalization through the existing runtime-validated application command;
- leave the Electron channel exclusively owned by that application-command adapter instead of also registering it in legacy upload IPC;
- resolve the default Project once and use the same ID for the Session mutation lane and repository call;
- isolate source Electron E2E data roots so parallel workers cannot clean another test's pending uploads.

## Scope and non-goals

- no new limiter, queue, or concurrency subsystem;
- no UI or normal interaction change;
- no new data fields, enum values, database schema, persistence format, or migration;
- `projectId` remains optional for compatibility, and omitted or blank values retain the existing default-Project behavior;
- historical Session upload recovery continues through its existing internal repository path and is not migrated by this command contract;
- the E2E-only storage override reuses the existing disposable-root mechanism and does not change production storage selection.

The request and mutation-lane regression tests were added before the fix. On the unmodified implementation, the same command failed deterministically twice with four invalid batches resolving successfully and the default mutation scope remaining empty. The IPC ownership regression also failed before its fix because the legacy adapter still registered `uploads:finalize-session`.

Windows PR CI then exposed a test-harness race twice: parallel source Electron workers shared `OpenScience-DEV`, so another worker's startup cleanup could remove `research-notes.md` from `.pending`; the failed attempt reported `ENOENT` and its isolated retry passed. The new source-E2E isolation expectation failed before the harness fix because `OPEN_SCIENCE_E2E_STORAGE_ROOT` was `undefined`, then passed after the fix.

- invalid finalize requests are rejected before owner execution -> `src/main/data-content-application-commands.test.ts` -> passed (4 cases);
- omitted `projectId` uses the default Project mutation lane and repository scope -> `src/main/uploads/command-owner.test.ts` -> passed;
- legacy upload IPC does not duplicate the runtime-validated Electron channel -> `src/main/uploads/ipc.test.ts` -> passed;
- Electron/Application Command inventories and Renderer catalog remain consistent -> composition, adapter, catalog, and wiring tests passed;
- final focused behavior set -> 8 files, 92 tests passed;
- source E2E launch isolation -> 7 tests passed;
- isolated workspace file upload and preview -> 2 E2E tests passed;
- `npm run build` -> passed;
- `npm run check:web-api-map` -> passed;
- changed-file ESLint, Prettier, and `git diff --check` -> passed;
- full `npm run lint -- --no-cache` -> passed before the final E2E-only edit; the affected files passed changed-file ESLint afterward;
- `npm run test:affected:explain -- --base origin/main --head HEAD` selected the full CI plan because `src/main/data-content-application-commands.ts` is not mapped to a module owner. Per maintainer direction, the full portable and platform suites remain authoritative in PR CI.

## Review focus

- the strict `UploadedAttachment` wire shape and duplicate-ID/path rules;
- the Renderer catalog's runtime-validated routing marker and exclusive Electron channel ownership;
- the single default-Project resolution shared by the mutation lane and repository;
- source E2E use of the existing disposable storage-root override.

### DIFF
--- a/e2e/fixtures/electron-app.ts
+++ b/e2e/fixtures/electron-app.ts
@@ -141,12 +141,10 @@ const launchEnvironment = (
   }
 
   environment.OPEN_SCIENCE_STORAGE_ROOT = storageRoot
+  environment.OPEN_SCIENCE_E2E_STORAGE_ROOT = storageRoot
   environment.OPEN_SCIENCE_E2E_HANDOFF_CAPTURE_ROOT = join(storageRoot, 'e2e-handoff-captures')
   environment.OPEN_SCIENCE_E2E_WINDOW_MODE = windowMode
   if (sessionPerformanceTrace) environment.OPEN_SCIENCE_PERF_SESSION_TRACE = '1'
-  if (environment.OPEN_SCIENCE_E2E_EXECUTABLE) {
-    environment.OPEN_SCIENCE_E2E_STORAGE_ROOT = storageRoot
-  }
   if (fakeRemoteItRoot) {
     environment.OPEN_SCIENCE_FAKE_REMOTEIT_STATE = join(storageRoot, 'fake-remoteit-state.json')
     environment.OPEN_SCIENCE_REMOTEIT_BIN = process.execPath
--- a/e2e/launch-environment.spec.ts
+++ b/e2e/launch-environment.spec.ts
@@ -11,9 +11,16 @@ test('normalizes a Windows-style Path before injecting the fake Agent directory'
   expect(environment.PATH).toBe(`fake-agent-bin${delimiter}system-bin`)
   expect(environment.Path).toBeUndefined()
   expect(environment.ELECTRON_RENDERER_URL).toBeUndefined()
-  expect(environment.OPEN_SCIENCE_E2E_STORAGE_ROOT).toBeUndefined()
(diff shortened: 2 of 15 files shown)


ewen-poch:
Addressed the branch-history reconciliation finding in commit 278ff697f. The runtime contract still enforces at most 10 attachments per request; reconcileBranchedAttachments now finalizes versionless legacy history sequentially in bounded batches and merges the results before the existing reconciliation step. The regression test models two individually valid historical messages with 10 + 1 attachments and validates every mocked call through the public application-command contract; it fails on the unbatched implementation with the reported Zod too_big error and passes after the fix. Final verification: workspace_runtime module slice (10 files, 365 tests), Web TypeScript check, and ESLint for the changed files all pass. This does not change architecture, persistence, data models, state, or user interaction.

PR #1911, opened 2026-08-30
feat(notebook-network): enforce domain allowlist and sandbox

ewen-poch:
## Problem

Notebook REPL, Python, R, and Bash processes can initiate network and filesystem access on behalf of an Agent. The application needs one enforceable boundary that limits outbound traffic to Open Science defaults and user-approved domains, prevents access to local/private targets, constrains filesystem access, and gives users an in-conversation approval path when a domain is blocked.

## Proposed change

- Add **Open Science domains** and **Allowed domains** management to Settings > Network, including validation, translated copy, runtime status, and Windows setup state.
- Route blocked Notebook destinations through the existing conversation approval surface with Deny, Allow once, and Always allow decisions. One-time decisions are bound to the conversation, runtime, and exact command so concurrent commands cannot consume another command's approval.
- Add a project-owned `notebook-network-sandbox` package that owns the authenticated command gateway, address policy, per-command lifecycle, filesystem policy, and platform adapters:
  - macOS: Seatbelt profiles;
  - Linux: bubblewrap, a private Unix gateway socket, and the exact Notebook RPC socket;
  - Windows: a capability-free AppContainer launched by the packaged native host.
- Deny direct/private/local traffic, IP-literal bypasses, DNS rebinding, and object-storage endpoint shapes that can redirect requests to user-controlled buckets or accounts.
- Apply write-deny-by-default filesystem policy while allowing declared session/data roots and the disposable Notebook workload-cache subtree. Protected application roots remain hidden.
- Project a minimal child environment and a validated complete CA bundle for package downloads behind TLS-inspecting proxies.
- Add packaging, installer cleanup, change-impact classification, platform smoke coverage, and module tests.

```mermaid
flowchart LR
  Settings[Settings > Network] --> Owner[Notebook network settings owner]
  Notebook[Notebook REPL / Python / R / Bash] --> Sandbox[Notebook sandbox owner]
  Owner --> Sandbox
  Sandbox --> Runtime[Per-command sandbox runtime]
  Runtime --> Gateway[Authenticated network gateway]
  Runtime --> Platform[macOS / Linux / Windows isolation]
  Gateway --> Policy{Domain and address policy}
  Policy -->|allowed| Network[Remote destination]
  Policy -->|blocked| Approval[Conversation approval]
  Approval -->|allow once| Runtime
  Approval -->|always allow| Owner
```

## Scope and non-goals

- Applies to Notebook execution surfaces; it is not a general sandbox for unrelated desktop processes.
- Does not inject credentials, implement AWS SigV4, inspect request bodies, or intercept TLS.
- Windows ships through current-user NSIS only. Portable ZIP and machine-wide installation are excluded because their deletion cannot reliably release per-user durable resources.
- No historical data migration is included because this feature has not shipped to users. Missing settings continue to resolve through current defaults.
- No persisted enum values are added. Approval request metadata reuses the existing Notebook runtime values.

## Durable Windows resources

- `NotebookNetworkSandboxOwner` and the packaged Windows host own the AppContainer profile, loopback exemption, firewall rules, process Job Object, and temporary ACL leases.
- The stable installation identity is `0f3cd2a44c3d4e4e9f1e2a5b`. Creation records a journal and ownership receipt under `%LOCALAPPDATA%\Aipoch\OpenScience\notebook-sandbox\<installationId>\`, binding a random ownership token, the token-suffixed profile, and exact firewall rule names.
- Setup and launch fail closed when ownership cannot be proved. Crash recovery reclaims receipt-backed ACL leases.
- Removal first stops owned processes, then removes only receipt-backed loopback/firewall resources and the owned profile. It is idempotent and preserves unproven or shared resources.
- The NSIS uninstall hook runs the same prepare/remove/finish lifecycle. UAC cancellation is surfaced as a cancellation instead of weakening isolation.

The following checks ran after the final material edit and rebase:

- `npm run typecheck:node`
- `npm run typecheck:web`
- 69 focused network policy, runtime configuration, filesystem policy, owner, trust-bundle, and shared-domain tests
- 30 shell/environment/settings architecture tests
- 2 focused persistent-kernel sandbox and workload-cache tests
- ESLint for the conflict-affected main-process and test files
- Prettier and `git diff --check`

The complete `npm test` suite was intentionally not run locally; PR CI remains authoritative for the full portable suite.

Uncovered platform risk: this managed macOS development host rejects test socket listeners with `EPERM`, so the real Unix-socket isolation integration case is committed but must execute in platform CI. Windows setup/remove, UAC, AppContainer, loopback/firewall cleanup, ACL recovery, concurrency, and package smoke behavior require confirmation on a Windows runner; a Windows agent will follow up on this PR with those results and any necessary test adjustments.

## Review focus

- Domain matching, private-address fail-closed behavior, DNS rebinding resistance, and live connection reset after settings changes.
- Approval attribution under concurrent Notebook sessions and commands.
- Filesystem visibility/write boundaries, especially the workload-cache exception inside the otherwise read-only runtime.
- Windows ownership proof, idempotent cleanup, and NSIS lifecycle integration.
- Settings ownership boundaries, persisted field defaults, renderer interaction, and translations.


ewen-poch:
## Windows validation report (2026-08-30)

Validated branch `feat/notebook-network-domain-allowlist`, updated with a normal merge from `origin/main` (no rebase/squash/force-push). Final pushed head is `89bc760186c9122e55f74fe654e3e7b1065487ca`.

### Verdict

The rebuilt x64 native host now passes the complete native ownership/filesystem/process/removal lifecycle. I also fixed several high-value Windows bugs and committed the exact x64 host produced and exercised by PR Gate.

This is **not a full sign-off for the requested Windows behavior**. Two important areas remain blocked/unverified:

1. Classic AppContainer did not honor deny ACEs for either an enabled capability SID or a separate deny-only capability SID when the same object was reachable through an allow. The implementation now fails closed for overlapping denied roots instead of silently exposing them. This is safe, but workloads whose `deniedReadRoots`/`deniedWriteRoots` overlap a granted root cannot launch. A complete nested carve-out needs a different filesystem isolation mechanism/identity model and would change the current security model.
2. GitHub's Windows 2025 runner reserves TCP ports `49699-49798` and `49799-49898`, which includes both required ports 49700/49701. The smoke now detects and reports this before setup and marks the real network boundary unverified rather than producing a false pass.

### Environment

- Disposable hosted runner: Microsoft Windows Server 2025 Datacenter `10.0.26100`, x64, image `windows-2025-vs2026` `20260824.214.3`.
- Node `v22.23.2`; npm `10.9.8`.
- Rust `1.98.0` (`x86_64-pc-windows-msvc`); Cargo `1.98.0`; LLVM `22.1.8`.
- Visual Studio `18.9.12112.369`; MSVC tools `14.51.36231`.
- Local machine used for Node/typecheck/contract checks: AMD64, Windows display version 25H2, build `26200.9168`, Node `v22.23.2`, npm `10.9.8`, Rust/Cargo `1.98.0`. Local MSVC Build Tools were unavailable, so authoritative Rust/MSVC build and native lifecycle evidence comes from the disposable hosted runner.
- ARM64 hardware was not available: **ARM64 unverified**.

Latest focused Windows evidence: [Windows Full Test run 33292377767](https://github.com/aipoch/open-science/actions/runs/33292377767), job [99206192247](https://github.com/aipoch/open-science/actions/runs/33292377767/job/99206192247). It used `mode=notebook-sandbox`; the complete npm test shards were skipped.

### Commands and results

| Command/check | Exit/result |
| --- | --- |
| Dependency/Prisma readiness | Dependencies were already present in the isolated worktree; the `scripts/check-prisma-client.mjs` pretest guard passed. I do not have a separate fresh-install exit-code log. |
| `npm run typecheck:node` | `0` |
| `npm run typecheck:web` | `0` |
| `npm run test:module -- notebook_network_sandbox` | `0`; 10 files: 8 passed, 2 platform-skipped; 63 tests: 55 passed, 8 skipped |
| `npx vitest run scripts/windows-release-workflows.test.ts scripts/electron-builder-config.test.ts scripts/ci/pr-gate-workflow.test.ts` | `0`; 3 files, 46/46 tests passed |
| `cargo test --locked --manifest-path packages/notebook-network-sandbox/vendor/windows-src/Cargo.toml` | `0` on MSVC; 3/3 tests passed, including the PowerShell UTF-8 BOM recovery regression |
| `node packages/notebook-network-sandbox/vendor/windows/build.mjs x64` | `0`; release build completed in 24.85 s |
| `smoke.ps1 ... -Mode Full` | `0`; `Windows Notebook AppContainer smoke test passed.` |
| i18n guard after merging current `main` | `0`; 743/743 tests passed |
| Affected ESLint, Prettier, `cargo fmt --check`, `git diff --check` | all `0` |

The rebuilt vendored host is from PR Gate artifact `windows-sandbox-host-x64` in [run 33292379474](https://github.com/aipoch/open-science/actions/runs/33292379474). PR Gate's Windows core passed source tests, build, Windows-specific tests, shell/timeout tests, native lifecycle, and path portability with that artifact. Vendored SHA-256: `FBFD751E91CD14147F3B50DD53FFF228D752C3FD21550143D79DB8CFF8EB7931`.

### Setup, ownership, and UAC

- Native setup and repeated setup were idempotent; no second profile/receipt/firewall set was created.
- The Full smoke changed `LOCALAPPDATA` to simulate elevation under another account while passing the original user's ownership root. The receipt remained under the original explicit root.
- The removal-cancellation checkpoint passed: after unprivileged `prepare-remove`, the profile, loopback exemption, firewall rules, and receipt still existed. Destructive profile/receipt removal happens only after privileged network-resource removal succeeds.
- Interrupted setup recovery passed after fixing the ownership journal reader to accept the UTF-8 BOM emitted by Windows PowerShell 5.
- I did **not** manually exercise the Settings UI or secure-desktop UAC prompts. Therefore UI `setupRequired`, install UAC cancel/retry copy/state, and actual uninstall-UAC cancellation are not claimed as manually verified. The native state transitions and cancellation checkpoint passed, but this is not a substitute for the UI/UAC walkthrough.

### Filesystem boundary

Real x64 AppContainer results:

- Token/profile verification passed. The token has no network capability; it does contain the intentionally generated per-command custom filesystem capability SID, so it is not literally capability-free.
- Required Windows system files were readable.
- Declared read-only root was readable and not writable.
- Declared read-write root was writable.
- A disjoint undeclared/denied user root was unreadable by AppContainer default-deny.
- ACL lease cleanup passed after normal exit and after forcibly killing the host; the next setup recovered the crashed lease receipt.
- Ownership records with missing/invalid proof were preserved rather than used to scan/delete unrelated resources.

Security finding: an exact denied file and its parent had explicit/inherited deny ACEs while a broader parent supplied an inherited allow ACE; the AppContainer still read the file. Reordering ACE creation and using a separate capability SID marked `SE_GROUP_USE_FOR_DENY_ONLY` did not change the result. The final implementation therefore rejects unenforceable overlap before launch. See Microsoft's [AppContainer implementation overview](https://learn.microsoft.com/en-us/windows/win32/secauthz/implementing-an-appcontainer) and [SID attributes in access tokens](https://learn.microsoft.com/en-us/windows/win32/secauthz/sid-attributes-in-an-access-token) for the underlying token model.

Actual concurrent Windows commands with distinct simultaneous ACL grants were not separately exercised; do not infer that result from the crash-recovery and installation-isolation passes.

### Process and installation lifecycle

- Process-tree removal passed: the command tree was stopped before profile deletion; repeated remove was idempotent.
- Two test installation IDs remained isolated; removing one did not alter the other.
- Crashed ACL lease recovery and interrupted creation-journal recovery passed.
- Missing-receipt/unowned-resource preservation passed.
- Final owned cleanup passed.
- The smoke did not separately drive every UI path for cancel, timeout, host-window close, and uninstall cancellation. PR Gate's focused Windows shell timeout tests passed, but those are not all equivalent to an end-to-end installed-app UI test.

### Network boundary and approvals

The runner could not bind either 49700 or 49701 before AppContainer setup because Windows excluded the encompassing ranges. Therefore the real loopback probes were intentionally skipped and emitted `NETWORK_BOUNDARY_UNVERIFIED`.

Consequently, I am **not** claiming real-Windows verification for:

- gateway-only access to 127.0.0.1:49700 versus 49701/other TCP ports;
- UDP, direct public/private sockets;
- allowed/unallowed domain traffic through the gateway;
- Deny / Allow once / Always allow UI results;
- concurrent-command approval ownership and credential separation;
- tearing down already-established CONNECT/SOCKS tunnels after an allowlist update.

The focused module suite for these policies passed (55 passed/8 platform-skipped), but the requested real Windows network behavior remains uncovered.

### NSIS/install path

- `electron-builder.yml` and the 46-test workflow/config contract suite confirm Windows targets NSIS only and has no portable ZIP target.
- I did **not** produce a real NSIS install/uninstall screenshot or log in this session. Actual custom-directory install, moved-app stable installation ID, UI uninstall/UAC cancellation, and installed-package cleanup remain unverified.

### Cleanup and residual resources

- The latest Full smoke reached `final owned cleanup` and exited `0`, proving cleanup of its owned profile, loopback exemption, firewall rules, ACL leases, receipts, and test directories on the disposable runner.
- A local post-check found no `notebook-appcontainer-host` process, no `OpenScienceNotebook-*` firewall rule, no `Aipoch.OpenScience.Notebook*` loopback/profile entry, and no receipt directory for `0f3cd2a44c3d4e4e9f1e2a5b`, `0123456789abcdef01234567`, or `89abcdef0123456789abcdef`.
- No ownership-proven local sandbox resource remains.

### Other current PR failures/risks

- `CI Integrity` fails because this PR changes protected `.github/workflows/pr-gate.yml` and `scripts/ci/change-impact.json`; the check explicitly requires a maintainer ruleset bypass.
- PR Gate's Linux Notebook isolation job failed on the hosted runner with `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted` (4 failed, 1 passed, 1 skipped). This is outside this Windows validation/fix scope and remains unresolved.
- ARM64, real Settings/UAC interaction, real fixed-port network enforcement, approval UI/concurrency/revocation, and installed NSIS lifecycle remain uncovered as noted above.

### Changes pushed during validation

High-value final changes include Windows firewall/GetTokenInformation fixes, enabling and verifying the per-command filesystem capability, fail-closed rejection of unenforceable denied-root overlaps, removal ordering that preserves recoverability after UAC cancellation, bounded native lifecycle diagnostics, UTF-8 BOM ownership-journal recovery, latest-`main` merge conflict resolution, and refresh of the exact tested x64 host binary. No rebase, squash, or force-push was used.

ewen-poch:
Final-head follow-up: PR head \89bc760186c9122e55f74fe654e3e7b1065487ca\ is mergeable, and its [Windows core job](https://github.com/aipoch/open-science/actions/runs/33292641821/job/99207118577) completed successfully after the vendored-host commit. Source test, x64 rebuild/artifact upload, Windows-specific tests, shell and timeout tests, AppContainer lifecycle, path portability, and the enforcement step all passed. The CI Integrity maintainer-bypass failure and the uncovered/blocked items in the main validation report remain unchanged.

ewen-poch:
Addressed the Windows kernel launch issue in `168508dc`. Windows command serialization now prefixes the single-quoted executable and arguments with PowerShell’s call operator (`&`), while preserving literal handling for apostrophes and command-substitution-looking input. Added a regression test covering a spaced executable path, an apostrophe, and a `$()`-shaped argument; the focused sandbox-owner, kernel-executor, and shell-process tests plus Node typechecking pass.

ewen-poch:
Addressed the stale PATH projection issue in the current head. `environmentPathRoots` now keeps only absolute entries that currently resolve to directories, so deleted or broken PATH entries are not sent to the Windows ACL lease. Added Windows-style stale-entry coverage; the process-environment and sandbox-owner suites pass (17 tests), the focused kernel cases pass, and Node typechecking succeeds.

The earlier firewall receipt finding is also addressed in the current head: validation now requires the exact installation-derived TCP and UDP rule names in their expected order, with a regression covering injected and swapped names. Receipt-controlled firewall names that only share a prefix are rejected before elevated setup or removal.

ewen-poch:
Addressed the latest review findings in 4b212947. Windows filesystem ACLs now use a per-installation shared lease state: the baseline is captured once, ACL mutations are serialized only under the short resource mutex, releasing one command rebuilds grants for every still-live command, and crash recovery reconciles validated receipts before restoring the final baseline. The Windows PR-gate smoke now keeps two commands alive on the same writable root, exits the first, verifies the second can still write, and then checks exact ACL restoration.\n\nOn Linux, the REPL RPC bearer token is no longer placed in the environment or command line. The parent sends it through inherited fd 3, bubblewrap preserves exactly that descriptor, and the trusted loop consumes and closes it before user code runs. Integration coverage verifies host.mcp still works and the token is absent from process.env and /proc/self/environ.\n\nAlso fixed the CodeQL reflected-XSS test fixture and made the Artifact provenance E2E producer lookup independent of compacted output previews. Focused module tests, full kernel/REPL files, x64/arm64 Rust checks, deterministic Windows host rebuilds, and the provenance Electron E2E pass locally.

ewen-poch:
The first Windows rerun reached the native smoke and exposed one exact-restore detail before the new concurrency assertion: ACEs matched, but restoring an unprotected DACL set the `SE_DACL_AUTO_INHERITED` control bit (`D:` became `D:AI`). The current head `601415b3` now snapshots both DACL auto-inheritance control bits, restores the effective ACL through `SetNamedSecurityInfoW`, then reapplies the original control bits to the restored descriptor before committing it. x64 and ARM64 hosts were rebuilt twice with identical hashes, both cross-target checks pass, and the resource hashes are updated. A new Windows core run is now pending.

ewen-poch:
The latest Windows core run reached the new concurrent-lease section. The first command failed before it became live because the smoke fixture passed a relative powershell.exe path, while sandbox launches use the absolute-executable contract already enforced by the product path. The fixture now resolves powershell.exe to its absolute application path before encoding the launch specification. Focused PR-gate contract tests and formatting checks pass; a new Windows core run is pending.

ewen-poch:
Rebased the branch onto current main and preserved both immutable working-file evidence cleanup and sandbox execution cleanup in the kernel conflict resolution.

The Linux review findings are addressed. inheritedFileDescriptorCount now reaches linuxLaunch, bubblewrap preserves fd 3, the gateway bridge explicitly forwards preserved descriptors, and the child receives the original ELECTRON_RUN_AS_NODE value while the bridge still runs through Electron as Node. Persistent kernels no longer bind process lifetime approval state to the first cell AbortSignal. Coverage now exercises the Electron-backed POSIX REPL, the Linux fd 3 bridge, wrapper-level descriptor forwarding, and includes runtime-config.test.ts in the notebook_network_sandbox module gate.

Focused results: typecheck:node passed; notebook_network_sandbox passed 72 tests with 1 platform skip; kernel plus real sandbox passed 97 tests with 2 skips; CI mapping and PR-gate contracts passed.

Correction to the earlier Windows note: resolving powershell.exe to an absolute path was necessary but was not the final concurrent-smoke failure. Later diagnostics proved both capability leases became live and the child completed, while Start-Process did not reliably populate ExitCode in the hosted PowerShell process object. The smoke now uses an explicit child completion marker, host-exit state, stderr, surviving-lease write access, and exact final ACL restoration instead of treating an unavailable ExitCode as failure.

ewen-poch:
The first real Linux rerun exposed a concrete CLI compatibility error in the fd fix: bubblewrap does not implement --preserve-fds. Its process model already passes inherited extra descriptors to the sandbox command. The unsupported option is removed; the validated descriptor count is now used only by the gateway bridge to construct the nested child stdio range. The real Linux test still requires fd 3 to reach the final command, so this remains an end-to-end assertion rather than a skipped compatibility path. The focused sandbox module passes 72 tests with 1 platform skip, and typecheck:node passes. A new Linux isolation run is pending on head 688f39f0.

ewen-poch:
Addressed the immutable-input and sandbox integration gap introduced by the latest Notebook file-generation work.

- `host.artifactPath(versionId)` and execution-scoped input resolution now stage an exact checksum-verified copy under the calling Notebook Session input root instead of returning the authoritative Upload/Artifact storage path.
- The stable input root is created before kernel launch and granted read-only access to persistent Python, R, REPL, and stateless Bash sandboxes. Authoritative `uploads/` and `artifacts/` directories remain inaccessible.
- Concurrent resolution reuses one staged copy; a corrupt derived copy is rebuilt atomically from the verified Version.
- Session and Project deletion remove only the owned derived input cache after runtime shutdown.
- This adds no database schema, persisted enum, historical migration, or user-interaction change. The directory is a derived filesystem cache.

Validation after rebasing onto `e3a5b5b7`:
- `npm run typecheck:node`
- 75 focused Notebook, Host Artifact, cleanup, architecture, and bundled-Skill tests
- `npm run test:module -- notebook_network_sandbox` (72 passed, 1 skipped)
- macOS production sandbox check proving staged content is readable and write attempts are denied
- ESLint on all changed TypeScript files and `git diff --check`

ewen-poch:
Addressed the approval-card spoofing finding in 5647fba1. Application-owned permission requests now carry a main-process appOwned provenance projection that provider requests cannot set, and the Notebook network presentation requires that trusted marker before hiding tool identity. A provider/MCP request with a notebookNetworkApproval-shaped raw payload now remains an External service request and displays its broker-resolved identity. The marker is transient and is not persisted. Focused Node/Web typechecks, 210 broker/presentation/render/interaction tests, targeted ESLint, and git diff --check pass.

ewen-poch:
Addressed the actionable review findings in `999937a7`.

- Windows firewall ownership is now unique per installation and AppContainer profile SID. Receipt/journal validation requires the exact derived rule names, setup validates the existing rule contents before replacement, and removal preserves any rule whose ownership cannot be proven.
- The Windows gateway no longer uses a fixed port. Setup asks Windows for an available loopback port, persists it in ownership receipt schema 3, exposes it through status, and the runtime binds the authenticated shared gateway to that recorded port. An occupied recorded port is reported as unavailable; setup can rotate it through a recoverable journal, and UAC cancellation restores the prior receipt state.
- Both vendored Windows hosts were rebuilt from the updated Rust source: x64 SHA-256 `3b1a25a741457e59542b926b9442049edbcc26a951a4023324e358a8a0ff6c7d`, ARM64 SHA-256 `29a27d4c812c1cdc2162e0ad7c29a3990b2b5ddd0cf4e1fc1e89a086a34c9bff`. The resource integrity test now pins these exact outputs.
- For AppImage, the existing supported provisioning path remains an explicit runtime gate: Debian declares `bubblewrap`; other Linux packages check for `bwrap` and Settings displays localized installation guidance before Notebook launch. AppImage cannot declare a host package dependency, and bundling a second bubblewrap binary would not make the required host user-namespace/setuid/kernel configuration portable, so no embedded binary was added.
- Compute submission now accepts only an exact immutable Notebook input file staged for the submitting Project and Session. It rejects cross-Session paths and symlink escapes while preserving the existing managed Artifact resolver. The composition stays behind the public Compute facade.
- The independent audit also identified that Network save failures were visually presented as neutral status; they now use the standard failure alert surface.

Focused validation:

- `npm run test:module -- compute_service`: 932 passed, 6 skipped
- `npm run test:module -- notebook_network_sandbox`: 73 passed, 1 skipped
- i18n catalog guard: 743 passed
- focused Compute/input/UI/architecture tests: 66 passed
- Node and Web typechecks, ESLint, Prettier, and `git diff --check`
- Rust tests: 7 passed; x64 and ARM64 MSVC targets compile successfully

This adds no database migration. The only persisted format change is the Windows sandbox ownership receipt's `gatewayPort`; the project is pre-release and intentionally does not retain compatibility with older receipts.

ewen-poch:
The latest Linux fd finding is not being applied because the proposed mechanism does not exist in bubblewrap. `bwrap` inherits already-open file descriptors across `exec`; it has no supported `--preserve-fds` option. An earlier implementation added that argument and the real Linux run failed immediately because the option is unsupported.

The current chain is intentional and end-to-end tested:

- `inheritedFileDescriptorCount` reaches `linuxLaunch`.
- bubblewrap inherits fd 3 into the in-sandbox gateway bridge.
- the bridge constructs the nested child `stdio` array through `3 + inheritedFileDescriptorCount`, preserving the same descriptor range for the REPL child.
- `network-enforcement.integration.test.ts` reads the MCP credential from fd 3 after both bubblewrap and the bridge.

The latest real `Linux Notebook isolation` job passed on head `999937a7`: https://github.com/aipoch/open-science/actions/runs/33321159179/job/99283991973

Adding a nonexistent descriptor-preservation argument would regress Linux startup, so no code change is appropriate for this review item.

ewen-poch:
Validated the current PR head on a real Windows 11 Pro x64 machine using Node 22.23.2 / npm 10.9.8 and Electron 39.8.10.

I found and fixed four deterministic Windows-only test-fixture failures in commit `ceb92fdfc`:

- skip the Linux mount implementation test on Windows;
- use a Windows named pipe for the local gateway RPC test instead of a filesystem Unix socket path;
- use a real temporary package-cache directory instead of hard-coded `/tmp`;
- use a directory junction for the staged-input realpath escape test, so the security assertion still runs without Developer Mode or administrator symlink privileges.

Verification completed successfully:

- `npm run typecheck:node`
- `npm run typecheck:web`
- `npm run lint`
- Prettier and `git diff --check`
- `npm run test:module -- notebook_network_sandbox`: 67 passed, 10 platform skips
- package/path portability fixtures: 132 passed
- Windows PR-gate-focused groups: 188 + 4 + 1 + 208 tests passed (2 platform skips)
- network permission/settings group: 385 passed
- compute/input group after the junction fix: 47 passed (68 unrelated skips)
- headless/task/application-command group: 116 passed
- Windows packaging/installer configuration group: 79 passed
- `npm run build:e2e`
- isolated Electron desktop validation: startup completed without UAC, `userData` and `OPEN_SCIENCE_STORAGE_ROOT` used separate `%TEMP%` roots, Settings > Network showed `Status: Not set up`, `Set up`, and the administrator-permission explanation, and shutdown/temporary-root cleanup completed.

Environment-limited items were not marked as passing:

- Exact `npm ci`, Rust tests, and rebuilding the vendored Windows host require MSVC/Windows SDK, which are not installed on this machine. `npm ci --ignore-scripts` plus patch integrity and Prisma generation succeeded; the checked-in x64/arm64 native binary hashes were unchanged after the failed build attempt.
- The official elevated AppContainer/WFP smoke was requested twice, but UAC was not approved. The smoke script never started. Native `status` checks for both fixed smoke installation IDs report `profileExists=false`, `networkFenceReady=false`, and `owned=false`, so no smoke resources remain.
- The isolated desktop profile had no configured agent provider and Windows protection was not installed, so live Bash/Python/R/package-download approval-card interaction was not claimed as tested.
- Headless permission cancellation/result propagation tests pass, but the current implementation has no reliable signal that distinguishes “no approver connected” from a headless web client that may respond later. A request with no responder can remain pending until its abort/session-cleanup signal fires; therefore “fast default deny when no approver exists” is not claimed as passing. Fixing that correctly would require an explicit approver-presence/lifecycle contract rather than a local guess.

One minor UI observation: the Not set up details currently repeat the administrator-setup reason several times before the gateway-port reason. It did not block setup-state validation, and I did not broaden this test-portability patch into a copy/state change.
Remote CI note: after this push, `CI Integrity` failed on the PR's complete diff, not on the four-file Windows fixture commit. The checker reports `protected-gate-control-plane` for `.github/workflows/pr-gate.yml` and `scripts/ci/change-impact.json`; both were already part of the PR before `ceb92fdfc`. By policy these changes require an explicit maintainer ruleset bypass. I have not attempted to weaken or work around that guard.

ewen-poch:
Addressed the repeated Windows setup explanation observed during isolated desktop validation in commit 7e5a720c. Multiple distinct native setup reason codes intentionally share the same user-facing administrator-setup copy; the Settings renderer now deduplicates those final display strings while preserving distinct diagnostics such as the unavailable gateway port. Added a focused regression test covering the four equivalent Windows setup reasons plus the independent gateway-port reason. Verified with `npm run typecheck:web`, the renderer i18n guard and focused render suite (751 tests), Prettier, and `git diff --check`. No status enums, persistence, setup lifecycle, or interaction semantics changed.

ewen-poch:
Windows 11 x64 native validation is complete and the fixes are included in e4c5cf704.

Fixed during real-machine validation:
- avoided Rust UB when Windows returns an empty AppContainer loopback list (count 0 with a null pointer)
- moved AppContainer profile creation into the elevated setup phase so cancelling UAC does not leave a profile that standard-user cleanup cannot remove
- preserved the automatic-inheritance ACL control flag when restoring filesystem ACL snapshots
- made concurrent smoke completion marker-driven while retaining localized PowerShell stderr for diagnostics
- made UAC cancellation handling independent of localized exception text
- fixed the generated PowerShell connection probe: joining try/catch/finally with semicolons produced invalid `}; catch` syntax and falsely reported windowsNetworkFenceMissing even though the native WFP fence was correct

Validation:
- VS Build Tools 2022 17.14.39; MSVC 19.44.35228; Windows SDK 10.0.26100.0; Rust 1.98.0; Node 22.23.2
- cargo fmt --check + cargo test: 7 passed
- notebook_network_sandbox module: 69 passed, 10 platform skips
- Notebook network settings render tests: 8 passed
- Windows installer/release workflow tests: 54 passed
- typecheck:node passed
- npm ci and build:e2e passed
- full elevated native smoke passed, including WFP boundary checks, ACL restoration, process-tree removal, interruption recovery, cancellation/idempotency, and final cleanup
- real desktop lifecycle passed in isolated storage: startup without elevation, first Python Notebook run through standard fallback, setup cancellation, setup approval, Active/ready verification, removal cancellation preserving Active resources, approved removal, and final setupRequired/unowned status
- no real provider credentials were available in the isolated store, so external-provider agent interaction was skipped rather than replaced with a fake provider

Rebuilt x64 host SHA-256: 8393f9e27fdd4266dadc0de0d214a3a1f99b473826381f4631da7c2fe5f43436

Repeated clean builds were functionally identical but not byte-identical because PE/Debug Directory timestamps and the RSDS/PDB GUID vary with the Windows toolchain. The committed binary was validated directly. Final product identity state is clean: profile=false, loopback=false, networkFenceReady=false, owned=false, ownershipState=unowned, gatewayPort=null.

ewen-poch:
Rebased the branch onto current main (a8bf5eb1) and compressed the PR history into two logical commits: the feature implementation and sandbox lifecycle hardening. The final tree also closes two non-architectural audit findings: the elevated Windows full-smoke now requires both concurrent native hosts to exit with code 0 after writing their completion markers, and the Windows ARM64 host was rebuilt from the current locked Rust source instead of shipping the pre-lifecycle-fix binary. The new ARM64 SHA-256 is `7e8d0afa82feab7bba5b7e2b5e73c8a95407b4c4fddf56dd180a7370ff22fc57`. Verified the notebook network module (78 passed, 1 platform skip), release workflow tests (22 passed), the resource hash test (2 passed), AArch64 PE identity, typechecks from the rebased tree, and diff checks. No architecture, persistence, status enum, or user-interaction semantics changed.

ewen-poch:
## CI follow-up (`f8e630f4`)

Located the remaining red checks on `b4703523` from the blob reports and Windows smoke log, then pushed the fixes onto this branch.

### Code-greenable failures

| Check | Root cause | Fix |
| --- | --- | --- |
| Full Module tests shard 1/2 | `reviewer-orchestrator.architecture.test.ts` timed out at 15s while AST-parsing every production source for Reviewer importers | Cache `SourceFile`s and skip files whose text cannot mention the target modules |
| Full Module tests shard 2/2 | `workspace-components.test.tsx` line cap `797 > 790` after kernel-stop confirmations on granted-root policy changes | Raise the presentation-owner cap to 810 |
| Windows core (`windows_sandbox`) | Concurrent-lease smoke: `HasExited=True`, complete marker present, `Start-Process` `ExitCode` still `$null` (`$null -ne 0`) | Refresh/wait for `ExitCode`; treat empty code as success only when the complete marker exists |
| Module tests and coverage / PR Gate | Cascade from the shard failures | Should recover once shards are green |

### Maintainer bypass (not code-fixable)

**CI Integrity** still fails because this PR changes protected control-plane files (`.github/workflows/pr-gate.yml`, `scripts/ci/change-impact.json`). That check requires an explicit maintainer ruleset bypass.

### Local Test Impact Set (after last material edit)

- `npx vitest run src/main/reviewer/reviewer-orchestrator.architecture.test.ts src/renderer/src/pages/workspace/workspace-components.test.tsx` → pass (37)
- Windows AppContainer smoke was not re-run here (no Windows runner)

### Explicitly unchanged in this follow-up

- No historical data migration
- No new persisted enum/state
- No additional durable persistence beyond the AppContainer resources already in this PR

PR #1914, opened 2026-08-30
fix(notifications): protect pending actions and native privacy

ewen-poch:
## Problem

The message center treated unresolved approvals, questions, and credential requests as ordinary history. Once 50 newer items existed, an active action disappeared from the renderer snapshot even though the unread count still included it. The 1,000-row retention pass could also delete the action permanently.

Native OS notifications separately exposed prompt snippets, approval/request details, and raw provider errors by default. Settings only represented the user's enable/disable preference, so it could not report whether Electron supported native delivery or whether an explicit test was shown, failed, or could not be confirmed.

## Proposed change

- Project every active `pending` action alongside the bounded recent-history window.
- Apply the 1,000-item limit to history only; active actions may temporarily exceed it, then history is pruned when an action settles, expires, or its target is invalidated.
- Default native banners to generic copy. Add an explicit task-content opt-in, while keeping provider errors generic even after opt-in.
- Expose portable `supported` / `unavailable` capability and an explicit test notification with `shown`, `failed`, `unconfirmed`, and `unavailable` results.
- Log delivery attempts, Electron-confirmed display, and failures with distinct semantics.
- Add the privacy toggle, delivery status, test action, and all seven renderer translations to General Settings.

## Scope and non-goals

- No database schema or migration is added.
- No inbox pagination or "load more" interaction is introduced; the existing history window remains bounded.
- The implementation does not invent a portable permission-denied state that Electron cannot reliably expose. The explicit test result is the observable diagnostic.
- Native diagnostics are Electron-only. Local Web may update the persisted privacy preference; remote Web is rejected by the existing local-only command policy.

- **Historical compatibility:** existing inbox rows are classified from the existing `actionState` and `targetInvalidatedAt` columns. Existing settings documents without the new field resolve to privacy-safe `false`; no rewrite or migration is required.
- **Persisted data:** adds one optional `showNotificationContent` boolean to `settings.json`. The notification inbox schema and stored rows are unchanged.
- **New states:** adds non-persisted `NotificationDesktopAvailability` (`supported | unavailable`) and `NotificationTestResult` (`shown | failed | unconfirmed | unavailable`) unions. No persisted enum gains a value.
- **Interaction:** General Settings gains one opt-in toggle plus a status row and test button. The test bypasses the ordinary enable/focus gates by design.

These are bounded fixes for reproduced user-visible loss and privacy failures; protecting active actions and exposing only Electron's portable delivery evidence avoids a broader queue or permission-state redesign.

The listed checks ran against the final material diff:

- Pending projection, retention, native privacy, and delivery diagnostics -> `npm exec --no -- vitest run src/main/notifications/notification-inbox-repository.test.ts src/main/notifications/notification-inbox-controller.test.ts src/main/notifications/task-notifications.test.ts src/main/notifications/electron-wiring.test.ts src/main/notifications/notification-inbox-ipc.test.ts` -> 5 files, 118 tests passed.
- Settings persistence, defaults, validation, and command routing -> focused Settings command -> 5 files, 104 tests passed (262 unrelated tests skipped by the filter).
- Renderer store, Settings UI, i18n, preload, Electron/Web contracts, surface profiles, and application-command wiring -> focused contract command -> 15 files, 1,034 tests passed after rebasing onto the latest `main`.
- Web argument-shape characterization -> `npm exec --no -- vitest run src/renderer/web/renderer-argument-shape-characterization.test.ts` -> 8 tests passed. (The file timed out when co-scheduled with the larger contract batch and passed when rerun in isolation.)
- Process types -> `npm run typecheck:node` and `npm run typecheck:web` -> passed.
- Static checks -> `npm run lint`, `npm run check:web-api-map`, and `git diff --check` -> passed.

`npm run test:affected:explain -- --base origin/main --head HEAD` conservatively selected the complete Vitest fallback because `src/main/ipc.ts` has no module owner. Per the requested workflow, the full portable suite was not duplicated locally; PR Gate remains authoritative for the exact-head full and platform lanes.

Uncovered local risk: real lock-screen, Focus/Do Not Disturb, and notification-policy behavior varies by OS. Unit coverage pins Electron's documented `show` / `failed` / `close` boundaries, while cross-platform PR CI remains the validation authority.

## Review focus

- The active-pending and history predicates are disjoint under SQL `NULL` semantics.
- Privacy fails closed for missing/invalid historical settings and for settings-read failures.
- Provider errors never enter native copy, including after content opt-in.
- Electron-only diagnostics and local/remote Web capability profiles remain intentional.


ewen-poch:
Addressed the P2 localization finding. Privacy-safe native bodies are now translated when each notification is described and carried through the existing genericBody field, so delivery no longer reverse-indexes already-localized titles. Added translations for all seven supported locales and regression coverage for completion, attention, approval, response, and plan-approval notifications.

Validation: npx vitest run src/main/notifications/task-notifications.test.ts src/renderer/src/i18n/resources.test.ts (822 passed); npm run typecheck:node; focused ESLint on the changed TypeScript files. This adds no state, persistence, schema, architecture, or interaction changes.

PR #1917, opened 2026-08-30
fix(remote-access): enforce browser authorization lifecycle

ewen-poch:
## Problem

- Opening Remote settings for the first time or after the renderer cache expired called the mutating `remoteAccess.detect()` path. In Browser mode that forced reconciliation, advanced the authorization generation, closed every remote WebSocket, and cleared pending and time-limited browser access.
- One-time and trusted-browser authorization was only checked during WebSocket upgrade. A socket that outlived its authorization could continue receiving replay, application, Task progress, and heartbeat traffic.

## Proposed change

- Add a read-only `remoteAccess.probe()` command across the service, IPC/host command, preload contract, and generated Web API map. Automatic Remote settings loads use `probe`; the existing explicit Detect action still uses `detect` for repair and public-link refresh. Probe retains its provider observation in memory so later snapshots stay current, without broadcasting or persisting it.
- Keep `probe` available to Electron and local Web callers while installing a rejecting stub on the remote Web surface.
- Represent accepted external WebSockets with a stable `principalId` and runtime `isCurrent()` check. Pairing expiration reports the expired principal so its sockets close immediately, and every outbound WebSocket path closes stale sockets before sending.

## Scope and non-goals

- No database or `remote-access.json` fields, migrations, or persisted authorization-principal model.
- No historical-data compatibility work is needed because no persisted shape changes.
- No new `RemoteAccessMode`, `RemoteAccessLifecycle`, pairing-decision, or other persisted enum values.
- No new UI interaction, copy, or i18n changes.
- No unrelated idempotency, caller-identity, or Remote Access refactor.

| --- | --- | --- |
| Remote settings automatic load uses read-only probe, not mutating detect | `rtk npm test -- src/renderer/src/pages/settings/SettingsPage.render.test.tsx -t "opens the Remote panel with three scenario-based access modes"` | Red twice before production change (`probe` called 0 times); green after implementation |
| Browser probe preserves a real one-time cookie and connection | `rtk npm test -- src/main/remote-access/service.test.ts -t "probes Browser access without closing or invalidating a one-time session"` | Red twice before production change (`closeExternalConnections` called once); green after implementation |
| Probe observations remain current without persistence or broadcast | `rtk npm test -- src/main/remote-access/service.test.ts -t "keeps provider observations from a probe current without persisting configuration"` | Red twice before production change (later snapshot remained stale); green after implementation |
| Natural expiry reports stable principal IDs and invalidates runtime authorization | `rtk npm test -- src/main/remote-access/pairing.test.ts -t "reports each naturally expired authorization by stable principal ID"` | Red twice before production change (expiry callback called 0 times); green after implementation |
| Stale real WebSockets close before application, replay, Task progress, or heartbeat traffic | `rtk npm test -- src/main/web-service/http-server.test.ts -t "closes stale"` | Red twice before production change (`leaked`, `replayed`, or `heartbeat`); final: 4 passed |
| Probe is absent from the remote Web RPC surface and remains callable only from local callers | `rtk npm test -- src/shared/renderer-contract-catalog.test.ts -t "keeps the Remote Access probe local-only"`; `rtk npm test -- src/main/host-application-commands.test.ts -t "keeps the read-only Remote Access probe available only to local callers"` | Each regression was red twice before its production change; both green after implementation |
| Remote Access, Web Service, Settings, IPC/host/preload, and contract impact set | `rtk npm test -- <29 targeted test files>` | 29 files passed; 481 passed, 3 skipped |
| Node types | `rtk npm run typecheck:node` | Passed |
| Web types | `rtk npm run typecheck:web` | Passed |
| Lint | `rtk npm run lint` | Passed |
| Generated Web API map | `rtk npm run check:web-api-map` | Passed |
| Changed-file formatting | `rtk npx prettier --check <5 changed files>` | Passed |
| Patch hygiene | `rtk git diff --check` | Passed |

Per contribution guidance, the full repository test suite was not run; validation used the final targeted Test Impact Set above.

## Review focus

- Confirm `probe()` only observes provider state and updates its in-memory observation; it never reconciles routing, advances authorization generation, closes connections, changes pairing/session state, broadcasts, or persists configuration.
- Confirm `probe` is callable from Electron/local Web but represented by a rejecting stub for remote Web, while explicit Detect retains its desktop-only repair and public-link refresh behavior.
- Confirm principal-scoped expiry closure and the defensive `isCurrent()` gate cover replay, application broadcasts (legacy and replay), Task progress, and heartbeat frames without affecting local WebSockets.

### DIFF
--- a/src/main/application-command-composition.test.ts
+++ b/src/main/application-command-composition.test.ts
@@ -309,6 +309,7 @@ describe('application command composition', () => {
     const firstSnapshot = vi.fn(() => snapshot)
     const firstOwner = {
       snapshot: firstSnapshot,
+      probe: vi.fn(),
       detect: vi.fn(),
       setMode: vi.fn(),
       disable: vi.fn(),
--- a/src/main/application-command-composition.ts
+++ b/src/main/application-command-composition.ts
@@ -241,6 +241,7 @@ const createRemoteAccessSlot = (): Readonly<{
   }
   const owner: RemoteAccessOwner = Object.freeze({
     snapshot: (...args) => current().snapshot(...args),
+    probe: (...args) => current().probe(...args),
     detect: (...args) => current().detect(...args),
     setMode: (...args) => current().setMode(...args),
     disable: (...args) => current().disable(...args),
--- a/src/main/host-application-commands.test.ts
+++ b/src/main/host-application-commands.test.ts
@@ -492,7 +492,7 @@ describe('Host application commands', () => {
         .filter((channel): channel is string => channel !== null)
     )
 
-    expect(localOnlyChannels).toHaveLength(28)
+    expect(localOnlyChannels).toHaveLength(29)
     for (const channel of localOnlyChannels) {
       await expect(
         router.dispatcher.invoke(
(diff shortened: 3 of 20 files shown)


ewen-poch:
Addressed the reported provider-observation consistency issue in `9f13b4447`.

`probe()` now retains the fresh Remote.It observation in service memory before returning the snapshot, so later snapshots cannot overwrite the renderer with stale provider data. It still does not broadcast or persist configuration, and it does not touch routing, authorization generation, connections, or pairing/session state.

Added `keeps provider observations from a probe current without persisting configuration`; it failed twice against the previous production code because the later snapshot remained stale, then passed after the fix. Final targeted impact set: 29 files passed, 470 passed, 3 skipped; node/web typecheck, lint, generated Web API map, and diff hygiene also passed.

ewen-poch:
Addressed the local-only probe review finding in 8aca3f294. remoteAccess.probe now uses the LOCAL renderer profile, remote Web installs a rejecting stub, and the host boundary accepts Electron/local Web while rejecting remote callers before the owner. Both new public-boundary regressions were red twice before production changes and are green; the final 29-file impact set passes with 481 passed and 3 skipped.

PR #1921, opened 2026-08-30
fix(compute): enforce remote operation lifecycle boundaries

ewen-poch:
## Problem

Four Compute boundaries treated an incomplete projection or a best-effort cleanup as a completed operation:

- An SCP timeout, cancellation, size limit, or output failure sent `SIGTERM` once and then waited indefinitely for `close`. A child that ignored `SIGTERM` could outlive the advertised 10/30-minute timeout, keep Session cancellation resources alive, and block Host deletion while active operations drained.
- `ComputeHostRepository.list()` silently omitted a Host row when its critical authentication data could not be decoded. Session hydration and permission reconciliation then interpreted the incomplete list as authoritative absence and could permanently prune enabled/selected Hosts and grants.
- Agent downloads used unowned operating-system temp directories. Transfer failures leaked the operation directory, and successful sensitive downloads had no Session/Project deletion lifecycle.
- All Compute approvals were presented as remote commands. Download approvals omitted `remote_path`; job approvals omitted resources, timeout, and remote workdir; remembered-scope confirmation did not describe the real operation class.

## Proposed change

- Extract a small SSH/SCP `BoundedChildTermination` helper: `SIGTERM`, a 2-second grace period, `SIGKILL`, then a final event boundary that settles and detaches even when Node never receives `close`.
- Make Host list decoding fail closed for the whole catalog. Existing hydration/reconciliation guards now preserve Session Host selections and grants because a corrupt/unsupported row is no longer projected as a missing Host.
- Add `SessionCacheOwner` under `dataRoot/compute/session-cache/<projectId>/<sessionId>/<operationId>/<filename>`.
  - A failed transfer immediately removes the exact operation directory.
  - Existing Session/Project deletion commits remove the exact owned cache subtree after Compute Job cleanup.
  - A complete Session catalog hydration reconciles crash leftovers inside the dedicated app-owned cache root.
- Replace the optional approval field bag with an `operation`-discriminated request union (`call_command | submit_job | download`) and render operation-specific titles, details, remembered-scope subjects, and code-execution disclosure. Job producers now include the existing resource request.

```mermaid
flowchart LR
  A[Agent download] --> B[Project / Session / operation cache]
  B -->|transfer fails| C[remove exact operation directory]
  B -->|Session or Project deleted| D[existing deletion commit]
  D --> E[remove exact owned cache subtree]
  F[complete Session hydration] --> G[reconcile crash leftovers in app-owned cache root]
```

## Scope and non-goals

- No database column, migration, or persisted approval format is added.
- No unavailable/corrupt Host state is added. A critical Host decode error makes the catalog request fail explicitly.
- Durable grant semantics remain `provider + operation + scope`; this change only makes the transient approval payload and presentation explicit.
- Existing `cs-session-*` operating-system temp directories are not scanned or deleted. Their ownership cannot be proven safely, so this PR does not introduce a historical sweep.
- Existing Session/Project identifiers remain compatible with the Session repository's path-segment rules.

## Persistence, compatibility, and state impact

- **Persistence:** successful Session-cache downloads now remain in the app-managed data root until their owning Session or Project is deleted. The cache is rebuildable and has no database index.
- **Historical compatibility:** no data migration is required. Old OS-temp cache directories remain a bounded legacy exception; new downloads use only the owned layout.
- **New states/enums:** `ComputeApprovalOperation` adds the transient IPC discriminator values `call_command`, `submit_job`, and `download`. No persisted lifecycle state or database enum changes.
- **Failure behavior:** critical Host decode failures now reject `list()` instead of returning a partial authoritative set. This may surface an error where the UI previously showed an incomplete list, intentionally preserving recoverable data.

- SCP hard timeout and size-limit termination -> `npx vitest run src/main/compute/scp-runner.test.ts src/main/compute/ssh-runner.test.ts` -> passed, including children that ignore `SIGTERM`.
- Corrupt/unsupported Host row handling -> `npx vitest run src/main/compute/repository.test.ts` -> passed.
- Session-cache allocation, transfer-failure cleanup, deletion composition, and crash reconciliation -> `npx vitest run src/main/compute/session-cache-owner.test.ts src/main/compute/compute-remote-operation-owner.test.ts` -> passed.
- Approval producer, broker, renderer, notification, store, and localization contracts -> focused 15-file Vitest run -> 15 files / 1243 tests passed; the final SCP/cache subset rerun passed 3 files / 120 tests.
- Main and shared TypeScript contracts -> `npm run typecheck:node` -> passed.
- Renderer/shared TypeScript contracts -> `npm run typecheck:web` -> passed.
- Changed Compute, notification, renderer approval, store, and shared TypeScript surfaces -> focused `npx eslint --cache ...` -> passed.

The full local `npm test` suite was intentionally not run. PR Gate is authoritative for the complete portable and platform-specific plan.

## Review focus

- Confirm every abnormal SCP exit path reaches the shared bounded terminator without changing normal exit/error results.
- Confirm Session cache cleanup occurs only from validated Project/Session/operation identities and only inside the dedicated app-owned root.
- Confirm a Host catalog decode failure cannot reach absence-based Session/grant pruning.
- Confirm approval disclosure and remembered-scope language match the discriminator while grant keys remain unchanged.

Uncovered local risk: platform-specific process signal/event behavior is left to the PR Windows/macOS/Linux CI lanes. The legacy OS-temp directories are deliberately not cleaned because this version cannot prove their ownership.


ewen-poch:
Addressed the two findings on head `4249bceb` in `bfb2d0a4`:

- Session-cache filenames are now validated before any operation directory is allocated. The regression test verifies that an invalid basename leaves no cache directories behind.
- Session-cache approval now completes before `connectionBroker.acquire()`. A denied request therefore performs no remote connection/authentication preflight. OS Downloads and Artifact download ordering are unchanged.

Validation: the two focused test files pass (59 tests), along with Node typecheck, focused ESLint, Prettier, and `git diff --check`.

The earlier in-flight cache deletion finding would require explicit per-Session/per-Project operation fencing and draining. I have not introduced that additional runtime lifecycle state in this focused follow-up because it is a broader design decision.

ewen-poch:
Implemented the in-flight cache deletion fix in `8ec77993`.

- `SessionCacheOwner` now registers an active operation before allocating its directory, and the download owner releases it in `finally` after transfer/stat/cleanup completes.
- Session and Project cache deletion now fences new operations, drains the matching active operations, and only then removes the directory.
- The fence is transient process state only; this adds no database field, persisted enum, migration, or historical-data requirement.

The regression tests reproduced the previous behavior at the public download/deletion boundary: both Session and Project deletion settled while the transfer was still blocked, and the download then lost its directory. They now verify that deletion remains blocked until release and that late Session/Project operations are rejected.

Validation: 63 focused tests pass, along with Node typecheck, focused ESLint, Prettier, and `git diff --check`.

ewen-poch:
Addressed the data-root migration gap in e33436e5.

- Registered `compute/` in the existing relocatable data-directory registry. The same registry already drives copy/verification, target adoption and validation, cleanup-journal allowlisting, and old-root source deletion, so Session Cache now follows future data-root moves without a parallel migration path.
- Added `compute-session-cache` to the Project-owned data catalog and locked its existing `SessionCacheOwner.removeProject` cleanup composition with an architecture test.
- Added a public-boundary regression at `runDataRootMigration`: before the fix it failed because `copyAndVerify` received no `compute` directory; it passes after registration.

Compatibility/persistence: this changes the migration policy for existing filesystem bytes under `compute/session-cache`, but adds no schema, persisted field, marker version, or enum state. A data-root move already completed by an older build is not retroactively backfilled; future moves preserve the cache.

Validation: full `src/main/storage` module (16 files / 385 tests), focused Compute and ownership tests (75 tests), Node typecheck, ESLint, Prettier, and `git diff --check` all pass.

ewen-poch:
Addressed the Session Cache/data-root migration race in 1fd22843.

- Wrapped the existing Session Cache allocation/transfer/stat/cleanup operation in the shared `withDataRootWrite` boundary. A migration now drains an already-running download before copying or switching roots, and a pending migration rejects a newly admitted cache write through the existing gate.
- Kept approval and remote connection acquisition outside the lease because neither writes to the data root; the lease begins immediately before `SessionCacheOwner.createOperationFile`.
- Added a public concurrency regression through `waitForDataRootWriters`. Before the fix, the drain resolved while the transfer was blocked; after the fix, it resolves only after the download finishes.

This introduces no new persisted field, schema, enum, owner state, or UI behavior. Validation: 166 focused Compute/migration tests, Node typecheck, ESLint, Prettier, and `git diff --check` pass.

ewen-poch:
Addressed the symlinked Session Cache parent issue in 0e57bc2e.

- Replaced recursive creation below the cache root with component-by-component directory creation and `lstat` verification. Existing symlinks/junctions and non-directory entries are rejected for the cache root, Project, Session, and operation directories before a download path is returned.
- Added cross-platform Project- and Session-parent regressions using symlinks on POSIX and junctions on Windows. Before the fix both cases returned a valid operation path and created UUID directories outside the data root; after the fix both reject and leave the external target empty.

This changes only filesystem validation behavior. It adds no schema, persisted field, enum, owner state, or UI interaction. Validation: 77 focused Session Cache/Remote Operation/migration-state tests, Node typecheck, ESLint, Prettier, and `git diff --check` pass.

ewen-poch:
Rebased onto main at ac6133f8 and addressed the remaining review findings on 0dabe12a.

- Session Cache now validates the compute root, cache root, and every Project/Session/operation component before creation, reconciliation, or recursive cleanup. Symlinked ancestors fail closed, and regressions verify that external files remain untouched.
- Older verified migration markers are rejected when their migrated directory set omits current data that exists in the source root. This preserves the old root and requires the move to be staged again; no marker schema or persisted enum was added.
- Compute Host JSON, shape, and author decoding now fails the Host catalog explicitly instead of collapsing corrupt or unsupported data into missing values. Job feeds and pending-notification recovery continue with provider IDs when Host display names are unavailable.
- SCP cancellation and output-stream failures now have ignored-SIGTERM/no-close regressions. Both tests fail against the pre-fix implementation for the reported unbounded-wait behavior.
- Renderer approval requests are projected from the existing snake_case transport DTO to a camelCase renderer model at enqueue time. Timeout copy now uses locale-correct plural forms, including 1 second.
- Compute module ownership now registers the bounded termination, repository, runner, and Session Cache owners and their tests.

Validation: compute_service 33 files passed and 1 skipped (908 tests passed, 6 skipped); Storage/Session Cache 140 tests; Renderer/i18n 850 tests; module-impact 79 tests; Node/Web typecheck, ESLint, Prettier, and git diff --check all pass.

ewen-poch:
Addressed the reconciliation lifecycle finding in e5d38926.

- Project deletion still records the permanent in-process deletion fence before draining active cache operations.
- Startup orphan reconciliation now calls the directory cleanup path directly, so removing cache for an existing Project with no active Sessions does not mark that Project as deleted.
- Added a public regression that reconciles an orphan-only Project and then creates a later Session cache operation under the same Project. Before the fix, it failed with: Session cache is being deleted and cannot accept new operations.

This adds no schema, persisted field, enum, data-model change, or UI behavior. Validation: the focused Session Cache/Remote Operation suites pass (73 tests), the full compute_service module passes (33 files passed, 1 skipped; 909 tests passed, 6 skipped), and Node typecheck, targeted ESLint, Prettier, and git diff --check pass.

ewen-poch:
Addressed the reconciliation admission race in 42a35db2.

- Added a transient Project-level reconciliation barrier before the active-operation drain snapshot and recursive cache removal.
- New cache operations now fail while that Project is being reconciled. The barrier is always released in finally, so a later Session under the still-existing Project remains usable.
- Explicit Project deletion still uses the existing permanent closedProjects fence.
- Added a deterministic concurrency regression that pauses recursive removal after the drain snapshot. Before the fix, a new operation was admitted; after the fix, it is rejected until reconciliation completes.

This adds only transient in-process state. It adds no persisted field, schema, enum, data-model relationship, or UI change. Validation: the focused Session Cache and Remote Operation suites pass (74 tests), the full compute_service module passes (33 files passed, 1 skipped; 910 tests passed, 6 skipped), and Node typecheck, targeted ESLint, Prettier, and git diff --check pass.

ewen-poch:
Rebased onto latest main at 8a16088e and resolved the Remote Operation conflicts.

- Preserved main PR #1925 three-way download size verification for Session Cache downloads.
- Preserved Project/Session-owned cache allocation, immediate failure cleanup, approval-before-connection ordering, and the data-root writer gate.
- Updated the concurrency tests to provide stable pre/post remote stat results while retaining their deletion-drain and migration-drain assertions.
- Range-diff shows the remaining eight commits replayed equivalently; only the three expected Remote Operation conflict integrations differ.

No additional schema, persisted field, enum, data-model relationship, or UI change was introduced by the conflict resolution.

Validation: compute_service 33 files passed and 1 skipped (927 tests passed, 6 skipped); Renderer/i18n 790 tests; Storage/Session Cache 127 tests; Node/Web typecheck, full ESLint, full Prettier, and git diff --check pass.

ewen-poch:
Addressed both AI review findings in 6fb978a3.

- Storage accounting: compute is now an explicit UsageCategoryKey, compute/session-cache bytes contribute to StorageInfo.totalBytes, and the Storage panel includes a translated Compute cache category with its own design token. The migration inventory already included compute, so no migration scanner change was needed.
- Crash reconciliation: new downloads are written under .partial-<uuid> and atomically renamed to <uuid> only after transfer and stable-size verification. Startup reconciliation removes only .partial-* directories for active Sessions. Existing UUID directories from older versions remain committed and are preserved.

Regression coverage was added at computeStorageUsage, SessionCacheOwner reconciliation, and the session-cache download integration boundary. Focused validation passed: compute_service (928 passed, 6 skipped), storage (387 passed), renderer/i18n (779 passed), Node and Web typechecks, ESLint, and Prettier. No database schema was changed.

ewen-poch:
Addressed the fresh-install cache-root finding in 1df44dda. SessionCacheOwner now creates and validates the configured data root with the existing safe-directory helper before creating dataRoot/compute. A regression test starts with an absent OpenScience root and verifies that Session cache allocation succeeds. Validation passed: compute_service 929 passed / 6 skipped, Node typecheck, ESLint, Prettier, and diff checks.

ewen-poch:
Rebased the branch onto origin/main@f37326f2 and resolved the Storage conflicts by preserving the shared STORAGE_USAGE_CATEGORY_KEYS contract and Notebook evidence accounting while adding the compute category. The rebased series was pushed with an exact force-with-lease; new head is 7f3def4e.

An independent standards/spec review found one high-value cleanup boundary issue: a replaced data root was not revalidated before recursive Session cache deletion. Commit 7f3def4e now validates storageRoot on every destructive/reconciliation path, with a regression that proves a symlinked data root is rejected and external data is retained.

The pre-rebase AI PR Review for 1df44dda completed as mergeable with no findings. Rebase validation passed: compute_service 930 passed / 6 skipped; Storage, Renderer, and i18n conflict tests 780 passed; the isolated Storage migration tests passed; Node and Web typechecks, ESLint, Prettier, and conflict-marker/diff checks passed.

PR #1923, opened 2026-08-30
fix(connectors): harden credential lifecycle

ewen-poch:
## Summary

- capture custom stdio MCP stderr through a bounded pipe (4 KiB per line, 64 KiB per transport) and redact configured credential values plus generic sensitive patterns before logging
- add local OAuth Disconnect and Reauthenticate lifecycle actions; disconnect closes runtime access, clears local OAuth tokens, and disables the Connector without claiming server-side revocation
- make local env edits explicit (`keep` / `replace` / `clear`), show saved variable names without values, and report malformed `KEY=VALUE` lines
- enforce centralized main-process limits for manual Connector counts, strings, args, scopes, env/header records, and credentials before encryption or persistence
- mask credentials with a fixed prefix-free display and harden historical masks at projection time without rewriting settings

## Reproduction and regression coverage

Each regression was first run against the unfixed implementation and failed for the reported reason:

- stdio transport exposed no piped stderr stream
- connected OAuth controls were disabled and `disconnectCustomServer` did not exist
- saved env names/actions were absent and malformed lines were silently dropped
- oversized args, 65 env entries, and a 65th custom Connector were accepted
- short/new and historical masks exposed credential prefixes

The same focused regressions pass after the fix.

## Test impact set

- Connector main owners/contracts: focused Vitest run across `mcp-client-manager`, Connector application/settings/template/workflows/IPC, crypto/provider projection, and settings architecture — **223 passed**
- Renderer/preload/i18n consumers: focused Vitest run across Connector form/panel/store, preload inventory, and locale guards — **910 passed**
- SettingsService OAuth and provider-mask boundaries — **3 passed** (258 unrelated tests skipped by pattern)
- `tsc --noEmit -p tsconfig.node.json --composite false` — **passed**
- `tsc --noEmit -p tsconfig.web.json --composite false` — **passed**
- ESLint on every changed TypeScript/TSX file — **passed**
- `node scripts/generate-web-api-map.mjs --check` — **passed**

These checks ran after the final material edit. Full portable/platform coverage is left to the PR Gate as requested. A diagnostic full run of `service.test.ts` also found seven unrelated Windows symlink fixtures blocked by local `EPERM`; the changed OAuth and mask cases pass independently.

- **Historical compatibility:** unchanged oversized public fields on existing Connectors are grandfathered; replacements must meet the new limits. Historical prefix-revealing `keyMask` values are hardened only when projected and are not bulk-rewritten.
- **New state:** adds only renderer-local `EnvironmentUpdateMode = keep | replace | clear`; no shared or persisted status enum is added.
- **Persistence:** no database/schema/template-version change. Disconnect removes `oauthRef` and stores `enabled: false`; explicit env clear removes env refs; future credential writes store the safer `keyMask` format. Server-side OAuth revoke is intentionally not performed.
- **Data model/API:** adds a renderer-safe optional `environmentNames` projection and a `disconnectCustomServer` IPC command. No persisted data field is added.


ewen-poch:
AI review follow-up for head b12ab26d0: the P1 finding about attaching stderr capture after transport startup does not apply to the locked @modelcontextprotocol/sdk 1.30.0. package-lock.json resolves 1.30.0; that StdioClientTransport constructor creates a PassThrough immediately when stderr is pipe, and its getter explicitly documents that listeners may attach before start() to avoid losing early output. The existing buildTransport regression asserts transport.stderr is non-null before start, and the complete mcp-client-manager suite passes 35/35. Moving capture after start would introduce an early-output race, so no code change is appropriate for this finding.

PR #1925, opened 2026-08-30
fix(compute): harden remote host reliability contracts

ewen-poch:
## Problem

Compute host reliability behavior was inconsistent at four public boundaries:

- SSH-config host creation accepted malformed ports and profile text that password mode rejected.
- a persisted successful Probe was exposed as the current fact `connected` even when old;
- OS Downloads, Artifact imports, and Session Cache downloads could accept size-unstable results;
- saving an empty Scratch root could pin an unrecoverable empty value.

## Proposed change

- Add one shared Host connection profile validator for SSH-config and password host creation, including strict integer ports in `1..65535` and normalized text constraints. Main-process validation remains authoritative.
- Replace the successful Agent summary status `connected` with `last_probe_ok`, and label Settings badges “Last probe succeeded”. Existing transient probing behavior remains unchanged. File browsing attempts the live operation directly instead of using a persisted Probe as a reachability gate.
- Verify every download with `pre-stat -> transfer -> local stat -> post-stat`; require a regular file and `preSize === localSize === postSize`, and clean partial/temp files on mismatch.
- Add an explicit `scratchClear` command through repository, service, Electron/Web contracts, renderer store, and Settings. It atomically writes the existing `{ scratchRoot: null, scratchPinned: false }` transition. Blank Scratch paths are rejected.

## Scope and non-goals

- One PR; Remote Job cancellation is intentionally excluded.
- No database schema change or migration.
- No Probe freshness TTL, heartbeat, background health monitor, or new persisted Probe state.
- No SHA-256 verification. Same-size remote content mutation remains possible and is a follow-up integrity option.

## Model, persistence, interaction, and compatibility impact

- **Enum/API:** `ComputeHostSummary.status` replaces non-persisted `connected` with `last_probe_ok`. In-repo Agent catalog consumers and fixtures are updated; external/internal consumers matching the old string must coordinate this change.
- **Persistence:** no fields are added. SSH validation applies only to new writes; historical host rows remain readable and are not rewritten. Scratch clear updates the existing nullable root and pinned Boolean in one repository operation.
- **Interaction:** invalid SSH-config ports cannot submit; old successful probes no longer claim a live connection; Browse remains available regardless of Probe history and reports the real operation result; pinned Scratch roots show “Restore auto-detection”, including historical empty-pinned rows, and clear failures remain visible.
- **Architecture:** one shared pure profile validator and one new renderer/main `scratchClear` command. Download validation is consolidated in the existing remote-operation owner.

Regression tests were first observed failing on the pre-fix behavior for the reported reasons:

- `22junk` left Add enabled and was parsed as port 22;
- negative/fractional/out-of-range ports and invalid profile text reached repository persistence;
- an invalid password-host display name reached connection validation;
- an old successful Probe rendered/reported `connected`;
- a blank Scratch root could be pinned and no clear operation/UI existed;
- shorter OS/Artifact transfers and a changing Session Cache remote size returned success.

Checks run after the final production-code edit and rebase:

- direct Vitest execution of the exact `compute_service` module plan: **846 passed, 15 skipped** across 32 files (31 passed, 1 skipped);
- Repository/Renderer/Preload/contract/i18n impact set: **971 passed** across 10 files;
- conflict-focused SSH/Scratch/remote-operation set: **238 passed, 5 skipped** across 6 files;
- Node and Web TypeScript checks: passed;
- ESLint on every changed TypeScript/TSX file: passed;
- generated Web API map check and `git diff --check`: passed.

`npm run test:module -- compute_service` generated the expected 32-file selective plan but its `pretest` stopped before Vitest because this isolated worktree has no local generated Prisma client. The same plan was then executed directly with the repository's Vitest runtime as reported above. PR CI remains authoritative for the standard generated-Prisma environment and complete cross-platform checks.

Uncovered risks: no real SSH/SCP endpoint was used locally; stable-size verification does not detect same-size content replacement; consumers outside this repository may still expect `connected`.

## Review focus

- Whether `last_probe_ok` is the right honest Agent contract without introducing TTL semantics.
- Whether the shared profile constraints match both authentication modes without rewriting historical rows.
- Whether all download destinations clean partial data and publish only after three-way size agreement.
- Whether `scratchClear` remains a minimal existing-column transition with no implicit Probe.

### Review follow-up validation

After the Agent-skill documentation follow-up (`3f887ae44`):

- Remote Compute Skill document test: **3 passed**;
- Prettier on the Skill/test, ESLint on the changed test, and `git diff --check`: passed;
- no production code, persisted model, or runtime contract changed in this follow-up.

After the second review follow-up (`cacaa9e1f`):

- Added red-first renderer regressions for an invisible Scratch-clear failure and Browse being gated by persisted Probe state; both failed for the reported reasons before the fix and now pass.
- Full Compute Host Detail and Panel render files: **52 passed**.
- Renderer i18n guard: **743 passed**.
- Web TypeScript, ESLint on changed TS/TSX, Prettier, and `git diff --check`: passed.
- No schema, persisted field, or Probe TTL was added.


ewen-poch:
Addressed the AI Review finding in 3f887ae44: the bundled Remote Compute skill now documents \last_probe_ok\ and explicitly states that it is a persisted Probe outcome, not live connectivity. Added a regression assertion in \skill-doc.test.ts\; the focused file passes 3/3.

ewen-poch:
Addressed both second-round findings in `cacaa9e1f`: Scratch clear errors now remain visible outside edit mode, and Browse no longer treats persisted `last_probe_ok` as a live reachability gate—the file-list operation itself is the current check. Added red-first renderer regressions; both focused tests and all 52 tests in the two render suites pass.

ewen-poch:
Rebased onto latest `origin/main` (`ac6133f81`). Resolved the overlapping Compute application-command test by preserving both upstream automatic-analysis commands and this PR's `scratchClear`; the merged public command count is 32. Follow-up architecture/catalog assertions now reflect the merged 35 Compute contracts. Focused contract tests pass 32/32, the full `compute_service` plan otherwise passed 849 tests with 15 skipped, the Renderer/contract impact set otherwise passed 976 tests, and Node/Web typechecks, ESLint, Prettier, and `git diff --check` pass after regenerating the local Prisma Client (no migration run).

ewen-poch:
Addressed the post-transfer stat finding in 3fd2d20b9. The new public download regression first failed because remoteFsError was undefined after the second SSH stat rejected; the stat boundary now preserves existing remote filesystem errors and AbortError while normalizing raw connection/authentication failures. The focused regression passes, the full compute-remote-operation-owner suite passes 56/56, and Node/Web typechecks, ESLint, Prettier, and git diff --check pass.

ewen-poch:
Addressed the file-browser host-switch finding in 2697983e0. A public interaction regression first showed the unprobed host chip was disabled; host chips now remain selectable and listDir is the live reachability check, while the gray probe indicator still communicates that no successful Probe snapshot exists. The focused test passes, both FileBrowserModal and ComputePanel suites pass 25/25, and Web typecheck, ESLint, Prettier, and git diff --check pass.

ewen-poch:
Addressed the Workspace Remote menu finding in 12b682d0e. The revised public menu regression first failed because the unprobed host carried data-disabled; the menu now opens FileBrowserModal for any configured host and lets listDir determine current reachability. Historical context remains visible as Probe failed or Not probed instead of the inaccurate Host unreachable claim. The focused test passes, the complete ProjectFilesView/FileBrowserModal suites pass 105/105, and Web typecheck, ESLint, Prettier, and git diff --check pass. A renderer-wide search found no remaining probeResult-based browse gates.

ewen-poch:
Follow-up CI fix in a38a55e01: Static checks correctly caught the now-orphaned Host unreachable renderer key in all seven locale catalogs after the Workspace copy was corrected. Removed those obsolete entries; the full i18n guard now passes 743/743 and Prettier plus git diff --check pass.

ewen-poch:
Rebased again onto the latest \origin/main\ at \c90f2780e\ (v0.23.0); the rebase completed without additional conflicts. Also addressed the latest review finding test-first: the new public download regression failed before the fix with \
ot_a_file / Path is not a regular file.\ for a disappeared remote file, then passed after splitting missing paths from existing non-regular paths and routing all download destinations through the shared file-stat classification. Post-rebase verification: remote-operation owner 57/57, Node typecheck, ESLint, Prettier, and \git diff --check\ all pass. No database migration, persisted field, or job lifecycle changes.

PR #1930, opened 2026-08-30
feat(onboarding): default Windows storage to a data drive

ewen-poch:
## Problem

On Windows, first-time onboarding shows the home-based data location, which is commonly on `C:`. Users with a separate data drive must browse away from the system drive. Selecting the drive only at the end of onboarding can also strand a Notebook runtime that was installed earlier against the old root.

## Proposed change

- Main derives `canAutoSelectDataDrive` from the current settings/filesystem snapshot. It is true only when the default data root is unconfigured and no data or managed runtime is present.
- Filesystem eligibility checks distinguish proven absence from permission/I/O errors and fail closed when the result is inconclusive.
- On Windows, eligible onboarding enumerates mounted drives and inspects `D:`, `E:`, `F:`, and later letters in order.
- Preselect the first candidate that is writable and can host a new `OpenScience` folder.
- Skip unusable candidates and drives that already contain Open Science data; adopting historical data remains an explicit Browse action.
- Move Data location directly after Environment, before Agent and Notebook runtime setup.
- Confirming a non-default location persists the existing `dataRoot` setting without completing onboarding, relaunches, and resumes at Agent from the configured non-default root.
- Notebook is now the final step and completes onboarding after any user-requested runtime setup has finished.
- Preserve the current location for legacy installs, configured roots, non-Windows platforms, interrupted setups that already contain runtime data, or when no candidate qualifies.
- Keep Continue disabled while eligible Windows storage information or drive inspection is pending; Browse and Back remain available, and a slow probe cannot overwrite an explicit choice.

## Scope and non-goals

- Adds derived cross-process `StorageInfo.canAutoSelectDataDrive` and `DataRootInspection.targetWasAbsent` booleans; neither is persisted.
- Reorders the existing onboarding steps and reuses the existing `setDataRootAndRelaunch` command.
- No new IPC command, persisted field, database migration, data relationship, or enum value.
- Drive type and capacity are not ranked. The existing mounted-drive API does not distinguish fixed, removable, or network drives.

All local checks below ran after the last material edit:

- Storage eligibility, fail-closed path checks, onboarding flow, Windows selection/fallback, relaunch resume, final completion, i18n guards, and regression coverage -> `node ..\..\node_modules\vitest\vitest.mjs run src/main/storage/path-presence.test.ts src/main/storage/ipc.test.ts src/renderer/src/pages/onboarding src/renderer/src/i18n/resources.test.ts` -> 10 files / 991 tests passed.
- Related host/storage consumers -> `node ..\..\node_modules\vitest\vitest.mjs run src/main/host-application-commands.test.ts src/renderer/src/stores/storage-info-store.test.ts src/renderer/src/pages/settings/StoragePanel.render.test.tsx` -> 3 files / 42 tests passed.
- Main/shared contracts and types -> `node ..\..\node_modules\typescript\bin\tsc --noEmit -p tsconfig.node.json --composite false` -> passed.
- Renderer contracts and types -> `node ..\..\node_modules\typescript\bin\tsc --noEmit -p tsconfig.web.json --composite false` -> passed.
- Repository lint -> `node ..\..\node_modules\eslint\bin\eslint.js --cache .` -> passed with no warnings.

The module-impact manifest has no Onboarding owner entry, so its local affected-test planner falls back to the complete Vitest suite. Per task scope, that full suite is left to PR CI rather than run locally.

## Review focus

- Confirm Storage is applied before Notebook runtime installation and that a non-default root resumes at Agent after relaunch.
- Confirm eligibility fails closed for legacy, non-empty, runtime-containing, or inconclusive roots before renderer drive probing begins.
- Confirm automatically selecting the first usable `D:`-or-later drive is the intended fresh-install Windows default.
- Confirm existing Open Science folders should continue to require explicit adoption.

- Historical compatibility: legacy and already configured data roots are not probed or overwritten, even after the legacy-move prompt was dismissed. Interrupted runtime data stays on its current root, and existing empty/runtime-only candidate roots require explicit Browse selection rather than automatic reuse.
- Resume compatibility: an incomplete onboarding with an existing non-default `dataRoot` resumes after Storage at Agent.
- New states or enum values: two transient cross-process booleans (`canAutoSelectDataDrive` and `targetWasAbsent`) plus one renderer-only boolean tracking probe completion; no enum changes.
- Persistence: no new persisted field. The existing `settings.dataRoot` is written earlier in onboarding, after restart confirmation, while onboarding remains incomplete until the final Notebook step.


ewen-poch:
Good catch. Addressed in 9bfa58122. Main now derives canAutoSelectDataDrive from the current settings and filesystem snapshot, and authorizes automatic drive selection only for an unconfigured root with no relocatable user data. A legacy in-place root therefore remains ineligible even after its move prompt was dismissed. The renderer now requires that main-owned signal before probing drives, with regression coverage for the dismissed legacy-prompt case.

ewen-poch:
Addressed both findings in 92a4b0fae.

- The storage owner now uses async stat checks that return empty only when every managed path is proven absent. Permission, I/O, and encoding errors propagate to the owner boundary, leaving canAutoSelectDataDrive false. Runtime presence is included in the eligibility check, with direct regression coverage.
- Data location now runs immediately after Environment, before Agent and Notebook setup. Confirming a non-default drive persists dataRoot without completing onboarding, relaunches, and resumes at Agent from that root. Notebook is the final step, so any app-managed environment is installed directly under the selected drive instead of being stranded by a later pointer switch.

No new persisted field or enum was added; resume is derived from the existing configured non-default dataRoot.

ewen-poch:
Addressed in ac7a214dd. The recommendation probe now starts only while Location is active, and leaving Location in either direction synchronously freezes the displayed draft before changing steps. The effect cleanup also cancels any in-flight result, so a late drive inspection cannot mutate the shell-owned choice after the step unmounts. Added a regression test that continues with the C: default while the D: probe is pending, resolves the probe afterward, and verifies Back still shows C:.

ewen-poch:
This finding does not apply to the production ownership boundary, so no code change is warranted. DATA_ROOT_DIRS intentionally inventories the relocatable data root. The listed managed agent directories are owned by the fixed config root: production constructs SettingsService from SettingsRepository(resolveStorageRoot()) without a storageRoot override, and AgentRuntimeManager receives that same SettingsService storageRoot. Those installations therefore remain available when settings.dataRoot moves from C: to D:/E:/F:. Adding their names to the data-root eligibility check would probe paths the agent installers do not use and would unnecessarily suppress the intended fresh-install drive recommendation. Notebook environments are the managed runtime that follows dataRoot, and runtime/ is already included in the fail-closed inventory.

ewen-poch:
Addressed the asynchronous default-selection bypass in 27113d2b6. On eligible fresh Windows onboarding, Continue now remains disabled while StorageInfo or the automatic drive inspection is pending, so a fast click cannot preserve C: before the D:/E:/F: recommendation arrives. Browse and Back remain available: an explicit Browse choice cancels the recommendation, while Back cancels the current result and retries when Location is revisited. Storage-info and drive-probe failures still fall back without trapping onboarding, and non-Windows flows do not wait. Added regression coverage for pending StorageInfo, pending drive inspection, and the enabled Browse/Back controls; the final targeted set passes 10 files / 987 tests, with Web typecheck and lint clean.

ewen-poch:
Addressed in dd8fb8f1e. A late non-default StorageInfo response can now resume at Agent only before the user has expressed Location intent. Browse suppresses automatic resume synchronously before opening the directory picker, and leaving Location suppresses it as well, so neither an in-flight picker nor a completed selection can be overridden by delayed storage metadata. A fresh wizard after the intended relaunch still initializes with no interaction intent and resumes at Agent as before. Added a regression test that keeps the picker pending, resolves non-default StorageInfo, verifies Location remains active, then completes the F: selection. The final targeted set passes 10 files / 988 tests; Web typecheck and lint are clean.

ewen-poch:
Addressed in b1ad17fcb. Automatic post-relaunch resume now requires the configured non-default data root to be available. When `dataRootMissing` is true, onboarding retains the normal Environment → Location path instead of jumping to Agent, so the existing Browse control remains available for recovery. Available non-default roots still resume at Agent unchanged. Added regression coverage for a missing persisted E: root; the final targeted set passes 10 files / 989 tests, with Web typecheck and lint clean.

ewen-poch:
Addressed in 4d7cbfe51. The main-owned inspection result now reports `targetWasAbsent` for `move` candidates using the existing fail-closed path-presence check. Automatic Windows selection requires `kind === "move"` and `targetWasAbsent === true`, so an existing empty or runtime-only OpenScience directory is skipped while later drives are still considered. Manual Browse keeps the existing `move` behavior, allowing an explicit user choice of such a target. Added main contract coverage for absent versus runtime-only targets and renderer coverage that skips runtime-only D: in favor of absent E:. The final targeted set passes 10 files / 991 tests, related consumers pass 3 files / 42 tests, and Node/Web typechecks plus lint are clean.

ewen-poch:
Updated the first-run step order to Environment → Data location → Agent runtime → Model provider → Notebook runtime in the numbered setup guide, FAQ, and feature matrix. The English README and all seven localized README copies now match the implemented onboarding flow. This is documentation-only, so the existing code validation remains applicable.

PR #1934, opened 2026-08-30
fix(web-service): bound client caller retention

ewen-poch:
## Summary

- validate Web client nonces as one ASCII token of at most 64 characters on both HTTP and WebSocket boundaries
- cap retained clients at 64 per authenticated principal, evict idle callers by LRU, and release HTTP-only callers after a five-minute idle TTL
- bind HTTP abort/connection-close events to scoped caller leases while preserving callers that still have concurrent requests or WebSockets
- keep the change inside the Web transport adapter; no database fields, migrations, renderer models, or persistent configuration are added

## Reproduction

The regression test sends 65 successful RPC requests for one authenticated principal while rotating valid client nonces. Before the fix, all 65 caller signals remained active, demonstrating that the caller/lease maps grew with every nonce until server shutdown. The test failed consistently on the unmodified implementation and passes after the bounded retention change.

Additional public-boundary tests confirm that the old implementation accepted oversized, malformed, and repeated HTTP client headers, allowed malformed WebSocket client identifiers to upgrade, and retained HTTP-only callers after disconnect or idle expiry.

## Behavior changes

- malformed or repeated HTTP client identifiers return `400`
- malformed WebSocket client identifiers are rejected during upgrade with `400`
- if all 64 client slots for a principal are actively leased, new HTTP RPC calls return `429` and new WebSockets close with `1013`
- disconnected HTTP-only callers are cancelled through the existing caller `AbortSignal`; successful HTTP-only clients retain cross-request ownership for five idle minutes

## Testing

- `npx vitest run src/main/web-service/http-server.test.ts src/main/application-command-client.test.ts src/main/caller-lifecycle.test.ts src/main/caller-context.test.ts` — 63 passed
- `npx eslint src/main/web-service/http-server.ts src/main/web-service/http-server.test.ts` — passed
- full `npm test` reached 21,246 passing tests; 153 unrelated Windows-environment tests failed because this host lacks symlink privileges/Python and the worktree path exceeds existing path-budget fixtures

### DIFF
--- a/src/main/web-service/http-server.test.ts
+++ b/src/main/web-service/http-server.test.ts
@@ -1796,6 +1796,246 @@ describe('startWebHttpServer', () => {
     expect(directSignals[0]?.aborted).toBe(true)
   })
 
+  it('bounds retained caller leases when one principal rotates HTTP client nonces', async () => {
+    const callerSignals: AbortSignal[] = []
+    const remoteInvoke = vi.fn(async (_channel, invocation) => {
+      callerSignals.push(invocation.callerLease.signal)
+      return []
+    })
+    const server = await startTestWebHttpServer({
+      host: '127.0.0.1',
+      port: 0,
+      token: 'local-token',
+      staticRoot: '/unused',
+      rpc: {
+        channels: () => ['projects:list'],
+        invoke: vi.fn(),
+        dispose: vi.fn()
+      },
+      applicationCommands: {
+        localWeb: { commandNames: () => [], invoke: vi.fn() },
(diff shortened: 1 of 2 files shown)


ewen-poch:
Fixed in 4dc381931. The WebSocket capacity-rejection path now removes the socket from externalSockets before closing it with code 1013. Added a regression test that fills the per-principal socket capacity, rejects a second external socket, revokes the principal, and verifies only the retained socket is processed. The complete http-server suite passes (49 tests), along with ESLint and the Node typecheck.

PR #1942, opened 2026-08-30
fix(main-lifecycle): handle confirmed quits and renderer load failures

ewen-poch:
## Problem

- A delegated-work confirmation at the final quit boundary discarded the user's choice. The ordinary-quit race recheck did the same, so choosing **Quit** never re-issued `app.quit()`.
- `BrowserWindow.loadURL()` / `loadFile()` promises were not observed. A rejected renderer entry load could reach Node's fail-fast unhandled-rejection behavior, while runtime `did-fail-load` events only logged and recovery reload rejections bypassed the existing retry budget.

## Proposed change

- Reuse one confirmed-quit helper for the ordinary and delegated recheck paths, and keep an earlier confirmation untrusted when delegated work is already active at the final boundary.
- Return the renderer load promise, observe initial and recovery attempts, destroy an unusable initial window, and route runtime document failures/rejected recovery loads through the existing two-attempt recovery budget and reload/close dialog.
- Avoid double-counting one Chromium failure when both `did-fail-load` and the explicit load promise report it.

## Scope and non-goals

- No database, settings, session, or shared data-model changes.
- No persistence changes, migrations, historical-data compatibility work, or new enum/state values.
- No new UI copy, options, or dialog. The existing renderer recovery dialog can now appear for repeated document-load failures as well as renderer process exits.
- No dependency or lockfile changes.

Red reproduction on the unfixed implementation was stable across repeated runs:

- `vitest run src/main/app-lifecycle.test.ts -t "reissues"` -> 2 failures: the second confirmed choices did not call `quit()`.
- `vitest run src/main/windows.test.ts -t "initial renderer load promise rejects|recovers a main-frame document load failure|counts rejected renderer recovery loads"` -> 3 failures: no startup-window destroy, no runtime reloads, and no recovery dialog.

- delegated quit rechecks and final teardown -> `vitest run src/main/app-lifecycle.test.ts src/main/windows.test.ts src/main/app-startup.test.ts` -> 146/146 passed.
- Main-process TypeScript contracts -> `npm run typecheck:node` -> passed.
- changed source/tests lint -> `npm run lint` -> passed.
- changed-file formatting -> `prettier --check src/main/app-lifecycle.ts src/main/app-lifecycle.test.ts src/main/windows.ts src/main/windows.test.ts` -> passed.

The complete `npm test` suite was intentionally not run per maintainer direction; PR CI remains authoritative for the portable and platform lanes. Local Vitest used the canonical repository dependency tree directly because the isolated worktree has no generated Prisma client and the main checkout has an unrelated dirty `package-lock.json`.

## Review focus

- Whether `quitConfirmed` now authorizes exactly the re-issued quit after the final delegated-work check without weakening the race guard.
- Whether initial-load destruction versus bounded runtime recovery is the right split for renderer load promise failures.
- Whether `rendererLoadsInFlight` correctly prevents a single failure from consuming two recovery slots.

### DIFF
--- a/src/main/app-lifecycle.test.ts
+++ b/src/main/app-lifecycle.test.ts
@@ -1161,6 +1161,44 @@ describe('installAppLifecycle', () => {
     expect(app.exit).toHaveBeenCalledWith(0)
   })
 
+  it('rechecks delegated work added after the delegated confirmation', async () => {
+    let active: ActiveSessionInfo[] = []
+    let resolveFirst: ((choice: CloseConfirmChoice) => void) | undefined
+    const confirmClose = vi.fn(() => {
+      if (confirmClose.mock.calls.length === 1) {
+        return new Promise<CloseConfirmChoice>((resolve) => {
+          resolveFirst = resolve
+        })
+      }
+      return Promise.resolve<CloseConfirmChoice>(
+        confirmClose.mock.calls.length === 2 ? 'quit' : 'cancel'
+      )
+    })
+    const { app, quit, prepareForQuit, flushSessionPersistence, shutdownBackends } = setup({
+      detectActiveSessions: () => active,
+      confirmClose
+    })
+
+    app.emit('before-quit')
+    active = [{ projectId: 'demo', sessionId: 'child-1', kind: 'delegated' }]
+    resolveFirst?.('quit')
+    await flush()
+
+    expect(confirmClose).toHaveBeenCalledTimes(2)
+    expect(quit).toHaveBeenCalledTimes(1)
+
+    active = [...active, { projectId: 'demo', sessionId: 'child-2', kind: 'delegated' }]
+    app.emit('before-quit')
+    await flush()
+
+    expect(confirmClose).toHaveBeenCalledTimes(3)
+    expect(confirmClose).toHaveBeenLastCalledWith('quit', active)
+    expect(prepareForQuit).not.toHaveBeenCalled()
+    expect(flushSessionPersistence).not.toHaveBeenCalled()
+    expect(shutdownBackends).not.toHaveBeenCalled()
+    expect(app.exit).not.toHaveBeenCalled()
+  })
+
   it('rechecks after a saved close preference resolves and safely minimizes for new delegated work', async () => {
     let active: ActiveSessionInfo[] = []
     let resolveSavedPreference: ((choice: CloseConfirmChoice) => void) | undefined
--- a/src/main/app-lifecycle.test.ts
+++ b/src/main/app-lifecycle.test.ts
@@ -1130,6 +1130,37 @@ describe('installAppLifecycle', () => {
     expect(app.exit).not.toHaveBeenCalled()
   })
 
+  it('reissues quit when the delegated work recheck is confirmed', async () => {
+    let active: ActiveSessionInfo[] = []
+    let resolveFirst: ((choice: CloseConfirmChoice) => void) | undefined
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the P1 in e8565d0bf. An initial renderer-load rejection now destroys the startup window only when no recovery attempt has begun; once 
ender-process-gone starts recovery, the superseded initial rejection is logged but leaves the recovery-owned window intact. Added a regression test that starts a replacement load, lets it resolve, then rejects the original load and verifies the window survives. The test failed before the fix and the final focused suite passes 147/147, with node typecheck, lint, and formatting checks also passing. This reuses the existing recovery history and adds no new state or interaction.

ewen-poch:
Addressed the P2 in 260117170. The window now tracks a per-window main-frame navigation generation. Each observed renderer load owns the next expected generation, and its rejection is treated as stale when a newer top-level navigation has advanced beyond it, so an initial load rejection cannot destroy a valid reload. Added a regression test for initial navigation followed by a newer reload, and tightened the unsuperseded failure test to preserve the destroy boundary. The focused suite passes 148/148, with node typecheck, lint, and formatting checks also passing. The generation is transient window-local state only; persistence, data models, and interaction are unchanged.

ewen-poch:
Addressed both actionable findings in 9d548c67c.

- Delegated-work confirmation now records the exact project/session snapshot shown to the user. The final before-quit boundary independently checks for unconfirmed delegated Sessions, so work admitted after either confirmation is blocked and confirmed again; the already-confirmed snapshot still proceeds without a duplicate dialog.
- Explicit renderer loads are now tracked by their expected main-frame navigation generation. did-fail-load is suppressed only when the current generation has a pending Promise owner, so a stale earlier load can no longer hide failure recovery for a newer navigation.

Both regressions were reproduced first with public lifecycle-event tests: the tests failed on the previous head because the third confirmation was skipped and the recovery load was not issued. The focused module suite now passes 134/134, along with Node typecheck, ESLint, Prettier, and git diff --check. No shared data fields, persistence, data model, or user interaction were changed.

ewen-poch:
Addressed the renderer recovery P1 in 443a17f07. Pending explicit loads now carry a window-local handled marker grouped by navigation generation. When render-process-gone owns recovery for the current navigation, or for the next expected generation when the process exits before did-start-navigation, the associated load rejection remains logged but cannot consume another retry slot.

The regression was reproduced first through the existing WebContents lifecycle boundary: the previous head issued a fourth renderer load and opened the crash-loop path after only two actual crashes. Both before-navigation and after-navigation timing cases now stay at three loads with no dialog. The complete windows module passes 72/72, along with Node typecheck, ESLint, Prettier, and git diff --check. This is transient per-window state only; no persistence, shared data model, architecture, or interaction changes.

ewen-poch:
Addressed the superseded-navigation P2 in c39f5be32. Main-frame did-fail-load now treats Chromium ERR_ABORTED as a canceled navigation: it remains logged for diagnostics but does not enter renderer recovery, so a late failure from an old document cannot overwrite the replacement navigation or consume retry budget. Other failures such as NAME_NOT_RESOLVED continue through the existing generation-aware recovery path.

The regression was reproduced first through the WebContents lifecycle boundary: the previous head issued a second renderer load after a newer top-level navigation had already superseded the aborted document. The complete windows module now passes 73/73, along with Node typecheck, ESLint, Prettier, and git diff --check. No new state, persistence, shared data model, architecture, or interaction changes were introduced.

PR #1943, opened 2026-08-30
fix(workspace): surface review and file action failures

ewen-poch:
## Problem

- Manual review requests discarded both rejected promises and `started: false` results, and the button remained clickable until a later `running` event arrived.
- Local Open, Download, Save as artifact, and clipboard failures were silent or console-only.

## Proposed change

- Add a renderer-local synchronous request latch plus transient per-session pending/error state for manual reviews.
- Surface retryable Workspace feedback for rejected and non-started review requests; treat `already-in-flight` as non-failure.
- Reuse `ActionToast` for retryable local-file action failures, while keeping download cancellation silent.
- Add all new renderer copy to the seven translated locale catalogs.

## Scope and non-goals

- No IPC or main-process contract changes.
- No architecture, data-model, persisted-state, or migration changes.
- No new shared or persisted enum values; only component-local renderer state was added.

Pre-fix regression proof:

- Double-clicking Request review called `reviewer.run` twice while the first promise was pending.
- `started: false` and rejected review promises produced no Workspace error.
- Open, Download, Save as artifact, and clipboard failures produced no toast; clipboard rejection was unhandled.

Post-fix checks:

- `npm test -- WorkspacePage.send-gate.test.tsx PreviewFileSurface.test.tsx PreviewFallback.test.tsx` — 74/74 passed.
- `npm test -- src/renderer/src/i18n/resources.test.ts` — 743/743 passed.
- `npm run test:module -- workspace_page` — 621/621 passed.
- `npm run test:module -- project_files_view` — 193/193 passed.
- `npm run lint` — passed.
- Prettier check for all changed files — passed.

Per request, the complete `npm test` suite was not run locally; PR CI is the authoritative full-suite check. Local dependency installation used `npm ci --ignore-scripts` plus Prisma generation because this Windows host lacks the Visual Studio C++ toolchain required by an unrelated native install script.

## Review focus

- The synchronous ref latch closes the pre-render double-click window without changing reviewer IPC semantics.
- `already-in-flight` clears local pending state without showing a false failure.
- OS error detail is shown with a Retry action; user-cancelled Download remains silent.


ewen-poch:
Addressed the review-startup race in 694cb33eb. The existing manual-review pending latch is now folded into a single derived \isReviewBusy\ condition and passed to the conversation controller, so message revision and branch switching are gated before the \
unning\ event arrives. The regression test now asserts both revision availability and the branch-switch guard while \
eviewer.run()\ is pending. Validation: WorkspacePage send-gate 25/25, workspace_page module 621/621, typecheck:web, lint, and Prettier all pass.

ewen-poch:
Addressed in a5b756752. Main now shares the authoritative per-turn start outcome with duplicate callers instead of immediately returning `already-in-flight`. A duplicate remains pending until the owning attempt either broadcasts a running Review and resolves `started:true`, or resolves the same pre-start failure; `runReview` is still invoked only once and the in-flight entry remains until the background run finishes. This does not change the IPC contract, result enum, data model, or persistence. Regression evidence: the two focused tests failed before the fix with immediate `already-in-flight`; after the fix, `src/main/reviewer/ipc.test.ts` passes 36/36 and the reviewer module loaded 448 passing tests.

PR #1945, opened 2026-08-30
fix(lifecycle): harden system shutdown and recovery

ewen-poch:
## Problem

Three P3 lifecycle/recovery paths lacked an observed owner boundary:

- OS shutdown events and headless `SIGTERM`/`SIGINT` could bypass the only bounded application
  shutdown sequence, so Renderer persistence, ACP/Notebook disposal, Web state removal, Remote Access
  invalidation, and log flushing were not attempted consistently.
- A `reachable` renderer network result had no freshness boundary, so focus/visibility recovery could
  keep showing stale green status after Wi-Fi, VPN, proxy, or DNS changes.
- The main-window external navigation handler discarded a rejected `shell.openExternal()` Promise,
  allowing an OS open failure to reach the main process's intentional fail-fast rejection policy.

## Proposed change

- Add a transient `system` shutdown trigger and route Windows `query-session-end`, Linux/macOS
  `powerMonitor.shutdown`, and dedicated-headless `SIGTERM`/`SIGINT` through the existing
  `app.quit()` -> bounded lifecycle owner. Windows `session-end` remains an explicitly best-effort
  fallback because it cannot delay session termination.
- Install those native sources immediately after the primary-instance lock. A one-bit startup relay
  coalesces any request received during database verification/runtime composition and forwards it only
  after the existing migration guard and shutdown owner are ready. If preparation never finishes, a
  five-second startup-only budget falls back to `app.exit(0)` rather than indefinitely delaying OS
  shutdown; the timer is cancelled as soon as the normal owner is available.
- Track the last completed reachability probe time in memory and silently re-probe an aged
  `reachable` result when the renderer regains focus or becomes visible. The TTL is 60 seconds and
  adds no polling timer.
- Catch `shell.openExternal()` rejection at the existing call site and emit only the existing
  fixed-vocabulary diagnostic category.

## Scope and non-goals

- No hidden `BaseWindow` is created for Windows headless mode. Service-manager signals are covered;
  a pure Windows logoff/shutdown of a headless daemon still has no preventable window callback.
- No generic `suspend`/`resume` framework is added. Suspend is not exit, so ACP/Notebook work is not
  destroyed; visible renderer recovery is covered by the focus/visibility TTL.
- No background network polling, UI, copy, or user-flow changes.
- No database, Settings, Session, shared wire-model, data-relationship, or migration changes.
- Historical compatibility: none required. No serialized format is changed.
- New state: `ApplicationShutdownTrigger` gains transient `system`; the probe timestamp is also
  process-memory-only.
- Persistence: no new field or file. Existing Renderer flush and Web state-file removal now run on
  more graceful exit paths.

Each regression was first run twice on the unfixed implementation and failed for the reported reason:

- Windows session end: event was not prevented and shutdown was not requested.
- Stale reachable result: focus produced one probe call instead of the expected second call.
- Rejected external open: no warning was recorded and the Promise had no observed rejection path.
- Headless signals / power shutdown: fake native sources had no registered bridge to `app.quit()`.
- Startup delivery (run twice after review feedback): the startup orchestrator installed the native
  listener callback zero times, so a request during `prepare()` was never queued for the lifecycle owner.
- Blocked startup (run twice after review feedback): after the startup budget elapsed, the force-exit
  callback had been called zero times, confirming that a queued request could hang indefinitely.
- Electron readiness chain (CI plus two local reproductions): `app-branding.test.ts` rejected the
  detached `app.whenReady().then` continuation. The power listener now registers immediately after the
  existing awaited readiness barrier and still precedes database verification/runtime composition.

Final evidence below was collected after the last material edit:

- System/session/signal/power shutdown policy ->
  `vitest run src/main/app-lifecycle.test.ts` -> 67 passed.
- External navigation failure handling -> `vitest run src/main/windows.test.ts` -> 65 passed.
- Reachability TTL/coalescing -> `vitest run src/renderer/src/stores/network-store.test.ts` ->
  21 passed.
- Startup/composition consumers -> `vitest run src/main/application-command-wiring.test.ts
  src/main/app-startup.test.ts` -> 25 passed.
- Electron readiness policy -> `vitest run src/main/app-branding.test.ts` -> 4 passed.
- Combined final targeted run of the six files above -> 182 passed.
- Main process types -> `npm run typecheck:node` -> passed.
- `npm run test:affected -- --base origin/main --head HEAD --explain` classified the lifecycle files
  as an unknown Module owner and therefore selected the full suite. Per maintainer direction, the
  complete local `npm test` was not run; PR Gate and platform lanes are authoritative.

Local delivery-event coverage remains bounded to the documented Electron API boundary; unit tests do
not simulate a real OS shutdown, service manager deadline, or unpreventable forced termination.

## Review focus

- The native handlers should only select the `system` policy and delegate to the existing shutdown
  owner; teardown ordering must remain single-owned.
- System shutdown intentionally bypasses interactive confirmation and conflict cancellation because
  those paths have no reliable interaction budget.
- `session-end` is not described as guaranteed cleanup, and Windows headless does not add a hidden
  native resource.
- Network focus/visibility stays silent, coalesced, and timer-free.
- External-open diagnostics must not retain URLs, OS error text, stacks, or local paths.

### DIFF
--- a/scripts/ci/change-impact.json
+++ b/scripts/ci/change-impact.json
@@ -190,6 +190,7 @@
         "src/main/settings/skill-catalog.ts",
         "src/main/storage-root.ts",
         "src/main/storage/**",
+        "src/main/system-lifecycle-adapters.ts",
         "src/main/tray.ts",
         "src/main/update/create-strategy.ts",
         "src/main/update/electron-updater-strategy.ts",
--- a/scripts/ci/change-impact.json
+++ b/scripts/ci/change-impact.json
@@ -190,7 +190,6 @@
         "src/main/settings/skill-catalog.ts",
         "src/main/storage-root.ts",
         "src/main/storage/**",
-        "src/main/system-lifecycle-adapters.ts",
         "src/main/tray.ts",
         "src/main/update/create-strategy.ts",
         "src/main/update/electron-updater-strategy.ts",
--- a/src/main/app-lifecycle.test.ts
+++ b/src/main/app-lifecycle.test.ts
@@ -56,25 +56,6 @@ const makeFakeApp = (): FakeApp => {
   }
 }
 
-const makeFakeSignalSource = (): NonNullable<AppLifecycleDeps['headlessSignalSource']> & {
-  emit: (signal: 'SIGTERM' | 'SIGINT') => void
-} => {
-  const listeners = new Map<'SIGTERM' | 'SIGINT', () => void>()
-  return {
-    once(signal, listener): void {
-      listeners.set(signal, listener)
-    },
-    removeListener(signal, listener): void {
-      if (listeners.get(signal) === listener) listeners.delete(signal)
-    },
-    emit(signal): void {
-      const listener = listeners.get(signal)
-      listeners.delete(signal)
-      listener?.()
-    }
-  }
-}
-
 type FakeWindow = {
   destroyed: boolean
   minimized: boolean
@@ -163,6 +144,7 @@ type Harness = {
   showMainWindow: () => void
   getMainWindow: () => import('electron').BrowserWindow | undefined
   isMainWindowHidden: () => boolean
+  onSystemShutdown: () => void
   closeOpts: CapturedCloseOpts[]
   confirmClose: ReturnType<typeof vi.fn>
 }
@@ -183,8 +165,6 @@ const setup = (
       | 'isMigrationInProgress'
       | 'platform'
       | 'createInitialWindow'
-      | 'headlessSignalSource'
-      | 'subscribePowerShutdown'
       | 'onAppearanceChanged'
       | 'initialWindow'
       | 'configureMainWindow'
@@ -217,41 +197,40 @@ const setup = (
   )
   const detectActiveSessions = overrides.detectActiveSessions ?? ((): ActiveSessionInfo[] => [])
 
-  const { showMainWindow, getMainWindow, isMainWindowHidden } = installAppLifecycle({
-    app: app as unknown as AppLifecycleDeps['app'],
-    createMainWindow: (opts) => {
-      closeOpts.push(opts)
-      const w = makeFakeWindow()
-      windows.push(w)
-      return asWindow(w)
-    },
-    initialWindow: overrides.initialWindow,
-    configureMainWindow: overrides.configureMainWindow,
-    createTray: (handlers) => {
-      trayHandlers = handlers
-      return tray as unknown as import('electron').Tray | undefined
-    },
-    shutdownBackends,
-    prepareForQuit,
-    abortQuitPreparation,
-    flushSessionPersistence,
-    log: overrides.log,
-    flushLogs: overrides.flushLogs,
-    logFlushTimeoutMs: overrides.logFlushTimeoutMs,
-    rendererFlushTimeoutMs: overrides.rendererFlushTimeoutMs,
-    shutdownTrigger: overrides.shutdownTrigger,
-    isMigrationInProgress: overrides.isMigrationInProgress ?? ((): boolean => false),
-    quit,
-    countWindows: () => windows.filter((w) => !w.destroyed).length,
-    createInitialWindow: overrides.createInitialWindow,
-    headlessSignalSource: overrides.headlessSignalSource,
-    subscribePowerShutdown: overrides.subscribePowerShutdown,
-    onAppearanceChanged: overrides.onAppearanceChanged,
-    platform: overrides.platform ?? 'linux',
-    detectActiveSessions,
-    hasActiveReviewerWork: overrides.hasActiveReviewerWork ?? (() => false),
-    createConfirmClose: () => confirmClose
-  })
+  const { showMainWindow, getMainWindow, isMainWindowHidden, onSystemShutdown } =
+    installAppLifecycle({
+      app: app as unknown as AppLifecycleDeps['app'],
+      createMainWindow: (opts) => {
+        closeOpts.push(opts)
+        const w = makeFakeWindow()
+        windows.push(w)
+        return asWindow(w)
+      },
+      initialWindow: overrides.initialWindow,
+      configureMainWindow: overrides.configureMainWindow,
+      createTray: (handlers) => {
+        trayHandlers = handlers
+        return tray as unknown as import('electron').Tray | undefined
+      },
+      shutdownBackends,
+      prepareForQuit,
+      abortQuitPreparation,
+      flushSessionPersistence,
+      log: overrides.log,
+      flushLogs: overrides.flushLogs,
+      logFlushTimeoutMs: overrides.logFlushTimeoutMs,
+      rendererFlushTimeoutMs: overrides.rendererFlushTimeoutMs,
+      shutdownTrigger: overrides.shutdownTrigger,
+      isMigrationInProgress: overrides.isMigrationInProgress ?? ((): boolean => false),
+      quit,
+      countWindows: () => windows.filter((w) => !w.destroyed).length,
+      createInitialWindow: overrides.createInitialWindow,
+      onAppearanceChanged: overrides.onAppearanceChanged,
+      platform: overrides.platform ?? 'linux',
+      detectActiveSessions,
+      hasActiveReviewerWork: overrides.hasActiveReviewerWork ?? (() => false),
+      createConfirmClose: () => confirmClose
+    })
   return {
     app,
     windows,
@@ -265,6 +244,7 @@ const setup = (
     showMainWindow,
     getMainWindow,
     isMainWindowHidden,
+    onSystemShutdown,
     closeOpts,
     confirmClose
   }
@@ -381,14 +361,11 @@ describe('installAppLifecycle', () => {
     expect(app.exit).toHaveBeenCalledWith(0)
   })
 
-  it('respects Windows session end while routing best-effort cleanup through the shutdown owner', async () => {
-    const { app, windows, shutdownBackends, flushSessionPersistence, quit, confirmClose } = setup({
-      platform: 'win32'
-    })
+  it('routes a system request through the existing shutdown owner', async () => {
+    const { app, onSystemShutdown, shutdownBackends, flushSessionPersistence, quit, confirmClose } =
+      setup()
 
-    const event = windows[0].emit('query-session-end')
-
-    expect(event.defaultPrevented).toBe(false)
+    onSystemShutdown()
     expect(quit).toHaveBeenCalledTimes(1)
 
     app.emit('before-quit')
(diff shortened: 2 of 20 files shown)


ewen-poch:
Addressed the startup-order finding in abcbce96d.

- Added a regression that requests system shutdown from inside prepare(). It failed twice before the
  fix because the early-listener installer was never called (expected one call, observed zero).
- Native sources are now installed immediately after the primary-instance lock. Requests received
  before composition finishes are coalesced and delivered after the migration guard is released and
  the existing lifecycle shutdown owner is installed.
- The startup Window is bound to Windows session-end events as soon as it is created; later windows
  reuse the same idempotent binder. Headless SIGTERM/SIGINT listeners are also installed before
  prepare(), and the powerMonitor continuation is registered before prepare() awaits app readiness.
- Final targeted result after the change: 5 files / 177 tests passed; node/web typechecks and lint pass.

The Windows Headless OS-logoff point is an explicit approved non-goal, rather than an unobserved
implementation omission. Electron documents powerMonitor shutdown for Linux/macOS only, while Windows
query-session-end/session-end are BrowserWindow events. A headless instance intentionally creates no
BrowserWindow, and the approved scope explicitly rejected adding a hidden BaseWindow solely for this
callback. Detached/service-manager SIGTERM and SIGINT are covered. The PR body records this remaining
platform limitation and does not claim guaranteed cleanup for Windows Headless OS logoff/shutdown.

Electron API references:

- https://www.electronjs.org/docs/latest/api/power-monitor#event-shutdown-linux-macos
- https://www.electronjs.org/docs/latest/api/browser-window#event-query-session-end-windows

ewen-poch:
Addressed the blocked-startup finding in e5eb50aaa.

- Added a regression with a prepare() Promise that never resolves. Before the fix it failed twice:
  after the configured budget, forceExit was expected once and observed zero times.
- The startup relay now starts one five-second fallback timer when an owner-less system request is
  queued. Binding the normal lifecycle owner cancels that timer before forwarding the request.
- If preparation stays blocked past the budget, the fallback calls app.exit(0), so a prevented native
  shutdown or consumed signal cannot hold the process indefinitely. It does not add a second cleanup
  sequence; full cleanup remains owned by the existing lifecycle whenever that owner becomes ready.
- Final targeted validation: 5 files / 178 tests passed; node/web typechecks, lint, formatting, and
  git diff checks pass.

ewen-poch:
Resolved the failed macOS shard in b82a9bdc5.

The blob report identified one deterministic failure in app-branding.test.ts: the branch had added a
detached app.whenReady().then continuation, while the startup policy requires Electron readiness to
remain in the awaited startup chain. I reproduced the same assertion twice locally before changing it.

The non-Windows powerMonitor listener now registers immediately after the existing awaited
app.whenReady() barrier. It still precedes database verification and runtime composition, while keeping
readiness failure/ordering inside orchestrateAppStartup's prepare() Promise.

Final local validation after the change: 6 related files / 182 tests passed; node/web typechecks, lint,
Prettier, and git diff checks pass.

ewen-poch:
This remains an intentional scope boundary rather than an unaddressed wiring gap.

Electron exposes Windows logoff/shutdown notifications through `BrowserWindow` session events, while `powerMonitor.shutdown` is limited to Linux and macOS. A true headless launch deliberately creates no window. Covering Windows OS logoff would therefore require either a hidden window solely as an event source or a new native Windows service-control bridge.

We are keeping the approved design: no hidden window and no new native lifecycle subsystem. The common detached/service-manager stop paths (`SIGTERM` and `SIGINT`) route through the existing startup relay and shutdown owner. Windows headless OS logoff/shutdown remains documented as a platform limitation and is not claimed as guaranteed cleanup.

This trade-off adds no persistence, data-model, UI, or protocol changes and avoids expanding the headless runtime architecture for a P3 best-effort guarantee.

ewen-poch:
Addressed the preventable system-shutdown intent finding in 53bf9d53f. Two lifecycle-boundary regressions cover a system request arriving while an ordinary quit aborts on a Renderer conflict and a migration guard temporarily owning the prevented quit. Both tests failed twice before the fix for the reported reasons. The existing systemShutdownRequested latch now records the stronger intent during an in-flight ordinary shutdown, replays quit after an abort, and remains intact until the migration guard reissues quit. This adds no state, persistence, dependency, architecture, data-model, or UI changes. Validation: 6 files / 185 tests passed, along with node/web typechecks, lint, Prettier, and git diff checks. The Windows true-headless OS logoff item remains the previously documented platform boundary: this scope does not add a hidden window or native Windows lifecycle bridge.

ewen-poch:
Addressed the guarded-cancellation leak in 339934459. When the existing shutdown trigger is system-owned, the migration quit guard now skips the Keep waiting / Quit anyway prompt and uses its existing crash-safe cancel-and-join path before reissuing app.quit(). Ordinary user quits retain the existing migration confirmation behavior. The regression failed twice before the fix because confirmQuit was invoked and the migration remained active; it now passes. No enum, persistence, data-model, or data-relationship changes were added. The only interaction change is that an OS-owned shutdown during an active data-root move proceeds without an interactive migration choice. Validation: 7 files / 197 tests passed, plus node/web typechecks, lint, Prettier, and git diff checks.

ewen-poch:
Not adopting this recommendation within this PR. Electron does not expose a process-wide Windows logoff/shutdown event for a true headless process: the documented session-end events belong to BrowserWindow, while powerMonitor shutdown is Linux/macOS-only. Implementing this would require either a hidden window used solely as an event source or a new native Windows service-control bridge, both of which are outside the approved architecture for this P3 change. Detached/service-manager SIGTERM and SIGINT remain covered through the startup relay and existing shutdown owner. The PR continues to document Windows true-headless OS logoff/shutdown as a platform limitation rather than claiming guaranteed cleanup.

ewen-poch:
Rebased onto origin/main at 1f26f4597. The only content conflict was in app-lifecycle.ts with the newly merged confirmed-quit hardening from #1942; the resolution preserves both the upstream delegated-session confirmation boundary and this PR's system-shutdown entry point without changing either state model. Post-rebase validation: 7 files / 209 tests passed, node/web typechecks passed after regenerating the local Prisma client for the new main-branch schema, and lint, Prettier, and git diff checks passed. No generated files were committed.

ewen-poch:
Addressed the Windows session-end finding in 6f8b463b6.

query-session-end no longer calls preventDefault(), so Windows logoff/shutdown is not vetoed. The handler still immediately signals the existing shutdown owner to start best-effort cleanup, and session-end remains the fallback when the query event is not delivered.

This intentionally treats Windows OS-owned termination as best effort: the OS may terminate the process before asynchronous cleanup completes. Replaying the original OS action would require a new native/system-command path and cannot reliably preserve shutdown versus restart intent.

Regression evidence: the updated public event-boundary test failed twice before the fix with "expected true to be false", then passed after the change. Targeted verification: 7 files / 209 tests, Node and Web typechecks, ESLint, Prettier, and git diff --check all pass.

ewen-poch:
No code change for this repeated Windows true-headless finding.

Electron exposes query-session-end and session-end on BrowserWindow, while the powerMonitor shutdown event is not available for Windows session termination. Providing an equivalent source for a true headless Windows process would require either a hidden BrowserWindow used solely for lifecycle events or a native Win32/service-control bridge. Both options expand the approved architecture and were explicitly excluded for this P3 scope.

External Headless stops through SIGTERM and SIGINT remain routed through the startup relay and existing shutdown owner. Windows OS logoff/shutdown in true-headless mode remains an acknowledged platform limitation rather than a guaranteed-cleanup path.

ewen-poch:
The failed PR Gate was caused by retry-only Windows workspace E2E flakes, not by a deterministic branch regression.

- Attempt 1: message-tool-layout-stability briefly measured a 116 px tool-card shift; its Playwright retry passed.
- Attempt 2: workspace-conversation briefly failed to find the previous message after revision navigation; its retry also passed.
- The two attempts failed in different tests, and this PR does not modify either E2E file or the message layout/revision implementation.

No code was changed for these unrelated flakes. A third failed-job rerun completed successfully: Windows E2E passed and the dependent PR Gate passed.

ewen-poch:
Rebased onto the latest origin/main at a8bf5eb10.

The rebase applied all seven PR commits without conflicts. git range-diff reports every rebased commit as content-equivalent to its pre-rebase counterpart.

Post-rebase verification:
- 7 targeted test files / 209 tests passed
- Node and Web typechecks passed after regenerating the local Prisma client from the current schema
- ESLint, Prettier, and git diff --check passed
- no generated or tracked files were added

ewen-poch:
Addressed the independent worktree audit findings with the approved minimal scope:

- Extracted the production OS lifecycle wiring into a directly tested adapter. The tests now exercise the real Windows session-end, Headless SIGTERM/SIGINT, and non-Windows power-shutdown bridges while preserving the existing shutdown owner.
- Added a powerMonitor resume notification for every live Renderer and wired the network store to force a silent reachability probe even when the previous reachable result is still fresh.
- Confirmed the Renderer regression test failed twice before the fix because no resume listener was registered, then passed after the preload contract and monitor subscription were added.

Verification: 12 focused test files / 339 tests passed; Node and Web typechecks passed; renderer contract map, ESLint, Prettier, and git diff checks passed.

This adds no persistence, schema/data-model changes, enum values, historical-data compatibility work, or user-visible interaction changes.

ewen-poch:
Fixed the Static checks failure in 315dbe8ab. The new Electron-only network.onSystemResume contract was correctly exposed by preload, but the pinned GENERATED_SOURCE_OMISSIONS inventory had not been updated. I reproduced the exact renderer-surface-inventory assertion locally, added the missing path, and reran the focused interface-contract suite: 5 files / 130 tests passed. check:web-api-map, ESLint, Prettier, and git diff checks also pass. No runtime behavior changed in this follow-up.

ewen-poch:
Addressed both network-resume findings in 16b7ab85e.

- The resume callback now refreshes navigator.onLine through the store's existing recheckOnline boundary before scheduling the forced silent probe. The regression starts with cached isOnline=false, restores navigator connectivity without an online event, and verifies resume reaches reachable.
- A forced resume request that arrives while another silent probe is in flight is now retained and replayed once from the current probe's finally path. Multiple overlapping requests remain coalesced.

Both regressions failed twice before the fix with the reported symptoms (cached offline state unchanged; expected three probes but observed two). Final focused validation: 13 files / 345 tests passed, plus Web typecheck, check:web-api-map, ESLint, Prettier, and git diff checks.

No persistence, data-model, enum, or user-interaction changes were added. The repeated Windows true-headless recommendation is not adopted: it still requires the previously excluded hidden window or native Windows lifecycle bridge and remains the documented platform limitation.

ewen-poch:
Fixed the deterministic macOS shard failure in 8dfd6ef08. The new src/main/system-lifecycle-adapters.ts contains an explicit win32 runtime branch, so the classifier coverage test correctly required it to be listed in the existing windows_sensitive overlay. The manifest now routes future changes to this adapter through the established Windows runtime/path and functional/workspace E2E lanes.

Focused validation: classify-pr-changes and PR Gate workflow tests passed (2 files / 85 tests), with Prettier and git diff checks passing.

The Windows workspace E2E failure was unrelated: message-tool-layout-stability measured the known transient 116 px shift on its first attempt and passed on retry; --fail-on-flaky-tests then failed the job. No product code was changed for that flake. The coverage and PR Gate failures were downstream of the deterministic shard failure.

ewen-poch:
Superseded the protected-manifest change in 8d02a2a42 after CI Integrity correctly rejected ordinary PR edits to scripts/ci/change-impact.json.

The platform decision now remains in src/main/index.ts, which is already covered by the existing windows_sensitive overlay. The lifecycle adapter receives two explicit capabilities (window session-end events and power-shutdown events) instead of inspecting process.platform itself. Runtime behavior is unchanged, and the protected change-impact manifest has no net diff from main.

Regression evidence: with the manifest restored and before the capability seam changed, classify-pr-changes failed twice because system-lifecycle-adapters.ts was uncovered. After moving the platform policy to index.ts, the classifier and lifecycle set passed (5 files / 160 tests). Node typecheck, ESLint, Prettier, and git diff checks pass. The local CI Integrity command now reports pass with no CI-sensitive files changed.

ewen-poch:
Addressed the startup-ordering finding. The existing awaited app readiness step and power-monitor listener installation now run before startup shell module imports, so a shutdown event cannot be missed while those imports are pending. I kept readiness in the startup promise chain (no detached continuation) and added a source-order regression test that failed consistently before the change and passes afterward. Focused startup/lifecycle tests: 32 passed; rebased changed-module suite: 297 passed.

PR #1952, opened 2026-08-31
fix(figure-composer): harden delegated figure workflow

wen2zhou:
## Problem

The bundled Figure Composer skill left important orchestration rules implicit. A run could treat the latest completed child as the whole result, accept missing structured output, compose in the main agent, skip formal review, or return a stale composite after a retry. The existing tests did not launch the packaged Electron application, so those failures were not covered end to end.

## Proposed change

- Define exact pixel geometry and a complete per-panel task contract.
- Wait on the same delegated handles, validate structured panel evidence, and keep current panel artifact versions explicit.
- Require independent compositor and reviewer Attempts, with one to three complete compose/review rounds.
- On revision, redelegate a fresh full panel task and compose only from the current accepted panel versions.
- Add focused Python-kernel coverage and an opt-in real-Electron Playwright gate that inspects persisted Attempts, structured output, final artifact lineage, and the root response.
- Make the Electron fixture select its provider through a discriminated provider configuration instead of a display string.

## Scope and non-goals

This PR changes only the Figure Composer skill, its kernel helpers/tests, and the Electron E2E fixture/gate. It does not change OpenCode behavior, generic delegation authentication, or generic ACP process shutdown; those application concerns are isolated in separate PRs.

- Exact geometry, wait semantics, required review, retry freshness, and final-return contract: `npx vitest run resources/skills/figure-composer/kernel.test.ts` — 4/4 passed.
- Real packaged workflow: built the E2E application and ran `e2e/figure-composer-live.spec.ts` with the locally configured Codex subscription provider and `gpt-5.6-luna` — passed in 40.0 minutes.
  - The run completed three review rounds and 12 delegate frames: two panels, an independent compositor, and an independent reviewer per round.
  - Review requested revisions twice; both retries produced fresh panel and composite versions.
  - The final reviewer accepted the third composite, and the persisted lineage and root response matched the accepted artifact.
  - No delegated frame failed and no `acp connection closed` occurred.
- `npm run lint` — passed.
- `npm test` — 1291 files / 21706 tests passed; 16 files / 238 tests skipped.
- Independent Standards and Spec review after fixes: no remaining P1/P2 findings.

Uncovered risk: the opt-in live gate was exercised on macOS with Codex subscription authentication. Other operating systems and providers remain covered by the normal unit/integration suite rather than this local live run.

## Review focus

- Whether the prose contract is strict enough to prevent main-agent composition and skipped review without over-prescribing model behavior.
- Whether retry freshness and final composite provenance are enforced at the correct durable Attempt/artifact boundary.
- Whether the 60-minute opt-in E2E gate is isolated enough from the default test suite.


wen2zhou:
fixed by https://github.com/aipoch/open-science/pull/2981

PR #1963, opened 2026-08-31
feat(connectors): add structured credential fields

ewen-poch:
## Summary

- add a shared Fields/Text editor for local environment variables and remote static headers
- bind each name directly to a device Credential and auto-select newly created Credentials
- preserve bulk text entry, validation, edit-time keep/replace/clear behavior, responsive layout, and translated copy

## Impact

- Historical compatibility: no migration is required; existing saved environment and header names, encrypted values, and keep/replace/clear behavior remain supported
- New state: adds only renderer-local `fields | text` view state; no shared or persisted enum is added
- Persistence: request shapes, database schema, IPC contracts, and stored Credential references are unchanged

## Test plan

- `npx vitest run src/renderer/src/pages/settings/ConnectorAddForm.render.test.tsx src/renderer/src/pages/settings/ConnectorNamedCredentialEditor.render.test.tsx src/renderer/src/i18n/resources.test.ts --silent=passed-only`
- `npm run typecheck:web`
- targeted ESLint for changed renderer files
- targeted Prettier check for changed renderer and locale files
- Hallmark 8-state visual review at desktop and 320px; no horizontal overflow


ewen-poch:
Addressed the review findings in 011fba3e:

- Field names are trimmed before source updates and credential-binding moves, so parser and binding keys stay canonical.
- Unbound credential selectors now receive an explicit empty value, preventing stale selections when rows are renamed or removed.
- Fields mode now surfaces validation for malformed non-empty-name source lines such as =secret and : value, while exact blank draft rows remain neutral during entry.

Focused verification: 899 renderer and i18n tests passed, along with web typecheck, targeted ESLint, Prettier, and diff checks.

PR #1965, opened 2026-08-31
fix(concurrency): fence stale operations and terminal races

ewen-poch:
## Problem

The repository-wide concurrency audit found 20 reachable race classes across Main owners, renderer projections, authentication, updates, uploads, Notebook, Reviewer, Compute, Projects, and multi-client transports. This PR implements the first migration-free batch: issues where a stale completion, duplicate live request, or cancel/finish race could be fixed without choosing a durable idempotency or historical-data policy.

## Proposed change

- Serialize or coalesce duplicate live side effects for completion handoff retries, Reviewer fix loops, update apply, and upload terminalization.
- Add post-approval cancellation fencing to Compute submission and preserve all concurrently active Notebook runs with deterministic cwd ownership.
- Make Settings mutations publish through one ordered Main snapshot owner across Electron IPC, application commands, Web RPC, and native close confirmation. Add an optional volatile snapshot revision so older RPC responses cannot overwrite newer events.
- Serialize isolated Codex logout with account mutations and reject stale login finalization after edit, delete, logout, or cancellation. Apply targeted validation patches instead of replacing a stale provider record.
- Add per-resource/per-intent generations for Project and Compute stores, local-file navigation, Session details/export, and ACP action metadata.
- Add deterministic regression tests for every changed race. These tests reproduced the pre-fix interleavings during the red/green inner loop.
- Let explicit Vitest shards tolerate an empty per-project partition while preserving the normal no-tests failure outside `--shard` runs.

## Scope and non-goals

This first batch intentionally defers changes that need a product decision, durable identity, CAS contract, or historical compatibility policy:

- Web RPC mutation idempotency and restart-safe receipts.
- End-to-end durable Compute submit identity.
- Granted-folder access CAS and cross-client events.
- Crash-safe upload completion replay.
- Project pin CAS, Tags revision admission, atomic bookmark commands, and User Skill edit conflicts.
- Stable operation identity for app-generated Artifacts and Project creation.

Existing protections were confirmed for Session autosave/revision rebase, Project Files/preview queues, ordinary ACP prompts across all four framework paths, permission/plan/elicitation/delegation owners, Compute cancel/finish, remote access/OAuth generation fencing, and Artifact MCP writes.

Compatibility and state impact:

- Historical data: no migration and no stored-format change. Older snapshots may omit `revision`; the renderer accepts them until it has observed a revisioned Main snapshot.
- New state: no enum values. New state is process-local only: promise owners, generations, refcounts, active-run sets, and `SettingsSnapshot.revision`.
- Persistence: no database schema, migration, or new persisted field. Existing Settings provider records are updated through targeted serialized mutations; upload/update arbitration remains process-local.

All listed checks ran after the final material edit:

- Rebased direct race contracts outside the full Settings module -> focused Vitest set -> 16 files, 430 tests passed.
- Settings owner, repository, IPC, application-command, and consumer coverage -> `npx vitest run src/main/settings` with loopback enabled -> 93 files passed, 1 skipped; 1720 tests passed, 4 skipped.
- Module ownership and trusted CI contracts -> focused CI suite -> 4 files, 80 tests passed.
- Vitest shard/config contract -> red test reproduced the empty-project exit; rebased direct suite passed; latest-base CI Integrity check passed.
- Node and renderer interfaces -> `npm run typecheck` -> passed.
- Source quality -> `npm run lint`, Prettier check, and `git diff --check` -> passed.
- Independent read-only review -> confirmed final fixes and found no remaining regression in this batch.

The final impact classifier selects the full PR Gate because `scripts/ci/module-impact.json` is a global gate input. Per maintainer direction, the complete local `npm test` was not run; exact-head PR CI is the authority for the full suite and platform lanes.

## Review focus

- Ordering and failure recovery in `SettingsSnapshotCommitOwner`, especially event-versus-RPC settlement.
- The non-persisted optional revision compatibility rule.
- Terminal claims/coalescing in completion handoff, update apply, and upload finish/abort.
- Per-resource generations in renderer stores and intent fencing in Session/local-file flows.
- Provider login/logout identity checks and preservation of concurrent provider edits.

### DIFF
--- a/scripts/ci/module-impact.json
+++ b/scripts/ci/module-impact.json
@@ -617,6 +617,7 @@
       "ownerPaths": [
         "src/main/settings/service.ts",
         "src/main/settings/network-proxy-settings-owner.ts",
+        "src/main/settings/settings-snapshot-commit-owner.ts",
         "src/main/settings/reviewer-model-owner.ts",
         "src/main/settings/subagent-model-owner.ts",
         "src/main/settings/vision-model-owner.ts"
@@ -632,6 +633,7 @@
       "testFiles": {
         "owner": [
           "src/main/settings/settings-backend.architecture.test.ts",
+          "src/main/settings/settings-snapshot-commit-owner.test.ts",
           "src/main/settings/service.test.ts",
           "src/main/settings/service.providers.test.ts",
           "src/main/settings/service.connectors.test.ts"
--- a/src/main/agents/completion-handoff-lifecycle.test.ts
+++ b/src/main/agents/completion-handoff-lifecycle.test.ts
@@ -275,6 +275,27 @@ describe('CompletionHandoffLifecycle', () => {
     expect(await repository.get(context)).toMatchObject({ stage: 'continued' })
   })
 
+  it('joins concurrent retries so handoff side effects run exactly once', async () => {
+    const repository = new InMemoryCompletionHandoffRepository()
+    const { runtime } = createRuntime({
+      reconfigure: vi.fn().mockRejectedValueOnce(new Error('target unavailable'))
+    })
+    const lifecycle = new CompletionHandoffLifecycle(repository, runtime)
+
+    await lifecycle.approve({ context, targetName: 'Approved Specialist', generation: 1 })
+    await lifecycle.capture(context, { kind: 'returned', value: 'captured' })
+    await expect(lifecycle.run(context)).resolves.toMatchObject({ stage: 'failed' })
+
+    await expect(
+      Promise.all([lifecycle.retry(context), lifecycle.retry(context)])
+    ).resolves.toEqual([
+      expect.objectContaining({ stage: 'continued' }),
+      expect.objectContaining({ stage: 'continued' })
+    ])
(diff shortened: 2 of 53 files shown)


ewen-poch:
Addressed the actionable synchronization findings in `5cda84d0`:

- Compute-host and Project projections now advance ownership only after a mutation commits. A failed newer mutation therefore cannot suppress an older committed result, while a successfully committed newer mutation still fences stale projections.
- Both the application-command and legacy IPC paths for `settings:set-notebook-network` now commit through `SettingsSnapshotCommitOwner.projectAfter`, preserving the Notebook network return value while publishing the authoritative revisioned snapshot.
- Active and queued reviewer fix loops now register their abort capability before lifecycle serialization. Cancellation aborts every admitted loop for the session, and a queued loop that was cancelled while waiting does not emit a false start/end lifecycle pair.

Added regression tests that failed against the previous implementation. Focused validation passes: 5 files / 161 tests, TypeScript typecheck, ESLint, Prettier, diff checks, and CI Integrity.

PR #1972, opened 2026-08-31
fix(onboarding): bound Windows storage recommendation

ewen-poch:
## Problem

On fresh Windows onboarding, automatic data-drive recommendation can remain pending indefinitely when drive enumeration or candidate inspection reaches a disconnected mapped, network, or sleeping drive. Location treated that opportunistic recommendation as a hard Continue gate, so first launch could not proceed even though the default path was already available.

## Proposed change

- Bound the whole Windows storage recommendation attempt to 2 seconds and keep the loaded default location on timeout or failure.
- Keep Continue disabled while Windows storage information itself is loading, but allow it while only the optional recommendation is pending; preserve the existing non-Windows behavior.
- Connect the recommendation to the Location effect lifecycle so leaving stops renderer-side waiting and prevents a late drive list from starting candidate inspections.
- Mark the automatic attempt resolved on step exit so Back/return loops do not accumulate uncancellable IPC requests.
- Ignore late recommendation results after the user continues or chooses a location.

## Scope and non-goals

- No Windows drive-type classification; fixed, removable, network, and optical drive metadata remain outside this PR.
- No concurrent or per-drive probing.
- No IPC, shared type, database, persisted settings, historical-data compatibility, enum, or i18n changes.
- An already-issued Electron IPC/filesystem request cannot be cancelled by the renderer. This PR limits it to one attempt per wizard, prevents follow-on candidate checks after exit, and removes all onboarding dependency on its completion.

The final regression tests were run red before their fixes. They reproduced the reported pending drive-list and candidate-inspection paths, plus review-found cross-platform gating and stale follow-on probing.

- pending drive listing remains manually continuable; pending candidate inspection falls back to the default; leaving prevents follow-on probes and retries; successful recommendation, Windows storage-info loading, and non-Windows behavior remain intact -> npm test -- src/renderer/src/pages/onboarding/OnboardingWizard.render.test.tsx src/renderer/src/pages/onboarding/LocationStep.render.test.tsx -> 47 passed
- renderer contracts and lifecycle signal usage -> npm run typecheck:web -> passed
- source lint -> npm run lint -> passed
- changed-file formatting and whitespace -> Prettier check plus git diff --check -> passed

The affected-test planner selected the complete Vitest fallback because onboarding has no module owner. Per maintainer direction, the complete local npm test run was intentionally left to PR CI.

## Review focus

- Whether a 2-second opportunistic recommendation budget is the right UX tradeoff.
- The separation between mandatory Windows storage-info loading and optional drive recommendation.
- The lifecycle boundary: already-issued IPC is not cancellable, but cannot trigger later candidate checks or accumulate on step re-entry.
- The explicit non-goal of identifying fixed local disks in this focused P2 fix.

### DIFF
--- a/src/renderer/src/pages/onboarding/LocationStep.render.test.tsx
+++ b/src/renderer/src/pages/onboarding/LocationStep.render.test.tsx
@@ -91,10 +91,10 @@ describe('LocationStep', () => {
     expect(onBack).toHaveBeenCalledOnce()
   })
 
-  it('disables Continue, but not Back or Browse, while resolving the Windows default', async () => {
+  it('keeps navigation available while resolving the optional Windows default', async () => {
     await renderStep(true)
 
-    expect(findButton(/^continue$/i)?.disabled).toBe(true)
+    expect(findButton(/^continue$/i)?.disabled).toBe(false)
     expect(findButton(/back/i)?.disabled).toBe(false)
     expect(findButton(/browse/i)?.disabled).toBe(false)
   })
--- a/src/renderer/src/pages/onboarding/LocationStep.tsx
+++ b/src/renderer/src/pages/onboarding/LocationStep.tsx
@@ -68,7 +68,7 @@ const LocationStep = ({
 }: LocationStepProps): React.JSX.Element => {
   const { t } = useTranslation()
   const { chosenParent, chosenDataRoot, chosenKind } = locationDraft
-  const isLoadingDefaultLocation = dataRootInfo === null && dataRootError === undefined
+  const isLoadingDefaultLocation = isResolvingDefaultLocation && dataRootInfo === null
   const [locationError, setLocationError] = useState<string | undefined>(undefined)
   const [confirmRestart, setConfirmRestart] = useState(false)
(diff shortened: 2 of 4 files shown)


ewen-poch:
Addressed the review in 341bd9f26:

- P1: restored the prior non-Windows behavior by applying the storage-info loading gate only when the wizard reports the Windows default-resolution state. Added a red/green macOS pending-storage-info regression test.
- P2: connected the recommendation to the Location effect lifecycle. Leaving now resolves the renderer wait immediately, aborts before any later candidate boundary, marks the single automatic attempt complete, and prevents Back/return loops from launching another pending IPC. Added a red/green test proving a late drive list cannot call inspectDataRoot and re-entry does not call listDrives again.

Electron invoke itself remains non-cancellable without a new IPC protocol; this focused PR now limits that residual request to one, prevents downstream probes, and makes onboarding independent of its settlement. The PR scope and non-goals have been updated accordingly. Final targeted tests: 47 passed; typecheck:web, lint, Prettier, and diff checks passed.

PR #1981, opened 2026-08-31
fix(notebook): honor injected runtime platform

ewen-poch:
## Problem

Notebook owners already accept injected platform/environment values, but downstream helpers sometimes read `process.platform`, `process.env`, or `homedir()` again. An injected Windows policy on a macOS test host can therefore combine Windows decisions with POSIX interpreter layouts, provenance rules, recovery guidance, or managed-runtime path checks.

The failure was reproduced before the fix with opposite-host platform injections at public behavior boundaries: provisioner recovery guidance and interpreter verification, runtime discovery candidates/provenance, and the managed-runtime mutation guard all failed for the reported mixed-decision reason.

## Proposed change

- Thread the existing platform policy through provisioner, runtime-path readiness/layout helpers, kernel execution, relocation, runtime service admission, and execution mutation checks.
- Let runtime discovery use one injected platform, process environment, and home directory for command lookup, standard install roots, interpreter layout, probes, and provenance classification.
- Use the injected platform's `node:path` implementation for managed-runtime guard normalization and containment.
- Keep production defaults backward-compatible by falling back to the real process when no dependency is injected.

## Scope and non-goals

- This is dependency propagation, not a new application architecture layer. It intentionally does not add a broad `RuntimeEnvironment` object containing packaged/resources/temp/shell state because the reproduced failures do not require that interface.
- `micromamba-cache` host ACL/PowerShell adapters continue to use the real host process. They perform host I/O and must not execute Windows security commands merely because a test injects `win32`; its logical cache selection paths already accept injected platform/environment dependencies.
- No user interaction, persisted data, schema, data model, historical migration, or enum/status value changes.

The following checks ran after the last material edit and after rebasing onto the latest `origin/main`:

- Provisioning, discovery, runtime path readiness, and relocation behavior -> `npm test -- src/main/notebook/provisioner.test.ts src/main/notebook/environment-discovery.test.ts src/main/notebook/runtime-paths.test.ts src/main/notebook/runtime-relocation.test.ts` -> 161 passed.
- Mutation policy plus kernel/runtime-service consumers under injected platforms -> `npm test -- src/main/notebook/managed-runtime-guard.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/runtime-service.test.ts -t 'uses the injected platform|drops and respawns the kernel when Windows cancellation|uses the PowerShell runtime-write policy on Windows|activates the complete Windows conda PATH|carries an external Windows conda R own activation prefix|does not infer Windows conda activation'` -> 7 passed.
- Formatting/diff integrity -> Prettier check and `git diff --check` -> passed.

Per the requested workflow, the full portable `npm test` was not run locally; exact-head PR Gate and platform lanes are authoritative. The local sandbox also denies a loopback-listen kernel test and one macOS Seatbelt cache-write test; both failures reproduce unchanged on `main`, while the focused regression and consumer tests pass.

## Review focus

- Confirm every changed decision uses the owner-injected platform/environment consistently through nested helpers.
- Confirm real host-only micromamba ACL/PowerShell adapters remain outside the simulated platform policy.
- Confirm the optional parameters preserve existing production callers and do not imply persistence or compatibility work.

### DIFF
--- a/src/main/notebook/data-execution-admission.ts
+++ b/src/main/notebook/data-execution-admission.ts
@@ -46,6 +46,7 @@ type NotebookDataExecutionAdmissionOwnerOptions = {
   ensureRecovered: () => Promise<void>
   resolveRuntimeEnablement: (language: NotebookLanguage) => Promise<RuntimeEnablement | undefined>
   repairPolicy: Pick<NotebookRuntimeRepairPolicy, 'blockKey' | 'requirement'>
+  platform?: NodeJS.Platform
 }
 
 const defaultEnvironment = (language: NotebookLanguage): string =>
@@ -132,7 +133,9 @@ class NotebookDataExecutionAdmissionOwner {
     const recoveryBlocked =
       (binding?.runtimeId && this.options.recovery.isRuntimeIdBlocked(binding.runtimeId)) ||
       (!isExternal &&
-        this.options.recovery.isPrefixBlocked(envPrefix(runtimeRoot, route.environment))) ||
+        this.options.recovery.isPrefixBlocked(
+          envPrefix(runtimeRoot, route.environment, this.options.platform)
+        )) ||
       (isExternal && this.options.recovery.isGloballyBlocked())
     const repairRequired =
       this.options.environmentOperations.isRepairBlocked(
@@ -173,7 +176,8 @@ class NotebookDataExecutionAdmissionOwner {
       source: cell.code,
       surface: cell.language,
       runtimeRoot: session.runtimeRoot,
-      cwd: session.cwd
+      cwd: session.cwd,
+      platform: this.options.platform
     })
     if (blockedMutation && rejection === undefined) {
       rejection = new Error(`MANAGED_RUNTIME_MUTATION_BLOCKED: ${blockedMutation.message}`)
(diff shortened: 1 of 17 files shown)


ewen-poch:
This finding targets `SessionMessageLink.tsx`, which is unchanged by this PR and was introduced on `main` in #1524. PR #1981 is limited to Notebook runtime-environment injection. Applying the recommendation here would expand the scope into renderer network policy and alter the established source-preview interaction, so no change is included in this PR. The favicon and preview admission behavior should be evaluated separately against the intended source-preview design.

ewen-poch:
Addressed two in-scope runtime-injection gaps found by the independent review:

- Interpreter discovery now uses the injected subprocess port and environment for Python probes, and preserves the injected environment for external R probes.
- The environment startup lifecycle now forwards its injected platform to startup planning and default-prefix detection, preventing Windows simulations on non-Windows hosts from mixing path layouts.

Both fixes were added test-first. The discovery regression failed before the fix because the injected probe was bypassed and returned no version; the startup regression failed because a ready Windows runtime incorrectly triggered Python repair. Both pass after the changes.

Validation: 201 targeted Notebook tests passed, Node typecheck passed, lint passed, and formatting/diff checks passed. These changes only extend existing internal dependency propagation; they add no persisted data, status values, data-model changes, architecture-owner changes, or user-interaction changes.

ewen-poch:
Rebased onto the latest main (156a9166). I also inspected the previous CI failures before rerunning: the Ubuntu shard failed in remote-access temp-directory cleanup with ENOTEMPTY, and the macOS workspace journey was a retry-passed flaky timeout. Neither failure is in this PR's diff, and the exact remote-access test passes locally (19/19). The rebased Notebook scope passes 421 targeted tests with 1 existing skip, plus Node typecheck, ESLint, and Prettier. I did not broaden this PR into unrelated flaky-test changes; the new CI run will provide a clean latest-main signal.

PR #1983, opened 2026-08-31
fix(message-links): gate remote loads by network policy

ewen-poch:
## Problem

Session message links rendered a remote favicon for every HTTP or HTTPS hostname before any user activation. Any model-produced link could therefore disclose network metadata to its domain. HTTPS source activation also created the in-app remote preview without consulting Settings > Network > Allowed domains.

The reported 350 ms hover-to-iframe behavior does not occur on current main: hover opens only a locally composed title, hostname, and URL summary. The regression coverage keeps that boundary explicit.

## Proposed change

- Reuse the effective Notebook network domain policy, including enabled Open Science domain groups and custom Allowed domains.
- Keep the local globe icon for every HTTP or HTTPS link, and load a remote favicon only for an admitted hostname.
- Keep hover and keyboard-focus details local without activating the preview workbench.
- Allow admitted HTTPS sources to use the existing in-app preview.
- Route unadmitted HTTPS source activation through the existing external-link safety confirmation instead of creating a remote iframe.
- Update the renderer design contract to document the shared policy boundary.

## Scope and non-goals

- No new settings fields, enum values, IPC methods, dependencies, or persisted data.
- No schema migration or historical-data conversion. Existing messages are evaluated against the current Network settings when rendered.
- No per-link authorization state or automatic mutation of Allowed domains.
- No change to the explicit Open source in browser action.
- No iframe fix for hover because current main does not create an iframe on hover.

The regression tests were first run against the unmodified implementation and failed deterministically twice:

- an unapproved hostname still inserted https://tracking.example/favicon.ico;
- activating an unapproved HTTPS source did not show the safety confirmation.

Final evidence after the last production edit:

- favicon and source-preview policy behavior, markdown title propagation, and shared domain matching -> npm test -- src/renderer/src/components/streamdown/SessionMessageLink.test.tsx src/renderer/src/components/streamdown/AgentMarkdown.test.tsx src/renderer/src/pages/workspace/SessionMessageMarkdown.integration.test.tsx src/shared/notebook-network.test.ts -> 4 files and 35 tests passed
- renderer types -> npm run typecheck:web -> passed
- lint -> npm run lint -> passed
- impact explanation -> npm run test:affected:explain -- --base main --head HEAD -> selected full mode only because docs/design.md has no module owner

The complete local npm test suite was intentionally not run per maintainer direction. PR Gate remains authoritative for the complete portable and platform-selected suites.

Electron and Web use the same Renderer component and policy helper, so the changed behavior is shared. No main-process, OS-specific, database, or transport path changed.

## Review focus

- Confirm that enabled Open Science domain groups and custom Allowed domains are the intended trust boundary for message-link favicons and in-app source previews.
- Confirm that disabled built-in groups or domains remain excluded through notebookNetworkSettingsAllowDomain.
- Confirm that unadmitted source links retain the local hover summary and existing external-link safety interaction without entering the preview workbench.


ewen-poch:
Not changing this behavior. The external-icon action is an explicit user gesture labeled "Open source in browser", so it is not a passive or incidental network load. This PR gates automatic favicon requests and in-app iframe creation behind the effective Allowed domains policy, while unapproved source-link and hover-URL activation continue through LinkSafetyModal. Adding a second confirmation to the explicitly external action would add interaction friction and conflicts with the scoped non-goal documented in the PR. No code change is needed for this finding.

Issue #1776, opened 2026-08-26
[Feedback]: Allow users to relocate or reinstall Python and R runtimes


FerryLF:
### Feedback type

Improvement

### What would you like to share?

Python and R runtimes may initially be installed in a default location that later becomes unsuitable due to limited disk space, storage organization, or environment issues.

Currently, users may not have a convenient way to change the runtime installation location after setup, or reinstall the Python/R environment directly from Open Science.

This makes runtime maintenance difficult and may require manual file operations or a full application reinstall.

Proposed solution

Add runtime management options in Settings > Runtimes that allow users to:

Choose a new installation location for Python and R
Migrate an existing managed runtime to the new location
Reinstall Python or R independently
Remove and reinstall a broken runtime environment
View the current runtime path and installation status
Verify the runtime after migration or reinstallation

Before migration, Open Science should check available disk space and clearly indicate whether the existing environment will be moved, copied, or reinstalled.

After the operation is completed, Open Science should automatically update the runtime path and continue using the new environment without requiring manual configuration.

<img width="1012" height="725" alt="Image" src="https://github.com/user-attachments/assets/9636b892-dbde-4cbc-befa-111ff7850ba1" />

PR #1984, opened 2026-08-31
feat(runtime): add safe managed runtime reinstall

LiuZhaoyan:
## Summary

Addresses #1776 within the current unified `<dataRoot>/runtime` ownership model.

- allow users to reinstall healthy app-managed default Python and R runtimes from Settings → Runtimes
- drain running and idle Notebook kernels, invalidate stale bindings, and durably rebind sessions before releasing the repair gate
- reuse the existing data-root write gate, serialized lifecycle, exclusive mutation lock, operation journal, crash recovery, interpreter verification, and ready-marker flow
- keep destructive reinstall unavailable for external, user-owned, and agent-created interpreters
- show target filesystem capacity and scan-derived copy requirements during data-root migration
- explain that runtime environments are rebuilt after restart, may need additional unquantified disk space, and do not guarantee restoration of pip-only or CRAN-only packages

## Runtime lifecycle

```text
Reinstall request
→ validate the exact managed default runtime identity
→ block runtime use and persist repair-required bindings
→ stop affected running and idle kernels
→ run the existing repair lifecycle
→ delete and reprovision the prefix under the existing mutation lock
→ verify the interpreter
→ rediscover and durably persist replacement bindings
→ clear repair markers and refresh the UI to Ready
```

Preparation and completion failures remain fail-closed: destructive mutation does not start when usage coordination cannot be persisted, and a binding-refresh failure removes the ready marker so the existing recovery path remains authoritative.

## Testing

- `npm run typecheck:node`
- `npm run typecheck:web`
- `npm run lint`
- `npx vitest run src/main/notebook/runtime-repair.test.ts src/renderer/src/pages/settings/RuntimesPanel.render.test.tsx src/renderer/src/i18n/resources.test.ts --maxWorkers=1` — 792 passed
- relevant runtime/storage test batches — 1,096 passed
- `npm run build:e2e`
- real Electron Python reinstall with an active cell, followed by successful execution on the rebuilt runtime
- real Electron R setup/reinstall, followed by successful Notebook execution
- real Storage migration staging against a separate filesystem, including available/copy-required display and safe staged-copy discard

A full local `npm test` run was also attempted. It is not fully green in this environment because sandboxed localhost listeners fail with `EPERM`, the escalated run has unrelated workflow fixtures requiring `jq`, and one pre-existing provenance verification case remains failing. The affected runtime, renderer, storage, IPC, lifecycle, and i18n suites pass in isolation.

## Scope

This intentionally keeps Python and R under the existing shared runtime root. Independent Python/R installation roots would require a separate architecture change to ownership, cache, locking, and relocation semantics.


ewen-poch:
Thanks for laying down the careful lifecycle here. I pulled the branch locally and walked through the actual failure paths rather than relying on the green CI alone. I found four edges that were not being exercised, and pushed e96382a to this PR.

A non-runnable but correctly identified app-managed default was being rejected before repair, and the current-session lookup only recognized root lanes, which meant delegated Frame kernels could survive prefix deletion. The update now admits only the exact broken managed default and checks the complete lane identity before quarantining and terminating kernels.

I also covered two recovery and compatibility cases. Historical Windows run.json bindings that still point at the legacy default-python/default-r prefixes are now lazily and durably rebound to the verified short managed prefixes. That migration is deliberately limited to exact app-managed defaults, so named and external runtimes stay untouched. If rebuilding fails after the old prefix has been removed, the durable repair marker is now reflected through the existing recovery status and the store refreshes status after both success and failure, so Settings offers Reset immediately instead of asking the user to trigger another failed setup first.

This keeps dataRoot/runtime as the single storage owner, adds no new persisted enum or schema, and reuses the existing runtimeBindings and repair marker. The five focused regressions pass, the affected Notebook/renderer/i18n batch passes with 1,160 tests and 5 skips, and Node typecheck, web typecheck, and lint are green. The complete provisioner test file hits its own Windows path-budget guard from this unusually long local worktree path; its changed focused test passes, and the normal PR checkout will still exercise the full file in CI.

ewen-poch:
Small follow-up from the first CI pass: the new Windows-only legacy binding migration made runtime-binding.ts Windows-sensitive, but our change-impact manifest did not know that yet. I added the file there so future edits keep the Windows runtime/path and GUI E2E lanes in scope; the classifier suite is now 68/68 locally. The other failure in that macOS shard was an unrelated Subagent rendering timing test, so I have not changed that area. A fresh CI run is underway and I’ll keep an eye on it.

ewen-poch:
One more correction after CI Integrity did exactly what it is meant to do: change-impact.json is protected control-plane and should not be edited in a normal PR. I restored it unchanged and moved the Windows-only legacy path matching into windows-runtime-binding.ts instead, which the existing filename overlay classifies automatically. The behavior is unchanged, the platform boundary is clearer, and no ruleset bypass is needed. The classifier plus runtime-service modules pass 297 tests locally (5 skipped), with Node typecheck and lint also clean.

LiuZhaoyan:
Thanks a lot for the quick and thorough review — really appreciate you digging into the failure paths and following up with the fixes so quickly. I’m going through the extra commits now to understand the edge cases better. Thanks again for taking the time on this!

ewen-poch:
Everything is green now. The first Windows core attempt failed only while the runner was opening the WFP loopback test socket; rerunning that job passed the AppContainer smoke, path portability, runtime, shell, and timeout checks without any further code changes. Both full macOS shards, Module coverage, Static, CI Integrity, Windows E2E, macOS E2E, and the final PR Gate are all passing as well.

LiuZhaoyan:
CI Integrity is failing because a71459c modifies the protected scripts/ci/change-impact.json. Maybe you could land those two Windows-sensitive entries on main through a separate maintainer control-plane change with the required ruleset bypass, then sync main back into this PR. That should preserve the correct Windows coverage while allowing CI Integrity to pass.

PR #1994, opened 2026-08-31
fix(runtime): preserve managed runtimes on failed updates

ewen-poch:
## Problem

Managed Claude extracted a replacement directly over the stable runtime path and only ran its
`--version` check afterward in `AgentRuntimeManager`. Managed OpenCode likewise extracted and smoke
tested at the stable path, deleting that path when the candidate failed. A failed update could
therefore remove or overwrite the last working runtime.

The regression tests reproduced both destructive outcomes before the fix:

- an incomplete Claude archive removed the existing binary (`ENOENT`);
- a Claude candidate that could not report a version replaced `WORKING-CLAUDE` with
  `BROKEN-CLAUDE`;
- an OpenCode smoke failure removed the existing binary (`ENOENT`).

## Proposed change

- Download and extract Claude and OpenCode candidates into UUID-scoped sibling staging directories.
- Apply executable permissions and run `--version`/smoke checks before publication.
- Rename the existing runtime to a UUID-scoped backup, publish the staged directory, and restore the
  backup automatically if publication fails.
- Stop retrying registries after a local publication failure, while preserving registry fallback and
  OpenCode's AVX2 baseline fallback before publication.
- Mark installer-created runtime trees with an exact product-specific ownership marker. Uninstall
  removes UUID-scoped staging/backup siblings only when both the name and marker prove ownership.
- Before each install, reconcile ownership-marked staging siblings left by an interrupted attempt,
  independently of persisted runtime settings; retain backups because one may be the only manual
  recovery copy after a failed restore.
- Convert staging-directory setup failures into the installers' existing structured failure result
  and system log instead of rejecting the install call.
- Move Claude candidate version validation into the installer transaction by injecting the manager's
  existing version probe, while retaining the manager's established final-path validation.

## Scope and non-goals

- No database/schema, persisted settings record, data-model, enum, IPC, or renderer changes.
- Stable managed runtime paths and stored path compatibility are unchanged; no data migration is
  needed.
- No generic installer framework is introduced.
- UUID-scoped staging/backup directories and `.open-science-managed-runtime` ownership markers are
  operational app-owned files, not business data. Successful installs clean temporary directories;
  retries clean ownership-marked interrupted staging directories; an unrecoverable restore retains
  the marked backup path in the error for manual recovery, and managed uninstall owns cleanup of
  marked orphans.
- A pre-existing managed runtime receives the marker immediately before its first atomic backup; no
  historical runtime or settings conversion is required.

All listed checks ran after the last material edit:

- Claude extraction/version/setup failure preserves the prior runtime and returns structured errors
  -> `npm test -- src/main/settings/managed-claude.test.ts` -> 33 passed.
- OpenCode smoke/publication/setup failure preserves or restores the prior runtime ->
  `npm test -- src/main/settings/managed-opencode.test.ts` -> 40 passed, 2 skipped.
- Manager validates Claude before persistence and keeps prior settings on failure ->
  `npm test -- src/main/settings/agent-runtime-manager.test.ts` -> 30 passed.
- Existing service contract for final-path version validation -> targeted test passed.
- Main-process interfaces compile -> `npm run typecheck:node` -> passed.
- Source and tests satisfy lint/format rules -> `npm run lint` -> passed.

`npm run test:affected:explain -- --base main --head HEAD` selected full mode because
`agent-runtime-manager.ts` has no module owner. Per the requested local validation scope, the complete
Vitest suite was not run locally; PR Gate is authoritative for that fallback and platform lanes.

## Review focus

- Backup restoration and error preservation when the staged rename fails.
- OpenCode standard-to-baseline fallback remaining entirely inside staging.
- Marker-backed orphan cleanup preserving unowned UUID-shaped directories.
- Pre-install reconciliation removing only ownership-marked interrupted staging directories.
- Staging setup failures staying inside the structured installer result boundary.

### DIFF
--- a/src/main/settings/agent-runtime-manager.test.ts
+++ b/src/main/settings/agent-runtime-manager.test.ts
@@ -1,6 +1,9 @@
+import { createHash } from 'node:crypto'
 import { chmod, mkdir, mkdtemp, readFile, readdir, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { dirname, join, posix } from 'node:path'
+import { Readable } from 'node:stream'
+import { gzipSync } from 'node:zlib'
 
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
@@ -40,7 +43,7 @@ const { AgentRuntimeManager } = await import('./agent-runtime-manager')
 const { SettingsRepository } = await import('./repository')
 const { getAppClaudeConfigDir } = await import('./provider-env')
 const { connectorSkillSourceDir } = await import('../connectors/provision')
-const { managedClaudeDir } = await import('./managed-claude')
+const { installManagedClaude, managedClaudeDir } = await import('./managed-claude')
 const { managedOpencodeDir } = await import('./managed-opencode')
 const { managedCodeBuddyDir, verifyCodeBuddyVersion } = await import('./managed-codebuddy')
 
@@ -60,6 +63,27 @@ const createInventory = (): RuntimeInventory => ({
   codexAdapter: new Map(),
   codexNative: new Map()
 })
+
+const buildManagedClaudeTgz = (content: Buffer): Buffer => {
+  const header = Buffer.alloc(512)
+  header.write('package/claude', 0, 'utf8')
+  header.write('0000755\0', 100, 'ascii')
+  header.write('0000000\0', 108, 'ascii')
+  header.write('0000000\0', 116, 'ascii')
+  header.write(`${content.length.toString(8).padStart(11, '0')}\0`, 124, 'ascii')
+  header.write('00000000000\0', 136, 'ascii')
+  header.write('0', 156, 'ascii')
+  header.write('ustar\0', 257, 'ascii')
+  header.write('00', 263, 'ascii')
+  header.fill(0x20, 148, 156)
+  const checksum = header.reduce((sum, byte) => sum + byte, 0)
+  header.write(`${checksum.toString(8).padStart(6, '0')}\0 `, 148, 'ascii')
+  const padding = Buffer.alloc((512 - (content.length % 512)) % 512)
+  return gzipSync(Buffer.concat([header, content, padding, Buffer.alloc(1024)]))
+}
+
+const sha512 = (data: Buffer): string =>
+  `sha512-${createHash('sha512').update(data).digest('base64')}`
 
 const createClaudeDeps = (inventory: RuntimeInventory): ClaudeDetectDeps => ({
   env: {},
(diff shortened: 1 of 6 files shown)


ewen-poch:
Latest Codex Review finding is already covered on the current head: agent-runtime-manager.ts lines 562-568 call detectDeps.getVersion(outcome.resolvedPath) after installer success and return a logged failure before setClaudeInfo when the probe is empty. I also reran the existing contract test: npm test -- src/main/settings/service.test.ts -t 'logs a version error and rejects an incompatible managed runtime' -> 1 passed. No code change is needed for this finding.

ewen-poch:
Addressed both symlink findings in 9e6540faa. Claude and OpenCode now use lstat before publication and reject either an existing managed runtime root or ownership marker when it is a symbolic link, before any marker write or rename. Public installer regression tests reproduce the prior behavior by pointing the managed root at an external directory: before the fix the external marker changed from EXTERNAL-DATA to the product ownership marker; after the fix both installers return structured failures and leave the external target unchanged. Validation: affected modules 75 passed / 2 skipped, typecheck:node passed, lint passed, and git diff --check passed.

ewen-poch:
Addressed both cleanup ownership-marker findings in a6eee1897. Orphan discovery now uses lstat and accepts the product receipt only when the marker is a regular file; symlinks and other non-regular entries fail closed before readFile, so UUID-shaped unowned directories are preserved. Regression coverage exercises the public uninstall behavior with a filesystem-boundary no-follow result: before the fix both product directories were deleted, and after the fix their payloads remain. Validation: affected modules 77 passed / 2 skipped, typecheck:node passed, lint passed, and git diff --check passed.

ewen-poch:
Addressed the hard-linked ownership-marker issue for both Claude and OpenCode.

- Existing markers are now validated and never rewritten.
- Missing markers are created exclusively with the wx flag to avoid overwriting a concurrently introduced entry.
- Publication rejects symbolic links, non-regular files, unrecognized contents, and markers with multiple hard links.
- Orphan cleanup applies the same link-count ownership rule before deleting staging/backup directories.

Regression coverage reproduces both reported effects through the public install/uninstall boundaries: the old code overwrote an external hard-linked inode during install and deleted an orphan directory whose marker had multiple hard links. Both cases now preserve the external data.

Validation: affected module suites passed (81 passed, 2 skipped), plus npm run typecheck:node, npm run lint, and git diff --check.

ewen-poch:
Addressed the ownership-marker inode race for both Claude and OpenCode.

- Marker content is now read through the same open file handle used for handle-level stat validation.
- Platforms that expose O_NOFOLLOW use it when opening the marker.
- The path lstat identity and opened-handle identity must match by device and inode; regular-file and single-link checks are repeated on the handle.
- Any open, identity, type, link-count, or read failure fails closed for orphan cleanup and aborts publication validation.

The regression tests deterministically replace an unowned marker with an apparently valid marker between the old lstat and path-based readFile calls. Before the fix, both public uninstall paths deleted the orphan payload and failed with ENOENT; after the fix, both preserve it.

Validation: affected module suites passed (83 passed, 2 skipped), plus npm run typecheck:node, npm run lint, and git diff --check.

PR #1995, opened 2026-08-31
fix(notebook-network): restore protected kernel startup

ewen-poch:
## Problem

In the packaged Windows x64 app (`0.24.0`), enabling Notebook network protection can leave the sandbox status at `ready` while every kernel exits before dispatch. A minimal Python cell reports `Notebook kernel process exited.`; the same session's Bash/package path exposes `restore filesystem ACL for C:\Program Files\GitHub CLI: Access is denied. (0x80070005)`.

Two related behaviors cause and preserve the failure:

1. The narrowed Notebook environment omits `ProgramFiles`, `ProgramFiles(x86)`, and `ProgramW6432`. The AppContainer host therefore cannot classify Program Files PATH entries as intrinsically readable and attempts an unnecessary unelevated ACL grant.
2. A failed transaction can leave an owner-recorded ACL snapshot with no active lease. Recovery then attempts a privileged restore even when the current ACL already exactly matches that snapshot, so affected installations continue to fail before a kernel starts.

## Proposed change

- Preserve the three non-secret Program Files directory variables in the Windows Notebook process environment.
- Make ACL restoration idempotent: skip the privileged write only when the current captured ACL exactly equals the owned snapshot; keep exact fail-closed restoration when it differs.
- Rebuild the checked-in x64 and ARM64 AppContainer hosts and update their integrity hashes.

## Scope and non-goals

- No database fields, schema migration, application data-model changes, or new enum/status values.
- No UI or interaction change; the generic kernel-exit presentation is intentionally unchanged.
- No network-policy, allow-list, or sandbox architecture change.
- No release architecture expansion; the existing ARM64 runtime selection now receives the same native fix as x64.

Historical compatibility is intentional. Existing `acl-state.json` files keep the same format and owner. On the next owner recovery, an unchanged inactive snapshot can be pruned without `WRITE_DAC`; a genuinely changed ACL is still restored exactly or recovery fails closed. This is the only persisted-state impact.

The regression tests were first run against behavior-preserving test seams and failed for the reported causes, then passed after the implementation.

- Windows environment projection -> `npm test -- src/main/notebook/process-environment.test.ts -t "preserves Windows system-directory metadata required by the protected launcher"` -> 1 passed after the final edit.
- Windows shell consumer contract -> `npm test -- src/main/notebook/shell-process.test.ts` -> 10 passed, 5 platform-skipped after the final edit.
- Historical ACL recovery and changed-ACL fail-closed branch -> `cargo test --locked --manifest-path packages/notebook-network-sandbox/vendor/windows-src/Cargo.toml acl_snapshot_restore_skips_only_unchanged_acl -- --nocapture` -> 1 passed after the final edit.
- Sandbox owner/resource behavior -> `npm run test:module -- notebook_network_sandbox` -> 69 passed, 11 platform-skipped after the final edit.
- Bundled host integrity -> `npm test -- packages/notebook-network-sandbox/src/resources.test.ts` -> both x64 and ARM64 resources passed after the final edit.
- ARM64 native build -> [Temporary Notebook ARM64 host build](https://github.com/aipoch/open-science/actions/runs/33454574464) -> passed; the artifact was built from Rust source and lock files byte-identical to this PR.
- Electron main-process contracts -> `npm run typecheck:node` -> passed after the final edit.
- Repository lint -> `npm run lint` -> passed after the final edit.
- Rust formatting and patch whitespace -> `cargo fmt --manifest-path packages/notebook-network-sandbox/vendor/windows-src/Cargo.toml -- --check` and `git diff --check` -> passed.

The full Rust suite reached 7 passed / 1 failed locally. The existing `acl_snapshot_restore_preserves_inheritance_control` test could not modify a newly created test directory DACL on this workstation (`Access is denied`) and failed before calling the changed restoration path. The clean Windows PR lane remains authoritative for that ACL integration test and native-host smoke coverage. The installed application's durable state was not mutated for a post-fix live probe.

## Review focus

- Confirm the Program Files variables are the minimum metadata needed by the existing AppContainer classifier.
- Confirm equality is checked against the full ACL snapshot before skipping restoration and that different snapshots still invoke the restore path.
- Confirm the compatibility behavior stays inside the native ACL-state owner and does not weaken cleanup of changed ACLs.


ewen-poch:
Addressed the Windows shell contract in cca2250a5. Preserving ProgramFiles is intentional for both kernel and shell processes because both cross the protected AppContainer launcher boundary; shell-process.test.ts now asserts that metadata remains available while host secrets remain excluded. The focused consumer test and the second full PR Gate pass.\n\nThe ARM64 finding is valid for an ARM64 package: runtime selection supports that architecture even though the current Windows release matrix only emits x64. Rebuilding and checking in the ARM64 executable would expand the current changed-file set to the ARM64 binary and its integrity hash, so I am pausing that scope expansion for maintainer confirmation before changing the PR.

ewen-poch:
Addressed the ARM64 P1 in 9f0fccb4b. The bundled ARM64 AppContainer host was rebuilt with the repository's documented cargo-xwin path in [this isolated build](https://github.com/aipoch/open-science/actions/runs/33454574464). The build used main.rs, Cargo.toml, and Cargo.lock files byte-identical to this PR. The downloaded artifact reports AA64 machine (ARM64); its checked-in SHA-256 is 2ecca8baa0ce5062ba723d9a01888461019f04c051dadd0fbb38c078133f7f3. The two-architecture resource integrity test and the Notebook sandbox module suite pass locally.

PR #2001, opened 2026-08-31
fix(tool-activity): redact raw tool credentials

ewen-poch:
## Problem

Small ACP tool `rawInput` and `rawOutput` payloads were bounded to 8,000 serialized characters but otherwise retained verbatim. Credential fields such as `apiKey`, `authorization`, `password`, or `token` could therefore reach renderer runtime events, durable ToolActivity records, persisted permission context, and Session JSON.

## Proposed change

- Add one recursive raw-tool payload sanitizer beside the existing shared tool-detail sanitizer.
- Replace sensitive field values with `[redacted]` while preserving non-sensitive structure and the existing size bound.
- Apply the same sanitizer before runtime IPC, when rebuilding persisted ToolActivity records, and when rebuilding persisted permission tool input.
- Cover live ACP projection and historical Session decoding with regression tests that failed against the pre-fix implementation for the reported credential retention.

## Scope and non-goals

- No Session schema, envelope version, persisted field, relationship, status, or enum changes.
- No new renderer interaction; only sensitive activity-detail values change to `[redacted]`.
- No tool-level raw-payload persistence policy or provider-specific opt-out metadata.
- No eager rewrite of every historical Session file. Historical files are redacted when decoded before renderer hydration and are rewritten through the existing save path when subsequently saved.

The ACP sanitizer runs after protocol translation in the shared session-update projection. `claude-code`, `opencode`, `codex-response`, and `codex-bridge` therefore use the same boundary; this change does not alter framework wire shapes, correlation, lifecycle, replay, subscription, launcher, model, or transport behavior. No framework-specific duplicate suite was added because the security invariant is owned by that shared post-translation boundary.

- live ACP raw tool credential redaction and bounded detail behavior -> `npm test -- src/main/acp/runtime-events.test.ts src/shared/session-persistence.test.ts src/shared/tool-detail-sanitizer.test.ts` -> 3 files, 195 tests passed
- historical ToolActivity and permission-context decode redaction plus Session persistence consumers -> `npm run test:module -- session_persistence` -> 7 files, 372 tests passed
- main/shared TypeScript surfaces -> `npm run typecheck:node` -> passed
- renderer/shared TypeScript surfaces -> `npm run typecheck:web` -> passed
- linted source/test surfaces -> `npm run lint` -> passed

Local validation covered the shared framework path on macOS. No model-specific or platform-specific branch changed. Full portable and platform coverage is delegated to the exact-head PR Gate as requested.

## Review focus

- Sensitive-key matching and preservation of safe sibling fields.
- Redaction occurring before both runtime IPC and durable Session projection.
- Historical compatibility: decode-time redaction without a schema migration or eager disk rewrite.


ewen-poch:
Addressed in c96c03e0. The raw payload sanitizer now applies the existing credential-text policy to every string leaf, covering URL query credentials, authorization headers, CLI arguments, and serialized JSON while retaining recursive sensitive-field redaction. The diagnostic module re-exports the shared policy so existing diagnostic consumers and tool persistence use one implementation. Added a runtime-to-persistence regression test that failed before the fix because the URL token remained in the event and now passes.

ewen-poch:
Addressed in b4adcfda:\n\n- Permission publication now sends a sanitized request projection while leaving the broker-owned request unchanged. The runtime snapshot boundary also sanitizes raw payloads for directly synthesized tool events, so those events cannot bypass notification normalization.\n- Raw object keys now use isSensitiveDiagnosticKey, covering auth, privateKey, pat, and the rest of the existing policy while preserving token metrics such as inputTokenCount.\n- Tool content and terminal output now pass through the shared credential-text sanitizer before renderer IPC and Session persistence, including retained text, resource, and diff strings.\n\nAdded public-boundary regressions for the permission callback/snapshot, synthesized tool events, content/terminal output, and sensitive-key policy. Each exposed the reported credential before its fix and now passes.

ewen-poch:
No code change is needed for this finding: b4adcfda already imports and calls isSensitiveDiagnosticKey(key) in sanitizeRawToolPayload. That shared helper normalizes camelCase and separators before matching the privatekey suffix, so privateKey, private_key, and private-key are all redacted. The focused regression in src/shared/tool-detail-sanitizer.test.ts includes privateKey and passes; with the previous regex it failed by returning the original value.

ewen-poch:
Addressed in 348a063b after rebasing onto the latest main. The persisted permission request remains a redacted display projection, while the existing SHA-256 fingerprint remains the authority used to match the exact live provider replay. The incompatible check that recomputed the original fingerprint from the redacted request was removed only from presentation-candidate setup; broker authorization and final presentation aliasing still require the live replay fingerprint to equal the stored fingerprint. Internal Notebook correlation now derives its bounded input from the original provider notification rather than the renderer-sanitized event for Claude Code, OpenCode, and Codex. The cross-provider regression runs through sanitizeSessionPermissionRuntimeContext, confirms the secret is absent from the persisted projection, failed with tool-replayed before each fix, and now aliases to tool-original.\n\nThe repeated structured-key finding requires no additional change: sanitizeRawToolPayload already calls isSensitiveDiagnosticKey(key), with passing coverage for auth, pat, privateKey, and token-metric preservation.

ewen-poch:
Addressed in d05536ef. Permission requests now pass through the same sanitized projection before publication through the callback, permission event, runtime state, or full snapshot. The broker-owned request remains unchanged so authorization and replay fingerprint checks still use the original input. The regression test exercises the public publication boundary and failed before the fix because pendingPermissions exposed the credential-bearing rawInput.

ewen-poch:
Addressed both findings in f1d860ba. The shared text redactor now recognizes URL schemes with slash-escaped separators, normalizes them for credential parsing, and restores the escaped scheme form after redaction. URL fragments are preserved when they are ordinary anchors and removed only when their parsed key/value pairs identify a credential. The public sanitizer regressions failed before the fix by retaining the escaped sig value and dropping a non-sensitive #methods fragment; both now pass, alongside the existing sensitive-fragment logger coverage.

ewen-poch:
Addressed both findings in f66b2120. Renderer-facing permission callbacks, events, state, and snapshots now publish a bounded, credential-redacted title while the broker-owned request remains unchanged. The durable permission sanitizer applies the same policy to the existing title field, so historical Session files are redacted on load without a schema change or migration. Restored Notebook presentation matching now compares both the current bounded preview and the exact trusted input from the same live provider notification, preserving pre-existing fingerprints above the 7,500-character preview limit. This compatibility applies only to presentation aliasing; broker authorization continues to require the stored exact fingerprint. Each regression was confirmed failing before its fix and passing afterward.

ewen-poch:
Addressed both findings in 302dc74a. The coordinator now reuses the existing permission projection for its final pending-permission aggregation and for both delegated and per-runtime permission callbacks, so renderer IPC and notifications cannot bypass title/rawInput redaction. The delegated regression covers the initial snapshot, direct callback, and subsequent state publication while confirming the source request remains unchanged. Tool detail text is now capped both before redaction and after marker expansion; the regression reproduced a 16,000-character input expanding to 34,000 characters before the fix and verifies the established capped output afterward.

ewen-poch:
Addressed in a1a6a7cf. Runtime snapshot ownership now sanitizes and bounds event titles before returning or retaining them, while leaving the source event unchanged. Persisted ToolActivity reconstruction applies the same existing text sanitizer to titles, so historical Session files are redacted on load without a schema change or migration. Both public-boundary regressions failed before the fix by retaining the credential-bearing title and now pass.

ewen-poch:
Addressed the access-key credential gap reported by the latest AI review job in 2dd6e99d. The shared sensitive-key policy now recognizes accessKey/access-key variants, and the existing text redactor covers access-key assignments and CLI arguments. Public runtime publication and historical Session reconstruction regressions both failed before the fix by retaining the structured and CLI secrets and now pass. This changes only the existing redaction policy; it adds no schema, persisted field, enum, or migration.

PR #2007, opened 2026-09-01
fix(message-scroll): respect small upward scrolls

ewen-poch:
## Problem

While the conversation is following the bottom, a small upward scrollbar-thumb drag can remain inside the message scroller's default 8 px end threshold. If a tool group is expanding at the same time, its height tween emits repeated content resize frames and bottom-follow snaps the viewport back to the new end. A continued small drag can therefore alternate between moving up and snapping down.

The pre-fix regression reproduced this deterministically: after moving from `scrollTop = 100` to `96`, one animated-content resize moved the viewport to `140` instead of preserving `96` (three consecutive failing runs).

## Proposed change

- Default the shared `MessageScrollerProvider` end threshold to zero, while preserving explicit caller overrides.
- Add a public-boundary regression test using the real provider, viewport, content, and item components. It controls only DOM geometry and `ResizeObserver` so the reported scrollbar-drag/animated-growth sequence is deterministic.

## Scope and non-goals

- Interaction change: any real upward movement immediately releases bottom-follow; the reader's position is then preserved while content grows.
- No tool animation rewrite and no new state machine or test-only production seam.
- No data fields, architecture, data model, enum/state values, persistence, migrations, or historical-data compatibility changes.

- Small upward scrollbar drag followed by animated content growth preserves the reader position:
  - Before fix: focused test failed 3/3 with expected `96`, received `140`.
  - After fix: `npm test -- src/renderer/src/components/ui/message-scroller.test.tsx` — 3/3 passed.
- Workspace conversation behavior: `npm run test:module -- workspace_page` — 29 files, 709 tests passed.
- Shared consumers and tool animation: focused six-file Vitest run covering `WorkspaceCollapsiblePanel`, both `WorkspaceMessageScroller` suites, and both artifact provenance suites — 6 files, 163 tests passed.
- `npm run lint` — passed.
- `npm run build:web` — passed (existing build warnings only).
- Real-browser smoke test against the saved report session: tool group collapsed and expanded successfully; Notebook detail expanded successfully.
- `npm run test:affected -- --base origin/main --head HEAD --explain` selects the full Vitest suite because the shared UI test has no module owner. Per request, the full suite is left to PR CI.

## Review focus

- Confirm that zero tolerance is the intended shared behavior for both workspace conversations and artifact provenance: any nonzero upward distance should release bottom-follow and show the return-to-end affordance.

### DIFF
--- a/src/renderer/src/components/ui/message-scroller.test.tsx
+++ b/src/renderer/src/components/ui/message-scroller.test.tsx
@@ -12,12 +12,15 @@ import {
   MessageScrollerViewport
 } from './message-scroller'
 
+;(globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = true
+
 let container: HTMLDivElement | undefined
 let root: Root | undefined
 
 afterEach(() => {
   if (root) act(() => root?.unmount())
   container?.remove()
+  vi.unstubAllGlobals()
   root = undefined
   container = undefined
 })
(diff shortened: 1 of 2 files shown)


ewen-poch:
CI triage note for run [33456962932](https://github.com/aipoch/open-science/actions/runs/33456962932):

- All three full portable-test shards, module tests/coverage, macOS E2E, Windows E2E, Windows core, Linux Notebook isolation, CodeQL, PR policy, CI Integrity, and Codex review passed.
- `Static checks` failed only because `typecheck_node` reports pre-existing type errors in untouched files:
  - `src/main/web-service/http-server.test.ts:1662`: a message event is missing required `role`.
  - `src/main/web-service/public-task-event-stream.test.ts:101-102`: message events are missing required `role` and `text`.
- The PR commit changes only `src/renderer/src/components/ui/message-scroller.tsx` and its test; none of the failing files or node types/configuration changed. `npm run typecheck:node` reproduces the same three errors locally, while CI's formatting, lint, `typecheck_web`, interface contracts, CLI/SDK, and i18n checks all passed.

I am leaving that unrelated baseline correction out of this focused scroll-jitter fix.

PR #2012, opened 2026-09-01
feat(runtime): add environment creation and uninstall controls

ewen-poch:
## Problem

Agents can currently create named Notebook environments and lazily provision a missing app-managed default without a user policy switch. Settings also has no supported uninstall action for Agent-created named environments, while app-managed defaults and user-owned interpreters must remain outside that deletion authority.

## Proposed change

- Add a global **Allow Agent to create environments** setting.
- Enforce the setting for `manage_environments(action: "create")`, Agent-triggered lazy default provisioning, and Agent package operations that would create a missing default.
- Keep explicit user setup from Settings available while Agent creation is disabled.
- Persist immutable ownership evidence for newly Agent-created named environments and add a confirmation-based uninstall action only when fresh discovery validates that evidence.
- Keep `default-python`, `default-r`, other app-managed defaults, and `user-own` interpreters non-uninstallable from Settings.
- Reuse the existing named-environment removal owner for recovery admission, name validation, live-kernel and Runtime-binding checks, serialized mutation, exact internal-prefix deletion, and repair cleanup.

## Authority and lifecycle

The renderer submits only `(language, envId)`. The main-process workflow performs fresh discovery, requires an exact matching entry with `provenance === "agent-created"`, `agentOwned === true`, and a discovered `condaEnv`, and only then passes that environment name to the existing `NotebookEnvironmentManagementOwner`. Renderer paths never become deletion authority.

`agentOwned` is derived only when a protected per-environment receipt and an environment-prefix generation marker contain the same random ownership id, name, language, and exact canonical prefix. The existing owner independently consumes that receipt before physical deletion, restores it if physical deletion fails, and also refuses reserved app-managed names, environments with live kernels, and environments selected by active or revoking Runtime bindings. `DefaultRuntimeProvisioner.removeEnvironment` resolves the validated name under the app Runtime `envs` directory and never scans or deletes user-owned interpreter locations.

Authenticated Agent RPC execution sources are canonicalized to `agent`, so caller-controlled metadata cannot claim user authority to bypass the creation policy.

- **Historical compatibility:** the optional `agentEnvironmentCreationEnabled` setting defaults to `true` when absent, preserving existing creation behavior without a migration. Historical named environments have no ownership receipt: they remain discoverable, runnable, and path-classified as `agent-created`, but are intentionally **not uninstallable**. Manually placed environments under the internal `runtime/envs` directory are likewise not uninstallable. Interpreters outside app-managed storage remain `user-own` and cannot be uninstalled.
- **New enum/state values:** no enum values are added. Discovery adds the optional derived `agentOwned` boolean; absence/false means deletion is not authorized. The previous PR design's persisted `RuntimeOperationKind = "remove"` and managed-default removal state remain absent.
- **Persistence:** the optional boolean `agentEnvironmentCreationEnabled` is added to the existing settings JSON. Each newly Agent-created named environment also receives a protected receipt at `runtime/.agent-environment-ownership/<name>.json` plus `.open-science-agent-environment.json` inside its prefix. These files contain schema/kind, name, language, exact canonical prefix, and a shared random ownership id. No database rows, data relationships, operation-journal enum, or migration are added.

## User interaction

- Settings → Runtimes shows the creation-policy toggle globally.
- App-managed default cards continue to support setup/reset and never show **Uninstall**.
- Receipt-owned Agent-created named-environment cards show **Uninstall** with an irreversible-removal confirmation.
- Historical/path-inferred named-environment cards remain usable but do not show **Uninstall**.
- User-owned interpreter cards never show **Uninstall**.

## Validation

- Focused Notebook Runtime modules: **7 files, 134 tests passed**.
- Ownership receipt, discovery, removal owner, Settings workflow, and renderer modules: **5 files, 100 tests passed**.
- Receipt-related Runtime service cases: **5 tests passed** (224 unrelated cases skipped by filter).
- Adjacent Runtime IPC and shared contract modules: **5 files, 37 tests passed**.
- Focused Settings, renderer, i18n, preload, HTTP, and Web contract modules: **10 files, 1019 tests passed**.
- Focused contract rerun: **6 files, 172 tests passed**.
- Changed-file ESLint: **passed**.
- Generated Web API parity (`npm run check:web-api-map`): **passed**.
- `git diff --check`: **passed**.
- Node/Web typechecks are locally environment-blocked because the shared `node_modules` is missing the already-declared `@types/js-yaml` and `@types/micromatch`; all reported files are outside this change. PR CI remains authoritative.
- Per request, the complete `npm test` suite was not run locally.

## Review focus

- Only fresh receipt-backed `agentOwned` discovery evidence can reach named-environment removal, and the removal owner revalidates/consumes the receipt independently.
- App-managed defaults and user-owned interpreters have no Settings uninstall affordance and fail the main-process provenance guard.
- The creation switch covers explicit Agent environment creation plus implicit missing-default creation from execution and package operations, including an explicitly bound managed default.
- Explicit Settings setup remains available when Agent creation is disabled.


ewen-poch:
Addressed the lifecycle provisioning race identified in the latest review. Settings provision/repair and startup maintenance now acquire the same per-default-environment provision admission used by lazy provisioning before they enter the shared provisioner queue. Managed uninstall drains requests admitted before the barrier closes and rejects requests admitted afterward, preventing lifecycle work from recreating a Runtime after deletion. This does not change the Agent-creation policy: explicit user setup remains allowed when Agent creation is disabled. Focused Runtime validation: 340 passed, 7 skipped across six modules.

ewen-poch:
Addressed the stale package-request race. Package operations now capture a per-environment removal-generation token before target resolution and validate it atomically when they acquire a removal reservation after mirror/metadata resolution. If managed uninstall started or completed in that interval, the stale request is rejected even though the live removal flag has already cleared. Requests that acquire the reservation first are drained before deletion. Provisioning and package requests share the existing process-local removal-admission drain; the generation/token is also process-local, with no persisted schema or enum changes. Focused validation: 276 passed, 5 skipped across the four affected Runtime modules; final direct regression rerun: 30 passed.

ewen-poch:
Updated the Runtime ownership boundary: app-managed defaults (including default-python/default-r) and user-owned interpreters no longer expose or support Settings uninstall. Settings uninstall now accepts only a freshly discovered agent-created environment, derives its internal conda name in the main process, and reuses the existing named-environment removal owner. The Agent creation switch remains enforced for named creation and missing-default creation from Agent execution/package paths; explicit Settings setup is still allowed. This also removes the previous managed-default removal journal, remove operation kind, reinstall state, and related recovery/concurrency machinery. No uninstall persistence or new enum value remains.

ewen-poch:
Confirmed the ownership-evidence concern. The current EnvProvenance contract is path-derived, so hardening uninstall to require an immutable creation receipt would add persistent ownership data and change historical named-environment behavior. Per the product constraints for this change, I am pausing before introducing that data-model/persistence boundary. No code change has been made for this finding yet.

ewen-poch:
Addressed the ownership finding in 410fa71db. Newly Agent-created named environments now receive immutable per-environment ownership evidence: a protected receipt records the language and exact canonical prefix, and a prefix-generation marker shares a random ownership id. Fresh discovery exposes deletion authority only when both records match, and the removal owner independently consumes the receipt before deletion (restoring it only when physical deletion fails). Historical or manually placed named environments have no receipt, remain usable, and are not uninstallable.

PR #2019, opened 2026-09-01
fix(reviewer): recover interrupted Fix Loops

ewen-poch:
## Problem

Fix Loop ownership was held only in Main-process memory and one-shot renderer broadcasts. A refreshed or late renderer could miss the active-loop and auto-review suppression events, while an application restart after the initial flagged assessment could leave a non-stale completed Review with open findings and no reliable recovery or retry path.

## Proposed change

- Keep an initial Review `running` after warn/fail findings are atomically persisted and complete it only after the bounded Fix Loop settles.
- Derive the persisted active-loop signal from `lifecycle=running`, `outcome=null`, and persisted warn/fail findings; this stays within the existing database constraints.
- Recover that state on startup by recording existing `aborted/unaddressed` dispositions for remaining open findings and closing the root Review as `complete/flagged`.
- Restore the renderer send lock from hydrated Review data, while retaining live START/END broadcasts.
- Exclude correction responses from ordinary auto-review using their durable reviewer-correction attribution, even if the one-shot suppression event was missed.
- Offer an explicit re-run on terminal Reviews that still contain unresolved findings.

## Scope and non-goals

- No new table, column, SQLite constraint, migration, or lifecycle/disposition enum value.
- Existing `Review`, `Finding`, and `ReviewFindingDisposition` rows are reused; historical `complete/flagged` rows are not rewritten.
- Restart recovery safely terminates the interrupted loop and exposes retry. It does not resume an exact correction prompt or round; resumable execution would require a dedicated Fix Loop aggregate in a separate change.
- No ACP wire protocol, framework adapter, launcher, model, or platform-specific behavior changes. Correction attribution is already part of the shared persisted session contract.

The commands below were run after the last material edit.

- Active root Review remains durable until remediation settles -> `npm test -- --run src/main/reviewer/fix-loop.test.ts` -> 1 file, 9 tests passed.
- Restart distinguishes an active Fix Loop from an unfinished initial assessment and records terminal dispositions -> focused Reviewer regression command -> covered by `reviewer-resilience.test.ts`, passed.
- Hydrated active-loop state restores the composer lock -> focused Reviewer regression command -> covered by `WorkspacePage.send-gate.test.tsx`, passed.
- Durable correction attribution prevents duplicate ordinary auto-review -> focused Reviewer regression command -> covered by `workspace-events.test.ts`, passed.
- Terminal unresolved Reviews expose a re-run and active findings remain visible -> focused Reviewer regression command -> covered by `SessionReviewerPanel.render.test.tsx` and `ReviewerCard.render.test.tsx`, passed.
- Focused regression set: `npm test -- --run src/main/reviewer/repository.test.ts src/main/reviewer/reviewer-resilience.test.ts src/renderer/src/lib/acp/workspace-events.test.ts src/renderer/src/pages/workspace/WorkspacePage.send-gate.test.tsx src/renderer/src/pages/workspace/SessionReviewerPanel.render.test.tsx src/renderer/src/components/ReviewerCard.render.test.tsx` -> 6 files, 241 tests passed.
- Reviewer owner, contracts, consumers, and Windows-sensitive overlay: `npm run test:module -- reviewer_orchestrator` -> 27 files, 573 tests passed.
- Renderer translation contract: `npm test -- --run src/renderer/src/i18n/resources.test.ts` -> 1 file, 805 tests passed.
- `npm run typecheck:node` -> passed.
- `npm run typecheck:web` -> passed.
- `npm run lint` -> passed.
- `npm run db:schema:check` -> passed; the final diff contains no database/schema changes.

`npm run test:affected:explain -- --base origin/main --head HEAD` selected the full plan because `src/main/reviewer/repository.ts` has no registered module owner. Per the requested local test scope, the complete `npm test` suite was not run locally; PR Gate is the authority for the resulting complete portable and platform lanes.

## Review focus

- The invariant that only an initial Review with atomically persisted warn/fail findings can be `running` with findings.
- Startup recovery idempotency and preservation of the original flagged result.
- Renderer locking from hydrated state without making one-shot broadcasts authoritative.
- Whether `aborted` plus its persisted note is sufficient for restart interruption; this intentionally avoids a new enum and migration.

## Uncovered risks

- Exact-round automatic resume after process restart is intentionally unsupported.
- Complete portable and operating-system CI lanes are pending PR Gate.
- Independent review of the final evidence mapping is pending.

### DIFF
--- a/src/main/reviewer/fix-loop.test.ts
+++ b/src/main/reviewer/fix-loop.test.ts
@@ -399,6 +399,7 @@ describe('fix loop: all-pass on re-review ends the loop (resolved)', () => {
     const client = createProjectDbClient(temporaryRoot!)
     await migrateApplicationDatabase(client)
     const repository = new ReviewRepository(() => Promise.resolve(client))
+    const onReviewUpdate = vi.fn()
 
     const reviewRun = runReview({
       sessionId: 'session-1',
@@ -409,6 +410,7 @@ describe('fix loop: all-pass on re-review ends the loop (resolved)', () => {
       acpRuntime: runtime,
       artifactStorageRoot: temporaryRoot!,
       mainSessionId: 'main-session-1',
+      onReviewUpdate,
       agentTarget: {
         frameworkId: 'codex',
         providerId: 'provider-1',
@@ -436,6 +438,14 @@ describe('fix loop: all-pass on re-review ends the loop (resolved)', () => {
         outcome: null,
         checks: [expect.objectContaining({ status: 'fail' })]
       })
+      expect(onReviewUpdate).toHaveBeenLastCalledWith(
+        expect.objectContaining({
+          id: initialReview?.id,
+          lifecycle: 'running',
+          outcome: null,
+          checks: [expect.objectContaining({ status: 'fail' })]
+        })
+      )
     } finally {
       releaseCorrection()
     }
--- a/src/main/reviewer/fix-loop.test.ts
+++ b/src/main/reviewer/fix-loop.test.ts
@@ -462,6 +462,75 @@ describe('fix loop: all-pass on re-review ends the loop (resolved)', () => {
     await client.$disconnect()
   })
 
+  it('publishes the terminal review when the Fix Loop exits exceptionally', async () => {
+    const process = new FakeAgentProcess()
+    const shared = makeSharedSession(makeSession())
+
+    startFixLoopFakeAgent(
+      process,
+      {
+        mainSessionId: 'main-session-1',
+        initialChecks: [
+          {
+            status: 'fail',
+            claim: 'Agent claimed 42 results',
+            evidence: 'Tool output shows 0 results',
+            locator: { blockRef: { blockIndex: 1 }, contentHash: 'abc123' }
+          }
+        ],
+        reReviewChecksByRound: [
+          [
+            {
+              status: 'pass',
+              claim: 'Agent claimed 42 results',
+              evidence: 'Correction confirmed: results now correct'
+            }
+          ]
+        ]
+      },
+      shared
+    )
+
+    const runtime = createFixLoopRuntime(process, shared)
+    await runtime.createSession({ cwd: '/workspace' })
+    const client = createProjectDbClient(temporaryRoot!)
+    await migrateApplicationDatabase(client)
+    const repository = new ReviewRepository(() => Promise.resolve(client))
+    const onReviewUpdate = vi.fn()
+    const onFixLoopStart = vi.fn(() => {
+      vi.spyOn(repository, 'commitFindingDispositions').mockRejectedValueOnce(
+        new Error('disposition write failed')
+      )
+    })
+
+    await expect(
+      runReview({
+        sessionId: 'session-1',
+        turnMessageId: 'msg-2',
+        projectId: 'project-1',
+        getSession: () => shared.getSession(),
+        reviewRepository: repository,
+        acpRuntime: runtime,
+        artifactStorageRoot: temporaryRoot!,
+        mainSessionId: 'main-session-1',
+        onReviewUpdate,
+        onFixLoopStart,
+        fixLoopMaxRounds: 0,
+        agentTarget: {
+          frameworkId: 'codex',
+          providerId: 'provider-1',
+          model: 'model-1',
+          reasoningEffort: 'high'
+        }
+      })
+    ).rejects.toThrow('disposition write failed')
+
+    expect(onReviewUpdate).toHaveBeenLastCalledWith(
+      expect.objectContaining({ lifecycle: 'complete', outcome: 'flagged' })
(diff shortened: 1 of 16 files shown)


ewen-poch:
Addressed the P2 in 1bf3f98d. The orchestrator now emits the committed running Review, including persisted warn/fail findings, before onFixLoopStart so live renderers receive the same state as hydrated renderers. Added a held-correction regression assertion; the Fix Loop tests pass 9/9, the Reviewer module set passes 573/573, and the Node typecheck passes.

ewen-poch:
Addressed both reported P1 findings in 42c73e9f.\n\n- Startup recovery now bounds same-turn Review selection to the active running Review's creation boundary and later re-reviews, preserving open findings from older completed chains without adding a chain field or migration.\n- Fix Loop cleanup now reloads and publishes the terminalized Review from the finally path, so an exceptional exit cannot leave the renderer on its last running snapshot.\n\nBoth regression tests failed against the previous implementation for the reported reasons and now pass. The focused Fix Loop tests pass 10/10, Reviewer resilience passes 7/7, the Reviewer module set passes 574/574, and Node/Sandbox typecheck, ESLint, and Prettier checks pass.

ewen-poch:
Addressed the hydration fail-open P1 in c6d958da.\n\nWorkspacePage now treats the scoped Review snapshot as unavailable until its existing hydration succeeds, and also fails closed when the existing load-error state is set. The conversation controller receives an explicit turn-admission guard and applies it to both immediate sends and queued drafts; this avoids misrepresenting unknown hydration state as an active Fix Loop or exposing an incorrect cancel affordance.\n\nThe regression tests reproduced all three public behaviors against the previous implementation: immediate send was enabled before hydration, queueing was enabled before hydration, and a cached snapshot remained sendable after a load failure. They now pass. The affected files pass 75/75 tests, the Reviewer module set passes 577/577, and Web typecheck, ESLint, and Prettier checks pass. No persisted state, schema, enum, or IPC contract was added.

ewen-poch:
Addressed both findings in a22fcd61.

- Review-history admission now gates immediate revisions, queued revisions, and branch-from-message availability until the scoped Review snapshot loads successfully. This keeps every existing turn-mutation path fail-closed without presenting unknown hydration state as an active Fix Loop.
- Startup recovery no longer infers Review-chain membership from createdAt ordering. It traverses the existing review_submission dispositions from the running Review through causeReviewId, so causally linked re-reviews are recovered while an independent Review with the same timestamp remains untouched.

The public WorkspacePage regressions failed against the previous implementation for both idle and running revision paths, and the repository regression failed because the same-timestamp historical finding became unaddressed. They now pass. The affected renderer tests pass 75/75, Reviewer resilience passes 7/7, the Reviewer module passes 429/429, and Web, Node, Sandbox typechecks, ESLint, Prettier, and diff checks pass. No table, migration, field, enum, or new persistence contract was added.

ewen-poch:
Addressed both admission gaps in e509cf0c:

- Manual review now fails closed while the scoped review history is unavailable. The disabled state and the request handler both enforce the guard, so a stale callback cannot bypass it.
- Context compaction now shares the same review-history availability gate. Its existing callback already rechecks the computed capability before invoking the runtime.

The regression tests failed before the fix because reviewer.run and compactContext were called during hydration, then passed after the guards were added. Focused validation: 77 tests passed, Web typecheck passed, and targeted ESLint/Prettier checks passed.

This does not add persistence, schema fields, lifecycle values, IPC, or new test seams.

ewen-poch:
Addressed both ReviewerCard findings in f4b19e29:

- Unresolved rerun eligibility now uses the exact submitted-check projection. Tracked items require a warn/fail assessment, a non-resolved disposition, and a source Finding that is still unresolved, so tracked-only terminal reviews receive the intended rerun action without reviving resolved history.
- A successful rerun result is returned only after main has created and pushed the replacement running Review row. At that point the superseded card removes its rerun notice instead of retaining a permanent Re-running latch; the new card owns progress and any later retry.

Both public render regressions failed against the previous code for the reported reasons and now pass. Focused ReviewerCard and SessionReviewerPanel validation passes 84/84, with Web typecheck, targeted ESLint/Prettier, and diff checks passing. No new copy, enum, persistence, data relationship, or cross-component state was added.

ewen-poch:
Rebased onto origin/main at ab034ba0 and resolved the migration conflicts in 8af8e421f.

The conflicts came from the first historical commit that temporarily introduced a Fix Loop 0025 migration and the following commit that removed it. The rebased history keeps main current 0025-managed-file-version foundation unchanged and retains the approved no-DB Fix Loop design; the final PR diff contains no database or migration files.

Post-rebase validation: affected renderer tests 254/254, Reviewer module 429/429, Web/Node/Sandbox typechecks, targeted ESLint/Prettier, and diff checks all pass. The branch was updated with an explicit force-with-lease against the previous remote head.

PR #2021, opened 2026-09-01
fix(message-scroll): eliminate inline-image scroll jitter with latching, geometry cache, and header-probed dimensions

roxi3906:
## Problem

When a session message contains an inline image (e.g. ``), scrolling upward causes violent per-frame jitter exactly when the bottom edge of the image is about to enter the viewport. The scrollbar also visibly grows while the image is offscreen and shrinks when scrolled back to it. After the first fix layer, one residual jump remains: an image never loaded in this app run still grows from the ~100px placeholder to its real height on first load.

## Root cause

`SessionArtifactImage` (`SessionMessageMarkdown.tsx`) gates `useManagedPreviewResource` behind `useNearViewport` (IntersectionObserver, 240px rootMargin): as soon as the image leaves the observer margin, the resource is **released and the `` unmounts**, swapped for an alt-text placeholder `<span>`. The placeholder (~100px) is far shorter than the loaded image (hundreds of px), which creates a self-sustaining oscillation:

1. Scrolling up, the image enters the 240px margin → `` mounts, height jumps, scrollHeight changes;
2. The MessageScroller's ResizeObserver anchoring compensation adjusts `scrollTop`, moving the image relative to the viewport;
3. The compensation pushes the image back outside the observer threshold → it unloads and the layout collapses → another anchoring pass → it lands inside the threshold again → remounts…

Each cycle also costs an acquire/release IPC round-trip, so it presents as per-frame jitter. The placeholder↔image height delta is exactly the observed scrollbar grow/shrink.

## Fix (three layers, additive)

**1. Lazy-load, but never unload (latch)** — once an image has been near the viewport, keep its resource acquired. Lazy loading is preserved (nothing loads before the first approach), but the unload → collapse → anchor → remount loop is gone.

**2. Cache image geometry, lock the placeholder size** — the latch fixes oscillation within a component's lifetime, but the transcript window only renders a window of messages: scroll far away and the whole message unmounts, so scrolling back collapses the placeholder to ~100px again. The `` `onLoad` records `naturalWidth` and aspect ratio per preview request key (which already embeds path + size + mtime, so in-place file replacements invalidate themselves), and the placeholder mirrors the final geometry with `width` + `max-width: 100%` + `aspect-ratio`, making the placeholder↔image swap scroll-neutral. The cache is a bounded module-level Map (200 entries, oldest evicted).

**3. Header-probed dimensions shipped with the preview resource** — layers 1-2 can't help an image never loaded in this run. The main process now probes pixel dimensions from image file headers and returns them as optional `width`/`height` on `ManagedPreviewResource`, so the renderer knows the exact geometry from the moment `acquire` resolves — before any image bytes are fetched or decoded:

- Converged with the shared raster-safety path: `acquire` already reads image headers (lease-safe, snapshot-verified) to enforce the 16MP decoded-pixel budget (`src/main/raster-image-safety.ts`, PNG/JPEG/GIF/WebP). This PR reuses that same header parse and attaches the dimensions as optional `width`/`height` on `ManagedPreviewResource` — zero extra I/O, no second parser. The shared JPEG parser gains EXIF orientation support (orientations 5-8 transpose the dimensions, matching the browser's default `image-orientation: from-image`), so phone-shot photos lock the box they actually display in. Non-raster types (SVG/TIFF/AVIF/BMP) ship no dimensions and fall through to the geometry cache. The preload contract adapter passes the result through untouched; the inline type in `web-service/task-api.ts` was extended to match.
- Dimensions are populated at both assembly points (`acquire`, including the trusted-lease path from #1204, and the Reviewer-only `acquireResolvedFile`) whenever the inferred MIME type is `image/*`. The preload contract adapter passes the result through untouched; the inline type in `web-service/task-api.ts` was extended to match.
- Renderer: the `` carries explicit `width`/`height` attributes (metadata first, geometry cache as fallback) with `h-auto` added to the image rule so `max-w-full` scaling never distorts the ratio; metadata also seeds the geometry cache, so a remounted placeholder locks the exact height from its first frame even if the previous instance's `` never decoded.
- Priority chain: metadata > geometry cache > the original ~100px placeholder. TIFF branch unchanged (fixed `aspect-ratio: 4/3` container).

### Artifact-metadata persistence: deliberately skipped

Persisting dimensions at artifact creation would require extending the Session-JSON `PersistedArtifact` schema, the `sanitizeArtifact` whitelist, `ArtifactFile`, and several main-process assembly/mapping points. The remaining unlock window without persistence is the local IPC round-trip between mount and `acquire` resolution (a few ms, no network, no decode) — not worth that blast radius. Revisit only if the flash proves visible in practice.

## Tests

- `raster-image-safety.test.ts`: EXIF orientation transposition (orientations 5-8, both TIFF endiannesses), non-transposing orientations, malformed EXIF segments degrading gracefully;
- `managed-preview-resources.test.ts`: acquire attaches probed dimensions for a real PNG header; forged content acquires cleanly without dimensions;
- `SessionMessageMarkdown.test.tsx`: the old "release on viewport exit" case became "stays mounted after viewport exit"; geometry-cache remount locking; metadata-driven `` attributes plus metadata-seeded placeholder locking **without any `load` event**;
- Related suites pass (managed-preview ipc/protocol, task-api, i18n guard — no new user-visible strings); `npm run typecheck` (node + web) and eslint are clean.

## Review-hardening follow-ups (folded into the third commit)

- **EXIF orientation**: JPEG SOF stores pre-rotation dimensions, but browsers apply EXIF orientation by default. The parser now reads the APP1 orientation tag and transposes width/height for orientations 5-8, so phone-shot photos lock the box they actually display in. On the renderer, the metadata seeds the geometry cache only when the entry is missing — a measured `onLoad` size always wins.
- **Never-throw fuzz**: 200 rounds of deterministic random buffers plus corrupt JPEG segment lengths, asserting the parser neither throws nor loops.

## Compatibility fix for the editable-versions identity requirement (fourth commit)

While validating this branch we hit a white screen when scrolling an artifact card into view. Root cause: the editable-artifact-versions work (#1204) made `createManagedPreviewRequest` throw when an acquire lacks the logical identity (`projectId` + managed file id), and the message artifact card path (`VisibleArtifactPreview`) never passed that identity into `ArtifactPreview`. The throw happens inside the reader's effect with no error boundary above it, so the whole React tree unmounts — a white screen. This affects main for any PDF/image/TIFF artifact card, not just sessions with old artifacts.

This PR therefore also:

- passes `resolvedProjectId` / `artifactId` through `VisibleArtifactPreview` and the offscreen card branch into `ArtifactPreview`;
- gates ArtifactPreview's reader-mounting branches (pdf/image/tiff) on identity presence, falling back to the static file-type tile for legacy artifacts persisted without it (text/csv/fasta branches render the fetched preview and never acquire, so they are intentionally not gated);
- gives inline session images the same identity in their acquire request, with an alt-text fallback when it is missing.

## Manual verification

1. Open a session containing a message with an inline image never opened in this app run;
2. scroll upward past the image: no jitter at the viewport edge, no flash-down/flash-up pair, scrollbar length stays constant;
3. scroll far away and back (transcript window recycling): the placeholder already has the final height, no jump.

---

*This PR absorbs the stacked follow-up (formerly #2051) so the whole image-scroll-jitter fix reviews as one unit.*

### DIFF
--- a/src/main/managed-preview-resources.test.ts
+++ b/src/main/managed-preview-resources.test.ts
@@ -273,6 +273,65 @@ describe('ManagedPreviewResources', () => {
     expect(close).toHaveBeenCalledOnce()
   })
 
+  it('attaches header-probed pixel dimensions when acquiring an image', async () => {
+    // Minimal PNG: signature + IHDR length/type + 640x400 dimensions.
+    const png = Buffer.alloc(33)
+    Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]).copy(png, 0)
+    png.writeUInt32BE(13, 8)
+    png.write('IHDR', 12, 'ascii')
+    png.writeUInt32BE(640, 16)
+    png.writeUInt32BE(400, 20)
+    const filePath = await createFile(png, 'plot.png')
+    const resources = new ManagedPreviewResources({
+      resolvePath: async () => filePath,
+      createId: () => 'resource-1'
+    })
+
+    const resource = await resources.acquire(17, { source: 'local', path: filePath })
+
+    expect(resource).toMatchObject({
+      mimeType: 'image/png',
+      width: 640,
+      height: 400
+    })
+  })
+
+  it('rejects forged image content whose header cannot be parsed', async () => {
+    const filePath = await createFile(Buffer.from('definitely not an image'), 'forged.png')
+    const resources = new ManagedPreviewResources({
+      resolvePath: async () => filePath,
+      createId: () => 'resource-1'
+    })
+
+    // The shared raster safety path fails closed for pixel-limited types: an unparseable header
+    // rejects the acquire rather than minting a dimension-less resource.
+    await expect(resources.acquire(17, { source: 'local', path: filePath })).rejects.toThrow(
+      /dimensions/i
+    )
+  })
+
+  it('skips the header probe for image formats the parser does not support', async () => {
+    // PNG magic behind an .svg name: SVG/TIFF/AVIF acquires must not pay for a probe that the
+    // parser cannot serve anyway.
+    const png = Buffer.alloc(33)
+    Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]).copy(png, 0)
+    png.writeUInt32BE(13, 8)
+    png.write('IHDR', 12, 'ascii')
+    png.writeUInt32BE(640, 16)
+    png.writeUInt32BE(400, 20)
+    const filePath = await createFile(png, 'vector.svg')
+    const resources = new ManagedPreviewResources({
+      resolvePath: async () => filePath,
+      createId: () => 'resource-1'
+    })
+
+    const resource = await resources.acquire(17, { source: 'local', path: filePath })
+
+    expect(resource.mimeType).toBe('image/svg+xml')
+    expect(resource.width).toBeUndefined()
+    expect(resource.height).toBeUndefined()
+  })
+
   it('inspects authoritative metadata without minting a resource capability', async () => {
     const filePath = await createFile(Buffer.from('office'))
     const createId = vi.fn(() => 'resource-1')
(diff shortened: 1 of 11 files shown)


roxi3906:
Reopening to force GitHub to recompute a stale mergeability state: the branch already contains the current main tip (c4fe46f75) and merges cleanly (verified with git merge-tree), so the CONFLICTING status cannot be real.

roxi3906:
Reopening to retrigger CI: the PR Gate run was cancelled mid-run with no replacement run in the concurrency group.

PR #2023, opened 2026-09-01
feat(settings): reveal resource details on interaction

ewen-poch:
## Problem

The About resource rows always showed supporting descriptions, giving primary labels and secondary guidance the same visual weight. The leading icons also stayed muted during interaction.

## Proposed change

- On hover-capable pointers, vertically center each resource title at rest and reveal its supporting description on pointer hover or keyboard focus.
- Change the leading icon from muted foreground to the semantic primary color in those interactive states.
- Keep descriptions visible on touch and coarse-pointer surfaces so hover is only progressive enhancement.
- Preserve row geometry by animating only transform, opacity, and color.

## Scope and non-goals

- Applies only to Help Center and Release notes in Settings > General > About.
- Does not generalize the pattern to ordinary settings rows.
- Adds no user-visible copy, dependencies, persisted data, enum values, or historical-data compatibility paths.

- About resource behavior -> npm test -- src/renderer/src/pages/settings/AppVersionSection.render.test.tsx src/renderer/src/pages/settings/GeneralPanel.render.test.tsx src/renderer/src/pages/settings/general-sections-i18n.render.test.tsx -> 29 tests passed.
- Source quality -> npm run lint -- --no-cache and focused Prettier check -> passed.
- Build coverage -> npm run build:web and npm run build:e2e -> passed.
- Interaction states -> local browser checks for pointer hover, keyboard focus-visible, touch/coarse pointer, light theme, and dark theme -> passed; row height remained stable.
- Settings visual baseline -> the focused visual test generated the updated settings-general baseline and passed in update mode; the baseline was manually inspected.

Consumer and platform lanes were excluded locally because this is a CSS-only behavior change inside AppVersionSection; ExternalTextLink interfaces, routing, application state, persistence, and native platform behavior are unchanged. Exact-head PR Gate remains authoritative.

Uncovered local risk: a non-update visual rerun stops before reaching Settings because the existing workspace-empty baseline lacks the current # sessions composer hint (2,265 pixels, about 1%). That unrelated baseline was not included in this PR.

Per project policy, verification remains pending until independent review confirms this impact mapping covers the final state.

## Review focus

- Whether hiding secondary descriptions only on hover-capable pointers preserves the intended information hierarchy.
- Keyboard focus parity and touch/coarse-pointer fallback.
- Semantic primary color behavior in light and dark themes.


ewen-poch:
Addressed the coarse-pointer finding in bf626c0b. The conceal/reveal rules now require both `(hover: hover)` and `(pointer: fine)`, so touch-primary and hybrid coarse-pointer devices keep the descriptions visible even if they also report hover support. The render test and Settings design note now cover that condition; the focused 29-test renderer set, `typecheck:web`, lint, Prettier, and `build:web` pass on the updated head.

ewen-poch:
Addressed the hybrid-touch finding in 70a73994. The desktop conceal rules still require `(hover: hover)` and `(pointer: fine)`, while an explicit `(any-pointer: coarse)` override forces the title translation to zero and description opacity to one. The override uses important only for those two presentation properties so it wins regardless of generated media-block order. The focused 29-test renderer set, `typecheck:web`, lint, Prettier, and `build:web` pass; the built CSS was also checked to confirm both coarse-pointer override rules are emitted.

PR #2024, opened 2026-09-01
test(e2e): suppress GitHub star nudge on macOS journeys

ewen-poch:
## Problem

macOS workspace E2E can fail under `--fail-on-flaky-tests` when the GitHub star nudge popover opens 5s after the workspace badge is visible. The popover is a Radix layer that can swallow the next pointer click, so tests such as `edits and navigates message revisions that persist after relaunch` stay on revision `2/2` after clicking Previous.

Windows E2E already records the nudge cooldown in the Electron fixture. macOS was left on the default cooldown so layout timing would match earlier CI; that leaves macOS journeys exposed. Observed on [PR 2018](https://github.com/aipoch/open-science/pull/2018) (retry passed; Playwright classified the case as flaky).

## Proposed change

- Record the same `open-science:github-star-nudge-last-shown-at` cooldown on every E2E platform after the window is ready.
- Also register it as a Playwright init script so later reloads keep the cooldown.
- Share that helper with the workspace project-switcher journeys instead of duplicating the key.

Production `GitHubStarBadge` behavior is unchanged.

## Scope and non-goals

- E2E fixture and the two project-switcher call sites only.
- No product UI, i18n, ACP, session, or persistence-format changes.
- Does not try to stabilize unrelated workspace flakes (for example tool-layout screenshot timing).

## Historical compatibility

None. No user data format, migration, or upgrade path is involved.

## New states / enum values

None. No production state machine or enum is extended.

## Data persistence

E2E-only. The fixture writes the existing renderer `localStorage` cooldown key inside the isolated Playwright `user-data-dir`. That profile is discarded with the test. Production profiles and SQLite/session storage are not read or written.

The listed checks ran after the last material edit:

| --- | --- | --- |
| Fixture cooldown helper and platform-wide suppression | `npx vitest run scripts/electron-app-fixture.test.ts` | 7 passed |
| GitHubStarBadge cooldown key contract (unchanged production) | `npx vitest run src/renderer/src/components/GitHubStarBadge.render.test.tsx` | 6 passed |
| Lint of the three changed files | `npx eslint e2e/fixtures/electron-app.ts e2e/workspace-project-switcher.spec.ts scripts/electron-app-fixture.test.ts` | pass |
| Node/e2e typecheck | `npm run typecheck:node` | pass |

Full `npm test` was not run locally. `e2e/fixtures/**` is unclassified in the change-impact map, so PR Gate should fail closed to the full plan, including macOS and Windows workspace E2E.

### Included / excluded lanes

- Included locally: fixture unit tests, related GitHubStarBadge render tests, eslint, `typecheck:node`.
- Excluded locally: Electron Playwright journeys (require `build:e2e` and the macOS CI runner). PR CI is the authority for `e2e_workspace_macos`, `e2e_workspace_windows`, visual, and accessibility.
- Visual darwin baselines already hide the GitHub link and do not include the nudge popover, so suppressing it should not require snapshot updates.

## Review focus

Whether recording the cooldown on macOS (not only Windows) is the right trade-off versus keeping the live nudge in macOS layout timing.

## Uncovered risks

- The original flake was not reproduced in a live Electron window here; CI workspace journeys are the confirmation.
- If a journey launches a window without `openMainWindow`, it would still need to call `suppressWorkspaceStarNudge` itself. Current Electron tests go through that helper.

### DIFF
--- a/e2e/fixtures/electron-app.ts
+++ b/e2e/fixtures/electron-app.ts
@@ -19,6 +19,21 @@ const FAKE_REMOTEIT_PATH = resolve(APP_ROOT, 'e2e', 'fixtures', 'fake-remoteit.c
 const FAKE_PROVIDER_NAME = 'Electron E2E provider'
 type E2eWindowMode = 'hidden' | 'normal'
 
+// Keep in sync with GitHubStarBadge. Workspace variant waits 5s, then opens a popover that can
+// swallow the next pointer click (revision navigation, project menus) on macOS and Windows CI.
+const STAR_NUDGE_LAST_SHOWN_STORAGE_KEY = 'open-science:github-star-nudge-last-shown-at'
+
+const recordStarNudgeCooldown = (storageKey: string): void => {
+  window.localStorage.setItem(storageKey, String(Date.now()))
+}
+
+const suppressWorkspaceStarNudge = async (
+  page: Pick<Page, 'addInitScript' | 'evaluate'>
+): Promise<void> => {
+  await page.addInitScript(recordStarNudgeCooldown, STAR_NUDGE_LAST_SHOWN_STORAGE_KEY)
+  await page.evaluate(recordStarNudgeCooldown, STAR_NUDGE_LAST_SHOWN_STORAGE_KEY)
+}
+
 const electronLaunchTarget = (
   userDataRoot: string,
   environment: NodeJS.ProcessEnv = process.env,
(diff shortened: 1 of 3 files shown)


ewen-poch:
Addressed the Codex P1 and the macOS/Windows workspace CI failure in 1d68c923.

The `addInitScript` path was too late for the first document *and* too broad: Playwright also injected it into isolated source-preview documents, where `localStorage` throws `SecurityError: Access is denied for this document`. That pageerror is what failed `shows the Electron failure reason when a source request fails` on both platforms.

The fixture now writes the cooldown only via `page.evaluate` on the app window (try/catch, matching `GitHubStarBadge`), then reloads so any already-scheduled workspace timer is dropped and the badge remounts with the cooldown already set. No production persistence, enum, or UI change.

PR #2030, opened 2026-09-01
fix(startup): exit Electron after UI startup failure

ewen-poch:
## Problem

A rejected normal UI startup only set `process.exitCode = 1`. That exit code is observed only when the event loop drains naturally, so a ready or partially initialized Electron process can retain windows, tray state, IPC resources, or the single-instance lock and remain resident without a usable window.

The main-entry regression test reproduces the reported termination gap. Before the fix, diagnostics reporting completed but `app.exit(1)` was never called. A second regression test covers failures after runtime composition; before the cleanup change, the startup promise rejected immediately without disposing the composed application surfaces.

## Proposed change

- Record the top-level startup failure before flushing diagnostics.
- After runtime composition has transferred ownership, await the existing bounded application-surface shutdown before propagating a later startup failure.
- Make the shared application-surface shutdown idempotent so concurrent lifecycle and startup-failure requests join one cleanup operation.
- Terminalize the pre-lifecycle OS-shutdown relay before failure cleanup so its success-code force-exit fallback cannot preempt diagnostics and `app.exit(1)`.
- Await the existing bounded diagnostics flush, then call `app.exit(1)` for the normal Electron UI path.
- Keep all MCP subprocess modes on their existing `process.exitCode` behavior.

## Scope and non-goals

- The startup coordinator gains one optional post-prepare failure-cleanup hook; normal lifecycle shutdown reuses the same ordered cleanup owner.
- No renderer copy or user interaction changes.
- No data relationship, persistence, schema, or model changes.
- No MCP subprocess lifecycle changes.

Local validation after the last material edit:

- Focused startup, runtime-shutdown, lifecycle, and diagnostics coverage -> `npx vitest run src/main/index-fatal-errors.test.ts src/main/index-startup-order.test.ts src/main/app-startup.test.ts src/main/application-runtime.test.ts src/main/app-lifecycle.test.ts src/main/diagnostics/startup.test.ts` -> 6 files, 114 tests passed.
- The OS-shutdown race regression test failed twice on the unfixed path because the success-code `forceExit` callback fired during failure cleanup, then passed after relay cancellation was restored.
- Main-process TypeScript remains valid -> `npm run typecheck:node` -> passed.
- Source and test lint rules remain valid -> `npm run lint` -> 0 errors; one pre-existing Prettier warning remains in the unchanged `scripts/ci/module-impact-shadow.test.ts`.
- Changed files match repository formatting -> focused `npx prettier --check ...` -> passed.
- Patch whitespace validation -> `git diff --check` -> passed.

The complete Vitest suite and cross-platform packaged behavior were not run locally; exact-head PR Gate remains authoritative for those risks.

## Review focus

- Post-composition startup failures await the same bounded ACP/Notebook, Web, Remote Access, and IPC cleanup used by normal lifecycle shutdown.
- Concurrent cleanup requests execute the disposer chain only once.
- A pending startup OS-shutdown fallback is cancelled before fatal cleanup begins.
- The fatal log entry and bounded diagnostics flush complete before `app.exit(1)`.
- Forced exit remains limited to the normal UI startup rejection path; MCP modes retain natural Node subprocess termination.

## Compatibility notes

- Historical data compatibility: none.
- New persisted fields or enum values: none.
- Data persistence: none.
- Data model or schema changes: none.
- User interaction changes: none.

### DIFF
--- a/src/main/app-startup.test.ts
+++ b/src/main/app-startup.test.ts
@@ -257,6 +257,81 @@ describe('orchestrateAppStartup', () => {
     expect(diagnostics.fail).toHaveBeenCalledWith(failure)
     expect(diagnostics.complete).not.toHaveBeenCalled()
   })
+
+  it('cleans the prepared runtime before rejecting a post-composition startup failure', async () => {
+    const failure = new Error('mark ready failed')
+    const cleanup = deferred<void>()
+    const events: string[] = []
+    const deps = makeDeps({
+      markReady: vi.fn(() => {
+        throw failure
+      }),
+      cleanupAfterStartupFailure: vi.fn(async () => {
+        events.push('cleanup:start')
+        await cleanup.promise
+        events.push('cleanup:end')
+      })
+    })
+
+    const startupOutcome = orchestrateAppStartup(deps).then(
+      () => undefined,
+      (error: unknown) => {
+        events.push('startup:rejected')
+        return error
+      }
+    )
+
+    await vi.waitFor(() => expect(events).toEqual(['cleanup:start']))
+    cleanup.resolve()
+
+    await expect(startupOutcome).resolves.toBe(failure)
+    expect(events).toEqual(['cleanup:start', 'cleanup:end', 'startup:rejected'])
+    expect(deps.cleanupAfterStartupFailure).toHaveBeenCalledWith({ tag: 'ctx' }, failure)
+  })
+
+  it('cancels the startup system-shutdown fallback before failure cleanup', async () => {
+    vi.useFakeTimers()
+    try {
+      const failure = new Error('mark ready failed')
+      const cleanup = deferred<void>()
+      const cleanupStarted = deferred<void>()
+      const forceExit = vi.fn()
+      let requestSystemShutdown = (): void => {}
+      const deps = {
+        ...makeDeps({
+          prepare: vi.fn(async () => {
+            requestSystemShutdown()
+            return { tag: 'ctx' }
+          }),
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed the P2 in `e434a2953`. Post-composition startup failures now await the same bounded application-surface shutdown used by the Electron lifecycle before the original startup error reaches the top-level `app.exit(1)` path. The shutdown callback is memoized, so concurrent lifecycle and startup-failure requests share one ACP/Notebook, Web, Remote Access, and IPC disposal operation. Added regression coverage for cleanup-before-rejection and idempotent concurrent shutdown; the focused six-file main-process suite passes 113 tests, with `npm run typecheck:node`, `npm run lint`, Prettier, and `git diff --check` also passing.

ewen-poch:
Addressed the P1 in `38b7b1818`. The startup system-shutdown relay now has a terminal cancellation operation, and `orchestrateAppStartup()` invokes it synchronously when startup fails, before diagnostics finalization or any awaited runtime cleanup. This clears the pending success-code force-exit timer and ignores later relay signals, so failure reporting and the final `app.exit(1)` cannot be preempted by the pre-lifecycle OS-shutdown fallback. The regression test failed twice on the unfixed path because `forceExit` fired during cleanup, then passed after the cancellation was restored. The focused six-file suite passes 114 tests; Node typecheck, changed-file Prettier, and `git diff --check` pass, and lint reports 0 errors with one unrelated existing warning in an unchanged CI test file.

PR #2031, opened 2026-09-01
fix(side-chat): include active replies in exit detection

ewen-poch:
## Problem

Side Chat owns an independent runtime and can be generating a reply while the main Agent, delegated work, Notebook, and Reviewer are idle. Its running state was not included in `detectActiveSessions`, so an ordinary quit treated the app as idle, skipped the existing confirmation, and interrupted the Side Chat turn during shutdown.

The existing shutdown path persists generated transcript content with an `interrupted` lifecycle, but the in-flight inference is still cancelled without advance notice.

## Proposed change

- Project Side Chat snapshots whose existing `running` flag is true into parent Session identities.
- Add those identities to the shared active-session aggregator as existing `agent` rows.
- Deduplicate root Agent and Side Chat work for the same parent Session, while preserving delegated work's stronger blocking policy.
- Feed the same projection to quit/close, storage migration, update, and archive activity checks.

## Scope and non-goals

- No second confirmation system or Side Chat-specific dialog.
- No new `ActiveSessionInfo.kind` value or other enum state.
- No new persisted field, database migration, lifecycle value, or serialized-format change.
- No ACP method, event, provider adapter, model path, or protocol behavior changes; all Side Chat frameworks use the same runtime-owner snapshot.

The regression test was added before the production fix and run three times on the unfixed code:

`npm test -- src/main/storage/detect-active.test.ts -t 'includes a running Side Chat in active Session detection'`

Each run failed deterministically because the detector returned `[]` instead of the Side Chat parent Session, matching the reported missing-confirmation condition.

- Side Chat-only active detection and source deduplication -> `npm test -- src/main/storage/detect-active.test.ts` -> 5 passed.
- Storage IPC exposes the Side Chat parent Session through the existing contract -> focused storage IPC regression -> passed.
- Quit/confirmation behavior, storage detection, and Side Chat shutdown persistence -> `npm test -- src/main/storage/detect-active.test.ts src/main/storage/ipc.test.ts src/main/storage/command-owner.atomicity.test.ts src/main/side-chat/runtime-owner.test.ts src/main/app-lifecycle.test.ts src/main/window-close-confirm.test.ts` -> 6 files, 267 tests passed.
- Main-process and sandbox types -> `npm run typecheck:node` -> passed.
- Source lint -> `npm run lint` -> passed with one pre-existing Prettier warning in unchanged `scripts/ci/module-impact-shadow.test.ts`.

`npm run test:affected:explain -- --base origin/main --head HEAD` selects the full fallback because `src/main/ipc.ts` has no declared module owner. Per maintainer instruction, the complete local `npm test` suite was not run; PR CI is authoritative for the remaining portable and platform lanes.

## Review focus

- Confirm that `running` Side Chat snapshots are the correct authority and dormant/open-but-idle Side Chats remain non-blocking.
- Confirm mapping Side Chat work to the existing parent Session `agent` row is preferable to expanding the shared enum and renderer surfaces.
- Confirm the shared aggregation coverage for quit, migration, update, and archive gates is appropriate.

### DIFF
--- a/src/main/application-command-wiring.test.ts
+++ b/src/main/application-command-wiring.test.ts
@@ -152,6 +152,19 @@ describe('production application command wiring', () => {
     )
   })
 
+  it('holds side chat admission through the in-place update handoff', () => {
+    const updateGate = compact(
+      between(ipcSource, 'const durableBackendHandoffGate', 'const detectResearchBlockers')
+    )
+    const updateStrategy = compact(
+      between(ipcSource, 'const updateStrategy', 'const updateCommandOwner')
+    )
+    expect(updateGate).toContain(
+      'shutdownCoordinator.runForUpdateGate(UPDATE_SHUTDOWN_BUDGET_MS, { holdSideChatAdmission: true })'
+    )
+    expect(updateStrategy).toContain('releaseInstallHandoff: abortUpdateHandoff')
+  })
+
   it('keeps native-only commands inside the Electron owner adapter and exposes only narrow views', () => {
     const electronOwner = compact(
       between(dependencyBlock, 'electron: {', 'events: applicationEvents')
--- a/src/main/application-runtime.test.ts
+++ b/src/main/application-runtime.test.ts
@@ -301,6 +301,10 @@ describe('application surface shutdown', () => {
         notebook: {
           dispose: async () => ({ reaped: true }),
           shutdownAll: async () => ({ reaped: true })
+        },
+        sideChat: {
+          shutdown: async () => undefined,
+          suspendAll: async () => undefined
         }
       })
       const runtime = await composeApplicationRuntime(async (modules) => {
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -4040,9 +4040,9 @@ const createApplicationModules = async (
     }
   }
 
-  // The shared coordinator remains the sole ACP + Side Chat + Notebook teardown owner. Register
-  // command routing after it so reverse disposal removes adapters, then the router, before any
-  // underlying owner stops.
+  // The shared coordinator remains the sole ACP + Notebook teardown owner.
+  // It also coordinates Side Chat suspension/shutdown. Register command routing after it so reverse
+  // disposal removes adapters, then the router, before any underlying owner stops.
   await modules.add({ shutdownCoordinator }, ({ shutdownCoordinator: coordinator }) => ({
     name: 'backend-shutdown-coordinator',
     capability: undefined,
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -3055,7 +3055,10 @@ const createApplicationModules = async (
     log: createLogger('shutdown')
   })
   const durableBackendHandoffGate = createDurableInstallGate(
-    () => shutdownCoordinator.runForUpdateGate(UPDATE_SHUTDOWN_BUDGET_MS),
+    () =>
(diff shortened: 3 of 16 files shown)


ewen-poch:
Addressed the migration lifecycle finding in 86facfd22. The shared backend handoff coordinator now invokes a reusable Side Chat suspension that cancels active replies, drains queued persistence, and saves the interrupted conversation before data-root copying can begin. A suspension failure makes the handoff readiness fail closed; unlike terminal shutdown, successful suspension leaves the Side Chat owner reusable if the handoff is aborted. The coordinator regression failed three consecutive times before the fix (suspendAll called 0 times) and now passes, alongside the runtime resume-after-suspension test.

ewen-poch:
Addressed the fail-fast teardown finding in 72bf1e8e1. The runtime/Side Chat aggregation now waits for both operations with all-settled semantics; a Side Chat failure maps to reaped:false, while a still-pending main runtime remains pending until the shared shutdown budget expires. The regression (Side Chat rejects first while main runtime hangs) failed three consecutive times before the fix because the coordinator settled before 24ms of a 25ms budget, and now passes. I also restored the production wiring test marker in f07359130 after CI identified that comment-level compatibility contract.

ewen-poch:
Addressed the data-root handoff admission race in 08e50d368. The Side Chat drain now accepts a data-root-only hold mode: after active turns are cancelled and persisted, new Side Chat prompts remain blocked until the handoff either commits/exits or the existing abort path releases the hold. Update-gate suspension keeps its prior reusable behavior. Preparation refusal, cancellation, failed migration, discard, and uncommitted return paths all release the hold. The regression failed 3/3 on the previous head because a prompt resolved after suspension instead of being rejected during handoff; it now passes. Focused verification: 5 files / 196 tests passed, typecheck:node passed, lint completed with 0 errors (one unrelated existing Prettier warning).

ewen-poch:
Addressed the in-flight Side Chat dispatch race in 477dda167. Public follow-up sends are now tracked until they settle, and the existing drain waits for the sends that were already admitted before the handoff. Dispatch rechecks the suspension/ownership boundary after queued persistence, reconnect, resume, and prompt persistence, so a send that has not installed a provider turn cannot continue against a closing runtime. If the user prompt was already durably appended, suspension preserves it with the interrupted lifecycle. The drain starts runtime suspension before joining dispatches, so an already-installed or never-admitted provider turn still follows the existing cancellation and bounded-timeout path instead of deadlocking teardown. The regression failed three consecutive times on the previous head because suspension completed while the admitted send still awaited provider admission; it now passes, along with 5 focused files / 197 tests, typecheck:node, lint (0 errors; one unrelated existing warning), and Prettier. No persistence schema, IPC contract, enum, or user-interaction changes were introduced.

ewen-poch:
Addressed in 76a6f1e46.

- The in-place update gate now holds Side Chat admission across backend teardown and renderer durability confirmation.
- ElectronUpdaterStrategy now releases an aborted install handoff after gate refusal or failure, interrupted preparation, a synchronous quitAndInstall failure, or an asynchronous updater error. A successful installer handoff keeps admission held through quit and relaunch.
- The production cleanup resumes Side Chat admission and notifies the renderer that the durability handoff was aborted.

Regression coverage verifies the installer-failure cleanup at the public strategy boundary, option forwarding through createUpdateStrategy, and the production update-gate wiring. Before the fix, the installer-failure assertion failed consistently with 0 cleanup calls; after the fix, all 142 related tests pass, along with Node/Sandbox type checks, targeted ESLint, and Prettier.

PR #2033, opened 2026-09-01
fix(ci): check PR formatting from the merge-base

ewen-poch:
## Problem

PR Gate on #2030 failed two selected lanes that the startup-exit change did not own:

- **Format** used a two-dot `git diff` of current `main` against the PR tip. Because that branch was not rebased, Prettier received unrelated files from later `main` commits, including deleted paths (`src/main/session-artifact-file-resolver.ts`) and an unformatted line in `scripts/ci/module-impact-shadow.test.ts`.
- **macOS native units** failed `allows a listed destination and blocks it after a live policy update`. The Seatbelt child invoked `/bin/bash -c`, which read `/Users/runner/.bashrc`. That filesystem deny ran before the policy gateway could annotate `OPEN_SCIENCE_NETWORK_DOMAIN_BLOCKED`. Closing the previous command's gateway is also asynchronous, so the next wrap could race the listener.

This PR does not change #2030. After it lands, a re-run of #2030's merge commit will pick up the fixes.

## Proposed change

- Format checks classify `merge-base...head` (the pull request's own files) and skip paths missing from the work tree.
- Reformat the already-landed `reasonChains` assertion so `main` itself is Prettier-clean.
- Launch bash/zsh under Seatbelt with `--noprofile --norc`.
- Wait for in-flight gateway close work before wrapping the next command.

## Scope and non-goals

- No historical-data compatibility requirement.
- No new state or enum values.
- No data-model, relationship, or persisted-format changes.
- No change to PR #2030 (`fix/electron-startup-fatal-exit`).
- Classifier/authority still uses two-dot `base..head` for lane selection; that remains a follow-up. This PR only fixes the format checker and the macOS sandbox failure that blocked #2030.

- Format selection uses merge-base and ignores missing work-tree paths -> `npm test -- scripts/ci/check-changed-format.test.ts` -> 1 file passed, 2 tests passed.
- Unformatted `reasonChains` assertion is wrapped -> `npx prettier --check scripts/ci/module-impact-shadow.test.ts` -> passed.
- macOS bash launch omits rc files; next wrap waits for gateway close -> `npm run test:module -- notebook_network_sandbox` -> 12 files passed, 81 tests passed, 1 skipped.
- Changed TypeScript remains valid -> `npm run typecheck:node` -> passed.
- Changed files satisfy ESLint -> `npx eslint` on the edited files -> passed.

The complete Vitest suite and cross-platform packaged behavior were not run locally; PR CI is authoritative for those risks.

## Review focus

- Confirm format checks match GitHub's three-dot PR file list rather than later `main` commits.
- Confirm Seatbelt still uses `localhost` as the only network host (sandbox-exec rejects `127.0.0.1` in `(remote ip ...)`).
- Confirm forced `app.exit` work in #2030 is untouched.

## Uncovered risks

- `classify-pr-changes.mjs` / `module-impact-authority.mjs` still two-dot-diff current `main`, so a stale PR can still select a full Gate plan. That is conservative over-testing, not a false pass.
- The Electron-as-Node launch integration test needs a fully installed `electron` binary; worktree `npm install` skipped that script here. CI installs it.

### DIFF
--- a/packages/notebook-network-sandbox/runtime/src/notebook-runtime.ts
+++ b/packages/notebook-network-sandbox/runtime/src/notebook-runtime.ts
@@ -164,6 +164,7 @@ const initialize = async (config: NetworkRuntimeConfig, ask: NetworkAskCallback)
 const wrap = async (
   request: NetworkWrapRequest
 ): Promise<{ argv: string[]; env: NodeJS.ProcessEnv }> => {
+  if (finishing.size > 0) await Promise.allSettled([...finishing])
   const config = runtimeConfig
   if (!config) throw new Error('Notebook process runtime is not initialized.')
   const filesystem = normalizeFilesystemLayout({
--- a/packages/notebook-network-sandbox/runtime/src/platform/macos-isolation.ts
+++ b/packages/notebook-network-sandbox/runtime/src/platform/macos-isolation.ts
@@ -1,4 +1,5 @@
 import { homedir, tmpdir } from 'node:os'
+import { basename } from 'node:path'
 
 import { findExecutable } from './executable.js'
 import {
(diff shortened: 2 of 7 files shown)


ewen-poch:
**[P1] zsh startup flags:** agreed. `--noprofile` / `--norc` are Bash-only; `zsh` rejects them (`zsh: no such option: noprofile`) and would exit before the wrapped command.

macOS Seatbelt now launches:
- bash: `--noprofile --norc -c`
- zsh: `-d -f -c` (`NO_GLOBAL_RCS` + `NO_RCS`, including `/etc/zshenv`)

The default sandbox shell is still `/bin/bash`. Added a zsh argv regression test in `filesystem-policy.test.ts`.

PR #2037, opened 2026-09-02
fix(windows): restore hosted full-test suite portability

ewen-poch:
## Problem

Scheduled [Windows Full Test](https://github.com/aipoch/open-science/actions/workflows/windows-full-test.yml) has failed on every `main` run since 2026-08-27. Latest run 33564211594 failed all three Vitest shards (2 + 101 + 27 tests). The Notebook sandbox job stayed green.

PR Gate does not run the complete portable suite on Windows, so host-specific regressions accumulated on `main`.

## Proposed change

Fix the Windows-host defects and fixtures that the advisory full suite actually executes:

- Split linux `PATH` with `posix.delimiter` / `posix.isAbsolute` even when the test host is Windows.
- Fsync exclusive file-evidence receipts on the write handle (Windows `FlushFileBuffers` rejects `O_RDONLY`, matching `file-durability.ts`).
- Close CLI launcher identity handles before NTFS `rename` / `unlink`.
- Give provisioner unit tests a tiny injected path budget so `%TEMP%` fixtures are not rejected as over-budget data roots.
- Canonicalize 8.3 vs long paths, skip Unix mode bits and parent-directory-swap TOCTOU cases that NTFS cannot reproduce, keep `PATH` for `node.exe` in the runtime-profile launcher test, and retry SQLite `rm` after disconnect.

## Scope and non-goals

- No workflow change, so no CI dry-run of `windows-full-test.yml`.
- No MAX_PATH budget or micromamba cache layout change.
- No Session JSON, SQLite schema, or migration change.

## Historical compatibility / new states / persistence

- **Historical compatibility:** none. CLI still recognizes the historical ownership marker; only the Windows replace/remove sequence changes (close the identity handle, then mutate).
- **New enum / status values:** none. Version-file parent-swap coverage stays POSIX-only; production error codes are unchanged.
- **Data persistence:** file-evidence on-disk format is unchanged. Exclusive writes now fsync a writable handle so Windows can actually flush receipts/blobs (same contract as `src/main/storage/file-durability.ts`). CLI launcher bytes and PATH journal format are unchanged.

| --- | --- | --- |
| PATH delimiter / CA relative path | `npx vitest run src/main/notebook/process-environment.test.ts src/main/notebook/trust-bundle.test.ts` | pass |
| CLI replace/remove | `npx vitest run src/main/cli-install/launcher.test.ts` | pass |
| Prompt snapshot modes | `npx vitest run src/main/acp/prompt-content-owner.test.ts` | pass |
| Filesystem physical paths | `npx vitest run packages/notebook-network-sandbox/src/filesystem-policy.test.ts` | pass |
| Runtime profile launcher | `npx vitest run scripts/performance/run-runtime-profile.test.ts` | pass |
| Provisioner path budget | `npx vitest run src/main/notebook/provisioner.test.ts` | pass |
| Version parent-swap skip | `npx vitest run src/main/managed-file-versions/version-file-operator.test.ts` | pass |
| Session projection cleanup | `npx vitest run src/main/session-persistence/projection.test.ts` | pass |
| File-evidence fsync consumers | `npx vitest run src/main/notebook/working-file-observer.test.ts src/main/notebook/input-registry.test.ts src/main/compute/harvest-engine.test.ts src/main/compute/job-dispatcher.test.ts src/main/compute/compute-job-workflow-owner.test.ts src/main/compute/compute-service.test.ts src/main/compute/concurrency-integration.test.ts` | pass |
| Lint | `npm run lint` | pass (no new issues) |

Full `npm test` was not run locally; PR Gate is authoritative for the portable suite. `windows-runtime-cache-uninstall.test.ts` is `skipIf` off Windows.

Uncovered risks: this machine cannot execute NTFS `fsync`/`rename` or 8.3 path expansion. Confirm on PR Gate Windows-sensitive lanes, and optionally `workflow_dispatch` Windows Full Test on this branch (`mode: full`).

## Review focus

- `resources/notebook/file_evidence_worker.js` write-handle fsync vs the existing durability contract.
- `src/main/cli-install/launcher.ts` close-before-rename/unlink on Windows.
- `environmentPathRoots` using the *target* platform's delimiter, not the host's.

### DIFF
--- a/packages/notebook-network-sandbox/src/filesystem-policy.test.ts
+++ b/packages/notebook-network-sandbox/src/filesystem-policy.test.ts
@@ -37,8 +37,8 @@ describe('Notebook filesystem policy', () => {
         deniedWriteRoots: []
       })
     ).toMatchObject({
-      readOnlyRoots: [realpathSync(root)],
-      readWriteRoots: [realpathSync(nested)]
+      readOnlyRoots: [absolutePhysicalPath(root)],
+      readWriteRoots: [absolutePhysicalPath(nested)]
     })
   })
 
--- a/resources/notebook/file_evidence_worker.js
+++ b/resources/notebook/file_evidence_worker.js
@@ -18,7 +18,6 @@ const {
   rmSync,
   statSync,
   statfsSync,
-  writeFileSync,
   writeSync
 } = require('node:fs')
 const { join } = require('node:path')
@@ -132,13 +131,22 @@ const readRegularFile = (name, maxBytes) => {
 const readJson = (name, maxBytes = MAX_INTERNAL_JSON_BYTES) =>
   JSON.parse(readRegularFile(name, maxBytes).toString('utf8'))
 const writeExclusiveFile = (name, contents) => {
-  writeFileSync(name, contents, {
-    encoding: 'utf8',
-    flag: constants.O_CREAT | constants.O_EXCL | constants.O_WRONLY | constants.O_NOFOLLOW,
-    mode: 0o600
-  })
-  const descriptor = openSync(name, constants.O_RDONLY | constants.O_NOFOLLOW)
+  // Windows FlushFileBuffers requires a writable handle. Reopening the exclusive file with
+  // O_RDONLY makes fsync fail with EPERM even when the file itself is writable — the same
+  // contract as src/main/storage/file-durability.ts.
+  const descriptor = openSync(
+    name,
+    constants.O_CREAT | constants.O_EXCL | constants.O_WRONLY | constants.O_NOFOLLOW,
+    0o600
+  )
   try {
+    const bytes = Buffer.from(contents, 'utf8')
+    let offset = 0
+    while (offset < bytes.length) {
+      const written = writeSync(descriptor, bytes, offset, bytes.length - offset)
+      if (written <= 0) throw new Error('Exclusive file-evidence write made no progress.')
+      offset += written
+    }
     fsyncSync(descriptor)
   } finally {
     closeSync(descriptor)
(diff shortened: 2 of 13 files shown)


ewen-poch:
### Codex P1 — launcher close-before-rename TOCTOU

Adopted on POSIX, not on Windows.

`MoveFileEx(MOVEFILE_REPLACE_EXISTING)` cannot replace an open destination on NTFS, which is why the hosted Windows suite failed with `EPERM` on rename/unlink. Closing the validated handle is required there after the identity checks.

POSIX can rename/unlink an open file, so the previous unified close-before-replace was a lock regression. `f968f836d` restores the original ownership lock on POSIX: the validated handle stays open through `rename` and `rm`. Windows still closes first.

The remaining Windows window is a local race against another writer in the same user bin directory. Closing that gap needs an exclusive share mode or `ReplaceFile`-style API that Node does not expose here. That would be a new host-specific replacement path, not a small follow-up to this portability fix, so it is out of scope for this PR.

ewen-poch:
### Codex P1 (round 2) — Windows replace/uninstall after close

Same residual NTFS constraint as the previous comment. No further code change in this PR.

POSIX again holds the validated handle through `rename`/`rm`. Windows still cannot `MoveFileEx(REPLACE_EXISTING)` onto an open destination, which is the `EPERM` the hosted suite hit.

The two recommended alternatives are out of scope here:

- A native exclusive-share / `ReplaceFile` path is a new Windows replacement primitive, not a portability fix.
- Aborting instead of replace/unlink would make CLI install and uninstall fail closed on Windows, i.e. the launcher could not be updated or removed.

Unlink-then-publish while the old handle is held is also crash-unsafe: if the second publish fails, the directory entry is already gone.

The remaining Windows window is a local race against another writer in the same user bin directory. Leaving that as a known NTFS limitation for this PR.

PR #2039, opened 2026-09-02
perf(database): batch migration recovery snapshots

ewen-poch:
## Problem

All 25 database migrations require an application-time backup. A multi-version upgrade therefore
ran `VACUUM INTO` before every pending migration, writing roughly one database-sized snapshot per
migration even though retention later removed most of those files.

As persisted task, review, usage, and file history grows, this multiplies upgrade time and disk I/O
by the number of pending migrations.

## Proposed change

- Create or reuse one recovery snapshot before the first backup-required migration in each
  migration-service invocation.
- If a migration batch partially commits and a later migration fails, create a new snapshot before
  the first remaining migration on the next invocation.
- Keep the existing two-file retention policy, backup validation, naming, and logging.
- Update database migration, logging, integration, historical upgrade, retry, and retention tests
  to assert the batch-level policy.

## Scope and non-goals

- Startup continues to run the full SQLite `integrity_check`; the proposed `quick_check` path was
  removed after review because its index-content coverage trade-off was too high.
- No Prisma schema, SQLite DDL, migration ID, checksum, ledger row, or application data-model change.
- No new persisted field, enum, crash marker, or lifecycle state.
- No renderer, IPC, startup-state, or other user-interaction change.
- Existing `.before-<migration-id>.backup` files, validation, naming, logging, and retention remain
  compatible.
- A continuous upgrade now retains its batch-start recovery point rather than an intermediate
  snapshot for every migration. Restoring it can require replaying more migrations.
- No automatic restore or destructive-migration checkpoint metadata is added.

All listed local checks ran after the last material edit.

- A multi-migration upgrade creates one batch snapshot; failed partial upgrades create a new
  snapshot on the next invocation; backup reuse, pruning, deletion, logging, and legacy contents
  remain covered -> `vitest run src/main/database/*.test.ts` -> 14 files, 151 passed.
- Packaged immutable migration identities and PR policy remain valid ->
  `vitest run scripts/database-migration-ledger-smoke.test.ts scripts/ci/check-pr-policy.test.ts`
  -> 2 files, 31 passed.
- Main-process types and notebook sandbox types -> `npm run typecheck:node` -> pass.
- Source lint -> `npm run lint` -> pass.
- Whitespace -> `git diff --check` -> pass.

The complete local `npm test` suite was intentionally not run per maintainer direction. PR Gate is
the final authority for the complete portable suite and selected platform lanes.

No renderer or Web consumer tests are included because this change does not alter IPC, shared
contracts, startup states, or UI behavior. Windows-specific local coverage is not available; the
backup path and SQLite behavior remain exercised by the existing cross-platform database CI lanes.

## Review focus

- Whether the per-invocation backup guard preserves the intended retry and recovery guarantees.
- Whether retaining the batch-start snapshot, rather than per-migration intermediate snapshots, is
  the right recovery-density trade-off.

### DIFF
--- a/src/main/database/application-database.integration.test.ts
+++ b/src/main/database/application-database.integration.test.ts
@@ -827,7 +827,7 @@ describe('application database (integration)', () => {
   it('backs up legacy data through the shared client on a portable storage path', async () => {
     storageRoot = await mkdtemp(join(tmpdir(), 'open science 数据 legacy backup-'))
     const databasePath = join(storageRoot, 'open-science.db')
-    const backupPath = `${databasePath}.before-0025_managed_file_version_foundation.backup`
+    const backupPath = `${databasePath}.before-0001_runtime_schema_baseline.backup`
     const seedClient = createProjectDbClient(storageRoot)
     try {
       await seedClient.$executeRawUnsafe(`CREATE TABLE "Project" (
@@ -864,10 +864,10 @@ describe('application database (integration)', () => {
         `
       ).resolves.toEqual([{ id: 'legacy-project', name: 'Preserved' }])
       await expect(
-        backupClient.$queryRaw<Array<{ id: string }>>`
-          SELECT "id" FROM "_open_science_migrations" ORDER BY "id" DESC LIMIT 1
+        backupClient.$queryRaw<Array<{ name: string }>>`
+          SELECT "name" FROM "sqlite_schema" WHERE "name" = '_open_science_migrations'
         `
-      ).resolves.toEqual([{ id: '0024_compute_job_file_evidence' }])
+      ).resolves.toEqual([])
     } finally {
       await backupClient.$disconnect()
     }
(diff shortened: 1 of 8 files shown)


ewen-poch:
The `quick_check` limitation is intentional and is called out in the PR's scope/non-goals rather than an accidental weakening hidden by the implementation.

The startup path still performs the project's exact table/column/foreign-key/index/CHECK schema audit on every launch. The SQLite content-level policy is:

- already-current ledger: `quick_check` as the routine health probe;
- any applied migration or adopted legacy baseline: full `integrity_check`;
- direct/explicit schema verification: full `integrity_check` by default;
- backup validation: full `integrity_check`.

Restoring full `integrity_check` for the already-current path would restore the startup cost this PR is specifically meant to remove. A version/abnormal-exit marker could preserve periodic authoritative checks, but that adds persisted lifecycle state and crash semantics that were explicitly kept out of this change. I therefore have not changed the code for this finding; the remaining index-content detection gap is an accepted, documented trade-off for routine startup and should receive maintainer review as a policy decision.

ewen-poch:
Addressed the integrity-audit finding in `066167a0`:

- restored full `PRAGMA integrity_check` for every startup, including an already-current migration ledger;
- removed the `quick`/`full` policy type and the quick-check startup test;
- retained only the per-invocation migration backup batching change.

This adds no persisted state, schema change, or UI behavior. Focused validation passed: 14 database test files / 151 tests, 31 migration-ledger and PR-policy tests, `npm run typecheck:node`, `npm run lint`, and `git diff --check`.

PR #2042, opened 2026-09-02
fix(workspaces): preserve managed workspace ownership

ewen-poch:
## Problem

Managed Session workspaces use opaque UUID directory names. Session and Project deletion deliberately
retain those directories but remove the Session JSON that previously supplied their Project and
Session association. Storage usage could therefore report only an opaque directory name and size,
with no durable ownership, last-use, or retained-after-delete evidence.

The regression was reproduced before the fix through the existing create-Session and storage-usage
boundaries. The new test failed consistently because the usage row contained only `name` and `bytes`.

## Proposed change

- Write one versioned, owner-only receipt per managed workspace under
  `workspaces/.ownership/<workspaceId>.json`.
- Record Project identity before invoking the Agent Session creator, then finalize the receipt with
  the returned Session identity.
- Inject a narrow workspace-ownership capability into Session deletion. Individual Session and whole
  Project deletion update the receipt before removing Session authority; a failed individual deletion
  restores the active marker.
- Backfill receipts only when a still-existing Session JSON authoritatively points at a direct child
  of the managed workspace root.
- Enrich workspace storage-usage rows with optional IDs, timestamps, and `retainedAfterDelete`.

All supported ACP backends and the Task adapter share the same create-Session workflow, so managed
workspace receipt publication is backend-neutral.

## Scope and non-goals

- No Prisma schema or database migration.
- No new lifecycle enum; the persisted addition uses a `retainedAfterDelete` boolean.
- No Settings UI, cleanup button, or automatic deletion behavior.
- Explicit external working directories remain reference-only and receive no receipt or mutation.
- Previously deleted historical workspaces cannot be attributed reliably. They remain unknown and
  must not authorize automatic cleanup.

- managed creation, deletion, historical backfill, external preservation, storage projection, and
  Project-owned-data catalog -> targeted Vitest command -> 45 passed.
- ACP managed-workspace IPC ordering and rollback -> targeted `src/main/acp/ipc.test.ts` selection ->
  6 passed.
- Session deletion, Project deletion, recovery, contracts, and consumers ->
  `npm run test:module -- session_persistence` -> 486 passed.
- Main/Sandbox type safety -> `npm run typecheck:node` -> passed.
- Renderer/shared-contract type safety -> `npm run typecheck:web` -> passed.

The exact-head affected-test explainer fails closed to the complete suite because the existing module
manifest does not own `src/main/acp/create-session-workflow.test.ts`. Per maintainer request, the full
local `npm test` was not run; PR Gate is the final complete and platform-specific authority.

## Review focus

- Receipt path validation rejects nested, external, reserved, and symlinked workspace directories.
- A provisional Project receipt exists before Agent Session creation; rollback removes it only after
  the provisional workspace directory was removed successfully.
- Corrupt or future receipts remain visible as unknown and never become cleanup authority.
- Project deletion performs bounded legacy backfill while Session JSON is still authoritative.

### DIFF
--- a/src/main/acp/create-session-workflow.test.ts
+++ b/src/main/acp/create-session-workflow.test.ts
@@ -9,7 +9,7 @@ type AcpCreateSessionWorkflowHarness = {
   createSession: Mock<(request: AcpCreateSessionRequest) => Promise<AcpCreateSessionResponse>>
   lease: ManagedSessionWorkspaceLease
   workspaces: {
-    acquire: Mock<() => Promise<ManagedSessionWorkspaceLease>>
+    acquire: Mock<(input: { projectId: string }) => Promise<ManagedSessionWorkspaceLease>>
   }
   dataRootWriteCalls: () => number
   events: string[]
@@ -28,13 +28,15 @@ const createHarness = (
   })
   const lease: ManagedSessionWorkspaceLease = {
     cwd: '/data/workspaces/managed-1',
-    commit: vi.fn(() => events.push('commit')),
+    commit: vi.fn(async () => {
+      events.push('commit')
+    }),
     release: vi.fn(async () => {
       events.push('release')
     })
   }
   const workspaces: AcpCreateSessionWorkflowHarness['workspaces'] = {
-    acquire: vi.fn<() => Promise<ManagedSessionWorkspaceLease>>(async () => {
+    acquire: vi.fn(async () => {
       events.push('acquire')
       return lease
     })
@@ -137,6 +139,8 @@ describe('ACP create-Session workflow', () => {
         projectId: 'project-1',
         permissionProfile: 'ask'
       })
+      expect(harness.workspaces.acquire).toHaveBeenCalledWith({ projectId: 'project-1' })
+      expect(harness.lease.commit).toHaveBeenCalledWith('session-1')
       expect(harness.events).toEqual([
         'guard:start',
         'acquire',
--- a/src/main/acp/create-session-workflow.test.ts
+++ b/src/main/acp/create-session-workflow.test.ts
@@ -7,6 +7,7 @@ import type { ManagedSessionWorkspaceLease } from './managed-session-workspace'
 type AcpCreateSessionWorkflowHarness = {
   workflow: ReturnType<typeof createAcpCreateSessionWorkflow>
   createSession: Mock<(request: AcpCreateSessionRequest) => Promise<AcpCreateSessionResponse>>
+  deleteSession: Mock<(request: { sessionId: string }) => Promise<unknown>>
   lease: ManagedSessionWorkspaceLease
   workspaces: {
     acquire: Mock<(input: { projectId: string }) => Promise<ManagedSessionWorkspaceLease>>
@@ -25,6 +26,10 @@ const createHarness = (
     events.push('session')
     if (createSessionResult instanceof Error) throw createSessionResult
     return { sessionId: 'session-1', cwd: request.cwd }
+  })
+  const deleteSession = vi.fn<(request: { sessionId: string }) => Promise<unknown>>(async () => {
+    events.push('delete-session')
+    return undefined
   })
   const lease: ManagedSessionWorkspaceLease = {
     cwd: '/data/workspaces/managed-1',
@@ -52,12 +57,13 @@ const createHarness = (
     }
   }
   const workflow = createAcpCreateSessionWorkflow(
-    { createSession },
+    { createSession, deleteSession },
     { workspaces, withDataRootWrite }
   )
   return {
     workflow,
     createSession,
+    deleteSession,
     lease,
     workspaces,
     dataRootWriteCalls: () => dataRootWriteCalls,
@@ -85,7 +91,10 @@ describe('ACP create-Session workflow', () => {
         admissionActive = false
       }
     }
-    const workflow = createAcpCreateSessionWorkflow({ createSession }, { withProjectAvailable })
+    const workflow = createAcpCreateSessionWorkflow(
+      { createSession, deleteSession: vi.fn() },
+      { withProjectAvailable }
+    )
 
     const pending = workflow.create({
       cwd: '/workspace',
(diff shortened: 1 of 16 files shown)


ewen-poch:

- A receipt finalization failure now compensates by deleting the newly created runtime session and releasing the provisional workspace; compensation failures are preserved in AggregateError.
- Deletion only updates an existing valid ownership receipt, so an explicit external workspace cannot be adopted from its path.
- Session and project deletion roll back retention markers when authoritative deletion fails.
- A missing managed workspace directory is treated as already removed, allowing authoritative Session/Project deletion to continue.
- Receipt operations validate .ownership with lstat as a regular, non-symlink directory before reads, writes, or removals.

Each case has a regression test that failed for the reported reason before the corresponding fix and passes afterward. Targeted tests, module tests, typechecks, lint, and the functional PR Gate are green.

ewen-poch:
Addressed the symlinked workspace ancestor finding in 30ad6962. Ownership reads, writes, and removals now require the configured workspaces root to be a regular non-symlink directory before touching .ownership. The regression test uses a redirected workspaces root and verifies that initialization rejects it and removal leaves an external receipt untouched; it failed before the fix because initialization succeeded, and now passes. The focused 35-test set, Node typecheck, lint, and diff checks are green.

ewen-poch:
Addressed both lifecycle recovery findings in fb6cb59b.

- If ownership publication and runtime Session rollback both fail, the workflow now keeps the workspace lease unreleased, preserving the cwd and receipt for the still-possibly-live runtime Session. The new double-failure regression test failed before the fix because release was called, and now passes.
- Session and Project deletion now register each ownership-retention attempt before awaiting the durable write. If a post-rename durability error is reported before authority deletion commits, rollback restores active ownership. New regressions for both deletion paths failed before the fix with retainedAfterDelete still true and now pass.

No persistence fields, states, data relationships, architecture, or user interaction changed. The focused 55-test set, Node typecheck, lint, and diff checks are green.

ewen-poch:
Addressed both findings in bdaa5f50.

- Managed workspace allocation now creates only the root itself first, validates that root as a regular non-symlink directory, and only then creates the UUID child. Rollback revalidates the same boundary before recursive removal. The regression uses a pre-existing external directory behind a workspaces symlink; it failed before the fix because the user directory was deleted and now remains untouched.
- Ownership metadata is excluded from user workspace usage. When an exact managed workspace is already missing during Session deletion, a valid receipt is removed only if its Project/Session identity matches; corrupt, unknown, or conflicting receipts remain fail-closed. The regression failed before the fix with a remaining receipt and nonzero hidden usage and now passes.

This does not add persistence fields, states, data relationships, architecture, or user interaction. The focused 56-test set, Node typecheck, lint, and diff checks are green.

ewen-poch:
Fixed the failing portable-test shard and module-coverage job in 1962466c. Both failures had the same root cause: src/main/acp/ipc.test.ts fully mocked node:fs/promises without the newly used lstat export, which caused 10 cascading failures and 2 unhandled rejections. The test mock now models a regular managed-workspaces root, and the affected assertions reflect the hardened allocation sequence (create the root recursively, then create the UUID child non-recursively). No production behavior, persistence schema, or UI changed. Verified locally with the six affected test files (101 tests), npm run typecheck:node, targeted ESLint, and git diff --check.

ewen-poch:
Addressed the provisional-receipt crash-recovery finding in 79ce17e6.

The first complete, authoritative Session hydration now reconciles managed workspace receipts that have no sessionId. A matching same-project Session finalizes the existing receipt. An unmatched receipt from an earlier process removes only its validated direct-child workspace and then its receipt. Conflicting, corrupt, retained, or current-process receipts are preserved fail-closed; the existing createdAt value and a process-start cutoff prevent cleanup from racing a newly allocated workspace.

The public-boundary regression failed before the change because reconciliation was never invoked (expected one call, observed zero), and now passes. Focused verification: 7 test files / 257 tests, Node typecheck, targeted ESLint, Prettier, and git diff check.

This adds no receipt field, enum, database migration, or UI change. Existing provisional receipts can now be recovered on the first authoritative startup load; historical unreceipted UUID directories and external working directories remain untouched.

ewen-poch:
Addressed the shared managed-workspace deletion finding in 2458266a.

Branched Sessions intentionally reuse the source Session cwd. When a valid receipt belongs to a different Session in the same Project, deletion now leaves that receipt unchanged and proceeds with deleting the branch. A Project mismatch remains a hard ownership conflict, so the cross-Project fail-closed boundary is unchanged.

The public deletion regression creates a source and branch sharing one managed workspace. It failed before the fix with the ownership conflict recovery path and now passes while confirming that the source Session remains live and its receipt remains active and unchanged.

This is deliberately the minimal relationship policy change: no multi-session receipt field, receipt-version change, migration, enum, database change, or UI change. Focused verification: 7 test files / 258 tests, Node typecheck, targeted ESLint, Prettier, and git diff check.

ewen-poch:
Addressed the incomplete Project catalog ownership finding in cd1f917b.

When Project Session authority is incomplete, deletion now supplements the readable-Session retention pass with a bounded scan of valid ownership receipts. Every receipt proven to match the deleting projectId is marked retained before Session authority is removed. Unrelated, malformed, future-version, and redirected receipts remain untouched. If Project authority deletion fails, only the receipts changed by this attempt are restored to active; partial marking failures also roll back internally before deletion can proceed.

The public Project deletion regression failed before the change because the unreadable Session receipt remained retainedAfterDelete false, and now passes. A second regression verifies rollback when authority deletion fails.

No receipt field, version, enum, database migration, or UI changed. This adds a Project-level persistence operation at the existing deletion-owner boundary. Focused verification: 7 test files / 260 tests, Node typecheck, targeted ESLint, Prettier, and git diff check.

ewen-poch:
Addressed the durable temporary-receipt recovery finding in 0c6e8806.

The ownership scan now reuses the existing durable JSON directory recovery before enumerating canonical receipts. Recovery still validates that the recovered target is the exact direct-child receipt for the workspace and that the receipt payload has the matching workspaceId; live-writer temp files remain protected by the existing PID check. A dead provisional temp receipt is therefore promoted and then handled by normal startup reconciliation instead of leaving its workspace orphaned.

Added an integration regression that failed before the fix because workspace-1 remained on disk, and passes after recovery removes both the provisional workspace and receipt.

Verification: 261 focused tests across 7 files, npm run typecheck:node, targeted ESLint, Prettier, and git diff --check. No persisted fields/version, enum, schema, architecture, data relationship, or UI behavior changed.

ewen-poch:
Addressed both review findings in 3486d9b8.

- Startup reconciliation now handles finalized as well as provisional receipts. A finalized, non-retained workspace is removed only when the complete authoritative catalog has no Session referencing its cwd and its lastUsedAt predates the process recovery cutoff. Any live Session sharing that cwd preserves it, including a branch whose source Session is absent. The regression failed before the fix because the finalized orphan directory and receipt remained, and now passes.
- create-Session now trims projectId once before Project admission and runtime dispatch. Whitespace-only input becomes undefined for admission, while managed workspace creation retains the existing default-project fallback. The table regression failed before the fix because both padded inputs reached the admission guard unchanged, and now passes.

Verification: 265 focused tests across 7 files, npm run typecheck:node, targeted ESLint, Prettier, and git diff --check. No persisted field/version, enum, schema, architecture, data relationship, or UI change.

ewen-poch:
CI follow-up: the initial Windows E2E failure was isolated to the existing responsive settings language-switch journey timing out on its first attempt; its built-in retry passed, and the ownership workspace journeys passed 22/22. Re-running the failed job completed successfully, including both functional and workspace journeys. No code change was needed for this unrelated flaky failure.

PR #2043, opened 2026-09-02
fix(session-contracts): validate runtime IPC payloads

ewen-poch:
## Problem

Several Session application commands relied on TypeScript-only payload types at runtime. Deletion accepted empty identifiers, archive and manifest writes had no strict command contract, and Session-valued results checked only four top-level fields. Malformed status, timestamps, revisions, conversation graphs, runtime context, and incomplete nested messages could therefore cross the application-command boundary or reach persistence.

## Proposed change

- require non-empty project and Session identifiers for deletion;
- add strict archive and manifest request schemas, aligned with the current `lastSessionId`-only manifest contract;
- promote the existing recursive Session file decoder/sanitizer to a shared runtime codec;
- apply that codec once at the main-process Session save boundary and to Session-valued save, archive, details, and delegation results; and
- preserve recoverable legacy upload paths across live command boundaries until the existing persistence owner upgrades them to immutable Upload Versions; and
- preserve object identity only for complete, already-valid payloads while normalizing recoverable historical and incomplete nested shapes and rejecting undecodable graphs.

## Scope and non-goals

- No new fields, enum values, database changes, migrations, or persisted-format versions.
- Historical Session files continue through the existing permissive disk decoder.
- Unversioned historical uploads retain their recoverable path in live projections until the existing live-save upgrade succeeds; persisted versioned uploads remain path-free.
- Historical version-1 manifests containing the removed `lastProjectId` field remain readable; the field is ignored on disk and rejected at the live strict command boundary.
- Valid archive, manifest, and Session-save behavior is unchanged.
- This does not add a second parallel Zod model for the full evolving Session graph, repeat validation inside repository layers, or change lifecycle-event publication architecture.

The regression tests were first run against the unfixed implementation and produced stable failures matching the reported boundary gaps. After the final material edit:

- malformed, incomplete, and legacy-compatible Session command requests/results -> `npm test -- src/main/data-content-application-commands.test.ts` -> 39 passed;
- Session persistence owner, contract, and representative consumer behavior -> `npm run test:module -- session_persistence` -> 487 passed across 9 files;
- main/shared TypeScript contracts -> `npm run typecheck:node` -> passed;
- renderer/shared TypeScript consumers -> `npm run typecheck:web` -> passed;
- linted source and tests -> `npm run lint` -> passed.

The complete local `npm test` suite was intentionally not run. The Session persistence module plan covers the changed owner, shared codec, delegation consumers, deletion integration, and Windows-sensitive overlay; PR CI remains the authority for the complete portable and platform suites. No local E2E claim is made.

## Review focus

- Confirm that strict request rejection is appropriate for empty identifiers, invalid archive timestamps, surplus fields, and the removed live `lastProjectId` manifest field.
- Confirm that the shared codec preserves historical disk compatibility while sanitizing renderer save inputs exactly once at the application-command boundary.
- Check that complete valid Session objects retain identity while graph-only and incomplete nested payloads normalize before persistence or return.
- Check that legacy upload paths survive live save/archive boundaries only until the persistence owner creates immutable Upload Versions.

### DIFF
--- a/src/main/data-content-application-commands.test.ts
+++ b/src/main/data-content-application-commands.test.ts
@@ -484,7 +484,14 @@ describe('Data and content application commands', () => {
       },
       {
         key: 'sessionUpdateArchive',
-        args: [request('session-update-archive')],
+        args: [
+          {
+            projectId: 'project-1',
+            sessionId: 'session-1',
+            archived: true,
+            expectedArchivedAt: null
+          }
+        ],
         owner: deps.sessions.updateArchive
       },
       {
@@ -1072,10 +1079,6 @@ describe('Data and content application commands', () => {
   it('sanitizes a renderer Session once at the main-process save boundary', async () => {
     const router = createApplicationCommandRouter()
     const deps = createDependencies()
-    deps.sessions.saveSession.mockImplementationOnce(async (session) => ({
-      created: false,
-      session
-    }))
     registerDataContentApplicationCommands(router.registrar, deps.dependencies)
 
     const malformedSession = {
@@ -1091,7 +1094,10 @@ describe('Data and content application commands', () => {
       expect.objectContaining({ status: 'idle', revision: 0, createdAt: 0 }),
       undefined
     )
-    expect(deps.sessions.saveSession.mock.calls[0]?.[0].runtimeContext).toBeUndefined()
+    const savedSession = (
+      deps.sessions.saveSession.mock.calls as unknown as Array<readonly [Record<string, unknown>]>
+    )[0]?.[0]
+    expect(savedSession?.runtimeContext).toBeUndefined()
   })
 
   it('dispatches every remaining Project and Session wrapper to its existing owner', async () => {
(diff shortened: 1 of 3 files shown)


ewen-poch:
Addressed the incomplete Session projection finding. The shared codec now preserves identity only when all required Session fields are present and every nested value already matches the sanitizer's canonical structure. Graph-only arguments/results and messages missing required nested fields are normalized instead of being cast through unchanged. Added regression coverage for both command argument/result paths and incomplete nested messages. The rebase also aligns the strict live manifest request with the current lastSessionId-only contract while retaining the existing historical v1 disk compatibility that ignores lastProjectId. Final focused validation: 38 command tests passed, 487 Session persistence module tests passed, node/web typechecks passed, and lint passed.

ewen-poch:
Addressed the legacy upload compatibility finding. The live Session codec now enables the existing preserveLegacyUploadPaths decoder option, so unversioned historical uploads retain their recoverable path through save arguments and Session-valued archive results. The existing persistence owner remains the sole upgrade authority: live saves convert those references to immutable Upload Versions, after which the durable projection is path-free. Added a regression that failed before the change and now covers both save and archive command boundaries. No enum, state, schema, persisted-format, or architecture change was introduced. Final focused validation: 39 command tests passed, 487 Session persistence module tests passed, node/web typechecks passed, and lint passed.

PR #2046, opened 2026-09-02
fix(notebook): honor runtime targets in gating and restart

ewen-poch:
## Problem

Notebook readiness and restart behavior were scoped more broadly than the runtime target the user was operating:

- The renderer disabled the whole Notebook whenever the managed Python runtime was unavailable, even when the Session had an active explicit Python or R runtime binding that the main process could execute.
- The “Restart R kernel” action invoked the historical Session-wide restart, which shut down Python, every R environment, and REPL state instead of the selected R environment.

Both behaviors were reproduced at public boundaries before implementation. The new renderer tests failed because the external-runtime terminal remained disabled and because the restart request had no target; the runtime-service test failed because the Session executor restarted and no exact R executor was terminated.

## Proposed change

- Treat an active `runtimeBindings[language]` entry as the current valid runtime target for renderer gating. Only a missing binding—the implicit app-managed default—depends on `pythonReady` / `rReady`.
- Extend the existing `notebook.restart` request with an exact `language` + `environment` target.
- Send the selected R target from the existing restart action.
- Reuse the runtime service's existing targeted executor termination path so sibling Python, R, and REPL kernels keep their state.
- Preserve the historical Session-wide restart when both target fields are omitted, and reject half-specified targets at runtime boundaries.

## Scope and non-goals

- No new API method, database schema, persisted field, migration, or runtime status enum.
- The two request fields are transient command data only.
- Existing untargeted restart callers remain compatible and retain Session-wide behavior.
- Historical runtime bindings that omit `status` retain their existing implicit-active meaning.
- No user-visible copy or i18n catalog changes.
- This does not add a generic kernel-management UI; it corrects the existing R restart action and readiness gate.

- Active external Python and R bindings remain interactive while managed runtimes report unavailable.
  - `npm test -- src/renderer/src/pages/workspace/provisioning-view.test.ts src/renderer/src/pages/workspace/NotebookPreview.gate.render.test.tsx` (65 tests passed)
- Restarting the selected R kernel terminates only that language/environment and preserves sibling kernel state.
  - `npm test -- src/main/notebook/runtime-service.test.ts src/main/notebook/ipc.test.ts src/main/notebook/local-rpc-notebook-adapter.test.ts` (263 tests passed)
- Target fields are forwarded through Electron IPC and local RPC; malformed half-targets fail closed.
  - Covered by the same main-process test command above.
- Static contracts remain valid.
  - `npm run typecheck:node`
  - `npm run typecheck:web`
  - `npm run lint`
  - `npm run check:web-api-map`
  - `git diff --check`

`test:affected --explain` conservatively selected the complete suite because the touched Notebook IPC test has no module owner. Per project guidance, focused module tests were run locally and the complete suite is left to this PR's exact-head CI.

## Review focus

- Whether an absent runtime binding should continue to mean the implicit managed default.
- Whether preserving untargeted Session-wide restart is the right compatibility boundary.
- Targeted restart cleanup: only the selected process key's termination and restart-recommendation state should be cleared.

### DIFF
--- a/src/main/notebook/ipc.test.ts
+++ b/src/main/notebook/ipc.test.ts
@@ -163,6 +163,7 @@ describe('notebook IPC handlers', () => {
     ])
 
     const session = { sessionId: 'session-1', workspaceCwd: '/workspace' }
+    const restartTarget = { ...session, language: 'r' as const, environment: 'default-r' }
     const namespace = { ...session, language: 'python' as const, environment: 'default-python' }
     const begin = { ...session }
     const append = { ...session, cellId: 'cell-1', writeId: 'write-1', delta: 'hello' }
@@ -194,7 +195,7 @@ describe('notebook IPC handlers', () => {
     await ipcHandlers.get('notebook:execute')?.(undefined, execute)
     await ipcHandlers.get('notebook:export-ipynb')?.(undefined, session)
     await ipcHandlers.get('notebook:export-ipynb-all')?.(undefined, session)
-    await ipcHandlers.get('notebook:restart')?.(undefined, session)
+    await ipcHandlers.get('notebook:restart')?.(undefined, restartTarget)
     await ipcHandlers.get('notebook:shutdown')?.(undefined, session)
 
     expect(service.state).toHaveBeenCalledWith(session)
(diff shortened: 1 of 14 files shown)


ewen-poch:
Follow-up on the latest automated review for head `b6216f00`:

- **Managed R provisioning with an external binding:** reproduced at the mounted renderer boundary. Before the fix, selecting the bound external R tab called `notebookEnv.provision('r')`; commit `ed7263fd` now skips managed provisioning for an active binding and keeps the regression assertion.
- **Targeted restart admission race:** not reproduced at the public runtime-service boundary. I held execution 1 open, started targeted restart until the target reported `restarting`, submitted execution 2 before releasing execution 1, and observed the stable order `execute-1 → terminate → execute-2` on the current implementation. Public execution requests must asynchronously complete Session `ensure` before they can enqueue, while the restart continuation calls `terminateExecutor` synchronously after the drain resumes. Because the reported failure mode did not occur, I did not add a new lock/barrier or retain a test that was already green before any fix.

The new exact head is `ed7263fd`; its CI and automated review are running now.

PR #2048, opened 2026-09-02
fix(skills): avoid catalog mutation deadlock

### LINKED ISSUE #2047: [Bug]: Bulk deleting Skills can deadlock the catalog and prevent Settings from loading Skills
### Preflight checklist

### What happened?

Bulk-deleting imported/personal Skills can leave the Skill catalog permanently blocked for the lifetime of the process. The deletion UI stops making progress, reopening Settings shows **“Open Science could not load Skills.”**, and direct calls to `window.api.settings.listSkills()` remain pending indefinitely. `settings.getSettings()` continues to resolve, and there is no useful renderer or catalog reconciliation error in the logs.

Expected behavior: each selected Skill is either deleted or reported as protected/failed, and the catalog remains readable throughout and after the operation.

This appears to be a circular wait involving the in-flight read cache in `src/main/settings/skill-catalog.ts`:

1. `SkillCatalogModule.listUserSkills()` stores and reuses `userSkillCatalogRead`.
2. During sequential bulk deletion, the filesystem catalog observer can start a user-Skill read that queues behind the shared Skill mutation owner.
3. A subsequent delete owns that mutation lock and invokes the Specialist deletion guard.
4. The guard builds the Specialist package catalog, which calls `listSpecialistSkillCatalog()` and eventually `listUserSkills()`.
5. `listUserSkills()` reuses the observer's cached Promise, but that Promise is waiting for the mutation lock currently held by the delete operation.

The wait cycle is therefore: delete owns mutation lock → guard waits for cached catalog read → cached catalog read waits for mutation lock.

Restarting Open Science clears the in-memory wait. Removing cross-context reuse of `userSkillCatalogRead` also resolved the issue locally; the existing `SkillMutationOwner` re-entrancy then allows the guard's catalog read to complete. A lock-context-aware cache would be another possible solution.

### Steps to reproduce

1. Import enough personal/imported Skills to make a multi-item deletion meaningful.
2. Open **Settings → Skills → Manage Skills**.
3. Select multiple deletable Skills and choose the bulk delete action.
4. Allow the sequential deletion loop and filesystem catalog observer to overlap. This is timing-sensitive; a larger selection makes it easier to reproduce.
5. Observe that deletion eventually stops progressing.
6. Close and reopen Settings → Skills.
7. Observe **“Open Science could not load Skills.”**
8. From an authenticated renderer, compare the calls:
   - `window.api.settings.getSettings()` resolves.
   - `window.api.settings.listSkills()` remains pending for more than 30 seconds.
9. Restart Open Science and observe that `listSkills()` works again with the on-disk catalog unchanged apart from Skills already deleted before the deadlock.

Suggested regression test: arrange for an observer `listUserSkills()` read to be queued behind the mutation owner, then invoke a delete whose Specialist guard reads the catalog. Assert that deletion and a subsequent `listSkills()` both settle. A higher-level sequential bulk-delete test should also verify that the UI remains loadable.

### Operating system

Linux

### App version

0.24.0

### Provider / model

Not applicable; the issue occurs entirely in Settings/catalog management without starting an agent session.

```shell
settings.getSettings(): resolved
settings.listSkills(): pending > 30s
UI: Open Science could not load Skills.

No "user skill catalog reconciliation failed", uncaught exception, or renderer request failure was logged while the IPC call remained pending.
```

mikemikimike:
Fixes #2047

## Problem

Bulk deletion can deadlock the Skill catalog. A filesystem observer may start a shared `listUserSkills()` Promise while a Skill mutation is held. When the deletion guard reuses that Promise from inside the same mutation-owner context, it waits on a read that cannot proceed until the deletion releases its own lock.

## Proposed change

- expose whether the current async call chain holds the existing Skill mutation owner through the repository transaction seam
- bypass only the cross-context in-flight catalog Promise while the current call already holds that owner
- cover the production `SettingsService` → `UserSkillRepository.delete()` → deletion guard → Specialist catalog read chain with a deterministic regression test

## Scope and non-goals

- preserve shared in-flight scans for ordinary callers
- do not change the bulk-delete UI or add a second lock/transaction owner
- no database, persisted settings, on-disk Skill format, migration, historical-data compatibility, enum, or state changes

- deadlock cycle and post-delete readability → `npm test -- src/main/settings/service.test.ts -t 'does not deadlock deletion on an observer read queued behind the mutation owner' --run` → **passed** (1 passed)
- regression sensitivity → the same test with the three-line cache bypass temporarily removed → **failed as expected** by timeout; the implementation was restored and the passing command above was rerun
- User Skill repository owner, contract, and declared consumer coverage → `npm run test:module -- user_skills_repository` → **796 passed, 9 local-environment failures outside the changed concurrency path** (`date not in range` or `canExport: false`); the same paths passed in clean exact-head CI
- Node/main-process types → `npm run typecheck:node` → **passed**
- lint → `npm run lint` → **passed**
- changed-test formatting → `npx prettier --check src/main/settings/service.test.ts src/main/settings/skill-catalog.test.ts` → **passed**
- patch whitespace → `git diff --check` → **passed**

The module impact plan includes the `windows_sensitive` overlay. Local verification covered the shared main-process contract on macOS; exact-head PR Gate run `33613082537` passed the complete portable suite, Linux isolation, Windows core/E2E, macOS build/E2E, static checks, PR policy, and Module tests/coverage. No new UI E2E was added because the regression is fully reproduced at the Settings service boundary without renderer timing.

Independent standards/spec review confirmed that the final integration test covers the reported observer/delete/guard cycle and that the listed impact set covers the changed behavior. No uncovered risk remains after the exact-head PR Gate passed; the nine local module-test failures did not reproduce in its clean full-suite and platform lanes.

## Review focus

- confirm that cache reuse is bypassed only from the currently held mutation-owner context
- confirm that the queued observer read, deletion, and a fresh `listSkills()` call all settle

### DIFF
--- a/src/main/settings/service.test.ts
+++ b/src/main/settings/service.test.ts
@@ -3996,6 +3996,34 @@ describe('SettingsService: skills', () => {
     await expect(service.getSkillDetail('personal-my-skill')).resolves.toBeDefined()
   })
 
+  it('does not deadlock deletion on an observer read queued behind the mutation owner', async () => {
+    const service = await createSkillService()
+    await service.createSkill({ name: 'my-skill', description: 'Mine.', body: '# Mine' })
+
+    let startObserver!: () => void
+    let markObserverStarted!: () => void
+    // Register the observer continuation outside the deletion's mutation-owner context.
+    const observerTrigger = new Promise<void>((resolve) => (startObserver = resolve))
+    const observerStarted = new Promise<void>((resolve) => (markObserverStarted = resolve))
+    const observerRead = observerTrigger.then(() => {
+      const read = service.listUserSkills()
+      markObserverStarted()
+      return read
+    })
+
+    service.setSkillDeletionGuard(async () => {
+      startObserver()
+      await observerStarted
+      await service.listSpecialistSkillCatalog()
+    })
+
+    await expect(service.deleteSkill({ id: 'personal-my-skill' })).resolves.toEqual([
+      expect.objectContaining({ id: 'demo' })
+    ])
+    await expect(observerRead).resolves.toEqual([])
+    await expect(service.listSkills()).resolves.toEqual([expect.objectContaining({ id: 'demo' })])
+  })
+
   it('uses the immutable name and reconciles references reported by the detail view', async () => {
     const service = await createSkillService()
     const b64 = (text: string): string => Buffer.from(text).toString('base64')
--- a/src/main/settings/skill-catalog.test.ts
+++ b/src/main/settings/skill-catalog.test.ts
@@ -17,7 +17,6 @@ import { SkillRegistry } from '../skills/registry'
 import type { FetchLike } from '../skills/github-import'
 import type { UserSkillRepository } from '../skills/user-skill-repository'
 import { SPECIALIST_PACKAGE_SKILL_METADATA } from '../skills/specialist-package-adapter'
-import { skillMutationOwnerFor } from '../skills/skill-mutation-owner'
 import { SettingsRepository } from './repository'
 import { SkillCatalogModule } from './skill-catalog'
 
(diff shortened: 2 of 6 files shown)


ewen-poch:
Thanks for the focused fix — the lock-context-aware cache bypass looks aligned with the reported deadlock. I found one regression-coverage gap before we can rely on it: the new test manually calls `listUserSkills()` inside `runExclusive()` with a mocked repository, so it does not exercise the production `UserSkillRepository.delete()` → Specialist deletion guard → catalog read path, nor the required post-delete `listSkills()` read described in #2047. That means the test could stay green if the production owner wiring or delete/guard integration were incomplete. I’ll add a deterministic integration regression for that real chain directly so we can validate the fix end to end.

ewen-poch:
Follow-up: I added commit `20bdac62` to replace the synthetic cache-branch test with a deterministic integration regression through the real `SettingsService` / `UserSkillRepository.delete()` / deletion-guard / Specialist catalog path. The test also verifies that the queued observer read and a fresh post-delete `listSkills()` call settle. As a sensitivity check, the same test times out when the cache-bypass lines are removed and passes once they are restored. The exact-head PR Gate is now fully green across all portable-test shards, static checks, module coverage, Linux isolation, Windows core/E2E, and macOS build/E2E. Thanks again for the solid root-cause fix.

PR #2050, opened 2026-09-02
fix(notebook): stabilize input and artifact file workflows

ewen-poch:
## Problem

A real “draw a pie chart” turn exposed four related failures:

1. The Agent first copied an uploaded CSV to `/tmp`, which Notebook shell correctly rejected, because the prompt exposed only an absolute turn snapshot and no usable Notebook-relative input path.
2. Matplotlib could not create its configured workload cache. The native sandbox allowed `<runtime>/cache/notebook`, but the Python audit hook rejected every runtime write, so Matplotlib fell back to a random temporary cache and prompted the Agent to try another forbidden directory.
3. Seven consecutive `write_artifact_file` calls used the identical empty `{}` arguments. The MCP parser returned the missing required `filename` error every time, but the model-facing guidance did not show one complete invocation or forbid repeating identical validation failures.
4. Generated Notebook cells could embed app-internal absolute upload snapshot paths, which are machine-specific and disappear when the prepared turn closes.

## Proposed change

- Allow Python libraries to write only inside the canonical `OPEN_SCIENCE_NOTEBOOK_CACHE_DIR`, while retaining the runtime mutation block everywhere else.
- Materialize exact registered Upload/Artifact Versions under the Notebook Session data directory as stable paths such as `inputs/groups-dbdc13461d5e.csv`.
- Return those paths from the Notebook input owner and append a transient `notebookPath` projection to the provider prompt before dispatch.
- Keep names one directory deep, content-addressed, collision-safe, portable within the 255-byte filename limit, and free of absolute or parent-relative components.
- Replace the Artifact guidance fragment with one complete `write_artifact_file` invocation containing `filename`, `source`, and `producerRunId`, plus a rule not to repeat identical failed arguments.
- Add an Artifact schema contract proving that `{}` is rejected specifically because `filename` is required.

## Scope and non-goals

- No database schema or persisted model changes.
- No new enum or retry state.
- No `/tmp` or home-directory permission expansion.
- No automatic Artifact filename inference or Artifact retry state machine.
- No migration or rewrite of historical Notebook cells that already contain absolute snapshot paths.
- Existing absolute attachment snapshots remain available for provider attachment transport; new Notebook code is directed to the materialized relative path.

Persistence impact: exact input copies now persist under the existing app-managed Notebook Session root at `data/inputs/`. They inherit the existing Notebook Session/Project deletion lifecycle and add no independent ownership record. Reattaching identical content reuses the verified file; same-name different content receives a different checksum suffix.

All listed checks ran against the final material implementation:

| Behavior | Framework/path | Command | Result |
| --- | --- | --- | --- |
| Matplotlib uses the configured workload cache while other runtime writes remain blocked | real `python_loop.py` | targeted 11-file Vitest command with `RUN_KERNEL=1` | 283 passed |
| Exact Upload/Artifact Versions become stable relative Notebook inputs | input registry + ACP prompt preparation | same targeted 11-file Vitest command | passed |
| Relative input remains readable after the turn snapshot is deleted | native Notebook sandbox | targeted platform-capability Vitest command | passed |
| Artifact prompt contains complete required arguments and stays inside the static context budget | Notebook MCP + Artifact MCP schema | targeted 11-file Vitest command | passed |
| Shared ACP behavior remains compatible | `claude-code`, `opencode`, `codex-response`, `codex-bridge` representative contract tests | two targeted Vitest commands | passed |
| Local RPC return projection and Notebook execution consumers | local RPC server suites | targeted platform-capability Vitest command | passed |
| Node and Notebook sandbox types | `npm run typecheck:node` | typecheck | passed |
| Changed TypeScript files | targeted ESLint over `origin/main...HEAD` | lint | passed with no warnings |

Test totals:

- Portable/module-focused set: 283 passed across 11 files.
- Platform-capability set (`ps`, loopback RPC, native Notebook sandbox): 124 passed across 4 files.
- `git diff --check`: passed.

`npm run test:affected:explain -- --base origin/main --head HEAD` conservatively selected the full fallback because `resources/notebook/python_loop.py` is not owned by the module-impact manifest. Per the requested local test scope, the complete `npm test` suite was not run; the real Python loop integration test directly covers that unknown resource, and PR CI remains the authority for the complete portable and platform suites.

Platform coverage: local macOS native sandbox passed. Path assertions use portable relative paths and run in the portable input-registry suite; Windows/Linux native sandbox and packaging behavior remain CI-covered risks.

## Review focus

- The canonical cache exemption must not permit writes elsewhere in the managed runtime.
- `NotebookInputRegistry` remains the sole owner of exact Version registration and model-facing materialization.
- `data/inputs` creation must reject symlinked directories, never overwrite mismatched content, and remain within the Notebook data root.
- The prompt projection must occur before provider dispatch without adding provider-specific branches.
- Artifact recovery remains guidance-only; the business layer must not guess which generated file to save.

### DIFF
--- a/resources/notebook/python_loop.py
+++ b/resources/notebook/python_loop.py
@@ -138,6 +138,23 @@ def _install_protected_paths_policy(entries):
     sys.addaudithook(audit)
 _runtime_dir_value = os.environ.get("OPEN_SCIENCE_RUNTIME_DIR", "")
 _managed_runtime_dir = _guard_path(_runtime_dir_value) if _runtime_dir_value else ""
+_cache_dir_value = os.environ.get("OPEN_SCIENCE_NOTEBOOK_CACHE_DIR", "")
+_managed_cache_dir = _guard_path(_cache_dir_value) if _cache_dir_value else ""
+
+def _path_is_within(path, directory):
+    return bool(directory) and (path == directory or path.startswith(directory + os.sep))
+
+if not (
+    _managed_runtime_dir
+    and _managed_cache_dir != _managed_runtime_dir
+    and _path_is_within(_managed_cache_dir, _managed_runtime_dir)
+):
+    _managed_cache_dir = ""
+
+def _runtime_path_is_blocked(path):
+    return _path_is_within(path, _managed_runtime_dir) and not _path_is_within(
+        path, _managed_cache_dir
+    )
 
 _package_mutation_command = re.compile(
     r"(?:\b(?:micromamba|mamba|conda|pip|pip3|pipx|uv|poetry)(?:\.exe)?\b.{0,160}"
@@ -234,15 +251,11 @@ def _text_references_managed_runtime(text):
 
 def _runtime_target_is_managed(value):
     text = str(value).strip().strip("\"'")
-    if _text_references_managed_runtime(text):
-        return True
     try:
-        resolved = _guard_path(text)
+        resolved = _guard_path(os.path.expandvars(os.path.expanduser(text)))
     except (TypeError, ValueError):
-        return False
-    return bool(_managed_runtime_dir) and (
-        resolved == _managed_runtime_dir or resolved.startswith(_managed_runtime_dir + os.sep)
-    )
+        return _text_references_managed_runtime(text)
+    return _runtime_path_is_blocked(resolved)
 
 def _runtime_write_targets(words, redirections=()):
     if not words:
@@ -362,9 +375,7 @@ def _protected_paths_audit(event, args):
             resolved = _guard_path(target)
         except (TypeError, ValueError):
             continue
-        if _managed_runtime_dir and (
-            resolved == _managed_runtime_dir or resolved.startswith(_managed_runtime_dir + os.sep)
-        ):
+        if _runtime_path_is_blocked(resolved):
             _blocked_environment_mutation()
 
     if event != "open" or not args:
@@ -386,9 +397,7 @@ def _protected_paths_audit(event, args):
     )
     if write_open and os.path.basename(resolved).casefold() == "pyvenv.cfg":
         _blocked_environment_mutation()
-    if write_open and _managed_runtime_dir and (
-        resolved == _managed_runtime_dir or resolved.startswith(_managed_runtime_dir + os.sep)
-    ):
+    if write_open and _runtime_path_is_blocked(resolved):
         _blocked_environment_mutation()
 sys.addaudithook(_protected_paths_audit)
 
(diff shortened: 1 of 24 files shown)


ewen-poch:
Addressed both review findings in 879c38d0:

- Native follow-up inputs are now materialized before provider dispatch and the steering prompt includes the exact `inputs/...` `notebookPath`. Turn registration is still committed only after the provider accepts the steering request, so refused or superseded turns do not leave a registration behind. The guidance also explicitly tells agents to ignore attachment resource URIs, paths, and basenames and to preserve the `inputs/` prefix.
- Prompt input materialization now validates each directory from the trusted storage root through `notebooks/<project>/<session>/data/inputs` with `lstat` and `realpath`, rejects symlinked ancestors, and revalidates the input parent immediately before the exclusive copy.

Regression coverage verifies pre-dispatch path advertisement, no registration commit on refusal, and rejection of a replaced `data` symlink before any out-of-scope write. Targeted tests, the affected socket/integration suites (652 tests), Node/sandbox typechecks, ESLint, and Prettier pass after rebasing onto current `main`.

ewen-poch:
Addressed the generated input-copy lifecycle concern in 3fe82cc6. Session deletion now removes that Session's exact data/inputs directory together with its staged input cache; Project deletion removes the same derived directory for each retained Session workspace while preserving every other Notebook working file and other Projects. Cleanup validates the storage path and does not follow symlinked directories. The Project-owned data catalog now records this narrow hard-delete exception inside the otherwise retained workspace. This intentionally does not add per-prompt ownership state: refused or cancelled prompt copies remain bounded by the existing Session/Project deletion lifecycle. Regression coverage first failed because the prompt copy survived deletion, then passed after the fix; the focused 265-test set, Node typecheck, and lint pass.

ewen-poch:
Thanks for calling this out. I am leaving this unchanged in this PR because data/inputs is a disposable convenience projection inside the Session-owned writable workspace, not the authority for input immutability. The canonical immutable Version remains under the input authority, and Notebook run leases, version identity, and provenance resolve from that authority rather than trusting the projection bytes. A kernel or shell can therefore disrupt its own later path-based reads, but it cannot mutate the canonical input or its recorded identity. Preventing unlink/rename robustly would require a cross-platform read-only projection boundary across sandboxed kernels, shell execution, and the unsandboxed fallback; directory chmod is not sufficient for a same-user process and is not portable to Windows. That architectural expansion and observable permission change are outside this focused path/lifecycle fix and should be handled separately if the product requires workspace paths to resist code running inside that same workspace.

PR #2051, opened 2026-09-02
feat(message-scroll): ship header-probed image dimensions with preview resources

roxi3906:
## Problem

Follow-up to #2021 (viewport-edge image jitter). That PR eliminated the unload/remount oscillation (latch) and made remounts scroll-neutral (geometry cache), but one jump remains: an image that has **never been loaded in this app run** still renders the ~100px alt-text placeholder first, then grows to its real height when the resource arrives and the `` decodes — a one-time layout shift (content flashes down a frame, anchoring compensation flashes it back up).

## Approach

Have the main process probe pixel dimensions from image file headers and ship them with the preview capability, so the renderer knows the exact geometry from the moment `acquire` resolves — before any image bytes are fetched or decoded.

**Main process** — new zero-dependency parser `src/main/image-header-dimensions.ts`:

- PNG (IHDR), JPEG (SOF0-15 marker scan, skipping non-frame markers), GIF (logical screen descriptor), WebP (all three variants: `VP8 `, `VP8L`, `VP8X`), BMP (including top-down negative heights);
- sniffs magic bytes rather than trusting extensions, so forged/mislabeled files simply yield `undefined`;
- reads at most the first 256 KiB (JPEG SOF can sit behind a large EXIF segment); any I/O or parse failure degrades to `undefined` and never fails the acquire;
- implausible dimensions (0 or absurd sizes) are rejected.

**Acquire path** — `ManagedPreviewResource` gains optional `width?: number; height?: number`, populated in both assembly points (`acquireFilePath` and the Reviewer-only `acquireResolvedFile`) whenever the inferred MIME type is `image/*`. The preload contract adapter passes the result through untouched, so no preload changes; the inline type in `web-service/task-api.ts` was extended to match.

**Renderer** (`SessionArtifactImage`):

- the `` now carries explicit `width`/`height` attributes (metadata first, geometry cache as fallback), so the browser reserves the exact box before and during decode; the stylesheet gains `h-auto` on the image rule so `max-w-full` scaling never distorts the ratio;
- metadata dimensions also seed the geometry cache, so a remounted placeholder (transcript window recycling) locks the exact height from its first frame **even if the previous instance's `` never decoded** (lazy loading keeps offscreen images undecoded);
- priority chain: metadata > geometry cache > the original ~100px placeholder. Both #2021 mechanisms (latch, geometry cache) remain untouched — this is additive;
- TIFF branch unchanged (fixed `aspect-ratio: 4/3` container).

## Artifact-metadata persistence: deliberately skipped

Persisting dimensions at artifact creation would require extending the Session-JSON `PersistedArtifact` schema, the `sanitizeArtifact` whitelist, `ArtifactFile`, and several main-process assembly/mapping points (provenance-repository, storage-access, reconciliation, task-runner). The remaining unlock window without persistence is the local IPC round-trip between mount and `acquire` resolution (a few ms, no network, no decode) — not worth that blast radius. Revisit only if the flash proves visible in practice.

## Tests

- `image-header-dimensions.test.ts` (13 cases): every supported format, truncated/forged/implausible headers, unknown formats → `undefined`, on-disk read, missing file → `undefined` (no throw);
- `managed-preview-resources.test.ts`: acquire attaches probed dimensions for a real PNG header; forged content acquires cleanly without dimensions;
- `SessionMessageMarkdown.test.tsx`: with metadata present, the `` carries exact `width`/`height` and a remounted placeholder locks the exact geometry **without any `load` event** (proving the metadata-seeded cache path);
- related suites pass (managed-preview ipc/protocol, task-api, i18n guard — no new user-visible strings); `npm run typecheck` and eslint are clean.

## Manual verification

1. Open a session containing a message with an inline image never opened in this app run;
2. scroll upward past the image: no flash-down/flash-up pair, scrollbar length stays constant;
3. scroll far away and back (transcript window recycling): the placeholder already has the final height, no jump.

## Relationship to #2021

Stacks on `fix/message-image-scroll-jitter` (branch base). Merge #2021 first; this PR will then diff cleanly against main.

### DIFF
--- a/src/main/image-header-dimensions.test.ts
+++ b/src/main/image-header-dimensions.test.ts
@@ -1,0 +1,148 @@
+import { mkdtemp, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+
+import { parseImageHeaderDimensions, readImageHeaderDimensions } from './image-header-dimensions'
+
+const pngBuffer = (width: number, height: number): Buffer => {
+  const buffer = Buffer.alloc(33)
+  Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]).copy(buffer, 0)
+  buffer.writeUInt32BE(13, 8)
+  buffer.write('IHDR', 12, 'ascii')
+  buffer.writeUInt32BE(width, 16)
+  buffer.writeUInt32BE(height, 20)
+  return buffer
+}
+
+const jpegBuffer = (width: number, height: number): Buffer => {
+  const app0Payload = Buffer.alloc(14)
+  const sof = Buffer.alloc(11)
+  sof.writeUInt16BE(11, 0) // segment length includes its own two bytes
+  sof[2] = 8 // sample precision
+  sof.writeUInt16BE(height, 3)
+  sof.writeUInt16BE(width, 5)
+  sof[7] = 1 // component count
+  return Buffer.concat([
+    Buffer.from([0xff, 0xd8]),
+    Buffer.from([0xff, 0xe0]),
+    Buffer.from([0x00, 0x10]),
+    app0Payload,
+    Buffer.from([0xff, 0xc0]),
+    sof
+  ])
+}
+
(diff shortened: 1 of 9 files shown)


roxi3906:
Folded into #2021 to keep the whole image-scroll-jitter fix in one reviewable unit. This branch's commits (header-probed image dimensions) now live on fix/message-image-scroll-jitter.

PR #2054, opened 2026-09-02
fix(workspace): defer generated previews until publication

ewen-poch:
## Problem

GENERATED artifacts can reach renderer preview surfaces before turn finalization publishes their first managed Artifact Version. The card starts both a thumbnail read and an availability probe for the pending path; inline Markdown images and links are a second preview entry point. `ManagedFileVersionService` correctly rejects these reads with `VERSION_NOT_FOUND`, so Electron logs `Error occurred in handler for 'artifacts:read-preview'` even though the renderer suppresses its own console error.

## Proposed change

- Recognize the exact cross-platform pending-file suffix: `.pending/<runId>/<filename>`.
- Keep pending GENERATED cards and inline Markdown references visible but inert; skip card previews, availability probes, image resources, TIFF previews, and click navigation until finalization replaces the artifact metadata.
- Share the pending-publication predicate across card and Markdown surfaces.
- Preserve the managed-version fail-closed behavior; do not swallow `VERSION_NOT_FOUND` in main.
- Keep a finalized artifact whose legal filename is `.pending` previewable.

## Scope and non-goals

- No schema, persisted field, enum, migration, data-model, or persistence changes.
- Historical compatibility requires no migration. The narrower suffix check avoids regressing existing finalized artifacts named `.pending`.
- Does not change artifact finalization or reconciliation. A path that remains pending indefinitely would be a separate lifecycle defect requiring its own reproduction.
- No new user-visible copy or i18n catalog entries.

- Card and Markdown preview boundaries -> `npx vitest run src/renderer/src/pages/workspace/SessionMessageMarkdown.test.tsx src/renderer/src/pages/workspace/WorkspaceMessageScroller.interaction.test.tsx --reporter=dot` -> 2 files, 61 tests passed.
- Workspace renderer consumers -> `npm run test:module -- workspace_page` -> 30 files, 736 tests passed.
- Prettier and `git diff --check` -> passed.
- Cross-platform pending paths -> regression coverage includes `/workspace/.pending/run-1/report.txt` and `C:\\workspace\\.pending\\run-1\\report.txt`.

Before the first production fix, the original tightened card regression failed identically on three runs: `artifacts.readPreview` was called twice for the pending path (32 KiB thumbnail plus 1-byte probe), matching the reported handler failure.

After Codex review identified the Markdown bypass and `.pending` filename collision, the two added public-boundary regressions also failed identically on three unmodified runs: the pending Markdown link remained enabled, and a finalized `.pending` file card remained disabled. Both pass after the shared exact-suffix guard.

A follow-up Codex review identified the PDF-specific nested viewport pipeline. With the observer harness dispatching every observed target, the pending-PDF regression failed identically on three pre-fix runs with `Managed preview requires a logical identity`; after pending cards stopped mounting `ArtifactPreview`, it passes without acquiring a preview resource.

`npm run test:affected:explain -- --base origin/main --head HEAD` selects the full CI fallback because `WorkspaceMessageItem.tsx` is not currently assigned to a module owner. Per maintainer direction, the full local `npm test` suite was not run; PR Gate is the final authority for that coverage.

## Review focus

- Confirm no card, inline image, TIFF, or Markdown link can acquire/open a pending Artifact Version.
- Confirm finalized artifacts, including a file literally named `.pending`, retain existing preview behavior.
- Confirm `.pending/<runId>/<filename>` remains the storage-layout invariant without adding publication state to persisted renderer data.

### DIFF
--- a/src/renderer/src/pages/workspace/SessionMessageMarkdown.test.tsx
+++ b/src/renderer/src/pages/workspace/SessionMessageMarkdown.test.tsx
@@ -146,6 +146,35 @@ describe('SessionMessageMarkdown', () => {
     expect(onPreviewArtifactModal).toHaveBeenCalledWith(artifact)
   })
 
+  it('keeps unpublished artifact links and inline images inert until publication', async () => {
+    const onPreviewArtifact = vi.fn()
+    const onPreviewArtifactModal = vi.fn()
+    const pendingArtifact = {
+      ...artifact,
+      path: '/managed/session/.pending/run-1/sin_curve.png'
+    }
+
+    await act(async () => {
+      root.render(
+        <SessionMessageMarkdown
+          content={'![Sine curve](sin_curve.png)\n\n[sin_curve.png](sin_curve.png)'}
+          artifacts={[pendingArtifact]}
+          onPreviewArtifact={onPreviewArtifact}
+          onPreviewArtifactModal={onPreviewArtifactModal}
+        />
+      )
+    })
+
+    const artifactLink = container.querySelector<HTMLButtonElement>('[data-session-artifact-link]')
+    expect(artifactLink?.disabled).toBe(true)
+    expect(previewResourceHarness.enabled).toBe(false)
+    expect(container.querySelector('[data-session-artifact-image]')).toBeNull()
+
+    await act(async () => artifactLink?.click())
+    expect(onPreviewArtifact).not.toHaveBeenCalled()
+    expect(onPreviewArtifactModal).not.toHaveBeenCalled()
+  })
+
   it('retains the existing safe-link component for external links', async () => {
     markdownHarness.href = 'https://example.com/sin_curve.png'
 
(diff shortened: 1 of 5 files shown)


ewen-poch:
Addressed both review findings in e8db8885. The pending-publication guard is now shared by GENERATED cards, Markdown artifact links, inline images, and TIFF previews, so none of those surfaces acquires or opens an unpublished Version. The predicate now matches the exact storage suffix `.pending/<runId>/<filename>`, which keeps a finalized artifact literally named `.pending` previewable. Added public-boundary regressions for both behaviors; the focused suite passes 60/60 tests.

ewen-poch:
Addressed the pending PDF thumbnail finding in e3799172. Pending cards now mount no `ArtifactPreview` at all, so `PdfThumbnail` cannot start its independent viewport-driven pipeline. The regression harness now dispatches entries for every observed element, which reproduced the issue consistently as `Managed preview requires a logical identity` before IPC acquisition; after the fix, the pending-PDF test and the full focused suite pass (61/61). No architecture, persistence, state-model, or interaction contract changes were introduced.

PR #2065, opened 2026-09-02
fix(notebook): bound concurrent shell execution

ewen-poch:
## Problem

Every `notebook_execute` shell request spawned a fresh process immediately. Concurrent agents in the same Project Session could therefore overlap writes to the same Notebook-owned roots, causing file evidence to fail closed with `observer-conflict`. Across Sessions there was no application-wide process ceiling, so a request burst could exhaust process, CPU, or memory resources.

The pre-fix regressions reproduced the reported behavior: a second same-Session process entered before the first released; a seventh cross-Session process started immediately; cancelling a queued candidate could not prevent its process from starting; and the MCP/local-RPC path did not forward shell cancellation.

## Proposed change

- Admit at most one shell process per `(projectId, sessionId)`, including root and child Frame lanes.
- Admit at most six shell processes across one Notebook runtime generation.
- Persist waiting shell Runs with the existing `queued` status, transition them to `running` only after admission, and expose the existing localized `queued` label in Notebook history.
- Forward cancellation from MCP through local RPC and the runtime service. Cancelling a waiter removes it from the in-memory queue and records a terminal `cancelled` Run without invoking the shell adapter.
- Register active and queued shell operations by Frame lane and Project Session. Direct, Session-wide, and global teardown close admission, abort matching work, and drain terminalization before shutdown returns, so subsequent input cleanup cannot race a late shell process.
- Keep the queue and fixed ceiling in the existing `NotebookExecutionOwner`; no new service, dependency, settings surface, or process registry is introduced.

## Scope and non-goals

- The limit is intentionally fixed at six. This PR does not add user configuration, queue priorities, or a persisted/resumable scheduler.
- The admission queue is generation-local. Existing startup reconciliation continues to mark persisted `queued`/`running` Runs as interrupted after a restart.
- No database schema, run-record field, wire schema, or enum value is added. `queued` already exists in `NotebookRunStatus` and the persisted `run.json` model.
- Historical data remains readable without migration. No compatibility adapter or backfill is required.

- Same-Session serialization across root/child Frames, six-process application ceiling, queued cancellation, Session teardown drain, non-conflicting back-to-back file evidence, terminal status transitions, RPC cancellation, and renderer status label -> `npm test -- src/main/notebook/runtime-service.test.ts src/main/notebook/run-terminalization.test.ts src/main/notebook/local-rpc-notebook-adapter.test.ts src/main/notebook/mcp-server.test.ts src/main/notebook/runtime-service.architecture.test.ts src/main/runtime-state-ownership.architecture.test.ts src/renderer/src/pages/workspace/notebook-cell-utils.test.ts` -> 7 files / 457 tests passed.
- Main-process and sandbox contracts -> `npm run typecheck:node` -> passed.
- Renderer consumer -> `npm run typecheck:web` -> passed.
- Source/style validation -> `npm run lint` -> passed.
- Final diff classification -> `npm run test:affected:explain -- --base origin/main --head HEAD` -> fail-closed full PR plan because `src/main/notebook/execution-owner.ts` has no declared module owner. Per project guidance, the full portable suite is left to blocking PR CI rather than duplicated locally.

The consumer set includes the local RPC and MCP adapters because their `executeShell` cancellation contract changed, plus the Notebook renderer label helper because `queued` is now observable for shell Runs. No platform-specific spawn behavior changed; existing shell-process cancellation/termination remains the platform boundary.

## Review focus

- Admission fairness and release on every terminal/error path.
- Session identity uses `(projectId, sessionId)`, not lane-specific storage paths, so child Frames cannot bypass serialization.
- Cancellation races around admission and the persisted `queued -> running -> terminal` sequence.
- Teardown ordering: active and queued shell work must terminalize before Session input cleanup can proceed.
- Confirm that a fixed ceiling of six is the desired product policy; configurability and queue-depth limits remain deliberate non-goals.

Uncovered local risk: the classifier selects the complete portable suite and platform overlays only in PR CI, so those blocking checks are authoritative for integration coverage.

### DIFF
--- a/src/main/notebook/execution-owner.ts
+++ b/src/main/notebook/execution-owner.ts
@@ -154,6 +154,66 @@ const cancelledExecutionResult = (cwd: string): NotebookSessionExecutionResult =
   status: 'cancelled'
 })
 
+const NOTEBOOK_SHELL_MAX_CONCURRENCY = 6
+const SHELL_CANCELLED_MESSAGE = 'Shell command was cancelled.'
+
+type ShellAdmissionWaiter = {
+  sessionKey: string
+  signal?: AbortSignal
+  resolve: (release: (() => void) | undefined) => void
+  onAbort: () => void
+}
+
+class NotebookShellExecutionAdmission {
+  private active = 0
+  private readonly activeSessions = new Set<string>()
+  private readonly waiters: ShellAdmissionWaiter[] = []
+
+  acquire(sessionKey: string, signal?: AbortSignal): Promise<(() => void) | undefined> {
+    if (signal?.aborted) return Promise.resolve(undefined)
+
+    return new Promise((resolve) => {
+      const onAbort = (): void => {
+        const index = this.waiters.indexOf(waiter)
+        if (index === -1) return
+        this.waiters.splice(index, 1)
+        signal?.removeEventListener('abort', onAbort)
+        resolve(undefined)
+      }
+      const waiter: ShellAdmissionWaiter = { sessionKey, signal, resolve, onAbort }
+      this.waiters.push(waiter)
+      signal?.addEventListener('abort', onAbort, { once: true })
+      if (signal?.aborted) onAbort()
+      else this.dispatch()
+    })
+  }
+
+  private dispatch(): void {
+    while (this.active < NOTEBOOK_SHELL_MAX_CONCURRENCY) {
+      const index = this.waiters.findIndex((waiter) => !this.activeSessions.has(waiter.sessionKey))
+      if (index === -1) return
+      const [waiter] = this.waiters.splice(index, 1)
+      if (!waiter) return
+      waiter.signal?.removeEventListener('abort', waiter.onAbort)
+      if (waiter.signal?.aborted) {
+        waiter.resolve(undefined)
+        continue
+      }
+
+      this.active += 1
+      this.activeSessions.add(waiter.sessionKey)
+      let released = false
+      waiter.resolve(() => {
+        if (released) return
+        released = true
+        this.active -= 1
+        this.activeSessions.delete(waiter.sessionKey)
+        this.dispatch()
+      })
+    }
+  }
+}
+
 // Root runtime ownership is Session-scoped, but every durable Run still belongs to the authenticated
 // conversation Frame that produced it. A renderer state read may have created the shared owner first.
 const runAgentFrameId = (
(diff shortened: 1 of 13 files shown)


ewen-poch:
Addressed the teardown finding in 5b030731. Shell operations are now registered by Frame lane and Project Session; direct, Session-wide, and global teardown close admission, abort matching active/queued work, and wait for terminalization before returning. Added a regression that failed before the fix because the active signal remained live and the queued child-Frame process started after shutdown; it now passes, and the final focused set is 7 files / 457 tests.

ewen-poch:
Addressed in 4077f2c6. POSIX shell children are now registered as owned process groups, and cancellation/timeout awaits the existing bounded process-tree terminator before the shell operation terminalizes. This removes the background SIGKILL path that allowed Session teardown to return while descendants were still alive.\n\nThe public runtime regression was changed first and failed on the previous code with the descendant still running after executeShell resolved (received true, expected false). It now passes, along with the shell-process module and the focused impact set (8 files, 470 tests), Node/sandbox/Web typechecks, and lint. No data fields, enum values, persistence, or UI behavior were added.

ewen-poch:
Addressed in 233978a5. Shell admission and operation tracking now occur before Session ensure starts. A request arriving while lane/Session/global teardown owns the gate is cancelled without entering the Session registry; a request already admitted is tracked and drained by teardown.\n\nThe public regression was added first and failed on the previous commit because a shell submitted behind the Session removal gate ran after shutdown returned (stdout was `must not run`). It now returns the existing cancellation result, does not invoke the shell adapter, and does not create a replacement Session executor. The focused impact set passes (8 files, 471 tests), along with Node/sandbox/Web typechecks and lint. No new fields, enum values, persistence, data-model changes, or UI changes.

ewen-poch:
Addressed in 9c901c83. Shell teardown now aborts and fully drains every tracked shell operation before invoking Session/lane/global teardown, so Session removal cannot race ahead of bounded process-tree termination. Admission gates remain held for the entire drain and teardown sequence.\n\nThe public regression was added first and failed on the previous commit because the Session executor was removed before the shell adapter completed cancellation (received true, expected false). It now passes together with the ensure-race and queued-drain regressions. Final focused validation: 8 files / 472 tests, Node/sandbox/Web typechecks, and lint. This changes teardown ordering only; there are no new fields, enums, persistence, data-model, or UI changes.

PR #2070, opened 2026-09-02
fix(file-save): publish exports atomically

ewen-poch:
## Summary

- publish desktop exports through one same-directory temporary-file helper, flushing complete bytes before atomic replacement and cleaning failed temporary output
- preserve managed-file self-copy protection while routing Blob, managed/version, project ZIP, conversation Markdown/PDF, notebook, Skill, Specialist, and Connector exports through that helper
- forward the shared logical managed-file identity from the Web adapter to preview-resource acquisition
- show Project Artifact Save As before opening sources and stream one verified Version lease at a time

## Reproduction

The regression tests were added first and run against the unmodified production paths. The focused run failed in four places (115 passed):

- a managed-file partial write changed the existing destination to partial bytes
- a conversation Markdown partial write changed the existing destination to partial bytes
- the Web RPC body omitted `projectId`, `fileId`, and `versionId`
- Project export opened both source leases before the save dialog and retained both concurrently

After the fix, those tests pass and assert the reported behavior directly.

The first independent Codex review then caught the real PDF-preview request shape: managed requests can carry a legacy `path` alongside logical identity. A follow-up regression failed against the first patch because it still selected `{source,path}`; the adapter now discriminates by `source`, so artifact/upload always use logical identity while local/notebook-input keep path-based behavior.

The second independent review caught that atomic replacement could widen an existing target from mode `0600` to the temporary file's `0644`. A deterministic regression reproduced that change; overwrite publication now copies the existing destination's permission bits before the durability barrier.

The third independent review identified transient Windows replacement failures. A fault-injected `EPERM` regression confirmed the publisher did not retry; it now shares the existing durable-file bounded retry policy (`25/50/100/200/400ms`) for `EPERM`, `EACCES`, and `EBUSY` rename failures.

The fourth independent review found that exclusive batch-export publication required hard-link support, which FAT/exFAT removable volumes do not provide. An injected `EOPNOTSUPP` regression failed on the hard-link-only implementation. Supported volumes still use atomic hard-link publication; unsupported volumes fall back to `COPYFILE_EXCL`, fsync the copied inode, and retain the no-overwrite guarantee. A race regression verifies that a destination appearing before fallback is preserved byte-for-byte. The Session Artifact batch IPC test also runs the production publisher with an injected no-hard-link filesystem and verifies both files are exported.

The fifth independent review found that applying a read-only destination mode before the temporary-file durability barrier made the required `r+` fsync fail with `EACCES`. A real mode-`0444` regression reproduced the failure. Complete temporary bytes are now synced before inheriting the existing mode and atomically replacing the destination.

The sixth independent review found two exclusive-publication gaps. First, direct `COPYFILE_EXCL` fallback could expose partial final output; a fault-injected copy reproduced it. The fallback now copies only to a random hidden sibling, syncs it, and uses the existing anchored native no-replace publisher for the final atomic rename. Second, same-inode batch destinations were rejected before collision retry; the public IPC regression now verifies that the source remains unchanged and export continues at `report (2).csv`.

The first Windows Gate for that follow-up compiled the native addon but reproduced a platform-specific publication failure in both public Session Artifact batch tests: the Win32 `SetFileInformationByHandle(FileRenameInfo)` wrapper rejected the handle-relative rename. The implementation now calls `NtSetInformationFile(FileRenameInformation)` directly with the already-open source and parent handles, `replace_if_exists=false`, and converts its NT status to the existing portable error codes. This keeps atomic no-replace rename semantics on local filesystems without requiring hard-link support.

## Behavior and compatibility

- Existing target bytes remain unchanged for failures before publication. If the final rename succeeds but directory `fsync` reports failure, the target contains the complete new file, never partial bytes.
- Exclusive publication is atomic on hard-link-capable volumes and on local FAT/exFAT volumes through native no-replace rename. Unsupported network filesystems fail without exposing the final destination.
- Atomic replacement creates a new inode. Existing permission bits are preserved; ownership, ACLs, and extended attributes are not guaranteed to carry over.
- Project export now prompts first. Canceling performs no source open/hash work; after confirmation, source failures can result in no ZIP being written.
- No product data fields, persisted formats, database schema, migrations, or enum/state values change.
- No historical-data compatibility path is required.

## Test impact

Final exact-head checks after the last material edit:

- desktop publisher, native no-replace publication, durable replacement policy, file-save, and conversation behavior: `npm test -- src/main/file-save.test.ts src/main/user-file-publisher.test.ts src/main/uploads/atomic-no-replace-publisher.test.ts src/main/storage/durable-json-file.test.ts src/main/session-persistence/conversation-export.test.ts` — 124 passed
- Web adapter/main parser contract: `npm test -- src/renderer/web/bootstrap.test.ts src/main/managed-preview-resources.test.ts -t 'path-only|logical managed version|streams a managed file'` — 4 passed, 53 filtered
- version lease and adjacent export consumers: `npm test -- src/main/managed-file-versions/version-file-operator.test.ts src/main/notebook/runtime-service.export.test.ts src/main/skills/export.test.ts src/main/specialist/package/electron-adapter.test.ts src/main/specialist/package/contribution-template.test.ts` — 74 passed
- `npm run typecheck:web` — passed
- `npm run lint` — passed
- `npx node-gyp rebuild --directory packages/safe-file-publisher-native` — passed on macOS arm64 after the NT rename change; Windows compilation and behavior are covered by the rerun PR Gate
- exact-head [PR Gate](https://github.com/aipoch/open-science/actions/runs/33653944346) — passed, including Windows core, Windows/macOS E2E, all three Ubuntu full-suite shards, static checks, Linux isolation, and module coverage enforcement

`test:affected:explain` selected the full suite because `src/main/file-save.test.ts` has no module owner in the impact manifest. Per maintainer direction, the full local `npm test` was not run; PR Gate is authoritative for the complete portable and platform plan.

## Uncovered risk

- FAT/exFAT hard-link rejection is fault-injected in the public batch-export boundary; CI validated the native rename fallback on Windows/NTFS, but does not mount a real FAT/exFAT volume. Microsoft documents rename support there independently of hard links.

### DIFF
--- a/packages/safe-file-publisher-native/src/safe_file_publisher_native.cc
+++ b/packages/safe-file-publisher-native/src/safe_file_publisher_native.cc
@@ -168,27 +168,6 @@ const char* WindowsErrorCode(DWORD error) {
   }
 }
 
-struct NativeIoStatusBlock {
-  union {
-    LONG status;
-    void* pointer;
-  };
-  ULONG_PTR information;
-};
-
-struct NativeFileLinkInformation {
-  BOOLEAN replace_if_exists;
-  HANDLE root_directory;
-  ULONG file_name_length;
-  WCHAR file_name[1];
-};
-
-using NtSetInformationFileFunction = LONG(NTAPI*)(
-    HANDLE, NativeIoStatusBlock*, void*, ULONG, ULONG);
-using RtlNtStatusToDosErrorFunction = ULONG(NTAPI*)(LONG);
-
-constexpr ULONG kFileLinkInformation = 11;
-
 napi_value PublishWindows(
     napi_env env,
     const std::string& root_utf8,
@@ -230,12 +209,6 @@ napi_value PublishWindows(
     return ThrowError(env, "Network storage roots are not supported for atomic publication.",
                       "ENOTSUP");
   }
-  bool supports_hard_links = false;
-  if (!QueryHardLinkSupport(root_handle, &supports_hard_links) || !supports_hard_links) {
-    CloseHandle(root_handle);
-    return ThrowError(env, "The storage root file system does not support hard links.", "ENOTSUP");
-  }
-
   FILE_ATTRIBUTE_TAG_INFO root_attributes{};
   if (!GetFileInformationByHandleEx(
           root_handle, FileAttributeTagInfo, &root_attributes, sizeof(root_attributes)) ||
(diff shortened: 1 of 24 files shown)


ewen-poch:
Follow-up on the Codex review for `4b0e5e09`: the four listed findings are already addressed on this exact head.

- `publishExclusive` catches unsupported hard-link errors, copies only to a hidden same-directory staging path, fsyncs it, then uses the anchored native no-replace publisher. The current Windows implementation uses handle-relative `NtSetInformationFile(FileRenameInformation)`.
- `publishUserFile` calls `syncFile(temporaryPath)` before inheriting the existing destination mode with `chmod`.
- overwrite publication calls the shared `retryFileReplacement` policy.
- batch publication no longer supplies managed self-copy validation; its public IPC regression verifies that a source-name collision is preserved and export continues at `report (2).csv`.

The exact-head [PR Gate](https://github.com/aipoch/open-science/actions/runs/33653944346) passed, including Windows native compilation, Windows-specific `file-save.test.ts`, Windows path tests, all three Ubuntu full-suite shards, and module coverage enforcement. No follow-up code change is needed for these repeated stale observations.

ewen-poch:
On SMB and other network volumes: no code change is planned in this PR. When hard links are unavailable, the existing anchored native publisher intentionally rejects remote roots because server-side atomic no-replace and path-resolution guarantees cannot be established consistently. The fallback fails before exposing the final destination, preserving the data-integrity requirement. Relaxing that boundary would change the existing storage/security design and should be considered separately rather than folded into this fix.

ewen-poch:
The POSIX P2 is an intentional unsupported-capability boundary, so no code change is planned. The fallback first attempts renameat2 with RENAME_NOREPLACE and then linkat; if the filesystem provides neither primitive, the native publisher returns ENOTSUP before exposing the selected destination, and the private staging file is removed. There is no portable atomic no-replace fallback once both filesystem operations are unavailable: opening the final path with O_EXCL and copying into it would expose partial bytes on failure and violate this PR’s data-integrity invariant. A capability probe would have to exercise the same operations and would not add support. The safe behavior here is therefore a clean failure with the destination unchanged.

PR #2080, opened 2026-09-02
fix(settings): recover Skill catalog inconsistencies

ewen-poch:
## Problem

- A failed user Skill catalog publication acknowledged the new filesystem fingerprint before Helper, Renderer, and Agent reload work completed, so unchanged scans never retried it.
- Connector Skill-document refresh failures were logged and converted into successful settlement after clearing the known materialized custom Skill names.
- Identity-conflicting on-disk Skills were filtered out of Settings together with the runtime catalog, leaving no visible diagnosis.

## Proposed change

- Commit the user Skill fingerprint only after publication succeeds and schedule a bounded retry after failure.
- Preserve the last materialized Connector Skill catalog, record an in-memory degraded state, propagate it to Settings, and add a full retry whose promise reports the real result.
- Keep runtime Skill identity filtering fail-closed while returning conflicting rows to Settings as unavailable, with ambiguous actions disabled.

## Scope and non-goals

- Issue 5 (unresolved Specialist package capability references) is intentionally unchanged. No fields or version changes are made to `specialists.json`.
- No database, settings JSON, migration, or other persisted-format change.
- Existing imports that already lost unresolved references are not reconstructed.
- New states are view/runtime-only: `SkillView.availability = 'identity-conflict'` and `ConnectorsSnapshot.skillProjectionStatus = 'degraded'`.

The report was first reproduced with public behavior tests on the unmodified implementation: failed catalog publication was not retried; failed Connector refresh fulfilled and discarded materialized names; duplicate-name Skills collapsed to one row; duplicate-ID Skills produced no Settings rows. The retained regression tests fail for those reported reasons before the fixes and pass afterward.

Final checks ran after the last material source edit:

- User Skill observer retry and conflict discovery -> `npm run test:module -- user_skills_repository` -> 19 files, 810 tests passed.
- Settings command/IPC/contract consumers -> `npm run test:module -- settings_service_facade` -> 30 files, 922 tests passed.
- Connector projection preservation/degraded retry plus focused Skill/Renderer/i18n behavior -> targeted `npm test -- <14 paths>` -> 14 files, 1,252 tests passed.
- Node and Renderer interfaces -> `npm run typecheck:node`; `npm run typecheck:web` -> passed.
- Lint and generated Web API contract -> `npm run lint`; `npm run check:web-api-map` -> passed.

The module selector requests the complete Vitest suite because `src/main/connectors/application.ts` has no declared module owner. Per maintainer instruction, the full local `npm test` run is deferred to PR CI. Cross-platform watcher behavior and packaged-app UI remain CI/runtime risks.

## Review focus

- Confirm failed full and targeted Connector refreshes keep the queue live while degraded state clears only on an appropriate successful refresh.
- Confirm Settings-only conflict rows cannot reach ID-based edit, export, delete, tag, bulk-management, or toggle mutations.
- Confirm no Specialist persistence/import/export behavior changed.


ewen-poch:
Addressed the forced catalog retry feedback in 25e4b730. The retry timer now enqueues a forced reconciliation, so a failed explicit catalog publication cannot be skipped when the filesystem fingerprint is unchanged. Added a regression test covering a successful initial scan followed by a failed forced notification and automatic retry. Verified with the user_skills_repository module suite (814 tests).

PR #2081, opened 2026-09-02
fix(notebook): launch protected Windows kernels directly

ewen-poch:
## Problem

On packaged Windows installs with the protected Notebook network sandbox ready, persistent kernels were launched through `powershell.exe -Command`. When the Notebook working directory was on a non-system drive such as `D:`, PowerShell could fail to initialize its FileSystem provider inside the AppContainer before Python started. The UI then reported only `Notebook kernel process exited.`, hiding the exit reason and stderr.

The production runtime differential was:

- managed Python launched directly in the AppContainer: NumPy and Matplotlib succeed;
- the same interpreter launched through the existing PowerShell wrapper: exits before Python starts with a missing FileSystem drive/provider error.

## Proposed change

- Carry an optional structured `executable` and `args` through the Notebook sandbox adapter.
- Use that exact argv for native executables in protected Windows AppContainer launches, avoiding the PowerShell bootstrap.
- Route `.cmd`/`.bat` interpreter shims through `cmd.exe` with pre-escaped, verbatim arguments; the native host preserves that prepared command line.
- Keep the existing command string for policy and diagnostic purposes.
- Include the exit code/signal plus the existing bounded, annotated stderr tail when a persistent kernel exits unexpectedly.
- Keep process liveness tied to the kernel's `exit` event, so a descendant that inherits stdio cannot leave a dead kernel registered while holding its pipes open.
- Annotate crash stderr before releasing the sandbox context that owns filesystem classifications and recorded violations.

## Scope and non-goals

- Direct or batch-shim argv behavior is limited to protected Windows AppContainer launches. Standard Windows execution and POSIX launch behavior are unchanged.
- No database schema, persisted-format, historical-data compatibility, state-enum, or renderer interaction changes.
- Existing execution-result persistence may store the richer error string through its current field; no new persistence path or field is introduced.
- `verbatimArguments` is an optional, default-false field in the internal transient host launch specification; it is neither persisted nor part of the user-facing data model.
- This does not attempt to repair syntactically escaped Python input such as `2 \\* np.pi`; such code remains an ordinary cell error and should not terminate the kernel.

The original three report-focused regression tests were added before the launch fix and failed consistently for the reported reasons: the Windows launch spec selected PowerShell, the owner dropped structured argv, and an early kernel exit lost its exit code/stderr. They pass after the fix.

The first AI review identified that waiting for `close` could hang if a kernel descendant inherited stdio. A public executor-boundary regression was added and failed before the lifecycle correction: the execution remained unsettled after the direct kernel exited. It passes after restoring `exit` as the liveness boundary.

The next AI review identified two more regressions. Before fixing them, focused tests confirmed both failures: a batch interpreter remained the direct `CreateProcessW` target, and sandbox cleanup ran before crash stderr annotation. The fixes route batch shims through safely escaped `cmd.exe` argv and keep the sandbox context alive until annotation is complete. A Windows-only end-to-end test also executes a generated `.cmd` interpreter shim with a metacharacter-bearing argument.

Final evidence mapping (all listed checks ran after the last material edit):

- native argv, `.cmd`/`.bat` shims, early-exit diagnostics, annotation ordering, and inherited-stdio lifecycle -> focused Vitest set -> 7 passed, 110 skipped
- Notebook network sandbox owner/contract/consumer coverage -> `npx vitest run packages/notebook-network-sandbox/src` -> 76 passed, 13 skipped
- native host command-line construction -> `cargo test windows_command_line` -> 2 passed
- main-process and sandbox TypeScript contracts -> `npm run typecheck:node` -> passed
- lint and formatting -> `npm run lint`, `cargo fmt --check`, and `git diff --check` -> passed
- production-equivalent Windows AppContainer repro -> managed NumPy probe and the reported Matplotlib program both completed; the latter printed `saved sin_plot.png` and returned a figure

The full native-host test binary compiled and ran locally: 8 of 9 tests passed. Its unrelated ACL inheritance test could not modify the current user's temporary-directory ACL (`Access is denied`); both command-line tests pass independently.

The complete `kernel-executor.test.ts` file was also attempted locally earlier. 46 tests across the two requested files passed and 32 were skipped; 31 kernel tests could not enter their test bodies because Windows denied the existing symlink-based Python fixture with `EPERM`. The focused regressions use the installed managed interpreter and passed. The affected-test planner conservatively selected the full suite because `kernel-executor.test.ts` has no declared module owner; per the requested local-test scope, the complete `npm test` suite was left to PR CI.

The exact-head PR Gate passed Windows core, Windows E2E, Static, Linux isolation, macOS, and portable shards 1/2. Portable shard 3 failed in unrelated `runtime-service.test.ts` shell/session shutdown tests: asynchronous repository writes continued after their temporary Session directories were removed, producing `ENOENT` plus two shell-run assertion failures; Module coverage failed only because it merges the shard result. These tests do not exercise the changed kernel/AppContainer files. The prior scheduler-route count failure disappeared after `main` advanced. The latest AI review also raised a generic orphan-process risk that already exists on the branch base; the protected Windows path here is supervised by the native host Job Object. A local heartbeat reproduction showed that the suggested post-exit `killChildTracked` call cannot rediscover a descendant after the parent PID is gone, so a real generic fix requires a separate cross-platform process-ownership design rather than an ineffective change in this PR.

## Review focus

- Confirm native executables avoid PowerShell only in protected Windows mode.
- Confirm `.cmd`/`.bat` shims preserve their effective argv through `cmd.exe`, including metacharacters.
- Confirm kernel liveness settles on direct-process `exit` and crash diagnostics are annotated before sandbox cleanup without waiting for descendant-held pipes.
- Exact-head PR Gate and its Windows lane remain authoritative for the uncovered platform/test-fixture risk.

### DIFF
--- a/packages/notebook-network-sandbox/runtime/src/notebook-runtime.ts
+++ b/packages/notebook-network-sandbox/runtime/src/notebook-runtime.ts
@@ -49,6 +49,8 @@ type NetworkAskCallback = (request: {
 
 type NetworkWrapRequest = Readonly<{
   command: string
+  executable?: string
+  args?: readonly string[]
   commandId: string
   shell?: string | WindowsShell
   cwd: string
@@ -226,6 +228,7 @@ const wrap = async (
     if (process.platform === 'win32') {
       const launchRequest = {
         command: request.command,
+        ...(request.executable ? { executable: request.executable, args: request.args ?? [] } : {}),
         ...(request.shell ? { shell: request.shell } : {}),
         gatewayPort: gateway.port,
         gatewayCredentials: credentials,
--- a/packages/notebook-network-sandbox/runtime/src/platform/windows-appcontainer.ts
+++ b/packages/notebook-network-sandbox/runtime/src/platform/windows-appcontainer.ts
@@ -16,6 +16,8 @@ type WindowsShell = Readonly<{ kind: 'powershell' | 'cmd'; path: string }>
 
 type WindowsLaunchRequest = Readonly<{
   command: string
+  executable?: string
+  args?: readonly string[]
   shell?: string | WindowsShell
   cwd: string
   gatewayPort: number
@@ -320,11 +322,13 @@ const windowsLaunch = (
     shell.kind === 'powershell'
       ? ['-NoLogo', '-NoProfile', '-NonInteractive', '-Command', request.command]
       : ['/d', '/s', '/c', request.command]
+  const executable = request.executable ?? shell.path
+  const args = request.executable ? [...(request.args ?? [])] : childArgs
   const layout = normalizeFilesystemLayout(request.filesystem)
   const specification = Buffer.from(
     JSON.stringify({
-      executable: shell.path,
-      arguments: childArgs,
+      executable,
+      arguments: args,
       cwd: request.cwd,
       readOnlyRoots: layout.readOnlyRoots,
       readWriteRoots: layout.readWriteRoots,
--- a/packages/notebook-network-sandbox/runtime/src/platform/windows-appcontainer.ts
+++ b/packages/notebook-network-sandbox/runtime/src/platform/windows-appcontainer.ts
@@ -30,6 +30,37 @@ type WindowsLaunchRequest = Readonly<{
   ownershipRoot: string
 }>
 
+const WINDOWS_BATCH_FILE = /\.(?:cmd|bat)$/i
+const CMD_META_CHARACTER = /([()\][%!^"`<>&|;, *?])/g
+
+// Ported from cross-spawn's Windows non-shell parser. Batch files require cmd.exe, while escaping
+// each token and asking the native host to preserve the resulting command line prevents cmd syntax
+// in an interpreter path or argument from becoming a second command.
+const escapeCmdCommand = (value: string): string => value.replace(CMD_META_CHARACTER, '^$1')
+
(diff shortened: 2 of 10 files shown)


ewen-poch:
The reported orphan risk is valid for a generic unsupervised kernel, but it is not introduced by this PR and the suggested post-exit teardown is not sufficient:

- The branch base (`7fb803911`) already removes a crashed kernel from `procs` on the direct child's `exit` without calling `killChildTracked`; this PR keeps that liveness boundary and only enriches the error/cleanup ordering.
- The protected Windows path changed here is supervised by the native AppContainer host. `launch_child` assigns the kernel to a Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`; when the kernel exits, the host drops that Job before its own exit is observed by `NotebookKernelExecutor`, terminating inherited descendants.
- I strengthened the descendant test locally with a heartbeat, added the recommended `killChildTracked(proc)` after direct-child exit, and reran it on Windows. It still failed: `taskkill /T` cannot rediscover the already reparented descendant from a dead parent PID, `shutdown().reaped` was false, and the heartbeat continued. I reverted that uncommitted experiment rather than land a false fix.

Closing the pre-existing generic gap requires a separate process-ownership design: owned POSIX process groups plus a Windows Job/supervisor for unprotected launches, with a cross-platform teardown contract. That is outside this protected-Windows launch fix and should be handled independently.

PR #2086, opened 2026-09-02
fix(acp): recover finalized Claude model call usage

ewen-poch:
## Problem

Claude Code can emit provisional zero token usage in live Assistant SDK messages when a turn runs through MiniMax-M3. The provider result still reports the correct aggregate usage and turn count, so Open Science persists Turn history but omits Calls because the zero-valued rows fail exact reconciliation.

Real provider reproduction confirmed that Claude writes finalized per-invocation usage to the same Session transcript before the prompt completes.

## Proposed change

- expose the active Claude config directory in the secret-free backend adapter view
- read the finalized Session transcript through the Claude SDK SessionStore boundary only when live call details fail exact reconciliation
- match only invocation ids observed during the current prompt, deduplicate repeated transcript fragments, and retain the existing fail-closed count and token-total checks
- preserve the fast path without transcript I/O when live call details are already exact

## Scope and non-goals

- no database, schema, persisted-field, enum, or renderer changes
- no historical Session backfill
- no synthesized or partial call rows
- no behavior changes for OpenCode, Codex, or CodeBuddy
- missing, malformed, late, or inconsistent transcripts keep the existing Calls unavailable behavior

All listed checks ran against the final commit.

- MiniMax-M3 provisional-zero regression and exact-call recovery -> npm test -- --run src/main/acp/claude-turn-adapter.test.ts src/main/acp/claude-transcript-reader.test.ts src/main/acp/backend-generation-owner.test.ts src/main/acp/provider-prompt-executor.test.ts -> 4 files, 33 tests passed
- Node adapter and SDK SessionStore types -> npx tsc --noEmit -p tsconfig.node.json --composite false -> passed
- changed ACP source and tests -> targeted npx eslint invocation -> passed
- committed patch whitespace -> git diff --check HEAD^ HEAD -> passed

An Electron production build also passed before the final test-only assertion adjustment. PR CI remains authoritative for complete portable and platform coverage.

## Review focus

- whether filtering transcript rows to invocation ids observed during the current prompt is sufficiently strict
- whether the read-only SessionStore adapter preserves the expected Claude config-directory isolation on every platform
- whether fallback failures retain enough aggregate usage while still refusing inexact call rows

### DIFF
--- a/src/main/acp/backend-generation-owner.test.ts
+++ b/src/main/acp/backend-generation-owner.test.ts
@@ -4,6 +4,26 @@ import { claudeCodeFramework, codexFramework } from '../agent-framework'
 import { AcpBackendGenerationOwner } from './backend-generation-owner'
 
 describe('AcpBackendGenerationOwner', () => {
+  it('publishes the Claude transcript root without provider credentials', () => {
+    const owner = new AcpBackendGenerationOwner(claudeCodeFramework)
+    const attempt = owner.prepare(
+      { epoch: 1, assertCurrent: vi.fn() },
+      {
+        framework: claudeCodeFramework,
+        executablePath: '/bin/claude',
+        env: {
+          CLAUDE_CONFIG_DIR: '/profiles/app-claude',
+          ANTHROPIC_AUTH_TOKEN: 'provider-secret'
+        }
+      }
+    )
+
+    const view = attempt.publish()
+
+    expect(view.adapter.claudeConfigDir).toBe('/profiles/app-claude')
+    expect(JSON.stringify(view)).not.toContain('provider-secret')
+  })
+
   it('publishes one immutable secret-free behavior view atomically', () => {
     const owner = new AcpBackendGenerationOwner(claudeCodeFramework)
     const sessionOptions = { settingSources: ['user'] }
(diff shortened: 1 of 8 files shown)


ewen-poch:
Addressed the P1 in c9e37a06. Transcript recovery now reads immediately and retries after bounded 50 ms, 100 ms, and 200 ms delays only when live call details have already failed exact coverage. It exits as soon as exact coverage is available, preserves the no-I/O live fast path, and still fails closed after 350 ms total backoff. The MiniMax-M3 regression now makes the first transcript read incomplete and the second read complete; the focused ACP suite passes with 33 tests.

ewen-poch:
Addressed the P2 in d23f97d9. A transient transcript read failure now consumes the same bounded retry budget instead of aborting recovery, while exact aggregate and invocation-ID coverage checks remain unchanged. The regression reproduces an incomplete trailing JSONL read as an initial SyntaxError and verifies that the next finalized read recovers all three MiniMax-M3 calls. Focused ACP tests pass: 4 files, 33 tests; TypeScript, ESLint, Prettier, and diff checks also pass.

ewen-poch:
Not adopting this suggestion in the current PR. The committed npm lockfile places @anthropic-ai/claude-agent-sdk@0.3.232 at the root node_modules path, matching the exact dependency pinned by @agentclientprotocol/claude-agent-acp@0.70.0, and the Electron packaging rules include the SDK package. The production require therefore resolves in the supported locked install and packaged layout. Adding a second direct declaration would expand package-manifest scope and couple the app to the ACP adapter's internal SDK version without fixing a currently reproducible failure. We can revisit a direct dependency if the SDK becomes an independently owned integration or the locked package layout changes.

PR #2087, opened 2026-09-03
fix(compute): harden remote transfers and host cleanup

ewen-poch:
## Problem

Remote downloads only compared file size before and after transfer, so a same-size replacement could publish bytes that did not belong to one stable remote version. Harvest also accepted successful bounded transfers without requiring the reported and local byte counts to equal the enumerated size.

Compute Host deletion treated every historical Job row as a permanent blocker. The removal dialog did not identify those Jobs or provide a safe way to clean their remote directories or explicitly abandon cleanup while retaining local history.

## Proposed change

- Capture remote file size, inode, and modification time before direct downloads and Harvest transfers, then require the post-transfer snapshot and exact local byte count to match before publishing the staged file.
- Persist a separate Compute Job remote-cleanup disposition: pending, cleaned, or abandoned.
- Keep Host credentials while any Job remains pending; mark cleaned only after owned remote cleanup and any active cancellation complete.
- Allow terminal Jobs to be explicitly abandoned while preserving Job history and local outputs. Active or queued Jobs must use cancellation plus remote cleanup.
- List blocking Jobs in the Host removal dialog, with per-Job cleanup, terminal abandonment, and bulk cleanup actions.

## Scope and non-goals

- No ComputeJobStatus values are added or changed.
- No individual Job-history deletion is introduced.
- No remote checksum is computed by default; size, inode, mtime, and exact received-byte checks provide the minimum stable-snapshot boundary without reading every large remote file twice.
- No cleanup timestamp or audit table is added.

## Persisted format and historical compatibility

- Migration 0026 adds ComputeJob.remoteCleanupDisposition as non-null text with a pending default.
- Historical Jobs intentionally migrate to pending and continue blocking Host deletion until cleaned or explicitly abandoned. The migration does not guess that old remote directories are gone.
- Invalid persisted disposition values are projected as pending so deletion fails closed.
- Job history and harvested local outputs remain intact after either cleanup decision.

Regression tests were first run against the unfixed behavior and reproduced same-size replacement acceptance, Harvest short-transfer acceptance, and the missing blocker-resolution UI.

Final checks after the last material edit:

- Remote transfer, Job cleanup/cancellation, IPC, and Compute consumers -> npm run test:module -- compute_service -> 49 files passed, 1 skipped; 1158 tests passed, 6 skipped.
- Database schema and historical migration -> npx vitest run src/main/database -> 15 files and 152 tests passed.
- Removal UI, preload, renderer/Web contracts, and translations -> targeted Vitest run -> 6 files and 931 tests passed.
- Main process and sandbox types -> npm run typecheck:node -> passed.
- Renderer and shared types -> npm run typecheck:web -> passed.
- Lint -> npm run lint -> passed.
- Generated database schema -> npm run db:schema:check -> passed.
- Generated Web API map -> npm run check:web-api-map -> passed.

The affected-test classifier conservatively selects the complete Vitest suite because prisma/schema.prisma is not assigned to a module. Per maintainer direction, the complete suite is left to PR Gate CI.

## Review focus

- Stable-snapshot comparison and failure cleanup before staged files are published.
- Fail-closed migration policy for historical Jobs.
- Stop-before-remove ordering for active Jobs and serialization between cleanup and abandonment.
- Electron, local Web, and remote Web command-contract coverage.


ewen-poch:
Addressed both automated-review P1 findings in 8bec2aa3: Harvest and stat mtimes now normalize to the same epoch-second representation, and owner deletion skips remote cleanup for Jobs already marked cleaned or abandoned. Added regressions for GNU find fractional mtimes and post-Host-removal owner deletion. The final Compute module run passes 49 files / 1158 tests, with 1 file / 6 tests skipped by existing conditions.

ewen-poch:
CI follow-up: Ubuntu portable shard 1 exposed an omitted packaged migration-ledger update for the new `0026_compute_job_remote_cleanup` migration. Commit c7215c9e pins the migration identity/checksum and updates legacy-schema replay fixtures to remove/reapply the new column. The exact smoke suite now passes locally (15/15), and lint passes.

ewen-poch:
Addressed the cancellation/cleanup race and active-process verification findings in 7bf4bf25. Manual active cleanup now pauses and drains the shared poller/cancellation runtime before requesting cancellation, cleanup settlement requires an unclaimed cancellation operation, and the remote command aborts without removing the directory when PID ownership or signal delivery is unverified. Regressions failed on the prior code for all three conditions and now pass. Validation: Compute module 49 passed / 1 skipped files, 1159 passed / 6 skipped tests; Node typecheck and lint pass.

ewen-poch:
Addressed the final automated-review lifecycle finding in 211434b0. Added a repository-boundary regression that first reproduced cleaned and abandoned jobs both being returned by terminal harvest recovery. The query now excludes only cleaned jobs, while abandoned jobs remain harvest-eligible. No schema, persistence field, or enum change was needed for this follow-up. Validation: job repository 29/29, Node typecheck, and lint pass.

ewen-poch:
Addressed the terminal-job data-loss finding in 0094acdf. Regressions first proved that the backend accepted cleanup for an unharvested terminal Job and the removal dialog enabled that action. Terminal success/failed/timeout Jobs now require durable harvest completion before remote cleanup; active Jobs still support the explicit cancel-and-clean path. The deletion-status response adds only an optional harvested projection so older payloads fail closed. No database field, migration, or enum was added. Validation: 5 focused files / 958 tests, Compute module 49 passed and 1 skipped files / 1161 passed and 6 skipped tests, Node and Web typecheck, lint, i18n guard, and Prettier all pass. The previous Windows E2E failure was an unrelated context-compaction timing flake (21 tests passed, 1 flaky) and was queued for rerun before this new head superseded it.

ewen-poch:
Addressed the latest automated-review findings in c759a42e. Regressions first reproduced both failures on the prior code: a remote-Web caller reached the remote-cleanup owner, and same-size/same-inode rewrites within one second were accepted by both direct download and Harvest. Remote cleanup is now local-only in the renderer contract and independently enforced by the application handler. Remote snapshots now request fractional epoch mtimes on GNU and BSD stat implementations and preserve normalized sub-second precision. No schema, persistence field, migration, or enum was added. Validation: 7 focused files / 152 tests; scoped Compute and renderer-contract suite 53 passed / 2 skipped files, 1050 passed / 7 skipped tests; Node and Web typecheck, lint, and Prettier all pass.

ewen-poch:
CI follow-up: Static checks exposed one additional exact renderer-surface inventory that had not been included in the focused contract set. Commit 9491913e adds only  to that expected local-only channel inventory. The failing CI assertion precisely reproduced the omission; the complete affected interface-contract set now passes locally (5 files / 48 tests), with formatting clean.

ewen-poch:
Correction: commit 9491913e adds only the compute:jobs:set-remote-cleanup channel to the expected local-only renderer inventory. The CI failure itself was the evidence for the missing exact-list entry; the affected interface-contract set passes locally (5 files / 48 tests), with formatting clean.

ewen-poch:
Addressed the two reproduced lifecycle findings in 0eb8f6ee. Red tests showed that a harvested Job with pending remote cleanup did not block either authentication-change guard, and that SQLite accepted an invalid cleanup disposition. Both guards now include pending cleanup. Migration 0026 now rebuilds ComputeJob and ComputeJobOperation with a named pending/cleaned/abandoned CHECK while preserving historical rows, operations, indexes, and foreign keys; historical rows still default to pending. The migration ledger and legacy fixtures were updated. The BSD stat finding was not reproducible: on the actual Darwin runner, stat -f with %.9Fm returns a numeric fractional epoch. A real-shell public download regression now exercises the BSD find-to-stat fallback and passes without reducing mtime precision. Validation: focused 150/150, database module 152/152, lint, Node/Web typecheck, and generated-schema check all pass.

ewen-poch:
Addressed the latest queued-cleanup race in 5ab15fce. Red-first regression: a Job first read as queued was promoted to running while owner pause drained an in-flight dispatch; the previous code read it once, ran no remote cleanup command, and still resolved with remote cleanup disposition cleaned. The cleanup path now pauses the affected queue owner, pauses polling, re-reads and revalidates the Job after dispatch drain, and derives cancellation plus remote cleanup from the refreshed status/handle before settling. Both pauses are released in finally. No schema, persisted field, or enum change. Validation: job-deletion-owner + concurrency-manager 67/67; full src/main/compute module 1031 passed, 7 skipped; Node/sandbox typecheck, lint, Prettier, and diff check passed.

ewen-poch:
Addressed the uninspectable-workdir finding in 6b602f79. Red-first regression executes the generated command with a real POSIX shell: when the expected Job path is a regular file, the previous command exited 0 and left it in place. The command now requires the scratch root and Job parent to be canonically inspectable, rejects a present non-directory or uninspectable workdir, validates the parent remains under the expected scratch layout, and verifies the original target path is absent after cleanup (including dangling symlinks). A companion assertion confirms an actually absent Job path under an inspectable parent remains an idempotent success. No schema, persisted field, enum, or interaction change. Validation: full src/main/compute module 1032 passed, 7 skipped; Node/sandbox typecheck, lint, Prettier, and diff check passed.

ewen-poch:
Addressed both latest findings in af32af53. Red-first proof was run by temporarily reverting only the production fixes while retaining tests: the snapshot command lacked dereferencing and the submitted-without-PID cleanup resolved cleaned; both tests failed for those exact reasons. Remote snapshots now use GNU find -L so metadata matches the target scp reads (BSD still falls back to dereferencing stat). Cleanup for submitted Jobs without a durable handle now requires either an absent workdir or a valid job.pid witness; an existing workdir without that witness fails closed and remote_cleanup_disposition remains pending. No new field, enum, migration, or interaction. Validation after restoration: focused 97/97; full src/main/compute 1033 passed, 7 skipped; lint, Prettier, diff check and Node/sandbox typecheck passed.

PR #2097, opened 2026-09-03
fix(settings-errors): hide backend diagnostics in alerts

ewen-poch:
## Summary

- replace raw renderer exception messages with stable translated error keys for Runtime, Connector retry/removal/Skill projection, and package mirror failures
- preserve Runtime rejection semantics while keeping backend paths and implementation details out of primary alerts
- add public rendering-boundary regressions using path-bearing diagnostics, including language switching after a mirror save failure

## Verification

- related Vitest suites and i18n guard: 922 passed
- npm run typecheck:web
- targeted ESLint and git diff --check

## Full-suite note

A non-sandbox npm test run completed with 23,509 passing tests and 4 unrelated existing failures: one WorkspaceMessageScroller preview-argument assertion, one session-persistence architecture count assertion, and two real-R image-capture assertions. Isolated reruns reproduced those failures outside the files changed here.

### DIFF
--- a/src/renderer/src/pages/settings/ConnectorsPanel.render.test.tsx
+++ b/src/renderer/src/pages/settings/ConnectorsPanel.render.test.tsx
@@ -453,6 +453,25 @@ describe('ConnectorsPanel (groups)', () => {
     expect(retryConnectorProjection).toHaveBeenCalledOnce()
   })
 
+  it('shows a safe Skill projection retry failure instead of backend details', async () => {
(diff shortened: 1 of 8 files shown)


ewen-poch:
The initial Windows E2E failure was an unrelated timing flake in `workspace-conversation.spec.ts`: the first attempt timed out after clicking the previous revision, while Playwright's retry passed. A clean rerun completed all 22 Windows workspace journeys without retries or failures, and the final PR Gate is green. No code or test changes were made because the failure is outside this PR's settings-error scope and was not reproducible.

PR #2098, opened 2026-09-03
fix(skills): support helper validation without audit hooks

### LINKED ISSUE #2093: [Bug]: Skill import fails — helper "figure-composer" validation errors on missing sys.addaudithook
### Preflight checklist

### What happened?

Environment: Open Science workbench (Windows), with a valid GitHub token configured.

Steps to reproduce

Configure a GitHub token (unauthenticated requests first fail with rate-limiting; after the token is set, the next error appears).
Import any skill directory from https://github.com/Yuan1z0825/nature-skills, e.g. skills/nature-polishing.
Confirm the preview and observe the failure.
Actual result: every skill in the repo fails to import with:
INVALID_REGISTERED_HELPER: helper "figure-composer" failed isolated callable export validation: AttributeError: module 'sys' has no attribute 'addaudithook'
Key diagnostic

sys.addaudithook was added in CPython 3.8, so the import-validation sandbox is running Python < 3.8.
Importing from an older commit (699f58885f925e08c8e5f3eb9f6d41203ccdc53e) that contains no figure-composer / addaudithook code (verified by full-text search) still fails identically — suggesting the helper is a built-in app capability, or the importer ignores the commit ref and always fetches main.
Expected: the skill should import successfully.

Suggested fix: run the import-validation sandbox on Python ≥ 3.8, or make the figure-composer helper guard sys.addaudithook behind hasattr(sys, 'addaudithook').

### Steps to reproduce

Configure a GitHub token (unauthenticated requests first fail with rate-limiting; after the token is set, the next error appears).
Import any skill directory from https://github.com/Yuan1z0825/nature-skills, e.g. skills/nature-polishing.
Confirm the preview and observe the failure.

### Operating system

Windows

### App version

24

### Provider / model

_No response_

```shell

```

ewen-poch:
## Problem

On Windows, a Python runtime without `sys.addaudithook` causes an otherwise valid Skill import to fail during promoted-catalog validation:

`INVALID_REGISTERED_HELPER: helper "figure-composer" failed isolated callable export validation: AttributeError: module 'sys' has no attribute 'addaudithook'`

The imported repository does not supply `figure-composer`. Promotion validates the merged catalog, including the application-bundled helper. A URL pinned to the reporter's older repository commit is parsed and fetched with that ref, so changing the ref cannot remove this built-in validation failure.

## Proposed change

- Mark helper source as trusted only when its catalog origin is `builtin`.
- Keep audit-hook-protected validation unchanged whenever the Python runtime provides `sys.addaudithook`.
- When the hook is unavailable, allow only trusted application-bundled helper source through the existing restricted-builtins/import environment.
- Reject personal or imported helper source before executing it when audit-hook protection is unavailable.
- Add focused regression coverage that simulates the missing hook. Before the fix it failed three consecutive runs with the exact reported `figure-composer` `AttributeError`; after the fix the trusted built-in validates, while external source remains fail-closed.

## Scope and non-goals

- No database or persisted-format changes.
- No historical-data migration or compatibility handling.
- No new fields, statuses, or enum values.
- No architecture, data-model, or UI interaction changes.
- No changes to GitHub token, rate-limit, commit-ref, or download behavior.
- No attempt to build a second Python interpreter for legacy runtimes; external helper validation still requires audit-hook support.

This is intentionally narrower than a general legacy-Python fallback. Application-bundled helper code is already part of the trusted app distribution, while imported and personal code retains the audit boundary.

All listed checks ran after the final material edit and after rebasing onto the latest `origin/main`:

- Missing audit hook and trust boundary -> `npm test -- src/main/notebook/python-command.test.ts` -> 12 passed.
- Registered helper generation and isolated validation -> `npm test -- src/main/skills/registered-helper-catalog.test.ts` -> 33 passed.
- Skill catalog promotion, rollback, and bundled helper execution -> `npm test -- src/main/settings/skill-catalog.registered-helpers.test.ts` -> 4 passed.
- GitHub ref parsing/pinning and import behavior -> `npm test -- src/main/skills/github-import.test.ts` -> 37 passed.
- Formatting -> `npx prettier --check src/main/notebook/python-command.ts src/main/notebook/python-command.test.ts src/main/skills/registered-helper-catalog.ts` -> passed.
- Patch whitespace -> `git diff --check` -> passed.

The two suites that launch a real Notebook kernel were run outside the outer agent sandbox because nested macOS `sandbox-exec` is rejected there. Both pass in their intended execution environment. PR CI remains the final cross-platform authority; no full local `npm test` was run.

## Review focus

- Only catalog entries whose origin is `builtin` may set `trustedSource`.
- External and personal helper source must fail before `exec` when the runtime lacks `sys.addaudithook`.
- The existing audited validation behavior must remain unchanged on supported Python runtimes.

Fixes #2093

### DIFF
--- a/src/main/notebook/python-command.test.ts
+++ b/src/main/notebook/python-command.test.ts
@@ -212,6 +212,22 @@ describe('validateNotebookHelperExports', () => {
     await expect(
       validateNotebookHelperExports(
         'figure-composer',
+        'int = 1\nclass FigureComposer(int):\n    pass\n',
+        ['FigureComposer'],
+        { python: legacyPython }
+      )
+    ).rejects.toThrow('legacy helper validation requires side-effect-free definitions')
+    await expect(
+      validateNotebookHelperExports(
+        'figure-composer',
+        'class FigureComposer:\n    int = 1\n    class Nested(int):\n        pass\n',
+        ['FigureComposer'],
+        { python: legacyPython }
+      )
+    ).rejects.toThrow('legacy helper validation requires side-effect-free definitions')
+    await expect(
+      validateNotebookHelperExports(
+        'figure-composer',
         'LEFT, RIGHT = (1,)\ndef compose_figure():\n    return None\n',
         ['compose_figure'],
         { python: legacyPython }
--- a/src/main/notebook/python-command.test.ts
+++ b/src/main/notebook/python-command.test.ts
@@ -156,7 +156,7 @@ describe('validateNotebookHelperExports', () => {
     await expect(
       validateNotebookHelperExports(
         'figure-composer',
-        'def compose_figure():\n    return None\n',
+        'META_GREY = "#888888"\ndef compose_figure():\n    return None\n',
         ['compose_figure'],
         { python: legacyPython }
       )
@@ -169,6 +169,22 @@ describe('validateNotebookHelperExports', () => {
         { python: legacyPython }
       )
     ).rejects.toThrow('missing or non-callable exports: compose_figure')
+    await expect(
+      validateNotebookHelperExports(
+        'figure-composer',
+        'def compose_figure():\n    return None\ncompose_figure = 1\n',
+        ['compose_figure'],
+        { python: legacyPython }
+      )
+    ).rejects.toThrow('missing or non-callable exports: compose_figure')
+    await expect(
+      validateNotebookHelperExports(
+        'figure-composer',
+        'open("validation-side-effect", "w")\ndef compose_figure():\n    return None\n',
+        ['compose_figure'],
+        { python: legacyPython }
+      )
+    ).rejects.toThrow('legacy helper validation requires side-effect-free definitions')
   })
 
   it('validates UTF-8 helper source when the interpreter defaults stdin to a legacy encoding', async () => {
--- a/src/main/notebook/python-command.test.ts
+++ b/src/main/notebook/python-command.test.ts
@@ -152,6 +152,17 @@ describe('validateNotebookHelperExports', () => {
       ...python,
       baseArgs: [...python.baseArgs, '-c', 'import sys; del sys.addaudithook; exec(sys.argv[-1])']
     }
+    const futureSource =
+      'from __future__ import annotations\ndef compose_figure(value: MissingType):\n    return value\n'
+
+    await expect(
+      validateNotebookHelperExports('figure-composer', futureSource, ['compose_figure'], { python })
+    ).resolves.toBeUndefined()
+    await expect(
+      validateNotebookHelperExports('figure-composer', futureSource, ['compose_figure'], {
+        python: legacyPython
+      })
+    ).resolves.toBeUndefined()
 
     await expect(
       validateNotebookHelperExports(
--- a/src/main/notebook/python-command.test.ts
+++ b/src/main/notebook/python-command.test.ts
@@ -156,7 +156,7 @@ describe('validateNotebookHelperExports', () => {
     await expect(
       validateNotebookHelperExports(
         'figure-composer',
-        'META_GREY = "#888888"\ndef compose_figure():\n    return None\n',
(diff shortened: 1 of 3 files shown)


ewen-poch:
Addressed the imported-name/attribute annotation finding in 0eb7dcb7, including safe imported defaults. I did not add `__future__`: the existing audit-hook path also rejects `from __future__ import annotations` because `restricted_import` only admits `allowed_modules`, which does not include `__future__`; compiling the directive emits an `IMPORT_NAME __future__`. Adding it would widen the existing helper import policy rather than restore legacy parity, so it is outside #2093 and should be considered separately if desired.

ewen-poch:
Follow-up c0425f6f now addresses the `__future__.annotations` point as well. The implementation adds only the compiler-directive module to the existing allowlist, compiles without executing on the legacy path to validate feature support and directive placement, and skips annotation resolution only when the directive is present. This keeps the host-access boundary closed while making both validation paths consistent.

ewen-poch:
Follow-up on the c0425f6f review:

- The safe-builtins compatibility gap was valid. Commit 362ae6a0 now derives legacy static references from the same `safe_names` tuple used by the audited execution path. A regression test first failed with `FunctionDef`, then passed after the change for `callable` and `range`.
- The `__future__` finding is not a Python runtime mismatch. Python executes `from __future__ import annotations` with an `annotations` global bound to `__future__._Feature` (and honors `as` aliases); the audited validator therefore also accepts `def f(value=annotations): ...`. The added parity test exercises both the audited and legacy paths and confirms both accept the same source. Removing that binding in the legacy path would create the mismatch described by the review, rather than fix one.

Focused validation after 362ae6a0: 14/14 notebook command tests, 37/37 GitHub import tests, 37/37 registered-helper integration tests, node typecheck, lint, Prettier, and `git diff --check` all pass locally.

ewen-poch:
The repeated `__future__` finding is contradicted by CPython's compiled runtime behavior. Disassembling the exact shape shows:

```text
IMPORT_NAME              0 (__future__)
IMPORT_FROM              1 (annotations)
STORE_NAME               1 (annotations)
...
LOAD_NAME                1 (annotations)
SET_FUNCTION_ATTRIBUTE   1 (defaults)
```

Executing `from __future__ import annotations; print(annotations)` likewise prints the bound `__future__._Feature` object. This is why both the audited `exec(compile(...))` path and the legacy static path correctly accept `def compose_figure(value=annotations): ...`; special-casing away the binding would make legacy validation reject valid code and diverge from kernel initialization. Commit 98211f89 documents the non-obvious `IMPORT_FROM`/`STORE_NAME` behavior beside the static binding logic, and the dual-path parity regression remains green.

ewen-poch:
Resolution of the final automated review: no code change is warranted for the repeated `__future__` claim.

The exact source used by the test was executed independently with CPython 3.9.6, 3.12.8, 3.13.1, and 3.14.3. Every version produced an `annotations` global of type `_Feature` and a callable `f`. CPython bytecode also explicitly contains `IMPORT_FROM annotations` followed by `STORE_NAME annotations`; the default then loads that name. The final PR CI passes on all platforms, including the modern audited test path and Windows, so the claimed guaranteed `NameError` is empirically disproven.

Accordingly, the requested special case would be a regression: it would make legacy static validation diverge from the real Python execution path. The review finding is closed as a false positive, with the runtime semantics now documented in code and covered by a dual-path regression test.

ewen-poch:
The final head `ffa65849` removes the AST fallback entirely, so the prior `__future__` review discussion no longer applies to the diff.

The fix is now trust-scoped and 35 added / 4 removed lines relative to `main`: on runtimes without `sys.addaudithook`, only application-bundled helpers (`origin === 'builtin'`) may use the existing restricted execution validator; personal/imported helpers are rejected before `exec`. This fixes the reported built-in `figure-composer` failure during an otherwise helper-free import without weakening the boundary for external code or implementing a second Python interpreter.

PR #2118, opened 2026-09-03
fix(notebook): preserve stdin in Windows standard launch

wen2zhou:
## Problem

On packaged Windows builds, a Notebook kernel can fail during startup with:

> Notebook kernel process exited with exit code 0.

This is reproducible on Windows 10/11 x64 with Notebook protection enabled, particularly when the application and user data are located on different local drives.

When protected Notebook execution falls back to the Windows standard-launch path, a structured kernel invocation (`executable` plus `args`) is serialized into a PowerShell command. PowerShell does not transparently relay redirected stdin to a long-running native child in this scenario. The kernel can therefore observe EOF and exit before the executor writes its first protocol frame, producing the misleading exit-code-0 error.

## Proposed change

- Preserve structured native invocations in Windows standard mode by launching the executable and arguments directly.
- Keep the existing PowerShell or `cmd.exe` path for serialized commands and batch-file shims.
- Add a Windows regression test that delays the initial stdin write and verifies that the structured child remains alive and responds.
- Add a platform-independent assertion covering the generated direct argv.

## Scope and non-goals

This change is limited to Windows standard-mode launch planning in the Notebook network sandbox.

It does not change:

- protected AppContainer launch behavior;
- proxy environment construction;
- local RPC socket handling;
- batch-file execution through a command shell;
- Notebook protocol framing or kernel lifecycle management.

The checks listed below were run after the last material code edit.

- Structured standard-mode launch preserves the executable and argument boundary
  - `npx vitest run packages/notebook-network-sandbox/src/windows-appcontainer.test.ts`
  - Passed: 8 tests

- A Windows structured child remains alive across delayed stdin and processes the first frame
  - Covered by the Windows-only regression test in `windows-appcontainer.test.ts`
  - Passed in the final-state targeted test run

- Sandbox TypeScript contracts remain valid
  - `npm run typecheck:sandbox`
  - Passed

- Changed files conform to repository formatting
  - `npx prettier --check packages/notebook-network-sandbox/runtime/src/platform/windows-appcontainer.ts packages/notebook-network-sandbox/src/windows-appcontainer.test.ts`
  - Passed

A later rerun through the documented `npm test -- <path>` wrapper was blocked before test collection by a local `std-env` ESM/CJS dependency-loading error. The failure occurred while loading `vitest.config.ts`, not in the changed tests.

Broader local validation remains constrained by existing environment issues:

- repository-wide typecheck: unrelated missing `waitForMcpServers` export and `ajv` dependency;
- lint: local `acorn` dependency unavailable;
- broader Windows kernel tests: fixture setup encounters symlink `EPERM` without elevated developer-mode privileges.

Cross-version packaged Windows certification remains covered by CI/Nightly. Independent review is pending PR review.

## Review focus

Please focus on:

- whether direct argv launch is the correct boundary for structured non-batch executables;
- whether retaining shell execution for `.bat` and `.cmd` files preserves the intended compatibility behavior;
- whether the delayed-stdin regression test adequately represents the persistent Notebook kernel failure mode.

### DIFF
--- a/packages/notebook-network-sandbox/runtime/src/platform/windows-appcontainer.ts
+++ b/packages/notebook-network-sandbox/runtime/src/platform/windows-appcontainer.ts
@@ -64,7 +64,14 @@ const windowsBatchInvocation = (
 type WindowsStandardLaunchRequest = Readonly<
   Pick<
     WindowsLaunchRequest,
-    'command' | 'shell' | 'gatewayPort' | 'gatewayCredentials' | 'env' | 'localRpcSocketPath'
+    | 'command'
+    | 'executable'
+    | 'args'
+    | 'shell'
+    | 'gatewayPort'
+    | 'gatewayCredentials'
+    | 'env'
+    | 'localRpcSocketPath'
   >
 >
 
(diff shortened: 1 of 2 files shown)


ipan118:
Thank you for the standard-launch fix. Our packaged Windows x64 v0.25.1 observations appear consistent with the failure mode described in this PR, but we have not yet validated a fixed build locally.

Two separately approved calls in dedicated empty sessions ran only:

```javascript
console.log("NOTEBOOK_GATE_A_0251_OK");
```

Both failed before dispatch: `kernelDispatched=false`, zero stdout, and `Notebook kernel process exited with exit code 0.` Actual run intervals were 1.530 s and 1.698 s; longer UI durations included permission waiting.

A metadata-only external trace observed the application starting PowerShell, then an application-binary child. PowerShell exited with code 0 after approximately one second. The child's stop event was not captured, and it was absent from the final snapshot. These observations do not independently prove the child's exit code or stdin EOF. No host API, connector call, or research data was involved.

We have now located the official nightly at source `3f72dc95e7893a49fe2d6b683cfd9f9371e3ec05`. Its commit history contains #2118 and #2122; release metadata and certification identify `aipoch-open-science-0.25.1-nightly.3f72dc9-win-x64-setup.exe`. We have not downloaded, installed, or tested it.

Could you confirm the recommended supported validation route for an existing Windows installation:

1. Is this nightly suitable for bounded standard-mode REPL validation, or should we wait for a stable release containing the fix?
2. Is there a supported way to isolate its configuration and database from the existing research installation, and what rollback constraints apply?
3. Is one fixed-marker REPL call followed, only on success, by one `host.capabilities()` call an appropriate initial gate?

We can report bounded results from that gate. We would prefer to preserve the existing protection policy and avoid local installed-package patches or repeated identical probes.

PR #2123, opened 2026-09-03
fix(provider-transport): bound successful response bodies

ewen-poch:
## Summary

- cap provider success responses at 64 MiB with Content-Length preflight and observed-byte accounting
- cap native Responses SSE lines and events at 8 MiB, canceling the upstream body on overflow
- apply the shared bounded reader to Responses, compatibility, xAI OAuth, and Skill-selector JSON paths
- preserve controlled 502 or protocol-level stream failures without changing persisted settings or renderer UI

## Reproduction

Regression tests first demonstrated that the previous implementation:

- returned 200 for oversized successful JSON and binary responses
- consumed oversized SSE lines/events instead of canceling the upstream body
- completed Responses SSE, Skill selection, and xAI compatibility calls after their configured response budget was exceeded

## Tests

- provider bridge regression suite: 110 passed
- Responses response-adapter suite: 9 passed
- npm run typecheck:node
- ESLint on changed files
- git diff --check
- full npm test: 23,686 passed, 3 unrelated existing Notebook/R failures
  - shell admission timeout passes when rerun alone
  - two real-R TIFF capture assertions remain reproducible when rerun alone

### DIFF
--- a/src/main/settings/bounded-response.ts
+++ b/src/main/settings/bounded-response.ts
@@ -9,21 +9,25 @@ class ResponseBodyLimitError extends Error {
   }
 }
 
-const readBoundedResponseText = async (
+const DEFAULT_MAX_PROVIDER_RESPONSE_BYTES = 64 * 1024 * 1024
+const DEFAULT_MAX_PROVIDER_SSE_LINE_BYTES = 8 * 1024 * 1024
+const DEFAULT_MAX_PROVIDER_SSE_EVENT_BYTES = 8 * 1024 * 1024
+
+const consumeBoundedResponseBody = async (
   response: Response,
   maxBytes: number,
-  label: string
-): Promise<string> => {
+  label: string,
+  consume: (chunk: Uint8Array) => void
+): Promise<void> => {
   const declaredLength = response.headers.get('content-length')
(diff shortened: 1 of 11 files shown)


ewen-poch:
Updated the PR title to follow the required scoped Conventional Commit format: fix(provider-transport): bound successful response bodies.

ewen-poch:
Rebased onto the latest main (426cf400) and reran the provider response regression suites: 6 files, 119 tests passed.

ewen-poch:
Addressed the P2 BOM compatibility finding. The bounded text reader now decodes with TextDecoder, preserving the previous Response.json()/TextDecoder behavior for a leading UTF-8 BOM. Added a public ResponsesBridge regression test that returned HTTP 400 before the fix and now returns 200; the related suite passes 120/120, with Node typecheck and ESLint also passing.

ewen-poch:
Addressed the P1 terminal-state ordering finding. Any captured stream error is now handled before stop/tool_calls completion, so a response that exceeds maxResponseBytes after a terminal frame emits response.failed and never response.completed. The public bridge regression test reproduced the old completed outcome before the fix and now also verifies upstream cancellation. Related tests pass 120/120; Node typecheck, ESLint, and diff checks pass.

PR #2125, opened 2026-09-03
fix(settings): surface provider errors and bound persistence

ewen-poch:
## Problem

Provider deletion and vendor-model refresh calls could reject without visible feedback, leaving users with an unchanged or ambiguous Settings surface and an unhandled Promise rejection. Separately, `settings.json` was read and decoded without matching the Provider limits enforced on writes, so a corrupt or locally modified document could retain oversized fields, unbounded collections, and duplicate Provider identities.

## Proposed change

- Catch rejected Provider deletion and model refresh operations at their existing UI boundaries and show localized inline errors.
- Add an optional pre-read byte budget to durable JSON reads and apply a 128 MiB read/write limit to `settings.json`.
- Reapply Provider identity, name, URL, model ID, Provider-count, and fetched-model limits while decoding.
- Keep the first valid Provider for each ID and stop after 64 unique valid Providers; stop each fetched catalog after 2,000 valid model IDs.

## Scope and non-goals

- No persisted fields or settings schema-version changes.
- No IPC, renderer contract, or domain-model changes.
- One component-local action discriminator, `delete`, is added for presentation only; it is not persisted.
- Historical valid settings remain compatible. Corrupt duplicate IDs are deterministically first-wins, over-limit collection entries and optional fields are dropped, and documents over 128 MiB are rejected rather than migrated or silently truncated.
- The generic durable reader only gains an optional limit. Existing consumers retain their current behavior unless they opt in.

The regression tests were first run twice against the unmodified implementation. Both runs failed on the reported symptoms: no delete alert, no refresh alert plus an unhandled rejection, all 66 Providers retained, oversized Provider fields/catalogs retained, and an oversized durable read proceeding to `readFile`.

All commands below ran after the last material edit:

- Provider delete and refresh feedback -> `npx vitest run src/renderer/src/pages/settings/ProvidersPanel.race.test.tsx src/renderer/src/pages/settings/SettingsPage.render.test.tsx` -> passed as part of 161 focused tests.
- Settings byte, collection, duplicate-ID, and field limits -> `npx vitest run src/main/settings/document-store.test.ts src/main/storage/durable-json-file.test.ts src/main/settings/record-codec.test.ts src/main/settings/document-codec.test.ts` -> passed as part of 161 focused tests.
- Settings Provider owner/contracts/consumers -> exact test list emitted by `test:module settings_provider_accounts` -> 22 files, 438 tests passed.
- Settings Repository owner/contracts/consumers -> exact test list emitted by `test:module settings_repository` -> 20 files, 744 tests passed.
- Renderer translations -> `npx vitest run src/renderer/src/i18n/resources.test.ts` -> 805 tests passed.
- Main process types -> `npm run typecheck:node` -> passed.
- Renderer/Web types -> `npm run typecheck:web` -> passed.
- Exact-head impact explanation -> `npm run test:affected -- --base origin/main --head HEAD --explain` -> selected full mode because `settings-resource-limits.ts` has no declared module owner; the full local suite was intentionally not run per maintainer direction and remains covered by PR CI.

The two module wrappers themselves stopped at their worktree-local Prisma preflight. The exact emitted test files were run directly instead; tests that bind localhost were rerun outside the sandbox after the sandbox-only `listen EPERM` failure.

Uncovered local risk: Windows-specific file behavior and the complete portable suite are delegated to required PR CI.

## Review focus

- Whether 128 MiB leaves sufficient room for legitimate Provider catalogs while bounding corrupt Settings documents.
- Whether first-valid-record wins is the preferred recovery policy for duplicate historical Provider IDs.
- Whether the optional durable-reader limit preserves recovery semantics for existing consumers.

### DIFF
--- a/src/main/settings/document-codec.test.ts
+++ b/src/main/settings/document-codec.test.ts
@@ -39,6 +39,28 @@ describe('settings document codec', () => {
     expect(settings.providers.at(-1)?.id).toBe(`provider-${PROVIDER_RESOURCE_LIMITS.providers - 1}`)
   })
 
+  it('preserves a Codex provider that follows the non-Codex provider cap', () => {
+    const providers = [
+      ...Array.from({ length: PROVIDER_RESOURCE_LIMITS.providers }, (_, index) => ({
+        id: `provider-${index}`,
+        type: 'custom',
+        name: `Provider ${index}`
+      })),
+      {
+        id: 'builtin-codex-shared',
+        type: 'codex-shared',
+        name: 'Legacy Codex'
+      }
+    ]
+
+    const settings = sanitizeSettings({ providers })
+
+    expect(settings.providers).toHaveLength(PROVIDER_RESOURCE_LIMITS.providers)
+    expect(settings.providers).toContainEqual(
+      expect.objectContaining({ id: CODEX_SUBSCRIPTION_PROVIDER_ID, type: 'codex-isolated' })
+    )
+  })
+
   it('preserves current durable settings families and drops retired Runtime selections', () => {
     const dataRoot = resolve('portable-settings-data')
     const settings = sanitizeSettings({
--- a/src/main/settings/document-codec.test.ts
+++ b/src/main/settings/document-codec.test.ts
@@ -4,6 +4,7 @@ import { describe, expect, it } from 'vitest'
 
 import { CODEX_SUBSCRIPTION_PROVIDER_ID, SETTINGS_FILE_VERSION } from '../../shared/settings'
 import { sanitizeSettings } from './document-codec'
+import { PROVIDER_RESOURCE_LIMITS } from './provider-resource-limits'
 
 describe('settings document codec', () => {
   it('exposes one pure document boundary', async () => {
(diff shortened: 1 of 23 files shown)


ewen-poch:

- Provider IDs are now validated against the same 128-character limit before persistence. The regression test verifies that an oversized ID is rejected and no provider is stored.
- Oversized recognized recovery candidates are now treated as recovery barriers. Oversized primary files are still rejected, while oversized temporary candidates are preserved instead of being cleaned up. The regression test verifies both the rejection and preservation behavior.
- Codex providers are now selected and migrated before the 64-provider cap is finalized. When a valid Codex provider exists after 64 non-Codex entries, one slot is reserved for the canonical Codex account. A regression test covers this ordering.

Focused codec/storage tests, Node type checking, and lint pass.

ewen-poch:
The failed Ubuntu shard was traced to an unrelated timeout in the runtime-service manual-interpreter restart test. The exact test passed locally twice in 2.41s and 1.96s, so no unrelated notebook code or timeout threshold was changed. The failed jobs were rerun: Ubuntu shard 3, module coverage, and PR Gate all pass.

PR #2136, opened 2026-09-04
fix(provider-validation): scope results to model targets

ewen-poch:
## Summary

- persist the model and API endpoint exercised by successful and target-specific provider validation
- keep authentication/connectivity failures provider-wide while model and protocol failures only affect the matching target
- apply the scoped result consistently to preflight, model catalogs, scenario model admission, and the provider status UI
- preserve legacy provider-level records and invalidate target records when credentials or endpoint configuration changes

## Reproduction

The regression tests were added and run before the fix:

- validating model A and selecting unvalidated model B left preflight ready (expected false, received true)
- a model-not-found result for model B removed model A from the configured catalog (expected model-a, received an empty catalog)

Both cases pass after the implementation.

## Test plan

- npx vitest run src/main/settings/provider-accounts.test.ts src/main/settings/provider-auth-lifecycle.test.ts src/main/settings/xai-provider-account-owner.test.ts src/main/settings/provider-runtime-projection.test.ts src/main/settings/agent-runtime-manager.test.ts src/main/settings/preflight.test.ts src/main/settings/repository.test.ts src/main/settings/settings-backend.architecture.test.ts src/main/settings/provider-runtime-projection.architecture.test.ts src/main/settings/provider-auth-lifecycle.architecture.test.ts src/main/settings/session-details-model-owner.test.ts src/shared/configured-model-catalog.test.ts src/shared/settings.test.ts src/renderer/src/pages/settings/ProviderList.render.test.tsx
- npx vitest run src/main/settings/service.test.ts -t "SettingsService: (Subagent|Reviewer|Vision) model"
- npm run typecheck
- ESLint on changed files
- git diff --check

### DIFF
--- a/src/main/settings/agent-runtime-manager.ts
+++ b/src/main/settings/agent-runtime-manager.ts
@@ -17,7 +17,11 @@ import type {
   Preflight,
   ValidateProviderResult
 } from '../../shared/settings'
-import { isProviderUsableByFramework } from '../../shared/settings'
+import {
+  isProviderUsableByFramework,
+  preferredEndpoint,
+  requiresChatCompletionsBridge
+} from '../../shared/settings'
 import {
   buildUnsupportedCodexAcpVersionMessage,
   isSupportedCodexAcpVersion,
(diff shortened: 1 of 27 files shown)


ewen-poch:
Addressed the validation-scope feedback in 3aab92f0:

- Shared Claude status checks now classify a supported-but-unauthenticated CLI as an auth failure, so the failure remains provider-wide instead of being attached to one model target.
- The provider verified indicator now receives the active model/endpoint target derived from the selected framework and requires the complete target to match. Switching between Anthropic and OpenAI protocols no longer leaves a stale green check.
- Added regression coverage for both cases. The focused suite passes 279 tests, and typecheck plus ESLint pass.

ewen-poch:
Resolved the effective-model mismatch in b7119b5e. ProviderList now derives the active target with the same fallback order used by the runtime: explicit active model, provider model, then defaultVendorModel for official providers. The endpoint is still derived from the active framework, so the verified indicator requires the complete effective model/protocol target to match. Added coverage for an official provider with no explicit active model; the focused suite passes 280 tests, and typecheck plus ESLint pass.

PR #2143, opened 2026-09-04
fix(workspace): avoid replaying finalized artifact claims

ewen-poch:
## Problem

A generated Artifact can be published successfully and displayed in its Session, then later show `Artifact run claim not found` when the retained artifact runtime event is replayed after the finalized claim's five-minute in-memory TTL. Session projection is idempotent by event id, but the finalization side effect was repeated.

## Proposed change

Treat a replayed artifact event as complete only when its event id is already attached to the same prompt response and every referenced Artifact has a non-pending path. Events whose files remain under `.pending` still call finalization, preserving retry after a real publication failure.

## Scope and non-goals

- No database or persisted-format changes.
- No new fields, states, or enum values.
- No renderer contract or interaction changes beyond removing the false error.
- No persistent Artifact claim registry or TTL change.
- Historical Sessions without the existing artifact event ids retain their prior recovery behavior; no heuristic legacy matching was added.

- Replaying a fully finalized artifact event does not invoke finalization again or set a Session error.
- Replaying an event after a genuine failed finalization remains retryable.
- `workspace_runtime -> npm run test:module -- workspace_runtime -> 11 files, 433 tests passed`
- `renderer type safety -> npm run typecheck:web -> passed`
- `linted source -> npm run lint -> passed`

All listed checks ran after the last material edit. Full `npm test` was intentionally omitted per the requested module-focused workflow; exact-head PR CI remains authoritative.

## Review focus

Please verify that the guard distinguishes durable finalized paths from `.pending` paths and therefore does not suppress legitimate publication retries.

### DIFF
--- a/src/renderer/src/lib/acp/workspace-events.test.ts
+++ b/src/renderer/src/lib/acp/workspace-events.test.ts
@@ -3677,6 +3677,46 @@ describe('workspace runtime events', () => {
     expect(reconcilePendingArtifacts).toHaveBeenCalledOnce()
     expect(finalizeRunArtifacts).not.toHaveBeenCalled()
   })
+
+  it('records native artifact reconciliation failures for retry', async () => {
+    const nativePendingArtifact = createArtifactFile({
+      id: 'native-version-reconcile-failure',
+      artifactId: 'native-artifact-reconcile-failure',
+      versionId: 'native-version-reconcile-failure',
+      versionNumber: 1,
+      path: '/Users/example/.open-science/managed/native-version-reconcile-failure/result.txt'
+    })
+    useSessionStore.getState().attachRunArtifacts({
+      sessionId: 'transport-session-1',
+      runId: 'run-native-reconcile-failure',
+      eventId: 'native-reconcile-failure-event',
+      artifacts: [nativePendingArtifact]
+    })
+    await applyWorkspaceRuntimeEvent(
+      createEvent({ id: 'stop-before-native-reconcile-failure', kind: 'stop' })
+    )
+
+    await expect(
+      applyWorkspaceRuntimeEvent(
+        createEvent({
+          id: 'native-reconcile-failure-event',
+          kind: 'artifact',
+          runId: 'run-native-reconcile-failure',
+          artifactClaimId: 'native-reconcile-failure-claim',
+          artifacts: [nativePendingArtifact]
+        }),
+        {
+          reconcilePendingArtifacts: vi.fn().mockRejectedValue(new Error('reconcile failed')),
+          saveSession: vi.fn().mockResolvedValue(undefined)
+        }
+      )
+    ).rejects.toThrow('reconcile failed')
+
+    expect(useSessionStore.getState().sessions[0]).toMatchObject({
+      status: 'error',
+      error: expect.stringContaining('Generated file finalization failed')
+    })
+  })
 })
 
 describe('loop guard: suppressNextAutoReview', () => {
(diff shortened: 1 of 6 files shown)


ewen-poch:
Addressed the P1 reconciliation error-path issue in 0720b1d7. Native artifact reconciliation now records failures through the existing artifact-error workflow before rethrowing, preserving retryable vs. terminal proof-error behavior. Added a regression test that failed before the change because the Session remained idle with no artifact error; the workspace runtime module now passes 436/436 tests, with web typecheck and lint also passing.

ewen-poch:
Addressed the P1 sibling-error issue in 4afa0a8d. Both replay-success paths now clear a Session artifact error only when every artifact reference in the Session belongs to the resolved event. A successful claim therefore cannot hide an unrelated claim's retry surface, while a single-artifact historical false error is still cleared. This reuses the existing Session error model and adds no per-claim state or persistence. The workspace runtime module passes 436/436 tests; web typecheck and lint pass.

ewen-poch:
Confirmed the P1 trade-off. Correctly distinguishing an unrelated historical artifact from a still-failing sibling claim requires associating the Session artifact error with its event/claim, or an equivalent authoritative state. The current two-file boundary does not contain that ownership information; another heuristic would either hide a real retry or retain a recovered error. I am pausing rather than introducing new persisted/per-claim state without design approval. The current behavior is conservative: it may retain the retry surface, but it does not clear an unresolved sibling failure.

ewen-poch:
Addressed the P1 ownership gap in 1fec44d1. Artifact finalization errors now persist an optional artifactErrorEventId in Session JSON. Replay success clears the error when it owns that event, regardless of unrelated historical artifacts, and refuses to clear an error owned by another event. Legacy Sessions without the field keep the conservative fallback. No database migration or new enum is involved, and the existing Retry UI is unchanged. Regression coverage includes both ownership directions plus Session normalization; workspace-events passes 98/98, session-persistence passes 168/168, web typecheck and lint pass.

ewen-poch:
Addressed the P2 sibling-error case in 4ee91b38. The persisted owner is now an optional artifactErrorEventIds set rather than a single event. Recording a failure adds its event; replay success removes only that event and clears the Session error only after no owned failures remain. This preserves the retry surface for unresolved sibling claims without letting unrelated historical artifacts block recovery. Legacy Sessions still omit the field and use the conservative fallback. No database migration, enum, or UI change is involved. Workspace runtime passes 437/437 tests, Session persistence passes 168/168, and web typecheck/lint pass.

ewen-poch:
I do not think this finding applies. sessionReferencesOnly only gates clearing an unowned legacy Session error; it does not gate finalized-event replay detection. In both the native and compatibility branches, once the current event is proven finalized, the function returns true unconditionally after the optional error-clear block. The compatibility proof itself is scoped to appliedMessage.artifactIds plus the current event's name/size/mtime, and the native proof is scoped to the current event's Version ids. A prior artifact elsewhere in the Session therefore cannot cause finalizeRunArtifacts to run again. No code change is needed for this point.

ewen-poch:
Addressed the P1 incomplete-reconciliation path in 3fef3745. A native replay now treats any result missing a requested Version id as an explicit unresolved-finalization error inside the existing event-owned error path, records it for retry, and throws instead of falling through to claim finalization. The regression test failed before the fix with the stale claim-not-found error and now receives the deterministic reconciliation error. Workspace runtime passes 438/438 tests; web typecheck and lint pass.

ewen-poch:
Addressed both findings in 8b67baf2. Native replay now awaits persistence of the latest attached Session before invoking reconciliation, so Main can validate message ownership from durable state. Generic artifact retry failures without an event id now preserve any existing artifactErrorEventIds instead of downgrading owned failures to legacy/unowned state. Added red-green coverage for the save-before-reconcile ordering and owner preservation through a generic retry failure. Workspace runtime passes 438/438 tests; web typecheck and lint pass.

Issue #924, opened 2026-08-07
[Feature]: Expose the existing @EXPLICIT environment lock as user-facing export and restore


carlkaub:
### Preflight checklist

- [x] I searched existing issues and discussions and this hasn't been proposed yet.

### What problem does this solve?

The Capability Map marks Environment Management 🟡 with *"portable lock export/restore is not built yet"*, and Phase 2 names portable environment restoration as a remaining target.

**Most of it is already built and simply not exposed.** In `src/main/notebook/micromamba.ts`:

- `normalizeExplicitLock` converts `micromamba list --explicit --md5` output into a valid `@EXPLICIT` lock — that is, it already produces a portable lock from a live environment.
- `createFromLockArgv` creates a new environment from a lock, offline.
- `installFromLockArgv` deterministically applies a lock to an existing environment.

These are used today only by `provisioner.ts` (bundled offline provisioning) and `runtime-relocation.ts`. No user-facing path reaches them.

So the gap between "we cannot share or restore an environment" and "we can" is wiring, not construction.

### Proposed solution

1. **Export.** An environment command that runs the existing `normalizeExplicitLock` path against a selected managed environment and writes a lock file the user can share or commit.
2. **Import/restore.** Create or update an environment from a lock via the existing `createFromLockArgv` / `installFromLockArgv` paths, reusing the current provisioning UI, journaling, and crash recovery.
3. **Attach the lock to provenance.** Record the lock alongside the observed inventory the provenance repository already captures. This is the step that converts "here is what was installed" into "here is how to get it back", and it is what makes the reproducibility claim real.

Scope note: this covers managed conda/micromamba environments. Bring-your-own interpreters and external R runtimes stay out of scope — package management there is already documented as manual.

### Alternatives considered

- **conda-lock or pixi as a dependency.** Better multi-platform solving, but adds a toolchain and duplicates machinery already present and already trusted for offline provisioning.
- **`pip freeze` / `environment.yml`.** Not reproducible — unpinned transitive dependencies and no checksums. `@EXPLICIT` locks pin URL + MD5, which is the property that matters.

### Additional context

Best value-per-effort item in #917, and a dependency for RO-Crate export and parameterized sweeps.

Suggested Capability Map correction once shipped: Environment Management's *"portable lock export/restore is not built yet"* is inaccurate today — the primitives exist; only the user-facing surface is missing.

---

[...]

PR #2151, opened 2026-09-04
feat(notebook): extract managed environment lock export primitive

LiuZhaoyan:
## Summary

Backend for #924 — managed-environment `@EXPLICIT` lock export and restore. Two commits, no IPC / renderer / provenance changes.

### Commit 1 — lock export primitive

- `environment-lock.ts`: `exportEnvironmentLock(env, deps)` (capture `list --explicit --md5` → `normalizeExplicitLock()` → require ≥1 package URL) and `managedEnvironmentRef()` (resolve a `DiscoveredInterpreter` to a trusted `{name, prefix}`; `user-own` / no-condaEnv → `undefined`, never a guessed default).
- `runtime-relocation.ts` now calls the shared primitive (one implementation, not two).

### Commit 2 — lock restore backend

- `parseExplicitLock(lock, {platform, arch})` — validates external input before any mutation: `@EXPLICIT` marker, md5-pinned URLs (sha256 locks rejected with a clear error), malformed entries, foreign-platform locks, duplicates. Unrecognized body lines are rejected, never silently dropped (a dropped package would silently restore a *different* env).
- `DefaultRuntimeProvisioner.restoreEnvironmentFromLock(name, lockText, {signal, onProgress})` — create and replace are one sequence:
  1. download every pinned URL into journaled `.incoming-*` staging (production: the shared resilient downloader — system/VPN proxy, Range resume, stall timeout),
  2. md5-verify each download,
  3. seed the shared pkgs cache via the existing `validateAndSeedPackIntoCache`,
  4. clear + rebuild the prefix **offline** via `createFromLockArgv` through the same `withJournaledPrefixWrite('materialize', phase 'restore')` machinery the relocation restore uses.

Properties:
- **Fully portable**: downloads verify before the prefix is cleared, and an empty package cache is fine — the downloads alone rebuild the env.
- **Crash-consistent with provisioning**: same journal kind/phase (startup recovery's `verifyOrRebuildEnv` arm), child-PID supervision, fail-closed `begin()`, download-journal staging cleanup, Windows MAX_PATH recovery, archive publication evidence.
- **Serialized** with all other mutations via `serializeProvisioner` + the per-env prefix lock + the shared pkgs-cache lock.
- **Cancellable** via `AbortSignal` (aborts download and create; a partial prefix is cleaned by the next restore or startup recovery).
- **Never looks ready on failure**: the ready marker is stamped only after the rebuilt interpreter verifies.
- Unsafe/unknown target names rejected (`assertSafeEnvName` + reserved defaults); user-own interpreters and external R out of scope; live kernels must be drained through the explicit-repair path by an in-session caller (the only sanctioned teardown).
- No renderer-supplied prefix; main process only.

## Test plan

Layer 1 (unit, Electron-free, no network):
- `environment-lock.test.ts` (20): parser accepts/throws per rule; downloader verifies md5, rejects corrupted files, propagates abort.
- `provisioner.test.ts` restore suite (11): exact create argv + staged lock path; broken-prefix replace with marker-only-after-verify; invalid lock rejected before download/mutation; unsafe/reserved names; unwired downloader; download-md5 failure leaves the existing env untouched; create failure leaves no marker and cleans staging; cancellation; download journal + materialize journal records observed mid-run and completed after; serialization behind an in-flight provision.
- All prior suites stay green (relocation, migration, ipc, lifecycle, foundation): 213 tests across 7 files.
- `tsc --noEmit -p tsconfig.node.json`, eslint, prettier clean.

Layer 2 (integration, env-gated behind `OPEN_SCIENCE_TEST_MICROMAMBA`, real micromamba 2.8.1 + network — never in `npm test`):
- `environment-lock.restore.integration.test.ts`: create minimal env (python=3.12 + six) → export lock → restore into a **pristine root** (no prefix, no cache) → re-export is **byte-identical to the input lock** → `import six` runs from the restored prefix. This is the portable-restore acceptance criterion.

Part of #924

### DIFF
--- a/src/main/notebook/env-ipc.test.ts
+++ b/src/main/notebook/env-ipc.test.ts
@@ -39,6 +39,7 @@ const fakeProvisioner = (over: Partial<RuntimeProvisioner> = {}): RuntimeProvisi
   upgradeIfNeeded: vi.fn().mockResolvedValue(undefined),
   repair: vi.fn().mockResolvedValue(undefined),
   restoreRelocatedEnvs: vi.fn().mockResolvedValue(undefined),
+  restoreEnvironmentFromLock: vi.fn().mockResolvedValue(undefined),
   cancel: vi.fn(),
   ...over
 })
--- a/src/main/notebook/environment-lifecycle-workflows.test.ts
+++ b/src/main/notebook/environment-lifecycle-workflows.test.ts
@@ -43,6 +43,7 @@ const fakeProvisioner = (over: Partial<RuntimeProvisioner> = {}): RuntimeProvisi
     await options?.onVerified?.()
   }),
   restoreRelocatedEnvs: vi.fn().mockResolvedValue(undefined),
+  restoreEnvironmentFromLock: vi.fn().mockResolvedValue(undefined),
   cancel: vi.fn(),
   ...over
 })
--- a/src/main/notebook/environment-lock.restore.integration.test.ts
+++ b/src/main/notebook/environment-lock.restore.integration.test.ts
@@ -1,0 +1,134 @@
+// Local integration test for issue #924's portable restore acceptance: an exported @EXPLICIT lock
+// must rebuild an equivalent managed environment WITHOUT the original prefix and WITHOUT a
+// pre-populated package cache (downloads re-fetch every tarball).
+//
+// Run with a real micromamba on PATH (or OPEN_SCIENCE_TEST_MICROMAMBA):
+//   OPEN_SCIENCE_TEST_MICROMAMBA=$(pwd)/.tmp-micromamba/micromamba \
+//     npx vitest run src/main/notebook/environment-lock.restore.integration.test.ts
+//
+// It is env-gated like the kernel suites (RUN_KERNEL/OPEN_SCIENCE_TEST_PY_ENV) so `npm test`
+// never touches the network. Setup, mirroring .github/workflows/runtime-certification.yml:
+//   node scripts/fetch-micromamba.mjs linux-64 .tmp-micromamba/micromamba
+//   .tmp-micromamba/micromamba create -y -p .tmp-int-env -c conda-forge python=3.12
+import { spawnSync } from 'node:child_process'
+import { existsSync, mkdtempSync, renameSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
+import { describe, expect, it } from 'vitest'
+
+import { exportEnvironmentLock } from './environment-lock'
+import { DefaultRuntimeProvisioner, type ProvisionerDeps } from './provisioner'
+import { DEFAULT_PY_ENV, envPrefix, pkgsCache, pythonBin, runtimeRoot } from './runtime-paths'
+import { captureMicromamba, verifyExecutable } from './provisioner-runtime'
+
+const mm = process.env.OPEN_SCIENCE_TEST_MICROMAMBA
+const gate = mm && existsSync(mm) ? describe : describe.skip
+
+const runMm = (args: string[], opts: { cwd?: string } = {}): string => {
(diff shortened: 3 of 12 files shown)


ewen-poch:
Thanks @LiuZhaoyan for contributing this work on #924!

We have been working in parallel on a broader Artifact Reproducibility integration. The environment-lock implementation is already in our local integration branch, and we plan to open the PR shortly. It connects captured locks to Notebook run provenance and Artifact versions, with Python/R lock coverage checks, isolated replay, lock bundle export/import, and reusable environments. This work is not on main yet.

There is useful overlap here, particularly the shared export primitive, downloading and verifying all pinned archives before replacing a prefix, and the real micromamba test that restores into a pristine root with an empty cache and checks the re-exported lock. That cold-cache acceptance test is especially valuable as we finish validating portability; our broader integration does not make those contributions redundant.

Thank you for the implementation and the careful restore coverage. This gives us concrete pieces to consider consolidating as we prepare the upcoming PR.

LiuZhaoyan:
@ewen-poch  Thanks for the context, and I’m glad to hear that the cold-cache restore test and the shared export/restore pieces are still useful.

I’d be very happy to see these parts consolidated into the upcoming Artifact Reproducibility work. If it would help, please feel free to tag me if you’d like any of the commits split out, adapted, or rebased against the integration PR once it is available.

Thanks again for taking a look at the implementation.

LiuZhaoyan:
@ewen-poch  Hi, just following up on the broader Artifact Reproducibility integration you mentioned earlier. I haven’t seen the follow-up PR appear yet, so I wanted to check where that work currently stands.

I’d also be interested to know whether the reusable pieces from this PR — particularly the shared export primitive, verified cold-cache restore flow, and micromamba acceptance test — are still being considered for consolidation into that integration.

If there’s anything you’d like me to adjust or split out on my side, I’d be happy to help.

Thanks!

ewen-poch:
Thanks @LiuZhaoyan for following up, and apologies that this has taken longer than our initial update suggested.

The broader Artifact Reproducibility implementation is now in our local integration branch. It connects environment locks to Notebook runs and Artifact versions, supports isolated Python/R replay, and presents output comparisons and verification history. We have also been strengthening cross-cell dependency and input/output lineage analysis, including random-state restoration and plotting configuration dependencies, to address issues found in real scientific workflows.

We have incorporated your pristine-root / empty-cache acceptance-test approach into our reproduction runtime tests, with an explicit reference to #2151 in the test source. We adapted it for both Python and R, covering lock round trips, reproduced outputs, corrupted archive rejection, and subsequent cache reuse without the archive server. Your export/restore work and concrete acceptance criteria have been valuable references for this integration—thank you for putting that coverage together.

We’re aiming to open the integration PR within the next day or two, after the final cleanup and review. Once it is available, we can review the overlap together and identify any remaining pieces worth splitting out or rebasing. There is no need to rework your branch in the meantime.

Thanks again for the contribution, your patience, and your offer to help with the integration.

ewen-poch:
Thanks @LiuZhaoyan for following up, and for your patience and willingness to help!

The broader Artifact Reproducibility integration has now landed on `main` in #2438. It covers the environment-lock capture and restore workflow needed for reproducibility, connecting captured locks to Notebook runs and Artifact versions and restoring isolated environments for replay.

Your pristine-root / empty-cache acceptance-test approach is included in `src/main/notebook/reproduction-runtime.integration.test.ts`, with an explicit credit to #2151. The Python and R coverage checks exact lock round trips, reproduced outputs, corrupted archive rejection, and cache reuse after the archive server is stopped.

The implementation uses isolated replay environments rather than the in-place managed-prefix replacement API proposed here, so this is not a verbatim integration of every backend change. The overlapping reproducibility workflow is now covered, however, and we can close this PR as superseded by the integrated approach. No rebase or split-out is needed at this point.

Thank you again for the thoughtful implementation and especially the concrete portability acceptance criteria—they were valuable to the integration.

PR #2161, opened 2026-09-04
fix(provenance): make message snapshots durable

ewen-poch:
## Problem

Message snapshot publication renamed its staging file and marked the database row ready without flushing either the file or the published directory. A power-loss window could therefore leave ready metadata pointing at bytes that never became durable. Startup reconciliation only handled staging rows and could not repair a missing or corrupt ready sidecar.

## Proposed change

- Flush each snapshot staging file, rename it, flush both directory trees through the storage root, revalidate the published bytes, and only then mark the row ready.
- Keep a valid staging row for retry when its durability barrier or database promotion fails, and recover directly from the pre-rename staging file when the final path does not yet exist.
- During startup, validate ready snapshots for Sessions whose persisted graph is still available and reconstruct missing or corrupt sidecars only when the rebuilt bytes match the existing checksum.
- Run ready-snapshot repair after Artifact reconciliation restores authoritative Session attachments.
- Batch ready-snapshot lookup by 400 distinct active Sessions to stay below SQLite expression and parameter limits.

- No database schema, field, migration, lifecycle-state, renderer, or interaction changes.
- Historical ready rows with an empty checksum remain readable and receive a checksum when their existing bytes validate structurally.
- Missing or corrupt checksumless snapshots remain unavailable because mutable Session state cannot prove the original history.
- Historical snapshots whose Session JSON has already been deleted remain unreconstructable and fail closed.

## Regression evidence

The tests were first run against the unfixed behavior and failed with the reported symptoms:

- Publication transitioned from staging to ready without file or directory sync.
- Startup left a removed ready sidecar missing with ENOENT.
- A valid staging-only sidecar was deleted instead of recovered.
- Recovery before Artifact reconciliation could not reconstruct attachment-backed snapshots.
- An unbounded 20,001-Session query exceeded SQLite's expression-tree limit.
- Cross-directory rename omitted the staging-directory durability barrier.
- Replacing staging bytes after initial validation could promote incorrect bytes to ready.
- Recursively created ancestor directory entries were not synced.

Final validation after rebasing onto current main:

- `npm run test:module -- artifact_provenance`: 27 files, 1665 tests passed.
- `npm run lint`: passed.

The full local `npm run test` suite was intentionally not run; PR CI remains authoritative for the complete portable and platform-specific plans.

## Review focus

- Durability ordering before the ready transaction.
- Fail-closed checksum preservation during startup reconstruction.
- Recovery ordering after Artifact attachment repair.

### DIFF
--- a/scripts/ci/module-impact.json
+++ b/scripts/ci/module-impact.json
@@ -788,6 +788,7 @@
         "owner": [
           "src/main/artifacts/provenance-dependency-read.test.ts",
           "src/main/artifacts/provenance-lifecycle-contract.test.ts",
+          "src/main/artifacts/provenance-message-snapshot-durability.test.ts",
           "src/main/artifacts/provenance-message-snapshot.test.ts",
           "src/main/artifacts/provenance-repository.architecture.test.ts",
           "src/main/artifacts/provenance-repository.test.ts",
--- a/src/main/artifacts/provenance-message-snapshot-durability.test.ts
+++ b/src/main/artifacts/provenance-message-snapshot-durability.test.ts
@@ -116,6 +116,7 @@ describe('Provenance Message snapshot durability', () => {
       'database:staging',
       `sync:${join(storageRoot, 'artifacts/project-1/session-1/.provenance/.staging/messages/snapshot-1.json')}`,
       `sync:${join(storageRoot, 'artifacts/project-1/session-1/.provenance/message-snapshots')}`,
+      `sync:${join(storageRoot, 'artifacts/project-1/session-1/.provenance/.staging/messages')}`,
       'database:ready'
     ])
   })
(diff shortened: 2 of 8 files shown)


ewen-poch:
Addressed the P2 finding about Provenance cleanup being skipped when Artifact recovery fails.

- Split Provenance startup reconciliation into independent Session/Review cleanup and Message snapshot repair phases.
- Run deletion tombstone and orphan Review cleanup before fallible Artifact recovery.
- Keep Message snapshot repair after Artifact attachment recovery, preserving the integrity ordering introduced by this PR.
- Added a coordinator contract regression test that failed before the fix because cleanup was never called after an Artifact recovery failure.

Validation: artifact_provenance module passed (27 files, 1665 tests), coordinator contract passed (12 tests), Node/sandbox typecheck passed, and lint passed.

Fixed in 4331847b.

PR #2165, opened 2026-09-04
fix(deletion): preserve compensation failures

ewen-poch:
## Summary

- preserve the original Project deletion setup failure when fence rollback also fails
- collect the original Session deletion failure and every failed compensation in `AggregateError`
- add public-boundary regression coverage for both double-failure paths

## Verification

- `npx vitest run src/main/projects/deletion-coordinator.test.ts src/main/session-persistence/coordinator.test.ts --reporter=dot` (204 passed)
- `npx eslint src/main/projects/deletion-coordinator.ts src/main/projects/deletion-coordinator.test.ts src/main/session-persistence/deletion-owner.ts src/main/session-persistence/coordinator.test.ts`
- `npm run typecheck:node`

### DIFF
--- a/src/main/projects/deletion-coordinator.test.ts
+++ b/src/main/projects/deletion-coordinator.test.ts
@@ -190,6 +190,32 @@ describe('ProjectDeletionCoordinator', () => {
     await expect(deletion).rejects.toThrow('intent unavailable')
   })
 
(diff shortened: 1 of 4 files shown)


ewen-poch:
Updated the PR title to the required scoped Conventional Commit format: `fix(deletion): preserve compensation failures`.

PR #2167, opened 2026-09-04
perf(session-persistence): bound Session resource usage

ewen-poch:
## Problem

Session snapshots repeatedly serialize and durably rewrite the complete conversation graph. Message content and streamed event IDs were unbounded, and the Session repository did not use the durable JSON reader's byte limit. Long-running conversations therefore accumulated increasingly expensive full-file writes, while a single oversized Session could continue consuming memory and disk without an actionable failure.

## Proposed change

- enforce a 256 MiB serialized Session limit on reads and before durable publication
- read limited JSON through one opened descriptor, verify it is a regular file, and consume at most `maxBytes + 1` bytes so a stat/read replacement race cannot bypass the limit
- preserve oversized files and recognized crash-recovery temporary files in place, and surface a typed, actionable storage-limit outcome instead of quarantining them as corruption
- propagate the `session-size-limit` code through legacy IPC, every application-command Session mutation, and ACP state/response boundaries
- validate projected Session files and recognized recovery temporary files before treating a ready SQLite catalog as authoritative
- retain existing SQLite reconciliation markers for oversized authority until the Session becomes readable or is removed, preventing a stale projection row from reappearing after recovery
- classify an oversized crash-recovery temporary file as `too-large` even when no primary Session file exists
- reject an oversized save before reserving SQLite projection metadata; new Sessions are admitted against a `Number.MAX_SAFE_INTEGER` number field so projection assignment cannot cross the boundary
- reject oversized existing authority before projection reservation and version-gated migration backup reads or copies
- cancel and gate only the affected active Session after the limit is reached, with a New conversation action
- when New conversation abandons an oversized Session, clear its completed failure from the ordered persistence queue so later flush/quit can complete while the abandoned Session remains write-blocked
- run cancellation from the renderer-lifetime runtime bridge so blocked runs are stopped on Home as well as Workspace
- retry one transient cancellation failure and release the deduplication marker if both attempts fail
- keep queued user/application work deferred for blocked Sessions, including through the always-mounted runtime bridge when Workspace is closed
- route Plan, permission, and elicitation response size-limit failures into the shared blocked/recovery flow
- route stable-message rearm and Session Memory preference persistence failures into that same per-Session recovery flow
- route the desktop ACP response contracts through the runtime-validated command adapter so Electron preserves the typed error envelope
- route Session archive updates through their existing runtime-validated application command instead of the legacy raw Electron handler
- route PDF context, branch, details, archive, and delegation-policy size-limit failures into the same per-Session recovery owner
- gate active and restored Plan previews by the target Session, including background Sessions; once blocked, their response surfaces become read-only
- keep unrelated write failures retryable when another Session is size-blocked
- compact terminal message and tool-activity event IDs to the latest 500 entries in both the graph and flat compatibility projection
- apply the same compaction when creating Session files, so terminal Sessions produced outside the renderer and historical terminal data are covered on their next write

## Scope and non-goals

- No persisted field, database schema, file-format version, or persisted Session status is added.
- Oversized projection recovery reuses the existing pending-reconciliation table; it adds no column, enum, or migration, but deliberately keeps that operational marker durable until authority is readable or removed.
- New discriminants are transient only: application command code `session-size-limit`, load warning kind `too-large`, and renderer recovery kind `oversized-authority`.
- Existing Session files at or below 256 MiB remain compatible. Files above the limit are preserved unchanged but cannot be opened by this version; users can move them out of the Session directory or use an older version to export them.
- Existing oversized `eventIds` arrays remain readable and are compacted when a terminal Session is next written.
- Normal ready-projection startup adds one metadata lookup per projected Session and one directory listing per projected Project; a full authority scan occurs only when a projected primary or recognized recovery temporary file exceeds the limit, or the projection already needs startup recovery.
- This does not introduce append-only storage, remove graph/flat duplication, or add separate message-count and per-message limits. The total serialized limit bounds those resources without expanding the persisted model.
- The write path still needs one complete serialization to measure exact UTF-8 bytes; it rejects before projection reservation, migration backups, temp-file creation, and durable I/O, but cannot prevent that first allocation.

- original oversized-read and terminal-event-ID reproductions failed consistently before the fix and pass after it
- earlier AI-review regressions reproduce ready-projection diagnostics, pre-backup admission, unrelated retry isolation, pre-projection reservation, and cancellation retry; each failed before its fix and passes afterward
- later review regressions cover number-assignment overflow, oversized existing authority, queued dispatch, restored Plan response, active Plan preview, Home-page cancellation, and descriptor-bounded reads; each was observed failing on the corresponding pre-fix path and passes afterward
- latest review regressions cover an oversized recognized recovery temporary file, ACP Plan/permission/elicitation responses missing the shared size-limit recovery flow, background Plan previews bypassing the target Session gate, the desktop ACP adapter losing the typed error envelope, direct Session mutations degrading to generic failures, a stale projection after moving oversized authority out, cold-start temp-only recovery losing the `too-large` diagnostic, and New conversation leaving a failed write that blocks flush/quit; each failed for the reported reason before the fix and passes afterward
- latest recovery-action run: 1 file / 24 tests passed; the new flush assertion failed with `session-size-limit` before the fix and passes afterward
- after merging current `main`, the public application-command composition boundary failed 10/10 because the three newly routed ACP response commands had no runtime codecs; adding their argument/result contracts and Electron inventory expectation made the 5-file composition/wiring/catalog set pass 60/60
- the exact PR merge ref reproduced the later Ubuntu shard failure in `application-command-electron-adapter.test.ts` (2/3 failed with `Electron Application Command adapter inventory mismatch`); aligning the explicit fixture with the three ACP response channels and Session archive channel made the related 5-file boundary set pass 101/101
- after `main` added Settings-install shutdown admission, the PR merge ref and merged branch both reproduced the same node typecheck failure in the delegation production-composition fixture; supplying the two new no-active-install dependencies made node+sandbox typecheck pass and the fixture's module file pass 62/62 outside the localhost-restricted sandbox
- the same Electron foundation E2E that closed during application composition before the codec fix now creates a project and reloads it after relaunch (1/1 passed)
- the latest three AI-review paths reproduced independently before their fixes: stable-message rearm and Memory preference persistence each omitted the size-limit callback, while archive remained on raw IPC and outside the validated Electron inventory; the focused 5-file run then passed 279/279, and the expanded preload/router/renderer run passed 443/443
- current recovery/error-boundary run: 3 files / 172 tests passed, including all 10 Session authority mutation commands through the public application-command boundary
- after merging current `main`, the 6-file persistence/Plan-preview impact set passed 184/184 across targeted runs; the SQLite projection file was rerun serially (35/35) after parallel migration contention caused three unrelated timeouts/transaction closures
- latest directly affected run: 20 files / 668 tests passed
- earlier broader affected-module run: 15 files / 1,570 tests; 1,569 passed in the serial aggregate, and the one existing Prisma-backed test that hit its 15-second timeout passed alone in 1.4 seconds while the host load average was 50–61
- Node, sandbox, and web TypeScript checks passed after the `main` merge; Node and sandbox checks passed again after the ACP codec integration fix
- changed source/test files -> ESLint -> passed with no warnings
- production Electron build -> passed after both command-routing changes; existing Vite chunk/CSS warnings only
- Electron foundation create/relaunch E2E -> passed again after moving archive ownership (1/1)
- changed files -> Prettier check -> passed
- patch whitespace -> `git diff --check` -> passed

Per maintainer request, the full local `npm test` suite was not run. Exact-head PR Gate and platform lanes are the authority for broader coverage.

## Review focus

- whether 256 MiB is the right product boundary for a single Session
- whether preserving oversized files while returning a recovery barrier is the desired historical-data behavior
- whether retaining the latest 500 terminal event IDs preserves all required diagnostics and UI behavior
- whether cancellation and per-Session admission gating cover every active-run, queued-message, application-message, and Plan-response path without blocking unrelated Sessions

Uncovered locally: platform-specific filesystem failure modes and complete-suite interactions. Durable write ordering and fsync behavior are intentionally unchanged.


ewen-poch:
Independent worktree review found and fixed two remaining size-limit boundary gaps in cd989ab1: revision-checked saves now perform typed existing-authority size admission before reading authority, and migration backup paths now reuse the descriptor-bounded reader instead of the unbounded reader. Both regressions were reproduced before the fix: one returned a generic revision-comparison error for oversized authority, and the other invoked the unbounded backup reader and returned a generic error. The focused tests are green after the fix; repository/storage validation passed 109/109, and the broader post-main integration set passed 1,282/1,282. This does not change the persistence schema, architecture, stored state, or user interaction.

PR #2168, opened 2026-09-04
fix(settings): own shutdown operations

ewen-poch:
## Summary

- register `SettingsService` as a named application module with rollback/dispose ownership
- abort active Codex, Claude isolated/shared, and xAI login controllers during shutdown
- expose in-flight Agent Runtime installation state to the existing quit confirmation flow, including pre-confirmed and post-confirmation race checks

## Impact

- no persistent fields, migrations, data-model changes, new UI copy, or new provider lifecycle abstraction
- reuses the existing close-confirm interaction; only the transient IPC activity flag is generalized from Reviewer-only to unlisted background work
- non-interruptible installs require explicit quit confirmation; authentication operations are cancelled during module disposal

## Verification

- regression tests were first run against unmodified `main` and failed because `dispose` / `hasActiveInstall` were absent and Settings work bypassed quit confirmation
- `npm run typecheck`
- targeted lifecycle/settings suites: 178 tests passed
- `npm test`: 23,918 passed; one unrelated Notebook shell timeout passed when rerun alone; two existing R TIFF capture tests still fail with missing captured images


ewen-poch:
Addressed the P1 Settings/delegated confirmation loop in `ee1a6238`.

- Both delegated confirmation paths now snapshot active Settings installation work immediately before showing the confirmation and pass that authorization into `requestConfirmedQuit`.
- Existing race tests now cover delegated work appearing during confirmation and the pre-confirmed Windows final-boundary path while a Settings install is active.

Validation: `src/main/app-lifecycle.test.ts` (74 tests), Node typecheck, ESLint, and Prettier all pass. This does not change architecture, persistence, data models, or user interaction.

ewen-poch:
Addressed the P2 in-place update gap in `b2cabec9`.

- Active Agent Runtime installation is now represented as `settings-install` and participates in the update gate's initial and final blocker checks.
- The in-place updater refuses `quitAndInstall` while that blocker is present and keeps the update retryable.
- The dialog shows a specific “wait for the installation to finish” message, localized across all supported renderer locales.

Regression coverage verifies the IPC wiring, installer refusal, blocker payload, and localized renderer behavior. Typecheck, ESLint, Prettier, and 1,052 targeted tests pass. The full suite has the same unrelated baseline failures: one flaky Notebook shell timeout that passes alone and two R TIFF capture failures.

ewen-poch:
Addressed both TOCTOU findings in b8133dce.

- The in-place update handoff now acquires Settings install admission before the first blocker snapshot, holds it through quitAndInstall, and releases it through the existing aborted-handoff path. A regression test verifies that a managed installer cannot enter while admission is held.
- Quit confirmation now carries the active install ID instead of a boolean. The final before-quit boundary requires the same ID and prompts again when the active install changes.

Validation: the two regression tests failed on the previous code for the reported reasons and now pass; 1,084 related tests pass; SettingsService tests pass 275/275; typecheck, ESLint, and Prettier pass. The required full suite retains only the same unrelated baseline failures: one flaky notebook concurrency timeout (passed alone) and two R TIFF capture tests.

ewen-poch:
Addressed the data-root handoff race in aa8ddf38. Data-root preparation now acquires the existing Settings install-admission hold before the blocker/durability gate, keeps it across copy and the staged commit decision, releases it through the existing abort notification on refusal/failure/discard, and leaves it held after a successful pointer commit until relaunch.

The regression wiring assertion failed on the previous code because data-root handoff had no admission hold and now passes. Validation: 1,204 related tests pass; typecheck, ESLint, and Prettier pass. The required full suite remains at 23,935 passed with only the same unrelated baseline failures (one flaky notebook concurrency timeout and two R TIFF capture tests).

ewen-poch:
Addressed the overlapping admission-lease defect in 0bf3c735. The shared Settings install admission is now reference-counted; each idempotent lease release removes only its own hold, so installs remain blocked until every update/data-root handoff releases.

The new overlapping-handoff regression test failed on the previous boolean implementation because releasing one lease admitted the installer while the second lease was still active, and now passes. Validation: 1,205 related tests pass; typecheck, ESLint, and Prettier pass. In the full suite, two unrelated async-highlighting failures passed alone; the two existing R TIFF capture failures remain.

ewen-poch:
Addressed the shutdown finding in b3325713. SettingsService.dispose() now aborts an active runtime install and waits for its cleanup to finish. The abort signal is propagated through registry requests, downloads, extraction, and version checks; shell/npm installs and the Codex ACP smoke check terminate and drain their process trees before settling. Managed installers retain their existing finally-based staging cleanup. Regression coverage now verifies the public SettingsService dispose boundary, process-tree drain behavior, and removal of an aborted download's staging directory. No persisted fields, data-model changes, or UI changes were introduced.\n\nThe focused suite passes (199 tests). The full suite completed with 23,939 passed and 3 unrelated failures: the two existing R/TIFF capture failures, plus one Notebook shell-concurrency timeout that passed immediately when rerun in isolation.

ewen-poch:
Addressed the three cancellation gaps in 9210d521:

- OpenCode staged-binary verification now uses an asynchronous, signal-aware process and checks cancellation before and after the probe.
- Codex adapter and CLI version checks are now asynchronous and receive the active installation AbortSignal.
- Claude, OpenCode, and Codex discovery now propagate the same signal through npm-directory lookup, version probes, native Codex discovery, and the ACP initialize smoke check. Successful unmanaged installs pass their active install signal into post-install discovery; an aborted Codex smoke check reaps its process tree before returning.

Regression coverage includes signal propagation at the installer, detector, and AgentRuntimeManager boundaries plus real slow-process abort checks. The focused suite passes 196 tests; typecheck and lint pass. The full suite completed with 23,946 passing and 226 skipped; its only two failures are the pre-existing R/TIFF capture cases in kernel-executor.test.ts.

ewen-poch:
Addressed both findings in 8ac2bce4.

- AgentRuntimeManager now owns a shutdown AbortController and tracks each public Claude, OpenCode, CodeBuddy, and Codex detection operation through its repository update. dispose() aborts standalone detection and active installation together, then drains both before returning. New detection and installation work is rejected after shutdown begins.
- The default Codex JavaScript adapter version probe now passes the effective AbortSignal to execFile, passes it through the non-JavaScript fallback, and preserves cancellation instead of converting it to a missing-version result.
- CodeBuddy discovery now propagates cancellation through candidate-directory and version probes, matching the other runtime detectors.

The regressions first reproduced the missing standalone abort/drain, post-dispose admission, one-second Codex probe delay, and dropped CodeBuddy signal. The related suite passes 203 tests; typecheck and lint pass. The full suite completed with 23,949 passing and 226 skipped. Its only failures were the two existing R/TIFF capture cases and one known Notebook shell-concurrency timeout that passed immediately in isolation. No persistence, data-model, or UI changes were introduced.

ewen-poch:
Addressed the data-root migration blocker gap in c4bfb6fb. Confirmed migration still permits interruptible agent/notebook work to be torn down, but now preserves settings-install alongside delegated/reviewer as non-interruptible blockers, so data-root teardown/copy cannot start during an active runtime installation. The regression case failed on the previous behavior by returning completed=true and invoking teardown; it now passes. Verification: 22 focused tests passed, typecheck passed, and targeted ESLint passed. The full suite completed with 23,950 passing / 226 skipped and only the two known local R/TIFF failures plus the existing Notebook concurrency timeout; that timeout passed immediately in isolation. No data fields, persistence, model, architecture, strings, or UI elements were added; the interaction now requires the active installation to finish before migration can proceed.

ewen-poch:
The latest finding is already covered by the current head (c4bfb6fb), so no additional code change is warranted. prepareDataRootHandoff routes through createDataRootResearchSafeInstallGate, whose confirmedInterruption branch explicitly retains settings-install at src/main/update/strategy.ts:65-69. createActiveResearchSafeInstallGate evaluates that filtered detector before teardown and again at the final boundary. The added parameterized regression test proves that confirmed migration returns blockedBy: ['settings-install'] and never invokes teardown. holdInstallAdmission separately closes only the new-install admission race; it is not being relied on to handle an existing install. This review comment repeats the pre-fix behavior rather than the code at its recorded head.

ewen-poch:
Merged current main in 91bbf6f2 and resolved the three AgentRuntimeManager overlaps by keeping main's configRoot rename together with the installation AbortSignal. The existing shutdown ownership and confirmed-migration settings-install blocker remain intact. No new behavior or state was introduced by the conflict resolution.

Typecheck and targeted ESLint pass. Runtime-manager, migration-gate, architecture, and translation tests pass. SettingsService tests encountered sandbox local-listener restrictions; the full suite is now running outside that restriction. I will report its result separately.

ewen-poch:
Addressed the preflight/environment lifecycle gap in 0f467ab4. Both public operations now use the existing shutdown-aware detection tracking. Parallel probes are also tracked individually so an early rejection cannot let disposal finish while a sibling probe is still cleaning up. Cancellation is checked before version/cache publication, and new calls are rejected after disposal starts.

The same signal now reaches configured-runtime probes, fallback discovery, memoized Codex version/ACP probes, and Codex component diagnostics. This reuses the existing lifecycle state; it adds no persisted fields, data-model changes, or user interaction.

Two public-boundary regressions first failed because the probes received no signal; they now verify abort, sibling draining, rejection after disposal, and suppression of late repository updates. A separate Codex diagnostic regression first resolved despite cancellation and now rejects before launching version probes. The related 69 tests pass. The final full suite is running with reduced concurrency because other local test runs are competing for resources; its result will be reported separately.

ewen-poch:
Final validation for 0f467ab4: typecheck, targeted ESLint, and all 69 runtime-manager/Codex detection tests pass. The full suite completed with 24,263 passed, 226 skipped, and 8 failed under substantial local contention. Two failures are the existing R/TIFF capture cases. The login probe and both Web bootstrap plus Notebook timing cases pass in isolated reruns (the last three passed in 2.45 seconds after the full run finished). The ACP REPL timeout also reproduced on the untouched pre-fix commit 91bbf6f2 with the same dependency layout; the ACP skill-edit case passed on rerun. The full run also reported a cleanup rejection after the timed-out login test, so this is not being presented as a completely green full suite.

The latest review raises a separate quit-admission window between the final installation snapshot and Settings disposal. The proposed change is to acquire the existing installation admission hold synchronously when quit preparation begins and release it if persistence cancels quitting. This requires new quit-stage hold/rollback wiring and would reject installation attempts during that interval. Implementation is paused for the owner's requested approval of lifecycle-state extensions; it has not yet been reproduced or fixed.

ewen-poch:
Addressed the quit-admission race in 28bbe459. Once quit confirmation is satisfied, the lifecycle synchronously acquires the existing Settings installation admission hold before any asynchronous persistence or teardown work. Repeated before-quit events reuse that attempt. Successful shutdown retains the hold; a cancelled quit releases it after rollback, including when another rollback step throws.

The capability is passed explicitly through the existing application composition, without widening the window Settings interface or adding persistent state. Regression tests failed on the previous lifecycle because admission remained open during the pending persistence flush. Both now pass, covering successful exit and failed persistence/rollback. Validation: 180 related tests passed via npm test; typecheck, targeted ESLint, and diff checks passed. This follow-up used focused regression checks; the preceding full-suite result and its remaining baseline/timing failures are documented above.

ewen-poch:
The quit-admission fix is complete in 28bbe459; all 180 related tests, typecheck, and targeted ESLint pass.

The latest finding identifies a cleanup-result propagation gap: runInstall currently discards the process-tree reap outcome and settles cancellation after bounded cleanup. However, preventing application exit when cleanup is degraded would change the existing shutdown policy. Current tests explicitly preserve bounded installer cleanup and degraded application shutdown; both policy checks still pass.

No further implementation has been made for this finding. Per the owner's instruction to pause on conflicts with the existing design, the policy choice needs confirmation: preserve bounded exit while accurately propagating/reporting degraded cleanup, or introduce a new refusal/confirmation path when process-tree cleanup cannot be confirmed. The latter affects shutdown interaction and must also define behavior for system shutdown and already-committed update/data-root relaunches.

ewen-poch:
Addressed degraded installer cleanup reporting while preserving the agreed bounded-shutdown policy.

- `reaped: false`, termination rejection, and the existing 8-second cleanup deadline now produce diagnostics and propagate a cleanup failure through SettingsService disposal. They are no longer reported as a clean module shutdown.
- SettingsService waits for authentication cleanup to settle before reporting an installation disposal failure.
- The cancellation response and bounded-exit behavior remain unchanged. Refusing to quit or introducing another confirmation would change the approved shutdown policy and is intentionally not included. No persistent state, data-model changes, or new user interaction were added.

Five regression cases failed against the previous implementation for the reported causes and pass with this fix. All 456 tests across the six related suites pass; typecheck, targeted ESLint, and diff checks pass. The full repository test suite was not rerun for this follow-up.

PR #2182, opened 2026-09-05
fix(connectors): release response bodies and enforce call budgets

ewen-poch:
HTTP failures in ParserEngine left response bodies unread, each retry reset `totalTimeoutMs`, and `Retry-After: 60` triggered another request after only five seconds. Release unfinished bodies, apply one deadline to the entire tool call, and respect server retry delays within the remaining budget.

- Cancel error bodies before retrying or throwing, and clean up interrupted, oversized, and failed body reads without letting cleanup failures replace the original error or block completion.
- Share the total deadline across retries, backoff, and every request in `run()`, while retaining the independent idle timeout and propagating caller cancellation.
- Accept numeric and HTTP-date `Retry-After` values without the five-second cap. If the remaining budget cannot accommodate the delay, return the HTTP status and suggested wait instead of retrying early.

Compatibility: `totalTimeoutMs` previously documented a per-attempt deadline; it now covers one complete `ParserEngine.call()`. The default 120 seconds is shared by all steps. Rfam sequence search uses the existing descriptor override to allow 600 seconds for submission and polling, preserving its default 300-second polling window; longer custom waits remain bounded by the 600-second whole-call cap. No persisted fields, schema migrations, dependencies, or UI changes are introduced. Arbitrary custom code still needs to cooperate with cancellation to stop its own work.

Validation:
- Added 16 regressions through the existing `ParserEngine.call()` / `fetchImpl` boundary. All fail against the unfixed source for the reported cleanup/deadline/retry behavior and pass after the fix; no new test seam.
- Added four Rfam regressions using the real engine and descriptor with fake timers. They failed twice at the old 120-second deadline before fixing, and now cover delayed submission, late successful results, the default polling timeout, and the hard call cap.
- `npm test -- src/main/connectors`: 58 files passed; 881 tests passed, 39 skipped (rerun with local socket/listener permissions).
- `npm run typecheck:node`, targeted ESLint, Prettier check, and `git diff --check`: passed.
- Full `npm test` was run: 23,374 passed, 627 failed, 226 skipped. Failures include widespread sandbox `listen EPERM` / `spawnSync ps EPERM` and timeout failures; this is not a clean full-suite result. The connector module passes after removing the local-listener restriction.

The body tests demonstrate the missing release path; they do not measure a production connection-pool leak.

### DIFF
--- a/src/main/connectors/descriptors/rna.test.ts
+++ b/src/main/connectors/descriptors/rna.test.ts
@@ -366,6 +366,96 @@ describe('rna / rfam', () => {
     })
   })
 
+  it.each([150_000, 295_000])(
+    'search_sequence preserves the default window for a job ready after %ims',
+    async (readyAfterMs) => {
+      vi.useFakeTimers()
+      vi.setSystemTime(0)
+      try {
+        const submissionMs = 20_000
+        const fetchImpl = vi.fn(async (_url: string | URL | Request, init?: RequestInit) => {
+          if (init?.method === 'POST') {
+            await new Promise((resolve) => setTimeout(resolve, submissionMs))
+            return Response.json({ jobId: 'job-1', resultURL: 'https://rfam.test/result/1' })
+          }
+          return Response.json(
+            Date.now() >= submissionMs + readyAfterMs
+              ? { hits: {}, searchSequence: 'GGUUCC' }
+              : { status: 'PEND' }
+          )
+        })
+        const result = new ParserEngine({ fetchImpl })
+          .call(tool('search_sequence'), { sequence: 'GGUUCC' }, {})
+          .then(
+            (value) => ({ value, finishedAt: Date.now() }),
+            (error: unknown) => ({ error, finishedAt: Date.now() })
+          )
+        await vi.advanceTimersByTimeAsync(submissionMs + readyAfterMs)
+        expect(await result).toMatchObject({
+          value: { job_id: 'job-1', num_hits: 0, search_sequence: 'GGUUCC' },
+          finishedAt: submissionMs + readyAfterMs
+        })
+        expect(vi.getTimerCount()).toBe(0)
+      } finally {
+        vi.useRealTimers()
+      }
+    }
+  )
+
+  it('search_sequence still stops an unfinished job at its default polling deadline', async () => {
+    vi.useFakeTimers()
+    vi.setSystemTime(0)
+    try {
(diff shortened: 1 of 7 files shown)


ewen-poch:
Rebased onto `main` at `d284a2dedfedd091079fdb998cf213fb79e4a3f5` and updated the PR head to `d6413bcaaee028f1c6e27d9e733139276d692938` using an explicit force-with-lease. Git replayed the change cleanly; the five connector files are byte-for-byte unchanged from the previous PR head.

Validation after rebasing:
- `npm test -- src/main/connectors`: 58 files passed; 877 tests passed, 39 skipped.
- `npm run typecheck:node`: passed, including the notebook sandbox typecheck.
- `git diff --check origin/main...HEAD`: passed.

The Rfam polling-budget finding remains pending approval to expand the file scope; this rebase does not address that finding.

ewen-poch:
Confirmed against the current PR head (`d6413bcaaee028f1c6e27d9e733139276d692938`) using the real `ParserEngine` and `RNA_TOOLS` descriptor with the existing `fetchImpl` injection and a fake clock. A job configured to finish at 150 seconds is rejected at exactly 120 seconds with `ConnectorRequestTimeoutError`, despite the descriptor's default `max_wait_s=300`.

The proposed correction is to use the existing descriptor-specific `totalTimeoutMs` override and add a regression test for the polling window. This preserves the whole-call deadline design and does not require persistent state, a data-model change, or a new UI. Editing `descriptors/rna.ts` and its test exceeds the currently authorized file scope, so implementation is paused pending approval. No repository files were modified for this reproduction.

ewen-poch:
Fixed the Rfam polling-window regression in `624b2f23631da85f7f20362938381382687547c4` by setting the existing `totalTimeoutMs` override on `rna/search_sequence` to 600 seconds. This preserves the default 300-second polling window with headroom for submission and network waits. The tool description now explicitly states the 600-second whole-call cap, which also bounds longer custom waits. The engine's shared deadline, idle timeout, and cancellation behavior are retained; no new state, configuration fields, or persistence are introduced.

Four regression cases use the real `ParserEngine` and Rfam descriptor with fake timers. Before the fix, all four failed twice because the call ended at 120 seconds. They now pass, including results available after 150/295 seconds of polling following a 20-second submission, the default unfinished-job timeout at 300 seconds, and enforcement of the 600-second cap for longer custom waits.

Validation: all 58 connector test files passed (881 tests passed, 39 skipped), node and sandbox typechecks passed, targeted ESLint passed, and the diff passed whitespace checks.

PR #2209, opened 2026-09-05
fix(sessions): respect newer metadata during lazy hydration

ewen-poch:
## Problem

An unopened Session starts from a revision-1 summary. When another client supplies a revision-2 full snapshot, lazy hydration previously kept the summary title, pin and archive while accepting the newer body/revision. A later unrelated title save could therefore submit the old pin with the current revision; Main correctly accepted that revision and passed the stale pin to its repository.

## Proposed change

Compare durable revisions in the existing lazy-hydration branch. A newer snapshot supplies title, pin and archive; only an explicitly pending local title (`unsavedTitle`) overrides it. Same/older lazy responses retain the existing metadata protection. This follows the confirmed title-only conflict policy: a pending local pin does not defeat a newer durable value.

## Scope and non-goals

- No new fields, transient markers, enum values, services, dependencies, IPC methods, UI flows, or durable schema/model changes.
- Persistence continues through the existing renderer saver and Main owner. This prevents stale pin submission; Main archive/runtime/details protections remain unchanged. Dedicated Session-details events are a separate projection path.
- No historical-data migration/backfill or compatibility adapter is needed. Already overwritten historical pin values cannot be reconstructed by this fix.
- A separate test-only commit supplies two now-required Settings lifecycle dependencies in the existing delegated-work fixture. Main had merged the Settings lifecycle change after that fixture was added. The unchanged baseline failed Node typechecking and seven cancelled-quit cases (`getActiveSettingsInstallId is not a function`); no lifecycle production behavior changes.

Before changing production code, the public store/saver regressions repeatedly failed in the reported way: three failures (untouched metadata, title-only metadata, stale pin at the Main write port) and one passing existing same-revision control. After the fix, the focused matrix passes all five cases, including the approved newer-revision pin conflict policy.

Final Test Impact Set (run after the final material edit):

| Behavior / boundary | Project-owned check | Result |
| --- | --- | --- |
| Lazy revision/metadata merge; same-revision edits; loaded title acknowledgements | `session-store.test.ts`, `session-store.architecture.test.ts`, `session_renderer` module contracts | PASS |
| Serialization, external-update no-echo, title-only saves, current-revision pin write | renderer `session-persistence.test.ts` and `.render.test.tsx`, real `SessionPersistenceStateOwner` through recording repository ports | PASS |
| Lifecycle events, archive undo, runtime/workspace/files/compute consumers | `useLifecycleSync.test.tsx`, `archive-undo-store.test.ts`, all declared tests of `workspace_runtime`, `workspace_page`, `project_files_view`, `compute_service` | PASS |
| Durable revision and Main field ownership | Main `coordinator.test.ts`, `coordinator-contract.test.ts`, `revision-conflict.test.ts` | PASS |
| Baseline lifecycle fixture repair | full `production-composition.test.ts` and `app-lifecycle.test.ts` | PASS |
| Process typing and style | `npm run typecheck:node`, `npm run typecheck:web`, `npm run lint`, changed-file Prettier check, `git diff --check` | PASS |

The test command is `npm test -- <deduplicated module testFiles plus explicit files above> --maxWorkers=2`; module lists come from `createModuleTestPlan` in `scripts/ci/module-test-impact.mjs`. Final result: **106 files passed / 1 skipped; 3,231 tests passed / 6 skipped**. Both typechecks, repository lint, formatting and diff checks passed.

The final automatic affected plan falls back to full at the existing production-composition test, which has no manifest owner; the original P01-only plan also fell back at the unowned renderer saver test. Per the maintainer request, no full local `npm test` was run and the manifest was not changed to narrow CI. Full PR CI remains authoritative. Socket-based consumer tests require localhost/Unix-socket permissions; sandbox EPERM failures were rerun with those permissions.

Independent Standards and Spec reviews found no actionable findings and confirmed the final impact mapping, including the two-line fixture repair. The new persistence regression is an in-memory composition test through real serialization/Main revision validation, not a real SQLite or two-window E2E run. No provider protocol, path handling, layout, or persisted format changes justify additional live-provider, visual, migration or generated-API checks; configured full/platform PR CI is still required.

## Review focus

Check that old summary metadata cannot acquire a newer durable revision, pending title protection stays field-specific, and unrelated saves retain the incoming pin. Confirm that the test-only fixture repair remains separate from the production fix.

### DIFF
--- a/src/main/delegation/production-composition.test.ts
+++ b/src/main/delegation/production-composition.test.ts
@@ -956,6 +956,8 @@ describe('production delegated-work composition', () => {
               window as unknown as ReturnType<AppLifecycleDeps['createMainWindow']>,
             createTray: () => undefined,
             shutdownBackends,
+            holdSettingsInstallAdmission: () => () => undefined,
+            getActiveSettingsInstallId: () => undefined,
             prepareForQuit,
             abortQuitPreparation,
             flushSessionPersistence,
--- a/src/renderer/src/lib/session-persistence/session-persistence.test.ts
+++ b/src/renderer/src/lib/session-persistence/session-persistence.test.ts
@@ -1,5 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
+import { SessionPersistenceStateOwner } from '../../../../main/session-persistence/state-owner'
 import { ARTIFACT_FINALIZATION_INVALID_PROOF } from '../../../../shared/artifacts'
 import {
   activateConversationBranch,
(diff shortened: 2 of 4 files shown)


ewen-poch:
Regarding **[P1] Preserve a pending local pin during lazy hydration**: the reported outcome is real, but it is the explicitly approved conflict policy for this change. The maintainer has reviewed this finding and confirmed that we should retain that policy.

- For a **newer durable revision**, incoming pin and archive values win; only an explicitly unsaved local title is preserved. This keeps lazy hydration aligned with the existing title-only protection for loaded Sessions.
- For a **same or older lazy-load response**, the existing protection for local title and pin edits remains intact.
- The pin-only case in `session-store.test.ts` explicitly tests the newer-revision policy. The renderer-to-Main composition regression separately verifies the reported defect: an untouched old summary must not retain its stale pin under a newer revision, and an unrelated title save must not undo that newer durable pin.

Preserving pending pins against newer snapshots would be a different conflict policy. It would require additional transient edit tracking and coordinated save/acknowledgement handling beyond the approved scope. We are therefore not adopting that recommendation in this PR. No code or data-model changes are needed for this disposition.

PR #2214, opened 2026-09-05
fix(notebook): correct environment repair and package outcomes

ewen-poch:
## Problem

Queued Python repair could quarantine a healthy runtime before its serialized operation started,
so cancellation stopped rebuilding but left bindings unavailable and a durable repair marker.
R package batches could change installed packages, fail verification, and discard restart advice.
Named-environment creation also accepted reserved default prefixes that removal then refused.

## Proposed change

- Consume queued repair cancellation before preparation. Pass preparation through the existing
  Provisioner repair options, stop active kernels before acquiring the exclusive environment lease,
  and retain the lease through rebuild, verification and binding finalization. Without preparation,
  lease-wait cancellation is checked before deletion. Preparation failure
  prevents deletion; finalization failure retains the existing fail-closed readiness behavior.
- Preserve restart advice independently of batch success. For R, confirmed package deltas and
  non-skipped installer attempts with mutation risk can recommend restart even when `ok=false`.
  Admission refusal and observations without mutation do not recommend restart. Protected-identity
  quarantine results retain their repair-only guidance instead of inferring a stale restart hint.
  Failed read-only R install/remove preflights record no mutation risk, while their original failure
  classification continues to govern fallback admission. Real unknown-risk failures retain restart advice.
- Use one reserved-name guard for creation and removal, including `default-python-` and `default-r-`.
- Add root/child/attempt/Project/stale-owner revocation regressions for E04. Its production fix
  already landed in #2204; this PR does not duplicate that implementation.

## Scope and non-goals

No new state enumeration, database schema, persisted field, file format, dependency, or renderer
interaction. `RuntimeRepairOptions.onStarting` is an optional in-process callback. E05 changes the
timing of existing repair-marker and binding writes. Existing quarantine, journal and directory
ownership protections remain in effect. R kernels are not restarted automatically.

Historical compatibility is deliberately bounded: existing repair markers are not automatically
cleared because false preparation cannot safely be distinguished from a real integrity failure.
Existing reserved-name conflicts remain protected; historical environment recovery requires a
separate ownership decision. This prevents new conflicts without claiming to repair old ones.
The current Settings reinstall flow already hides Cancel; no queue-only cancellation UI is added.

Before production changes, the eight E05–E07 behavioral regression cases failed twice for the
reported symptoms. The final tests were also rerun twice against original production: ten failures
(the original eight plus two repair callback/lease contracts), ten passing controls. E04's two
regressions fail twice against the pre-#2204 implementation and pass with main's existing fix.

Additional review regressions reproduce lease-wait cancellation through the real serialized Provisioner
and Lifecycle, and a stale restart hint after protected-R repair through the same RuntimeService.
The unprepared Provisioner cancellation and post-repair hint tests each failed twice before their fixes;
the cancellation control passed against main. Both entry paths and the post-repair hint now pass. R install/remove preflight regressions through
the real package manager also failed twice: only dry-run ran, but restart advice was published.
They now pass, and a real-installer unknown-risk control still requires restart advice.

Final local results, run after the last material edit:

| Behavior | Check | Result |
| --- | --- | --- |
| Revocation identities, repair queue/leases/failures, package outcomes, naming, binding/inventory and runtime consumers | Main owner/contract command below | 894 passed across module runs; see process-permission retry below |
| Authenticated local RPC dispatch | `npm test -- src/main/notebook/local-rpc-server.test.ts --maxWorkers=1` | 69 passed |
| Existing reinstall UI/store and package activity consumers | Renderer consumer command below | 148 passed |
| Main and sandbox type contracts | `npm run typecheck:node` | Passed |
| Source lint and patch integrity | `npm run lint`; `git diff --check` | Passed |

```sh
npm test -- src/main/notebook/{environment-operations,runtime-repair,package-operations,environment-management,environment-lifecycle-workflows,environment-operation-foundation,provisioner,provisioner.startup,provisioner.upgrade,package-manager,package-mutation,package-admission,runtime-paths,session-registry,lane-identity,runtime-repair-policy,environment-state-tracker,environment-lease-manager,runtime-service,runtime-service.architecture,mcp-server,env-ipc}.test.ts --maxWorkers=1
npm test -- src/main/notebook/package-manager.test.ts --maxWorkers=1
npm test -- src/renderer/src/pages/settings/RuntimesPanel.render.test.tsx src/renderer/src/stores/notebook-env-store.test.ts src/renderer/src/pages/workspace/workspace-tool-activity-details.test.ts src/renderer/src/pages/workspace/WorkspaceActivityGroup.i18n.render.test.tsx --maxWorkers=1
```

The combined main run passed 893 tests; one existing child-reaping test could not inspect processes
inside the local sandbox. All 95 package-manager tests passed when rerun with process inspection
permission, leaving 894 unique passing main tests. Including the RPC and renderer consumers,
1,111 unique tests passed after the final material edit.

The initial RPC run hit sandbox `listen EPERM`; all 69 tests passed when run with local socket
permission. Older runtime-service assertions that erased restart advice were updated to preserve it.
The dependency classifier reports an unknown Notebook module owner and selects full CI. Per maintainer
instruction, local checks use the explicit owner/interface/consumer set above rather than the complete
portable suite; CI ownership metadata was not changed. Real binding behavior is covered by
runtime-service and repair tests.

Real interpreter downloads, R namespace/DLL behavior and full cross-platform Electron flows are not
established by these substitute-based regressions. Existing path helpers and Windows provisioning
cases remain covered; exact-head PR CI and independent review are the final validation authority.
Framework protocol adapters, branch replay and subscription payloads are unchanged.

## Review focus

Queued cancellation must precede every repair side effect. Preparation must stop active execution
before waiting for its environment lease. Once preparation starts, per-language cancellation remains
unsafe; before preparation, waiting for a writer must remain cancellable. Failed package batches must preserve actionable restart
advice, while unchanged metadata and never-started installers must not create it. Existing reserved
prefixes must remain protected from deletion without proven ownership.

### DIFF
--- a/src/main/notebook/environment-lifecycle-workflows.test.ts
+++ b/src/main/notebook/environment-lifecycle-workflows.test.ts
@@ -40,6 +40,7 @@ const fakeProvisioner = (over: Partial<RuntimeProvisioner> = {}): RuntimeProvisi
   provisionR: vi.fn().mockResolvedValue(undefined),
   upgradeIfNeeded: vi.fn().mockResolvedValue(undefined),
   repair: vi.fn().mockImplementation(async (_language, _onProgress, options) => {
+    await options?.onStarting?.()
     await options?.onVerified?.()
   }),
   restoreRelocatedEnvs: vi.fn().mockResolvedValue(undefined),
@@ -159,6 +160,7 @@ describe('createNotebookEnvironmentLifecycle', () => {
     // UI repair is the user's Reset: it force-clears the quarantine (force: true).
     expect(provisioner.repair).toHaveBeenCalledWith('r', expect.any(Function), {
       force: true,
+      onStarting: expect.any(Function),
       onVerified: expect.any(Function)
     })
   })
@@ -169,7 +171,8 @@ describe('createNotebookEnvironmentLifecycle', () => {
       order.push('prepare')
     })
     const provisioner = fakeProvisioner({
-      repair: vi.fn(async () => {
+      repair: vi.fn(async (_language, _progress, options) => {
+        await options?.onStarting?.()
         order.push('repair')
       })
     })
@@ -198,14 +201,16 @@ describe('createNotebookEnvironmentLifecycle', () => {
 
   it('does not enter destructive repair when runtime coordination fails', async () => {
     const provisioner = fakeProvisioner()
+    const onRepairCompleted = vi.fn()
     const lifecycle = createLifecycle(provisioner, {
+      onRepairCompleted,
       onRepairStarting: async () => {
         throw new Error('binding persist denied')
       }
     })
 
     await expect(lifecycle.repair('r', 'default-r')).rejects.toThrow('binding persist denied')
-    expect(provisioner.repair).not.toHaveBeenCalled()
+    expect(onRepairCompleted).not.toHaveBeenCalled()
   })
 
   it('publishes a successful verified repair back to the runtime service', async () => {
@@ -417,6 +422,7 @@ describe('createNotebookEnvironmentLifecycle', () => {
     await lifecycle.repair('python', 'default-python')
     expect(provisioner.repair).toHaveBeenCalledWith('python', expect.any(Function), {
       force: true,
+      onStarting: expect.any(Function),
       onVerified: expect.any(Function)
     })
 
--- a/src/main/notebook/environment-lifecycle-workflows.test.ts
+++ b/src/main/notebook/environment-lifecycle-workflows.test.ts
@@ -160,7 +160,7 @@ describe('createNotebookEnvironmentLifecycle', () => {
     // UI repair is the user's Reset: it force-clears the quarantine (force: true).
     expect(provisioner.repair).toHaveBeenCalledWith('r', expect.any(Function), {
       force: true,
-      onStarting: expect.any(Function),
+      onStarting: undefined,
       onVerified: expect.any(Function)
     })
   })
@@ -422,7 +422,7 @@ describe('createNotebookEnvironmentLifecycle', () => {
     await lifecycle.repair('python', 'default-python')
     expect(provisioner.repair).toHaveBeenCalledWith('python', expect.any(Function), {
       force: true,
-      onStarting: expect.any(Function),
+      onStarting: undefined,
       onVerified: expect.any(Function)
     })
 
--- a/src/main/notebook/environment-lifecycle-workflows.ts
+++ b/src/main/notebook/environment-lifecycle-workflows.ts
@@ -148,11 +148,13 @@ const createNotebookEnvironmentLifecycle = (
           if (deps.waitForRecovery) await deps.waitForRecovery()
           await provisioner.repair(parsedLanguage, report, {
             force: true,
-            onStarting: () =>
-              deps.onRepairStarting?.(
-                parsedLanguage,
-                explicitRuntimeRepairTarget(parsedLanguage, runtimeIdentity)
-              ),
+            onStarting: deps.onRepairStarting
+              ? () =>
+                  deps.onRepairStarting?.(
+                    parsedLanguage,
+                    explicitRuntimeRepairTarget(parsedLanguage, runtimeIdentity)
+                  )
+              : undefined,
             onVerified: () => deps.onRepairCompleted?.(parsedLanguage)
           })
         }),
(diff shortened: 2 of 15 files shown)


ewen-poch:
The lease-wait cancellation finding is reproducible through the real serialized DefaultRuntimeProvisioner and EnvironmentLeaseManager. With an exclusive package-mutation lease held, cancelling a repair that has no preparation callback is ignored; after releasing the lease, the interpreter is replaced and the rebuild and completion callbacks run. The regression test failed twice on this PR and passed against the current main implementation. The test is local pending the cancellation-boundary decision.

Moving the guard after lease acquisition for every repair would conflict with the explicit-repair preparation contract: preparation persists isolation, invalidates bindings, and stops kernels before acquiring the exclusive lease. Those kernels can hold shared leases, so preparation cannot simply move inside the exclusive lease. Accepting cancellation after that preparation would leave the environment unavailable again, recreating E05.

The proposed minimal adjustment is to preserve cancellation during lease acquisition when no preparation has run, while retaining the non-cancellable boundary once explicit preparation begins. This stays within the existing files and adds no states, persisted fields, or UI. Maintainer confirmation is pending because the review recommendation crosses the previously agreed boundary.

The branch has been rebased onto main b5077745 and pushed as d4329f04. All 172 tests in the five directly affected regression suites passed after the rebase. The previous Windows E2E shard 2 failure occurred during dependency setup: node-gyp's Node header download failed with an undici Parser.finish assertion, before any build or E2E test ran. The rebased head has started fresh CI; no CI or dependency changes were made without evidence of a repository defect.

ewen-poch:
Implemented the maintainer-approved cancellation boundary. Repairs without preparation remain cancellable while waiting for an environment lease; the abort signal is checked immediately after acquiring the lease and before deletion. Explicit preparation remains non-cancellable because it has already invalidated bindings and persisted isolation. Lifecycle no longer supplies an empty preparation callback when no preparation owner exists. The serialized Provisioner regression failed twice before the fix and passed against main; both Provisioner and Lifecycle entry regressions now pass.

Also reproduced the protected-R restart-advice finding twice through the real RuntimeService: needsRestart was true, and the same service still projected restartRecommended after successful explicit repair and subsequent R execution. Restart inference now excludes repairRequired results and preserves the installer's repair-only guidance. Ordinary failed batches with confirmed or possible package mutations still recommend restart. No states, persistence formats, data models, or UI were added.

After the final material edit: 796 main owner/contract/consumer tests, 69 RPC tests, and 148 renderer consumer tests passed (1,013 total), along with typecheck:node, lint, and git diff --check. The PR body contains the exact focused test commands and remaining platform limitations.

The rebased Windows shard passed dependency setup but failed two workspace E2E cases at Send message / deterministic-reply waits, outside this PR's changed files. A targeted retry was requested for that shard. No unrelated test, workflow, or product code was changed; the new head's CI will also validate those cases.

ewen-poch:
The dry-run finding is valid. Regression tests now exercise the real package manager through NotebookPackageOperations for both R install and uninstall. Each invokes only the dry-run subprocess, executes no package mutation or fallback, yet the result reports needsRestart and publishes restart advice. Both cases failed twice before any production adjustment.

The existing attempt shape does not distinguish preflight from a real Conda transaction with unstructured output. Ignoring all unknown risks in PackageMutation would hide advice for real partial mutations. The proposed minimal fix is to report the existing mutationRisk: none for failed read-only R preflight attempts at the producer, while keeping the original failure classification for fallback authorization. This needs no new fields, enum values, schema, or persistence mechanism; historical attempt records would remain unchanged.

That adjustment touches package-manager.ts and its existing tests, outside the current PR file list. It is paused for maintainer approval under the requested scope rule. The failing integration regressions remain local and uncommitted. The already-pushed cancellation and protected-identity advice fixes are unchanged.

ewen-poch:
Implemented the approved producer correction for failed R install and uninstall preflights. Their attempts now record the existing mutationRisk: none because no prefix-writing transaction ran. The original failure classification still controls fallback authorization; unstructured diagnostics cannot enable fallback. Actual installer failures with unknown mutation risk still recommend restart.

Both public package-operation regressions use the real package manager, verify that the only executed command is dry-run, and verify that no fallback or restart recommendation is produced. They failed twice before the fix. No new fields, enum values, data models, or persistence formats were introduced; existing historical attempt records remain unchanged.

Final focused validation covers 1,111 unique passing tests: 894 main owner/contract/consumer tests, 69 RPC tests, and 148 renderer consumer tests. The existing package-manager child-reaping fixture initially could not inspect processes inside the local sandbox; all 95 tests in that module passed with the required process permission. typecheck:node, lint, and diff checking also passed after the final material edit. The PR body contains the exact command set and limitations.

The previous head ac8f85b7 completed all CI successfully, including the Windows E2E shard that previously failed. No workspace page, E2E, dependency, or workflow changes were needed. The new head will be checked independently before merge.

PR #2257, opened 2026-09-06
fix(storage): preserve migration references and recovery semantics

ewen-poch:
## Problem

Moving a data root can preserve relative symlink text while changing its referent to a missing or different external dataset. A failed directory-pointer save is also presented as a successful move needing a restart. Separately, reconnect recovery waits for a full usage scan, and the controlled migration dialog does not forward keyboard dismissal after errors.

## Proposed change

- Validate relative links against the actual migrated path set before copying and again before committing a staged copy. Check lexical and filesystem-resolved paths, preserve symlink/parent traversal ordering, and refuse unproven dangling aliases. Keep supported internal relative and external absolute references.
- Present `switchoverFailed` as an uncommitted location switch, explain that the original location is still active, and direct users to close and retry the move. Update all eight locales and correct misleading comments.
- Use a fresh `storage.getStatus()` for reconnect recovery. Keep usage errors in the existing storage page/store. No older-Web fallback is added, as requested by the maintainer.
- Forward Dialog close requests only in the existing error stage. Let Root own the Escape gate as well as outside dismissal; remove the redundant Escape handler that retained an earlier stage in real-browser validation.

## Scope and non-goals

No new public command, state/enum, persisted field, database schema, migration marker format or data model. No link rewriting or external-file ownership changes. Existing unsafe staged copies are refused by the commit recheck; already completed historical reference damage is not repaired automatically. Pointer-save ordering and rollback ownership are unchanged.

Reproduced before production changes: five expected-correct assertions failed repeatedly on the original code, covering two external-link outcomes plus the three renderer symptoms. The link reproductions used real copy, inventory verification, commit and source cleanup; they observed ENOENT and wrong dataset content while the original external file remained intact. Two real backend checks confirmed settings EIO preserves source data and that getStatus succeeds while getInfo rejects workspace EACCES. No production test seam was added.

| Behavior / boundary | Command | Result |
| --- | --- | --- |
| Copy/link safety, inventory/commit, status/usage, cleanup; settings pointer persistence; renderer consumers; locale catalogs; Electron/Web wiring | `npx vitest run src/main/storage src/main/settings/data-root-setting.test.ts src/renderer/src/stores/storage-info-store.test.ts src/renderer/src/components/LegacyDataMoveDialog.render.test.tsx src/renderer/src/components/DataRootMissingDialog.render.test.tsx src/renderer/src/pages/settings/StoragePanel.render.test.tsx src/renderer/src/pages/settings/StorageMigrationModal.render.test.tsx src/renderer/src/pages/settings/storage-i18n.render.test.tsx src/renderer/src/i18n/resources.test.ts src/preload/index.test.ts src/renderer/web/api-installer.test.ts src/shared/web-rpc-contract.test.ts scripts/ci/check-pr-policy.test.ts scripts/ci/classify-pr-changes.test.ts scripts/ci/module-test-impact.test.ts scripts/ci/validate-module-impact.test.ts scripts/ci/pr-gate-workflow.test.ts scripts/ci/evaluate-pr-gate.test.ts` | 37 files, 1650 tests passed |
| Main, sandbox package and renderer types | `npm run typecheck` | Passed |
| Source quality and formatting | `npm run lint`; Prettier check on changed code/tests and catalogs; `git diff --check` | Passed |
| Real rendered interactions | ego-browser with production components/styles and injected command outcomes | Error Escape/outside click close; committing Escape/outside click remain blocked; remote reconnect calls status once, usage zero times, and resolves. Final screenshot inspected. |
| Automatic impact selection | `npm run test:affected:explain -- --base c6e835a9 --head HEAD` | Full plan: storage tests have no registered module owner. Per maintainer instruction, full local suite was not run; PR CI remains authoritative. |

Consumers were included because the existing storage interface reaches legacy migration, Settings, startup recovery and both Electron/Web transports. Provider/agent implementations and database schema are unchanged and excluded from the focused local run. Tests use platform-aware paths and real temporary files; The parent-traversal test respects Windows syntactic normalization and POSIX link traversal. The existing Windows core lane does not execute these two migration suites: their native Windows execution is an uncovered risk, not a claimed result. Adding them to the protected PR Gate requires an explicit maintainer ruleset bypass; that attempted coverage-list change was withdrawn rather than bypassing the guard. Browser checks used an isolated component preview, not a live external drive or packaged Electron application.

## Review focus

Verify that rejected links cannot pass the staged-copy commit gate; ordinary relocatable links still work; only uncommitted switch failures get the original-location guidance; and Root's dismissal gate protects all non-error stages. Required exact-head CI and independent review remain the final verification authority.

## Review disposition

The automated review's sole finding asks for an older-Web `getInfo()` fallback ([review](https://github.com/aipoch/open-science/pull/2257#issuecomment-5560151256)). The maintainer explicitly declined older-Web compatibility after reviewing this plan. Current-backend reconnect correctness is covered by the regression and browser checks above; adding that fallback would contradict the agreed scope. The finding is therefore not adopted. The automated verdict/labels are left intact; no review result or CI gate is overridden.

The initial head had one Windows functional-journey startup timeout (database phase migrating exceeded the fixture budget). The test retried successfully, and the job correctly failed under --fail-on-flaky-tests. A strict rerun of that group passed with no source/assertion/timeout change. Subsequent validation-only changes correct the platform expectations and require fresh final-head CI; prior-head results are not reused as final proof.

Final validation follow-up corrects the parent-traversal expectation for Windows and documents the intentionally current-only reconnect API. The protected PR Gate test list remains unchanged. Local final selection: 1650 tests passed; typechecks, lint and formatting passed.


ewen-poch:
Regarding [the mixed-version recovery fallback recommendation](https://github.com/aipoch/open-science/pull/2257#issuecomment-5560573679): older Web/Main versions without `storage.getStatus` are intentionally outside the supported scope of this change. The maintainer explicitly confirmed that older Web compatibility is not required.

The reconnect flow therefore keeps the existing lightweight `getStatus()` API as its recovery boundary. This preserves Main's availability and recovery checks without making a successful reconnect depend on a recursive usage scan. The current code documents this compatibility decision, and the regression test covers successful recovery when `getInfo()` would fail with a usage-scan permission error.

No fallback will be added for this recommendation. Supporting mixed-version recovery would change the agreed compatibility scope; the existing review verdict has been left intact.

PR #2265, opened 2026-09-06
fix(cli): reject unused positional arguments

### LINKED ISSUE #2264: [Bug]: CLI silently ignores extra positional arguments
### Preflight checklist

### What happened?

Commands with a fixed positional-argument contract accept extra values, retain them in `positionals`, and continue to dispatch without consuming them. For example, `status unexpected` is parsed successfully even though `status` accepts no positional arguments. The same behavior affects state-changing commands such as bare `run`, `start`, and `update`, so a typo or stale script argument can trigger an unintended operation.

I expected the parser to reject surplus positional arguments with a usage error (exit code 2) before any daemon connection or command dispatch.

### Steps to reproduce

1. Check out current `main` at `33aae409` (package version `0.25.1`).
2. From the repository root, run:

   ```sh
   node --input-type=module -e "import { parseCliArgs } from './packages/open-science/cli.mjs'; console.log(parseCliArgs(['status', 'unexpected']))"
   ```
3. Observe that parsing succeeds and the returned object contains `positionals: ['unexpected']`.
4. Run the equivalent installed CLI command while the daemon is available: `open-science status unexpected`.
5. Observe that `status` proceeds instead of returning a usage error.

Expected: the extra argument is rejected before connecting to the daemon.

Actual: the extra argument is silently ignored by dispatch.

### Operating system

Windows

### App version

0.25.1 (main at 33aae409)

### Provider / model

Not applicable (CLI argument parsing)

```shell
Observed parser result on `main` before the fix:

{
  command: 'status',
  options: { open: true, json: false },
  positionals: [ 'unexpected' ]
}
```

QiuLsG:
## Problem

Fixed-arity CLI commands retained surplus positional arguments and then ignored them during dispatch. A typo such as `open-science run unexpected --project ...` could therefore start a run that the caller did not intend.

Closes #2264

## Proposed change

- Define positional-argument limits for every fixed-arity command path.
- Raise `CliUsageError` during argument parsing, before daemon connection or dispatch.
- Keep `project create`, `project update`, and `project session-defaults` unbounded so unquoted multi-word Project names continue to work.
- Add table-driven regression coverage for every fixed-arity path and the multi-word Project-name exceptions.

## Scope and non-goals

This only rejects surplus positional arguments. Validation for missing required positionals and unknown grouped subcommands is unchanged.

All passing checks below ran after the final material edit on Windows:

- Fixed-arity commands reject extra positionals -> `npx vitest run cli packages/open-science` -> 20 files passed; 364 tests passed, 21 skipped.
- Multi-word Project names remain accepted -> same CLI Gate -> passed.
- Node and renderer types -> `npm run typecheck` -> passed.
- Formatting -> `npx prettier --check packages/open-science/cli.mjs packages/open-science/cli.test.ts` -> passed.
- Patch whitespace -> `git diff --check` -> passed.

Full fallback attempt:

- `npm test` exited 1 locally after 1,235 test files and 20,593 tests passed; 242 files / 54 tests failed outside the changed CLI surface. Failures were associated with unavailable Windows symlink privileges, a missing `py` launcher, and resource exhaustion in the default high-concurrency run. No failing suite imports the modified CLI module, and the exact CLI Gate passes. Exact-head CI remains authoritative.

No E2E or platform-specific lane was added because validation happens in the pure argument parser before I/O.

## Review focus

- Completeness and accuracy of the positional-limit table versus the documented commands.
- Preservation of multi-word Project names on the intentionally unbounded Project command families.

### DIFF
--- a/packages/open-science/cli.mjs
+++ b/packages/open-science/cli.mjs
@@ -140,6 +140,30 @@ const VALUE_OPTIONS = {
 
 const TASK_COMMANDS = new Set(['project', 'run', 'session', 'settings', 'plan', 'artifacts'])
 const GROUP_COMMANDS = new Set(['codex', 'project', 'session', 'settings', 'plan', 'artifacts'])
+// Project create, update, and session-defaults intentionally remain unbounded because their
+// positional Project names may contain multiple unquoted words.
+const POSITIONAL_LIMITS = new Map([
+  ['start', 0],
+  ['stop', 0],
+  ['status', 0],
+  ['url', 0],
+  ['update', 0],
+  ['rollback-to-0.7.3', 0],
+  ['codex login', 0],
+  ['project list', 0],
+  ['run', 0],
+  ['run status', 1],
+  ['run cancel', 1],
+  ['session status', 1],
+  ['session config', 2],
+  ['settings agent-routing', 1],
+  ['plan show', 1],
+  ['plan approve', 1],
+  ['plan reject', 1],
+  ['plan revise', 1],
+  ['artifacts list', 1],
+  ['artifacts download', 1]
+])
 
 export class CliUsageError extends Error {
   constructor(message) {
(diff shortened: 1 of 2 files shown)


ewen-poch:
Thanks for the fix and the thorough regression tests! Rejecting extra positional arguments before dispatch helps prevent unintended operations, and preserving multi-word project names keeps existing workflows working.

PR #2276, opened 2026-09-07
fix(compute-host): preserve concurrent profile and authentication updates

ewen-poch:
Concurrent details saves could both report success while one document disappeared; probes could overwrite a newly pinned scratch path or write an old identity's path after authentication changed. Remote shell failures were also treated as successful resource detection, and editing an SSH-config host silently changed an inherited port to an explicit 22.

- Compare and write details by stable host ID and old text in one database update. Append retries this comparison at most five times, rereading and checking the length limit each time. Conflicts preserve the editor draft and offer a reload-and-manual-merge path.
- Persist probe results and automatic scratch paths in one transaction, matching stable host ID and authentication revision. Automatic paths additionally require the current stored pinned flag to be false. Rejected probes return an explicit conflict.
- Require a successful exit and usable OS/scheduler output. Missing scheduler checks remain unknown; only explicit negative checks mean no scheduler. Failed probes retain bounded diagnostics and do not change the existing host shape. The latest failed snapshot replaces resource metrics rather than presenting old metrics as newly verified.
- Keep SSH-config ports optional through the renderer, request type, authentication comparison, validation candidate, and persistence. A blank field restores SSH configuration inheritance; password authentication still requires a valid explicit port.

No database columns, migrations, dependencies, or architectural layers are added. This uses existing host identity, authentication revision, and document text. UI changes are limited to draft conflict recovery and inherited-port display/input; no automatic merge or history system is introduced.

Validation: deterministic regression tests were run on the original implementation twice and reproduced lost replacements/appends, length-limit races, both scratch races, invalid successful probes, and inherited-port changes. An additional authentication-owner test rejected an omitted SSH port on the original code. The draft recovery test failed because no reload action existed. All now pass. The concurrency tests use a real migrated SQLite database, existing public owner/repository methods, and the existing broker boundary; no production test seam was added.

Targeted tests (including all locale guards), TypeScript checks, and lint passed. Full-suite result recorded below.

The sandboxed full `npm test` run completed with 25,649 passing tests and 689 failures across 70 files; representative failures were `listen EPERM` for local TCP/Unix sockets and `spawnSync ps EPERM`. All changed-file regressions passed. All 70 failed files passed when rerun with the local socket/process permissions those tests require: 2,404 tests passed and 35 skipped. The initial full run plus this retry covers the complete suite; this is not a claim that the sandboxed run itself passed.

### DIFF
--- a/src/main/compute/compute-auth-owner.test.ts
+++ b/src/main/compute/compute-auth-owner.test.ts
@@ -958,3 +958,33 @@ describe('Compute authentication identity-change application handler', () => {
     })
   })
 })
+
+it('validates and persists SSH configuration without adding a port override', async () => {
+  const host = {
+    ...publicHost(),
+    sshOverrides: { user: 'before' },
+    authentication: { ...publicHost().authentication!, mode: 'ssh_config' as const }
+  }
+  const validateSshConfig = vi
+    .fn<(candidate: ComputeHost) => Promise<void>>()
+    .mockResolvedValue(undefined)
+  const changeAuthentication = vi.fn(async () => host)
+  const owner = new ComputeAuthOwner({
+    repository: authRepository({ get: vi.fn(async () => host), changeAuthentication }),
+    vault: credentialVault(),
+    passwordAdapter: {} as PasswordSshAdapter,
+    validateSshConfig
+  })
+  await owner.changeAuthentication({
+    providerId: host.providerId,
+    expectedRevision: 1,
+    operationId: 'inherit-port',
+    authenticationMode: 'ssh_config',
+    username: 'after'
+  } as ChangeComputeHostAuthenticationRequest)
+  expect(validateSshConfig.mock.calls[0]?.[0]).toMatchObject({ sshOverrides: { user: 'after' } })
+  expect(
+    (validateSshConfig.mock.calls[0]?.[0] as unknown as ComputeHost).sshOverrides?.port
+  ).toBeUndefined()
+  expect(changeAuthentication).toHaveBeenCalledWith(expect.objectContaining({ port: undefined }))
+})
--- a/src/main/compute/compute-auth-owner.ts
+++ b/src/main/compute/compute-auth-owner.ts
@@ -57,7 +57,7 @@ type ChangeComputeHostAuthenticationPersistence = Readonly<{
   requestFingerprint: string
   authenticationMode: ComputeAuthenticationMode
   username: string | undefined
-  port: number
+  port?: number
   identityFile?: string
   ciphertext?: Buffer
   verifiedAt: Date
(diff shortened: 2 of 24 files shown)


ewen-poch:
Fixed SSH-config port inheritance at the Electron IPC boundary in 48aaae62. The validator now accepts an omitted or explicitly undefined port for `ssh_config`, while password authentication still requires a port. Integer/range validation remains enforced by the authentication owner.

The regression tests invoke the installed Electron IPC handler rather than calling the owner directly. Both inherited-port cases failed on the previous implementation with `Invalid arguments for Electron IPC channel: compute:change-authentication`; they now pass. The password-port rejection test also passes. Validation: 163 related tests, Node/sandbox type checks, and lint passed.

The portable CI failure is separate: `App.test.tsx` supplies a `conversationGraph` fixture without its required `messages` array. The failure reproduces locally with coverage enabled: prompt preparation throws when accessing `.find()`, so the expected send never occurs. That test fixture is outside the approved file scope and has not been changed pending confirmation.

ewen-poch:
Fixed the portable-test failure in a46f97ae by adding the required `conversationGraph.messages` array to the background-analysis test fixture in `App.test.tsx`.

The original fixture reproduced the CI failure with coverage enabled: analysis preparation threw on `.find()`, preventing the expected runtime send. With the one-line fixture correction, all 65 tests in `App.test.tsx` pass with coverage instrumentation enabled; lint also passes. The focused coverage run does not enforce repository-wide coverage thresholds.

No production code, runtime state, persistence, or user interaction changed, and no assertions or timeouts were relaxed.

ewen-poch:
Rebased onto current main (`5c5bddc3`). The conflict was in the App test fixture: main now uses `createLinearConversationGraph`, which supplies the complete graph shape. Kept that implementation and dropped the superseded fixture-only commit. The compute-host concurrency, probe, authentication, and IPC fixes are preserved.

Validation: all 1,139 selected regression tests passed across the initial run and the isolated translation-guard rerun; full Node, sandbox, and renderer typechecks passed. The initial coverage run had one translation-scan timeout during concurrent local workloads; all 804 translation tests passed when rerun alone with the original timeout unchanged. No product behavior or test timeout was changed for this rebase.

ewen-poch:
Rebased onto main at `4310366e`; both compute fixes replayed without conflicts. All 1,139 selected regression tests and the full Node, sandbox, and renderer typechecks passed.

Investigated the previous CI failure: the only failed test was `LiteratureLibraryPage > filters and moves the current view to Trash`, which could not find the reference selection checkbox immediately after observing the search call. The module-test and PR gate failures propagated that result. The test passed individually under coverage, and all 106 tests in its file also passed under coverage. An asynchronous rendering race is suspected but has not been reproduced reliably; no speculative test or product change has been committed. This literature test is outside the agreed PR scope.

ewen-poch:
Fixed the literature test's asynchronous selection race. Observing the search call does not guarantee that the returned row has rendered; the test now waits for the selection checkbox before clicking it.

The existing search mock now holds the filtered response until after the request assertion, then resolves on the next timer turn. With the original synchronous lookup, this reproduced the exact CI missing-checkbox failure twice. With the awaited lookup, all 106 tests in the file pass under coverage. ESLint and formatting checks also pass.

Only the authorized test file changed. Production behavior, data models, and timeouts are unchanged.

Issue #2184, opened 2026-09-05
[Bug]: Session Plan creation fails and Artifact reconciliation remains unresolved


ipan118:
### Preflight checklist

- [x] I searched existing issues and this has not been reported yet.
- [x] I'm running a recent version of Open Science.

### What happened?

Generated file finalization failed: Error invoking remote method 'artifacts:reconcile-pending': Error: Native Artifact finalization remains unresolved.

### Steps to reproduce

Session Plan error:
Managed file has no published version.

Artifact error:
Generated file finalization failed:
Error invoking remote method 'artifacts:reconcile-pending':
Native Artifact finalization remains unresolved.

Date/time:
2026-09-04, Asia/Shanghai

Context:
在当前 Open Science Project 中尝试生成 Session Plan，
随后重试发布 Artifact，但 reconciliation 仍未完成。

### Operating system

Windows

### App version

0.25.1

### Provider / model

Codex subscription · gpt-5.6-sol

### Relevant logs or screenshots

```shell
Detected OS: Windows
Agent framework: Codex
Runtime: Electron 39.8.10, Chrome 142.0.7444.265, Node 22.22.1

Runtime log not attached automatically (it can contain local paths and prompts).
Reveal it from Settings → General → Diagnostics and attach after reviewing.
```

VishvakR:
had a dig at this one. it's `inferDurableFinalizationMessageId` in `src/main/artifacts/provenance-finalization-recovery.ts` — it only looks at agent Messages between the run's prompt and the next user Message on the Branch. but a Session Plan writes review feedback as a real user Message inside the still-open turn (`persistUserMessage` in `plan-service.ts`), so the agent Message that actually owns the run's Versions ends up past that boundary and the window matches nothing. the run stays in `unresolvedNativeFinalizationRunIds`, `reconciliation-owner.ts` throws, and since nothing about that state changes the retry button just fails the same way forever. the "Managed file has no published version" bit is downstream of the same thing — the run never gets to `activateFinalizedRun`, so `managedVisibleAt` stays null.

got it to reproduce in the existing recovery integration harness with your exact error, so it's not just a read of the code.

[...]

PR #2281, opened 2026-09-07
fix(artifacts): recover finalization across plan review feedback

VishvakR:
## Problem

Retrying a pending Artifact publication failed with `Native Artifact finalization remains unresolved`, and the generated file stayed unreadable with `Managed file has no published version` (#2184).

`inferDurableFinalizationMessageId` in `src/main/artifacts/provenance-finalization-recovery.ts` closes its candidate window at the next user Message on the Branch, so it only considers agent Messages between the run's prompt and that boundary. A Session Plan persists review feedback as a durable user Message inside the still-open turn (`persistUserMessage` in `src/main/session-plan/plan-service.ts`), so the agent Message that owns the run's Versions sits past the boundary and the window resolves to no candidate at all. The run stays in `unresolvedNativeFinalizationRunIds` and `reconciliation-owner.ts` throws. Nothing about that state changes between attempts, so the retry affordance fails identically forever. The second symptom is the same defect: the run never reaches `activateFinalizedRun`, so `managedVisibleAt` stays null and the lineage head never advances, which is exactly the state `managed-file-versions/service.ts` reports as having no published version.

## Proposed change

Resolve the owner from a unique, complete durable attachment claim before using the prompt-window heuristic. The owner must be an agent Message after the declared prompt on the same Branch, Frame, and Runtime Segment. All run Versions must be attached consistently in Session metadata and every present owner projection. Any partial, competing, or inconsistent claim blocks recovery instead of falling back to an unrelated commentary Message.

A complete claim remains authoritative across an intervening user Message, including an ordinary later prompt; the previous later-turn rejection contract is deliberately updated. This avoids making ownership depend on whether zero, one, or several commentary Messages precede Plan feedback. If no Message claims any run Version, retain the existing single-candidate window before the next user prompt and never infer an unclaimed owner beyond it.

`validateDurableMessageOwnership`, exact Version-set proof, existing ownership-conflict checks, and publication ordering remain in place.

## Scope and non-goals

One private recovery function and its integration tests. No new field, enum, API, IPC method, schema, migration, persisted format, dependency, or UI change.

Recovery still writes existing Version ownership/publication fields, compatibility publication files, and Lineage visibility through the existing owners. Pending historical runs with sufficient evidence may recover through existing retry/startup paths; already-finalized incorrectly bound Versions are not automatically reassigned or migrated.

The original regression failed on main with `Native Artifact finalization remains unresolved` and passes with the fix. Two additional message-order regressions failed consistently on the author's commit `f7d9a54c`: one commentary before feedback caused an empty retry result and a Version bound to that commentary; two commentary Messages left the Version unresolved. The tests use the public coordinator retry API, real SQLite, filesystem storage, and the public managed preview reader. No production test seam was added.

The expanded guard cases also ran red before the follow-up fix. All final checks below ran after the last material edit.

| --- | --- | --- |
| Recovery, ownership guards, later-prompt policy and public preview | `npm test -- src/main/session-persistence/artifact-finalization-recovery.integration.test.ts` | 39 passed |
| Provenance + Session persistence owners/contracts and direct read/preview consumers | Deduplicated module plan below | 39 files / 2,386 passed |
| Main-process types | `npm run typecheck:node` | Passed |
| Lint and formatting | `npm run lint`; Prettier check on both changed files; `git diff --check` | Passed |

<details>
<summary>Reproduce the selected module test run</summary>

```sh
node --input-type=module - <<'JS'
import { createModuleTestPlan, executeModuleTestPlan } from './scripts/ci/module-test-impact.mjs'
const plans = ['artifact_provenance', 'session_persistence'].map(id => createModuleTestPlan(id))
const extras = [
  'src/main/managed-file-versions/service.integration.test.ts',
  'src/main/managed-file-preview.test.ts',
  'src/main/managed-preview-resources.test.ts',
  'src/renderer/src/lib/session-persistence/session-persistence.test.ts'
]
const plan = {
  ...plans[0],
  modules: plans.flatMap(p => p.modules),
  testFiles: [...new Set([...plans.flatMap(p => p.testFiles), ...extras])].sort(),
  reasonChains: [...plans.flatMap(p => p.reasonChains),
    'Artifact publication -> managed read/preview consumers',
    'Session recovery -> renderer persistence consumer']
}
process.exitCode = executeModuleTestPlan(plan, {
  testArguments: ['--maxWorkers=2', '--no-file-parallelism']
})
JS
```

This uses the repository-owned module selector/executor, deduplicates shared tests, and adds the direct publication readers. No global configuration, ownership manifest, or public interface changed, so local selection stays with these two modules and their consumers.

</details>

The shared persistence owner is used by claude-code, opencode, codex-response, and codex-bridge. These checks exercise durable message shapes, not live provider turns; no ACP wire, permission, subscription, or launcher behavior changes. The original Windows packaged interaction has not been replayed locally. Windows/macOS build and E2E breadth is left to the selected PR CI lanes; skipped checks are not claimed as verification. No full local test suite was run.

An independent review confirmed the ownership checks and final Test Impact Set.

## Review focus

The explicit contract is complete durable ownership over the prompt-window heuristic, with no fallback from partial or conflicting evidence. Confirm that the ordinary-later-prompt acceptance, retained unclaimed-run boundary, and exclusion of historical finalized reassignment match the intended recovery policy.

### DIFF
--- a/src/main/artifacts/provenance-finalization-recovery.ts
+++ b/src/main/artifacts/provenance-finalization-recovery.ts
@@ -56,10 +56,9 @@ type ArtifactProvenanceFinalizationRecoveryOptions = {
   >
 }
 
-// Resolves the agent message owning the prepared prompt turn's output. It prefers a single message
-// before the next user prompt on the declared Branch and Runtime Segment, and past that boundary
-// accepts only a message the durable Session already claims; choosing a latest message (or
-// accepting multiple unclaimed candidates) could attach a crashed run to a later turn.
+// Prefer a complete durable claim on the declared Branch and Runtime Segment over the prompt-window
+// heuristic: an in-turn user feedback Message can separate commentary from the real output owner.
+// Only an entirely unclaimed run may fall back to a single Message before the next user prompt.
 const inferDurableFinalizationMessageId = (
   session: PersistedChatSession,
   context: PreparedArtifactFinalizationContext,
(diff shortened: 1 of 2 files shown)


ewen-poch:
Thanks, @VishvakR — the diagnosis and the regression test were really helpful. I could reproduce the failure, and your fix gets the original case working.

While checking a couple of nearby message sequences, I found two gaps: one commentary message before the feedback could take ownership of the file, and two commentary messages before it still left finalization unresolved. Both reproduced consistently through the public retry API with real SQLite/file storage.

I've pushed a follow-up commit to your branch. It checks for a unique, complete durable attachment claim before using the message-window heuristic, while keeping partial or competing claims blocked. With no claim, recovery still stays inside the original prompt window. This also makes the treatment of a complete claim after a later user message explicit; we aren't adding a new message type or changing the stored format.

The two regressions now pass, along with all 39 recovery tests and the selected module/consumer checks (39 files, 2,386 tests). Node typechecking, lint, and formatting also pass. No automatic reassignment of previously finalized files is included. The updated PR checks will cover the remaining platform validation.

Thanks again for digging into this and bringing a concrete repro — it made these remaining cases much easier to pin down.

PR #2282, opened 2026-09-07
fix(conversation-export): preserve reviewed content and export outcomes

ewen-poch:
A conversation export could succeed with an older saved answer than the one reviewed in the dialog, rewrite executable Markdown examples, or report the wrong PDF outcome. Reopening the same conversation could also retain selected turns that no longer exist.

- Flush pending session persistence before exporting and compare a SHA-256 digest of the reviewed export projection with the durable content. Save failures remain visible; changed content requires renewed review. Main continues to generate documents from its own stored Session.
- Preserve message whitespace and Markdown code tokens while retaining the filtering of actual reasoning blocks.
- Apply one PDF deadline to page loading, font readiness, and printing. Destroy the dedicated window on timeout and prevent late preparation tasks from continuing to print. Time spent choosing a Save As destination is excluded.
- Log temporary-directory cleanup failures through the central logger without replacing a successful publication or the original generation error.
- Clear turn selections on reopening, and derive counts, select-all state, button availability, and submitted IDs from the currently available turns. Canceling Save As still preserves the current selection.

The only request-contract addition is the optional `expectedContentHash` precondition, always supplied by the preview dialog. There are no persisted Session fields, database migrations, new dependencies, or global conversation locks. The existing reviewed error translation is shared through the common namespace in all eight translated locales.

Validation:

- Eleven behavioral regression cases failed consistently in two runs against the original production code, with failures matching the reported output, waiting, cleanup, and selection problems; they pass with the fixes.
- Export projection, real dialog/persistence/service integration, host-frame consumers, locale guards, and central logger checks pass after rebasing onto current main.
- Type checking, changed-file ESLint, formatting, and diff whitespace checks pass.
- Full `npm test` was run. The sandboxed run passed 25,701 tests but failed local socket, port, and subprocess-dependent suites, and caught a direct-console diagnostic that this patch subsequently corrected. Retrying all 71 failed files outside that sandbox passed 69 files. The runtime installer's 200 ms diagnostic timeout case then passed in isolation with both original and fixed production code. Three Micromamba package-cache integration cases still fail with the same filesystem access denial on the unmodified baseline; this PR does not change their sandbox policy.

The cleanup regression uses the real filesystem and file publisher with deterministic substitute print bytes. These tests do not claim real Electron IPC or PDF visual-rendering validation.

CI follow-up: rebased onto current main and fixed the Home background-analysis test race. A delayed durable claim reproduces the original zero-send assertion failure; the test now waits for the actual runtime dispatch rather than one zero-delay timer turn. App and export regression suites pass together (119 tests), with formatting and ESLint passing.

### DIFF
--- a/src/main/session-persistence/conversation-export.test.ts
+++ b/src/main/session-persistence/conversation-export.test.ts
@@ -344,6 +344,84 @@ describe('conversation export service', () => {
     expect(removeDirectory).toHaveBeenCalledWith('/tmp/open-science-conversation-export-test')
   })
 
+  it.each(['page loading', 'font readiness'])(
+    'bounds %s with the PDF generation deadline',
+    async (stage) => {
+      vi.useFakeTimers()
+      let release!: () => void
+      const blocked = new Promise<void>((resolve) => {
+        release = resolve
+      })
+      const dependency = stage === 'page loading' ? loadFile : executeJavaScript
+      dependency.mockReturnValue(blocked)
+      const outcome = vi.fn()
+      const promise = createService({
+        exportLimits: {
+          maxMessages: 10,
+          maxImageBase64Bytes: 1_000_000,
+          maxHtmlBytes: 1_000_000,
+          pdfPrintTimeoutMs: 20
+        }
+      })
+        .exportConversation({ projectId: 'project-1', sessionId: 'session-1', format: 'pdf' })
+        .then(outcome, outcome)
+      try {
+        await vi.advanceTimersByTimeAsync(120)
+        expect(dependency).toHaveBeenCalledOnce()
+        expect(outcome).toHaveBeenCalledWith(
+          expect.objectContaining({ message: expect.stringMatching(/timed out/i) })
+        )
+        expect(destroy).toHaveBeenCalledOnce()
+        expect(removeDirectory).toHaveBeenCalledOnce()
+        expect(printToPDF).not.toHaveBeenCalled()
+        release()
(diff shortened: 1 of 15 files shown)


ewen-poch:
Confirmed the blockquote case through the public export projection: `> ~~~xml\n> <think>literal</think>\n> ~~~` exports with the literal tag removed. The added local regression fails on the current implementation for that exact content loss. A simple list-indented fence passes, so the reproduced defect is specifically the unsupported nesting context, rather than all list fences.

The fix needs to preserve Markdown container context while locating code spans in the original source, without reserializing code or exposing actual reasoning blocks. Implementation is paused for scope confirmation before introducing additional parser bookkeeping. The reproduction remains local and has not been pushed as a failing test.

ewen-poch:
Preserved nested fenced code by using the existing Markdown block lexer and mapping container-stripped lines back to their original source positions. Sanitization now protects code inside blockquotes and lists without reserializing its contents. Actual reasoning remains filtered, including an outer reasoning block containing a fenced example with a literal closing tag.

The blockquote regression failed on the previous implementation with the literal tag removed. The updated tests cover nested quotes, lists containing quotes, backtick and tilde fences, CRLF, Unicode, unfinished fences, and reasoning around/containing code. All 77 export and host-frame consumer tests pass; formatting and ESLint pass. Changes are limited to the existing shared sanitizer and its test file, with no persistence, schema, architecture, or interaction changes.

PR #2284, opened 2026-09-07
fix(preview): restore tab action feedback and focus

ewen-poch:
## Problem

Preview tab downloads, clipboard writes, and artifact saves swallowed API failures, so the menu closed without visible feedback. Closing the focused tab left DOM focus on the body. A pending PDF reading-context request still appeared executable when its tab menu was reopened.

## Proposed change

- Propagate file-action rejection to the shared menu provider. A preview-local error carries the original operation to a panel-level, dismissible ErrorNotice with existing translated titles, file identity, diagnostics, and Retry. Inactive or subsequently closed tabs retain a usable retry within their original project; changing projects clears the notice, drops late failures from other projects, and blocks stale retry clicks. Retry progress belongs to its failure and cannot block another project; retry clicks and the corresponding tab menu command are disabled while awaiting completion. Download cancellation remains silent.
- Restore focus to the original tab, then the active remaining tab, then the existing composer focus event. Preserve PDF commands' composer handoff.
- Pass the existing PDF pending boolean to the tab action binding's disabled condition.

The initial full CI run exposed a pre-existing App conversation-graph fixture failure, also reproduced at the original base. Main now contains the complete fixture correction, so the duplicate test-only commit was dropped when resolving the branch conflict.

## Scope and non-goals

Four preview renderer source/test files. No shared-provider API change, new dependency, global task manager, database field, new business enum, persisted format, migration, or historical compatibility path. Only transient failure/retry UI state (including the originating project and tab IDs) and a derived pending property are added. Existing save and PDF persistence operations are unchanged.

Before production edits, five public-behavior cases in the existing real PreviewPanel/Radix harness failed identically in two runs: three missing failure titles after confirmed API calls/menu dismissal, body focus after confirmed store selection repair, and missing aria-disabled while a real hook request remained pending. No production test seam was added. All five now pass.

After the final retry-menu fix and rebase, the targeted Vitest files below were rerun together (1,161 passed), the workspace module was rerun separately (825 passed), and typecheck/lint passed:

| --- | --- | --- |
| Failure visibility on inactive tabs, retry after tab removal, retry failure/success and deduplication, cancellation, focus restoration/fallback and PDF handoff/pending | `npx vitest run src/renderer/src/pages/workspace/PreviewPanel.test.tsx src/renderer/src/pages/workspace/preview-tab-actions.test.ts` | 77 passed |
| Shared menu contracts, ErrorNotice, content-menu consumers, preview store, translations | `npx vitest run src/renderer/src/components/action-menu src/renderer/src/components/error-notice.test.tsx src/renderer/src/pages/workspace/PreviewFileSurface.test.tsx src/renderer/src/pages/workspace/PreviewFileSurface.freshness.test.tsx src/renderer/src/stores/preview-workbench-store.test.ts src/renderer/src/i18n/resources.test.ts` | 1,019 passed |
| App startup/background analysis fixture | `npx vitest run src/renderer/src/App.test.tsx` | 65 passed; failed identically on the exact base before the fixture correction |
| Workspace composition and representative consumers | `npm run test:module -- workspace_page` (plan inspected with `--explain`) | 825 passed across 30 files |
| Source lint | `npm run lint` | Passed |
| Patch whitespace | `git diff --check` | Passed |

Manual browser validation used the real PreviewPanel and styling in an isolated local fixture with injected file/API data: actual right-click, visible inactive-tab failure, successful retry dismissal, disabled pending PDF action, project-switch notice removal, and a disabled artifact-save menu command during retry. Screenshots were captured locally for the maintainer; this is not full Electron E2E coverage.

`npm run test:affected:explain -- --base origin/main --head HEAD` reports a full fallback because PreviewPanel.test.tsx has no module owner in the current manifest. The maintainer explicitly requested module-local validation rather than the full local suite; the final PR CI plan remains authoritative. Main/preload, iframe coordinate handling, persistence, and path normalization are unchanged, so no local native/platform or migration lane was run. Independent review identified project-switch leakage in the first implementation. Four additional public-behavior regressions reproduced retained notices, late rejections, stale retry clicks, and retry-progress leakage before their fix; all now pass. A subsequent independent review identified executable menu commands during a pending retry. Three additional real-menu regressions failed identically twice before the fix and now pass. CI and independent review of the final retry correction remain pending. The maintainer has explicitly requested that this PR not be merged.

## Review focus

Original-operation capture for retry, preservation of PDF composer focus, and visibility of failures from inactive tabs. Confirm the final behavior-to-check mapping before marking the change verified.

### DIFF
--- a/src/renderer/src/pages/workspace/PreviewPanel.test.tsx
+++ b/src/renderer/src/pages/workspace/PreviewPanel.test.tsx
@@ -1697,6 +1697,28 @@ describe('PreviewPanel', () => {
   })
 
   it.each(['download', 'copy-path', 'save-as-artifact'])(
+    'clears the failure after %s succeeds from the menu',
+    async (command) => {
+      const operation = vi.fn().mockRejectedValueOnce(new Error('File operation denied'))
+      Object.assign(navigator, { clipboard: { writeText: operation } })
+      if (command === 'download') window.api.saveManagedFile = operation
+      if (command === 'save-as-artifact') window.api.uploads.stageLocalPath = operation
+      usePreviewWorkbenchStore.getState().upsertAndActivateItem(createFileItem({ source: 'local' }))
+      await renderPanel()
+
+      await openTabContextMenu(0)
+      await clickMenuCommand(command)
+      expect(container.querySelector('[data-testid="preview-tab-action-error"]')).not.toBeNull()
+
+      operation.mockResolvedValue({ saved: true })
+      await openTabContextMenu(0)
+      await clickMenuCommand(command)
+      expect(operation).toHaveBeenCalledTimes(2)
+      expect(container.querySelector('[data-testid="preview-tab-action-error"]')).toBeNull()
+    }
+  )
+
+  it.each(['download', 'copy-path', 'save-as-artifact'])(
     'disables the same menu command while %s is being retried',
     async (command) => {
       let finish!: () => void
--- a/src/renderer/src/pages/workspace/PreviewPanel.test.tsx
+++ b/src/renderer/src/pages/workspace/PreviewPanel.test.tsx
@@ -1696,6 +1696,47 @@ describe('PreviewPanel', () => {
     }
   })
 
+  it.each(['download', 'copy-path', 'save-as-artifact'])(
+    'disables the same menu command while %s is being retried',
+    async (command) => {
+      let finish!: () => void
+      const operation = vi
+        .fn()
+        .mockRejectedValueOnce(new Error('Permission denied'))
+        .mockImplementation(
+          () =>
+            new Promise<void>((resolve) => {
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed the stale failure state: a successful file action from the tab menu now clears only the matching error for that file, command, and active project. Added regression coverage for download, copy path, and save as artifact.

ewen-poch:
Addressed concurrent retries by tracking pending operations with a project, tab, command, path, and version key. Completing one retry now removes only that operation's key, so every in-flight menu command remains disabled. Added a regression test with two simultaneous artifact-save retries.

ewen-poch:
Normalized the optional pending-key lookup to false so the disabled predicate always returns a boolean. The renderer typecheck and affected tab-action tests now pass.

PR #2291, opened 2026-09-07
feat(renderer): add source/rendered toggle to mermaid block actions

roxi3906:
## What

Adds a source/rendered toggle button to the mermaid block's top-right action bar (alongside download / copy / fullscreen). Clicking it swaps the rendered diagram for the fence's source in a scrollable `<pre>`; clicking again restores the diagram. The SVG stays mounted (`display: none`) while the source is shown, so toggling back costs no re-render.

Related: pairs nicely with #2142 — when both land, the height change from toggling is animated instead of jumping.

## How

Streamdown keeps the mermaid source only in React props — nothing in the DOM carries it. Two pieces:

1. **Source capture** (`mermaid-runtime.ts`, the app's single mermaid import site): wraps the upstream plugin so every `render(id, source)` call records the source in a small bounded registry (`mermaid-source-registry.ts`, 100-entry LRU) and stamps the render id onto the SVG root as `data-mermaid-render-id`. Sources are remembered even when rendering fails.

2. **Toggle installer** (`mermaid-view-toggle.ts`, wired into `installStreamdown`, same pattern as the existing `installCodeLanguageBadges`): a MutationObserver decorates each `mermaid-block-actions` bar inside `.agent-markdown-root` once the rendered SVG has exposed its source. The button mirrors streamdown's own action-button styling, reflects state via icon (code/eye), `aria-pressed`, and `title`/`aria-label`.
   - A re-rendered diagram (retry, streaming) refreshes a visible source view to the latest chart.
   - The button is disabled while an error panel replaces the diagram (the error surface already shows the source).
   - Labels reuse the existing `View source` key and add `View diagram`, translated for all eight locales and updated live on `languageChanged`.
   - Uninstall restores any visible source view and removes the injected buttons.

## Testing

- `mermaid-runtime.test.ts` (5): plugin identity passthrough, source recording + SVG stamping, source retained on render failure, registry LRU eviction and recency refresh.
- `mermaid-view-toggle.test.ts` (7): injection only after a sourced diagram exists, toggle round-trip, source refresh on re-render, disabled state on error panel, live translation on language change (en/de/zh-Hans), uninstall cleanup.
- i18n guard suite (`resources.test.ts`): 804 passed — placeholder parity and per-locale coverage for the new key.
- Full `vitest run`: 26,455 passed; the 46 failures (`kernel-executor`, `package-cache-sandbox`, `network-enforcement`, `literature-review kernel`) reproduce identically on an untouched `main` checkout — environment issues (system python3 is a broken Xcode CLT shim on this machine, sandbox network policy), unrelated to this change.
- `npm run typecheck` and ESLint clean.

### DIFF
--- a/src/renderer/src/components/streamdown/install-streamdown.ts
+++ b/src/renderer/src/components/streamdown/install-streamdown.ts
@@ -12,6 +12,7 @@ import {
   STREAMDOWN_TABLE_FULLSCREEN_SELECTOR
 } from './dom-selectors'
 import { resolveLanguageIconPath } from './language-icons'
+import { installMermaidViewToggle } from './mermaid-view-toggle'
 
 const saveBlobFile = (request: SaveBlobFileRequest): Promise<SaveBlobFileResult> =>
   window.api.saveBlobFile(request)
@@ -808,7 +809,8 @@ const installStreamdown = (): (() => void) => {
       installFullscreenDialogAdapter(),
       installTableActions(),
       installTableFullscreenFix(),
-      installCodeLanguageBadges()
+      installCodeLanguageBadges(),
+      installMermaidViewToggle()
     )
   }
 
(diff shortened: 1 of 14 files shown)


roxi3906:
Folded into #2142 as a single mermaid UX PR (cherry-picked as 985cb3894).

PR #2304, opened 2026-09-07
fix(office-preview): preserve worksheet visibility and session errors

ewen-poch:
## Problem

A normal blank XLSX worksheet is discarded, leaving first paint pending or causing a hidden worksheet to be shown instead. Spreadsheet errors after first paint never reach the existing retry card. DOCX internal bookmark links lose activation in the adapter.

## Proposed change

- Preserve blank worksheets and source identity in the existing spreadsheet patch. Select only visible sheets; zero/no-visible worksheets render an explicit empty state and complete readiness.
- Keep the spreadsheet error channel until disposal. Reuse `error` / `RENDER_FAILED`, cleanup and retry; reject stale runtime reports.
- Activate DOCX bookmarks only within the preview container, supporting pointer/Enter and retaining keyboard focus. External navigation remains disabled.

## Scope and non-goals

No persistence, migrations, historical-data compatibility layer, or new state enum. The transient runtime-start message gains an optional validated `locale`; omitted values default to English. The host supplies its active locale and the isolated runtime initializes translation before rendering. One optional callback is added to the local renderer options. Both Hidden=1 and Hidden=2 stay hidden. XLS controls pass, without generalizing to every XLS variant. PPTX partial-error recovery and local row-window retries are out of scope.

Before fixes, two consecutive runs of the real-parser regression suite yielded exactly four report-matching failures and three passing controls. Tests use generated XLSX/XLS/DOCX packages, real package validation and installed parsing/rendering code; only Worker transport and the existing canvas observation fixture are substituted in jsdom. No production test seam was added.

| --- | --- | --- |
| Blank/hidden/zero-sheet XLSX; legacy controls; post-ready failure; DOCX activation and containment | `office-behavior-regressions.test.ts` | Pass |
| Adapter startup/abort/cleanup and bounded caches | `office-renderers.test.ts`, `spreadsheet-cache-policy.test.ts`, `office-package.test.ts` | Pass |
| Session errors, stale events, IPC/host retry contracts | renderer `office-preview/`, main `office-preview/`, shared `office-preview.test.ts`, `office-preview-lease.test.ts`, `OfficePreview.test.tsx`, `locale.test.ts` | Pass |
| Host and targeted-reviewer consumers | `PreviewFileContent.test.tsx`, `PreviewFallback.test.tsx`, `PreviewToolContent.reviewer.integration.test.tsx` | Pass |
| New empty-state copy in all eight translated locales | `npx vitest run src/renderer/src/i18n/resources.test.ts` | 804 pass |
| Renderer/shared-contract types and lint | `npm run typecheck:node`; `npm run typecheck:web`; `npm run lint`; Prettier check | Pass |
| Reproducible dependency patch | Apply patch to clean vendor and compare to tested files | Pass |

The explicit Office/consumer slice above is 18 test files / 229 passing tests. No complete local `npm test` run; full/platform validation is delegated to PR CI as requested. The shared startup contract changed, so both process typechecks and host/runtime contract tests are included. No main-process logic changed, but its supervisor/protocol contracts are included because they consume runtime error states. Notebook/ACP execution paths are not modified.

Built Chromium integration checks use the actual bundled Worker and canvas renderer: blank and blank+hidden workbooks reach ready, no-visible sheets show empty copy, controlled Worker error reaches the existing retry card and Retry restores ready, and pointer/Enter bookmark activation retains target focus without URL navigation. This is a focused harness using real runtime and fallback components, not a full Electron screenshot or an OS-level Worker crash. The real-browser check caught and fixed Chromium losing focus when tabindex was removed immediately. A separate Chinese all-hidden-workbook capture confirms runtime initialization without any harness-side translation initialization. Locale regressions failed before the follow-up fix and pass after it.

## Review focus

Review blank-sheet metadata/visibility, first-paint versus session error races and disposal, and bookmark containment/focus cleanup. CodeGraph supplied baseline relationships; its index belongs to the main checkout, so final worktree relationships were verified against source and the listed tests. Independent PR review and exact-head CI remain the final authority.

The initial review correctly identified isolated-runtime translation initialization; the optional startup locale addresses it. Its toolbar finding does not reproduce: visible-sheet startup returns before the empty-branch toolbar hide. An added regression checks computed toolbar visibility and switches to a second visible sheet through the actual renderer.

The second review raised premature first paint on visible workbooks. A controlled first-window transport hold confirms the renderer remains pending, keeps the toolbar visible, and never inserts empty copy until the actual data window arrives. This passes without a production-code change. The patch now includes the previously omitted `return` context, so the no-visible-sheet branch is directly reviewable. After resolving translation-only conflicts with main and this test addition, the combined Office/consumer/i18n run passes 19 files / 1,033 tests; web types, lint and patch reproducibility were rechecked.

The third review assumed `useTranslation().i18n` has stable identity across language changes. The installed react-i18next implementation returns a new language-specific wrapper. A ready-host regression changes the actual i18next language, verifies the same iframe receives a new start locale, and confirms reattachment without reopening the host session. It passes without changing production code. This guards the observed dependency behavior rather than assuming older library semantics.

### DIFF
--- a/patches/@file-viewer+renderer-spreadsheet+2.2.3.patch
+++ b/patches/@file-viewer+renderer-spreadsheet+2.2.3.patch
@@ -213,7 +213,7 @@ index 3378e4a7..55292dc1 100644
          cached.loadingWindows.clear();
          virtualState = cached;
          errorMessage = '';
-@@ -1418,7 +1506,7 @@ const renderFileViewerSpreadsheet = async (buffer, target, type, context) => {
+@@ -1418,26 +1506,49 @@ const renderFileViewerSpreadsheet = async (buffer, target, type, context) => {
      };
      controller.onWorkerEvent('sheets', ({ sheets: list }) => {
          sheets = list;
@@ -222,7 +222,9 @@ index 3378e4a7..55292dc1 100644
          if (firstSheet) {
              sheetIndex = firstSheet.id;
              renderChrome();
-@@ -1428,16 +1516,39 @@ const renderFileViewerSpreadsheet = async (buffer, target, type, context) => {
+             startSheetSession();
+             scrollActiveSheetIntoView();
+             return;
          }
          sheetInitializing = false;
          loadingState = false;
--- a/patches/@file-viewer+renderer-spreadsheet+2.2.3.patch
+++ b/patches/@file-viewer+renderer-spreadsheet+2.2.3.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@file-viewer/renderer-spreadsheet/dist/spreadsheet.js b/node_modules/@file-viewer/renderer-spreadsheet/dist/spreadsheet.js
-index 3378e4a..d779036 100644
+index 3378e4a7..55292dc1 100644
 --- a/node_modules/@file-viewer/renderer-spreadsheet/dist/spreadsheet.js
 +++ b/node_modules/@file-viewer/renderer-spreadsheet/dist/spreadsheet.js
 @@ -4,6 +4,8 @@ import { buildRows, clampWindowStart, collectWindowStarts, createEmptyVirtualSta
@@ -39,6 +39,18 @@ index 3378e4a..d779036 100644
      let scrollDirection = 1;
      let lastScrollY = 0;
      let sheetSessionId = 0;
+@@ -636,9 +652,9 @@ const renderFileViewerSpreadsheet = async (buffer, target, type, context) => {
+     const getActiveSheet = () => sheets.find(sheet => sheet.id === sheetIndex);
+     const getSheetTabs = () => {
+         const visible = sheets.filter(sheet => !sheet.hidden);
+-        return visible.length ? visible : sheets;
++        return visible;
+     };
+-    const getActiveSheetId = () => { var _a; return sheetIndex !== null && sheetIndex !== void 0 ? sheetIndex : (_a = sheets[0]) === null || _a === void 0 ? void 0 : _a.id; };
++    const getActiveSheetId = () => { var _a; return sheetIndex !== null && sheetIndex !== void 0 ? sheetIndex : (_a = sheets.find(sheet => !sheet.hidden)) === null || _a === void 0 ? void 0 : _a.id; };
+     const getHostHeight = () => tableHost.clientHeight || 0;
+     const showBlockingLoading = () => !errorMessage && !hasInitialWindow && (loadingState || sheetInitializing);
+     const showStreamingLoading = () => !showBlockingLoading() &&
 @@ -905,7 +921,7 @@ const renderFileViewerSpreadsheet = async (buffer, target, type, context) => {
              endRow,
              direction: scrollDirection,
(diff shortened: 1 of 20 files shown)


ewen-poch:
The review points have been checked against the current implementation and behavior tests:

- **Runtime translation initialization — addressed.** The host passes an optional validated locale and the isolated runtime initializes i18n before rendering. Missing locale defaults to English. Startup-language assertions failed before the fix and pass afterward; the built browser harness also shows the Chinese empty state. This adds no persisted data or lifecycle enum.
- **Visible-sheet toolbar and premature empty readiness — not reproduced.** The visible-sheet branch starts its sheet session and returns before the empty-state code. The patch now includes that return in its context. The [first-window test](https://github.com/aipoch/open-science/blob/fef1f012e941df0b5c260abf01217129c6d0c8e1/src/renderer/src/pages/workspace/previews/office-behavior-regressions.test.ts#L175) holds the actual parser's window response and confirms pending readiness, no empty-state DOM, and a visible toolbar; releasing the response completes first paint with the expected data. The adjacent test verifies switching between visible worksheets.
- **Language changes after attachment — not reproduced with the locked dependency.** react-i18next 16.6.6 returns a language-specific i18n wrapper from `useTranslation`; its wrapper identity changes when the language changes. The effect dependency therefore changes. The [ready-host regression](https://github.com/aipoch/open-science/blob/fef1f012e941df0b5c260abf01217129c6d0c8e1/src/renderer/src/pages/workspace/previews/renderers/OfficePreview.test.tsx#L301) changes the real i18next language after readiness and verifies that the same iframe receives a start message with the new locale, with a second attachment and no new host open call. No production change is warranted for the reported failure; the behavior test guards this dependency contract.

At `fef1f012`, the local Office/consumer/i18n selection passes 1,033 tests, affected-process typechecks and lint pass, and PR Gate—including the full portable suite and Windows/macOS E2E—passes. The static review still reports `needs changes`; the evidence above explains why its remaining language-lifecycle finding is not being adopted.

ewen-poch:
Rebased onto `main` at `3cc81113`; the updated head is `c2183f2e`. Resolved the eight locale-catalog conflicts by retaining both main's new entries and the Office empty-state translations. The Office production code, dependency patch, and regression tests are unchanged from the previous head.

After the rebase, the targeted Office/consumer/i18n selection passes all 1,033 tests. Node and web typechecks, lint, formatting, and diff checks also pass. No additional architecture, interaction, state, or persistence changes were introduced.

ewen-poch:
Rebased onto `main` at `5b110b6c`; the new head is `44f8b4e7`. Resolved the eight locale-catalog conflicts by preserving both sets of translations. The targeted Office/consumer/i18n suite passes 1,033 tests after the rebase; node/web typechecks, lint, formatting, and diff checks also pass.

The previous [Windows E2E failure](https://github.com/aipoch/open-science/actions/runs/34118752412/job/101731699550) occurred in `message-tool-layout-stability.spec.ts` while `configureFakeAgent` was restarting Electron, before the layout assertion. The reported error is `Electron E2E graceful close did not finish within 10000ms`, thrown from `e2e/fixtures/electron-app.ts:112` via `closeForCleanup` / `close` / `configureFakeAgent`. This identifies the failing phase, not its root cause; this rebase is not claimed to fix the timeout. No E2E fixture or shutdown behavior was changed, as that would extend beyond this PR's current file scope.

PR #2322, opened 2026-09-07
fix(projects): preserve drafts and authoritative project state

ewen-poch:
Project saves could accept edits that were discarded when an older submission closed the dialog, and delayed update replies could roll the renderer cache back after a newer list refresh. Changing one project's Agent Context also retired independent runtimes serving other projects, while initial loading incorrectly displayed the empty-project message.

This change:
- Disables all three shared create/edit form inputs while saving; failures retain the draft and restore editing.
- Applies accepted list snapshots to the existing per-project projection generations, preserving request ordering without relying on timestamps for pin/archive state.
- Passes the changed project ID through the production callback and retires only runtime generations owning sessions in that project. Shared generations retain the existing conservative retirement behavior.
- Uses the existing loaded flag and translated loading text before displaying the empty-project state.

No database schema, persistent fields, dependencies, or runtime architecture changes.

Validation:
- Added/extended public-boundary regression tests before production changes. Two runs on unfixed code produced the same seven failures matching PH-01 through PH-04.
- Covers create/edit pending and failed submissions, both list/update return orders, lifecycle projection control, management fields, context replacement/clearing, independent-runtime continuity, and initial loading-to-empty transition.
- Final focused suite: 1,023 tests passed across seven files, including the i18n guard.
- Full suite: `npm test -- --maxWorkers=4` — 27,012 passed, 246 skipped (1,494 files passed, 17 skipped). Run outside the outer agent sandbox because local HTTP/RPC tests require loopback listeners.
- Node, sandbox, and renderer typechecks passed; changed-file ESLint and diff checks passed.

### DIFF
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -3295,6 +3295,55 @@ describe('AcpRuntimeCoordinator', () => {
     }
   )
 
+  it.each([
+    { hasPriorOwner: true, changedProjectId: 'project-a' },
+    { hasPriorOwner: false, changedProjectId: 'project-a' },
+    { hasPriorOwner: false, changedProjectId: 'project-b' }
+  ])(
+    'scopes pending adoption retirement to $changedProjectId (prior owner: $hasPriorOwner)',
+    async ({ hasPriorOwner, changedProjectId }) => {
+      const adoption = createDeferred<void>()
+      const created: ReturnType<typeof createFakeRuntime>[] = []
+      const coordinator = new AcpRuntimeCoordinator((callbacks) => {
+        const fake = createFakeRuntime({
+          frameworkId: 'claude-code',
+          sessionIds: [`session-${created.length}`],
+          callbacks,
+          ...(created.length === 0 ? {} : { beforeResume: () => adoption.promise })
+        })
+        created.push(fake)
+        return fake.runtime
+      })
+      const sessionId = hasPriorOwner
+        ? (await coordinator.createSession({ projectId: 'project-a' })).sessionId
+        : 'restored-session'
+      const resume = coordinator.resumeSession({
+        sessionId,
+        projectId: 'project-a',
+        cwd: '/workspace',
+        previousFrameworkId: 'claude-code',
+        agentTarget: {
+          frameworkId: 'claude-code',
+          providerId: 'incoming-provider',
+          model: 'model',
+          reasoningEffort: 'high'
+        }
+      })
+      await vi.waitFor(() => expect(created[1]?.resumeSession).toHaveBeenCalledOnce())
+      await coordinator.requestProjectAgentContextReload(changedProjectId)
+      adoption.resolve()
+      if (changedProjectId === 'project-a') {
+        expect.soft(created[1].requestRetirement).toHaveBeenCalledOnce()
+        await expect(resume).rejects.toThrow('adoption was superseded')
+      } else {
+        expect(created[1].requestRetirement).not.toHaveBeenCalled()
+        await expect(resume).resolves.toMatchObject({ sessionId })
+        await coordinator.sendPrompt({ sessionId, text: 'Continue without another resume' })
+        expect(created[1].sendPrompt).toHaveBeenCalledOnce()
+      }
+    }
+  )
+
   it('publishes prompt ownership only from the runtime that currently owns the session', async () => {
     const retirement = createDeferred<void>()
     const created: ReturnType<typeof createFakeRuntime>[] = []
--- a/src/main/acp/runtime-coordinator.test.ts
+++ b/src/main/acp/runtime-coordinator.test.ts
@@ -3344,6 +3344,60 @@ describe('AcpRuntimeCoordinator', () => {
     }
   )
 
+  it.each(['project-a', 'project-b'])(
+    'scopes unpublished session creation retirement to %s',
+    async (changedProjectId) => {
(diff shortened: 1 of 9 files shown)


ewen-poch:
Confirmed the pending-adoption race with regression tests through the existing coordinator API. With resume held before attachment, editing the project's context does not retire the incoming runtime, and the pending resume still commits ownership. This reproduces both with an existing owner and during a cold restore.

The proposed fix is to retain the request's project ID alongside the incoming runtime in the existing pendingSessionAdoptions map and include matching pending runtimes in project-context retirement. The existing superseded-adoption guard can then reject the stale resume before ownership commits. Inferring the project only from live sessions misses cold restores before attachment.

This adds transient in-memory metadata, without persistence, schema changes, new UI controls, or changes outside runtime-coordinator.ts and its existing test file. The affected pending resume would use the existing superseded/retry behavior; independent projects remain unaffected. Implementation is paused for owner approval because adding state requires confirmation. No production change has been pushed.

ewen-poch:
Fixed in ff46523. Pending session adoptions now retain the request's project ID (falling back to the existing owner's project), and project-context reload includes matching incoming runtimes before they publish a live session. The existing superseded-adoption guard prevents a stale resume from committing ownership.

The regression failed twice before the fix for both an existing-owner handoff and a cold restore. Both now pass. An additional control verifies that changing another project does not retire the pending runtime and that its session can prompt immediately after adoption. All 138 coordinator tests, node/sandbox typechecks, and changed-file ESLint pass.

Changes are limited to runtime-coordinator.ts and its existing test file. The owner approved the additional transient project metadata; there are no persistence, schema, or new UI changes.

ewen-poch:
Confirmed the creation race through the coordinator's public createSession and requestProjectAgentContextReload APIs. Holding creation before it publishes a session, then reloading the same project's context, leaves the runtime unretired; creation completes and the next prompt is admitted without a resume.

The minimal proposal is to register each in-flight creation with its project ID and runtime, remove that registration on settlement, and include matching runtimes in project-context retirement. This would reuse the existing retirement/resume behavior and keep unrelated projects unaffected. Changes would remain in runtime-coordinator.ts and its existing test file, with no persistence, database/schema changes, or new UI controls.

The pending-adoption fix in ff46523 is already pushed and validated. The creation fix is paused because it needs an additional transient state registration beyond the previously approved pending-adoption metadata. A failing regression is prepared locally; no production change for this new finding has been pushed.

ewen-poch:
Fixed in 3711c66. Each pending create now registers its project ID and runtime before provider creation starts; the registration is removed in finally on success or failure. Matching project-context changes retire these runtimes even before a live session is published, so the existing prompt guard requires a fresh resume. Shutdown also clears the registrations.

The coordinator regression failed twice before the fix. All 141 coordinator tests now pass, including controls for unrelated project updates and cleanup after failed creation. Node/sandbox typechecks, changed-file ESLint, and diff checks pass.

The owner approved this additional transient registration. Changes remain limited to runtime-coordinator.ts and its existing test file, with no persistence, schema, or new UI changes.

ewen-poch:
Confirmed with a public store regression, reproduced twice: an update begins, a later list read returns the pre-commit row, then the update commits and its response is suppressed by the list's projection generation. The cache remains on the old row.

Allowing every completed update response to override the overlapping list would reintroduce PH-02, where that list already contains newer authoritative state than the delayed reply. The request start order alone cannot distinguish these two database execution orders.

The proposed reconciliation is to preserve stale-response suppression and obtain a fresh authoritative read after a successful mutation response is suppressed, rather than applying the delayed row directly. This avoids assuming either the list or the delayed reply must be freshest. The intent is to reuse the existing store loading/error path, without schema, persistence, or new UI controls, and keep edits within the existing store and test files.

Implementation is paused for owner confirmation because this refines the previously agreed list/update ordering design. The failing regression is prepared locally; no production change for this finding has been pushed.

ewen-poch:
Fixed in 45aa7e7. A successful update/archive response that loses projection ownership now triggers a fresh authoritative list read if a list request started while that mutation was pending. Deletion follows the same rule and refreshes its existing cleanup projection as well. Lifecycle-only supersession without an overlapping list does not add a read.

The delayed response is never applied directly, preserving PH-02's protection against rolling newer list contents back. A failed reconciliation retains the accepted projection and successful mutation result, and exposes the existing load-error/Retry path instead of misreporting the write as failed. No new store fields, persistence, schema, or UI controls were added.

Regressions reproduce the pre-commit list race for update, archive, and delete. They pass alongside the original newer-list protection, lifecycle control, cleanup refresh, and reconciliation-failure coverage. The combined seven-file suite passes all 1,033 tests; renderer typechecking, changed-file ESLint, and diff checks pass.

ewen-poch:
Rebased onto the latest main (5e4ec0e2) without conflicts; the rebased head is 614f1e44. The seven-file project regression suite passes all 1,038 tests after the rebase.

The failing Ubuntu shard and merged coverage report share the same two failures in src/main/notebook/runtime-service.test.ts: both time out after approximately 1,000 ms while polling for releaseInstall. The coverage job is failing on the merged test failures, not a coverage threshold. Both tests pass when run in isolation on the rebased branch.

A minimal proposed test-only change replaces the one-second polling with an explicit install-started promise, matching the adjacent environment-lock test. This preserves the lock-order assertions and does not change production behavior, state, persistence, or UI. Implementation is paused for owner confirmation because this test file is outside the original PR file scope. The rebase is pushed; no Notebook test changes are included yet.

PR #2327, opened 2026-09-07
fix(transcript): preserve reading windows and segment identities

ewen-poch:
## Problem

New timeline rows can displace history being read, closing find can discard a newly selected run, and standalone cards can split a named tool group into rows with duplicate identities. Completion placement also scans all timeline ownership entries once per completed turn.

## Proposed change

- Retain a scope-bound reading anchor and follow intent; preserve explicit run/end selections through find close until the user scrolls elsewhere, and reset on session/branch changes. Cover empty-to-populated initialization through both the hook and mounted component (120/240 input rows → 80 mounted rows), and preserve find navigation after explicit end-following has settled.
- Reuse the existing scroller after a bounded tail replacement with unchanged mounted row count, leaving ordinary growth to its native anchor policy. Commit the tail before the end button measures it and restore following after the primitive processes new prompt anchors.
- Derive contiguous tool-segment row IDs from the first activity while retaining business group relationships.
- Index each prompt's latest timeline position once, preserving every completion row and its order.

## Scope and non-goals

Renderer-only changes. No persisted fields, migration, historical-data backfill, business enum, new dependency, worker or replacement scrolling architecture. Existing message/activity IDs suffice for historical records; synthetic group IDs are transient UI identities. Main-process execution frameworks and subscription contracts are unchanged.

Before implementation, two repeated runs against unchanged production code failed on the same five new assertions: loss of 2/80 and 80/80 reading nodes, loss of a run selected during find, coupled group collapse/duplicate keys, and 125,000 ownership visits for 250 turns. Existing cases in those files passed.

Checks below ran after the last production/test edit:

| Behavior / boundary | Check | Result |
| --- | --- | --- |
| Window anchors, appends, find, scope changes; segment IDs and completion order; mounted scroller, Run Marks, annotation and provenance consumers | `npx vitest run src/renderer/src/pages/workspace/{use-transcript-window.test.tsx,WorkspaceMessageScroller.interaction.test.tsx,WorkspaceMessageScroller.annotations.test.tsx,WorkspaceMessageScroller.render.test.tsx,WorkspaceRunMarks.interaction.test.tsx,workspace-tool-activity-groups.test.ts,workspace-conversation-timeline.test.ts,ArtifactProvenancePanel.render.test.tsx,ArtifactProvenanceMessages.integration.test.tsx}` | 9 files, 288 passed |
| Workspace owner/contracts and session-store consumer coverage | `npm run test:module -- workspace_page` | 31 files, 873 passed |
| Repository lint | `npm run lint` | Passed |
| Real component scrolling, immediate find close, bottom following and independent segment collapse | Local Chromium fixture with actual WorkspaceMessageScroller/MessageScroller, real wheel input and Run Marks | Passed; screenshots supplied in the task |

The same one-warm-up/five-sample Node benchmark at 4,000 completed turns improved from 112.40 ms to 3.23 ms median, retaining all 12,000 rows. Regression assertions count the reported scan instead of using a hardware-dependent timing threshold.

The affected-test explain command routes this diff to the full suite because the manifest has no explicit owner for the scroller test file. The focused local set above was selected from actual owner/consumer paths; no full local portable suite was run, as requested by the maintainer. Exact-head PR CI remains authoritative. Browser validation uses fixture IPC/data and does not claim Electron native text-find end-to-end coverage. No main-process, migration or OS-specific behavior changed, so those local lanes were excluded.

## Review focus

For empty scopes, unmatched state still derives `start = max(0, items.length - 80)` and renders `items.slice(start, end)`; both hook and actual scroller tests exercise this without changing initialization logic. Review reading-anchor/follow transitions, scope cleanup, the ordering around the scroller MutationObserver, and synthetic row-ID consumers. Confirm the evidence mapping covers the final diff; do not treat unselected CI lanes as proof.

The latest review's initial-end find scenario was also exercised directly against head `a864b253`: viewport at 9,200 / 10,000 with height 800, open find, first programmatic scroll to the old message, then close. It retains the old message and clears following (10 hook checks passed including this temporary reproduction). Opening find stores prior follow intent in `snapshot.window.followEnd`; `snapshot.followEnd` stays unset until an explicit end selection during find. The review's premise that opening find initializes the latter to true does not match the code. No production change was made for this unreproduced finding. The temporary reproduction and output are retained in local verification evidence.

The exact-head macOS gate exposed a short-window regression: after scrolling away with fewer than 80 rows, the retained capacity could freeze at that smaller count and omit later replies. A mounted component regression failed on the missing appended node before the follow-up fix. Retained capacity now stays at least 80 while keeping the reading start unchanged. The two failing Electron scenarios (permission/file preview and representative conversation/project/recovery) each passed twice locally after the fix; renderer types, lint, the 288 focused checks and 873 workspace checks also passed. Full exact-head CI passed after the follow-up and formatting correction.

The formatting-only head `83b1726c` has the same production behavior as `2b806a14` (which received a mergeable review). Its review repeats the empty-hydration and initial-find claims addressed above; these are contradicted by the mounted-component/hook checks and direct reproduction evidence, rather than accepted as verified defects. Formatting now passes the repository Prettier check.

Final CI verification: all checks passed (or were intentionally skipped) on `83b1726ccd3f98556341b6d41789d42156768ca6`, including three full Ubuntu test shards, module coverage, static checks, CodeQL, Windows core and both E2E shards, macOS build/E2E/accessibility/visual, and Linux Notebook isolation. The automated review job completed, but its textual verdict remains needs changes for the two disproved claims documented above; this is not represented as review approval.

### DIFF
--- a/src/renderer/src/pages/workspace/WorkspaceMessageScroller.annotations.test.tsx
+++ b/src/renderer/src/pages/workspace/WorkspaceMessageScroller.annotations.test.tsx
@@ -429,11 +429,15 @@ describe('WorkspaceMessageScroller annotation prop sync', () => {
       container.querySelector<HTMLButtonElement>('[data-testid="tool-group-header"]')?.click()
     )
     await render([target, ...fillers])
-    expect(container.querySelector('[data-message-id="activity-group-target-group"]')).toBeNull()
+    expect(
+      container.querySelector('[data-message-id="activity-group-notebook-reveal-target"]')
+    ).toBeNull()
 
     await act(async () => requestAnnotationReveal(annotation))
 
-    const targetGroup = container.querySelector('[data-message-id="activity-group-target-group"]')!
+    const targetGroup = container.querySelector(
+      '[data-message-id="activity-group-notebook-reveal-target"]'
+    )!
     expect(
       targetGroup.querySelector('[data-testid="tool-group-header"]')?.getAttribute('aria-expanded')
     ).toBe('true')
--- a/src/renderer/src/pages/workspace/WorkspaceMessageScroller.interaction.test.tsx
+++ b/src/renderer/src/pages/workspace/WorkspaceMessageScroller.interaction.test.tsx
@@ -4241,6 +4241,27 @@ describe('WorkspaceMessageScroller artifact click behavior', () => {
     }
   )
 
+  it('leaves ordinary transcript growth to the native scroller anchor policy', async () => {
+    const { WorkspaceMessageScroller } = await import('./WorkspaceMessageScroller')
+    root = createRoot(container)
+    const messages = [createMessage({})]
+    const render = async (): Promise<void> => {
+      await act(async () =>
+        root.render(
+          <WorkspaceMessageScroller
+            activeSession={createSession({ status: 'idle', messages: [...messages] })}
+            onSendEditedMessage={vi.fn()}
+          />
+        )
+      )
+    }
+    await render()
+    scrollToEndMock.mockClear()
+    messages.push(createMessage({ id: 'ordinary-append', createdAt: 1710000000001 }))
+    await render()
+    expect(scrollToEndMock).not.toHaveBeenCalled()
+  })
+
   it('mounts the latest window before the end button measures its scroll target', async () => {
     const { WorkspaceMessageScroller } = await import('./WorkspaceMessageScroller')
     const messages = Array.from({ length: 240 }, (_, index) =>
--- a/src/renderer/src/pages/workspace/WorkspaceMessageScroller.interaction.test.tsx
+++ b/src/renderer/src/pages/workspace/WorkspaceMessageScroller.interaction.test.tsx
@@ -4241,6 +4241,34 @@ describe('WorkspaceMessageScroller artifact click behavior', () => {
     }
   )
 
+  it('mounts only the tail when an initially empty session receives long history', async () => {
+    const { WorkspaceMessageScroller } = await import('./WorkspaceMessageScroller')
+    root = createRoot(container)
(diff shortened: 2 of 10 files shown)


ewen-poch:
Regarding the two findings on `83b1726ccd3f98556341b6d41789d42156768ca6`:

- **Hydration after an empty session:** `stateMatchesScope === false` takes the bounded fallback: `start = Math.max(0, items.length - 80)`, `end = items.length`, followed by `items.slice(start, end)`. It does not select the entire history. The existing hook test covers 0 → 120 items and retains items 41–120; the mounted `WorkspaceMessageScroller` test covers 0 → 240 messages and retains only the final 80. Both pass on this head. An additional initialization flag is not needed for the reported behavior.

- **Find navigation from the end:** opening find saves prior follow intent in `snapshot.window.followEnd`; it does not set `snapshot.followEnd`. The latter records an explicit end selection made during find. Therefore the early end-restoration branch is not taken simply because find opened at the end. I exercised the stated sequence against this head: viewport height 800, scroll height 10,000, initial scrollTop 9,200; open find; first programmatic scroll to an old message at scrollTop 500 without a wheel/key event; close find. The old message remains in the bounded window and `isFollowingEnd` becomes false. Existing tests also cover navigation after an explicit end selection settles. This is hook/DOM-boundary evidence, not a claim of Electron native text-find end-to-end coverage.

The focused verification passed all 5 selected checks across the hook and mounted-component test files. The additional first-scroll reproduction was temporary, and its source/output were retained in local evidence; committed tests and production files are unchanged. I am not adopting either suggested code change because the reported failure is not reproduced and the relevant fallback/restoration paths already preserve the required behavior.

PR #2332, opened 2026-09-07
fix(literature): preserve import identities and author metadata

ewen-poch:
## Problem

An imported reference whose DOI belongs to one record and PMID to another silently reused whichever identifier appeared first. Filling missing fields could copy the other record's identity onto that target. Reviewed merges could retain a discarded record's primary DOI in citations even when a different survivor was selected. NBIB imports dropped collective authors and interpreted abbreviated personal names as organizations.

## Proposed change

- Resolve every strong-identifier owner, including Trash and earlier rows in the same file, through one batch resolver for preview and transactional commit. Conflicting values or owners block reuse/fill-missing before any writes. Explicit separate import remains available for the entire batch. Refresh conflict details if the database changes after preview.
- Define `isPrimary` as at most one explicit preference per scheme. Preserve the survivor's effective preferred value during reviewed merges and keep historical identifiers as nonpreferred. Use deterministic legacy fallback in citation projection and independent scheme choices in the editor.
- Preserve NBIB author field order: CN is an organization, AU/FAU are personal names, matching adjacent abbreviated/full forms represent one person, and initials remain separate during citation formatting. Preserve uncertain text in existing Extra metadata and show a preview warning.

## Scope and non-goals

No new database columns, tables, migrations, identifier uniqueness constraint, or automatic historical repair. Existing explicit metadata writes normalize primary flags; old stored records remain readable. Previously dropped authors require original source data to recover, and prior merges cannot reliably reveal the user's intended preferred identifier.

Transient contract additions: import `conflict` status and match details, plus `uncertain-author-name` warning. Existing duplicate-policy and creator-mode enums are unchanged. This does not introduce a per-record target picker or persistent review queue, and does not redesign Add/Inbox matching.

Before production changes, eight new assertions failed identically in two runs against real SQLite, the public parser, and merge/write/read/CSL boundaries. The failures reproduced the reported behavior; all 53 existing tests in that initial slice passed. A further public Vancouver regression failed twice when adjacent initials were treated as one given-name token; it now preserves both initials. No test seam was added.

Checks below ran after the final material edit:

| Behavior | Check | Result |
| --- | --- | --- |
| SQLite identity conflicts, all three policies, Trash, batch rollback, preview/commit race, identifier preferences, NBIB parsing/persistence, citation/export and catalog consumers, renderer interactions, shared contracts and all locale guards | `npx vitest run src/main/literature src/shared/literature.test.ts src/shared/literature-csl.test.ts src/shared/renderer-contract-catalog.test.ts src/renderer/src/pages/literature src/renderer/src/i18n/resources.test.ts src/main/acp/literature-reference-prompt.test.ts` | 1,260 passed |
| Main, renderer and sandbox types | `npm run typecheck` | Passed |
| Source/style checks | `npm run lint`; `git diff --check` | Passed |
| Import blocking/separate option, independent DOI/PMID preference and merge preference display | ego lite, actual components and project CSS with controlled data; page-scoped screenshots | Passed |

CodeGraph established baseline ownership; its index belongs to the primary checkout, so final worktree helper consumers were checked directly. The batch resolver stays private to the catalog. The affected-plan explain command requests full coverage because `src/main/ipc.ts` lacks a module owner. Per maintainer direction, local validation used the explicit owner/consumer slice above; full and platform validation is left to PR Gate. No database schema or platform-specific filesystem logic changed. Packaged Electron and external live providers were not exercised locally. Independent review of `764e1eaf93f7569639d4ce8d8e152cc47ef6dfa6` concluded [mergeable with no actionable findings](https://github.com/aipoch/open-science/pull/2332#issuecomment-5573645286). PR Gate supplies the full/platform results.

## Review focus

Identity ambiguity must never restore, supplement or link a record before the batch is accepted. Explicit separate records remain supported. Inspect the transient IPC warning/detail flow and preferred identifier selection after persistence. NBIB interpretation follows [NLM's author-field and ordering guidance](https://www.nlm.nih.gov/pubs/techbull/mj06/mj06_corp_author.html).

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -4620,8 +4620,14 @@ const createApplicationModules = async (
       get: (itemId) => literatureCatalog.get(itemId),
       importPdf: (request) => literaturePdfImporter.import(request),
       importRecords: async (request) => {
-        const parsed = await literatureCitationFormatter.parseReferences(request.content)
-        const entries = await literatureCatalog.inspectImportItems(parsed.items, parsed.errors)
+        const { warnings, ...parsed } = await literatureCitationFormatter.parseReferences(
+          request.content
+        )
+        const entries = await literatureCatalog.inspectImportItems(
+          parsed.items,
+          parsed.errors,
+          warnings
+        )
         if (request.mode === 'preview') return { ...parsed, entries }
         if (parsed.items.length === 0) throw new Error('No valid references were found.')
         return {
--- a/src/main/literature/catalog.test.ts
+++ b/src/main/literature/catalog.test.ts
@@ -11,9 +11,12 @@ import {
   literatureItemInputSchema,
   type LiteratureCandidateInput
 } from '../../shared/literature'
+import { toCslItem } from '../../shared/literature-csl'
+import { buildLiteratureMergeItem } from '../../renderer/src/pages/literature/literature-merge'
 import { migrateApplicationDatabase } from '../database/migration-service'
 import { createProjectDbClient } from '../projects/prisma-client'
 import { LiteratureCatalog, normalizeIdentifier } from './catalog'
+import { LiteratureCitationFormatter } from './citation-formatter'
 
 describe('Literature identifier normalization', () => {
   it('removes text joined to the end of a DOI before provider lookup', () => {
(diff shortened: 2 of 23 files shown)


ewen-poch:
Addressed the conflicting-row identity-index finding. `resolveImportItems` now returns a conflict before registering any identifiers from that row. Later rows therefore retain their independent preview status, and valid matches still point to the existing database record.

Added a regression through the public catalog preview API using real SQLite. It failed twice before the fix because the later row was incorrectly marked `existing`, and now passes. All 357 tests in the Literature main module and the import/library renderer test slice pass, as does targeted ESLint. The change only touches the existing catalog implementation and test file; it adds no states, persistence fields, or interaction controls.

ewen-poch:
Rebased onto main at `418f4a61` and resolved the CSL test conflict by retaining both the date-precision cases from main and the identifier-preference cases from this PR. The three fix commits retain their behavior; no new design or schema changes were introduced.

Validation covered 1,300 related tests. The initial run passed 1,294; all failing cases passed in focused reruns under reduced concurrency. Two additional UI timing failures during the rerun also passed in isolation. Full typechecking passed. GitHub now reports the branch as mergeable; the new CI run remains authoritative.

PR #2333, opened 2026-09-07
fix(literature): preserve metadata integrity during completion

ewen-poch:
## Problem

Metadata completion could delete editors/translators when replacing authors, omit identifier changes from review/readiness, and combine a retained publication year with a different source date. Failed task checkpoint writes could publish unscheduled work, while the error Retry only refreshed the task. The advertised response limit was checked after full decoding and counted characters instead of bytes.

## Proposed change

- Merge only author creators, preserving other roles and their relative order.
- Derive identifier review entries from normalized original and candidate values/primary flags, ignoring order-only changes while preserving the original list; require selection for replacements and primary changes. Identifier-only additions now make batch rows ready and single-item completion actionable.
- Review publication date/year together, infer missing years only from valid numeric dates, and preserve ambiguous text unless explicitly replaced.
- Persist command drafts before publication, coordinate worker checkpoints with commands, and retry the actual failed request even when saving its pending review draft failed. Changing the selection retires a stale failed request.
- Cancel metadata streams as soon as received bytes exceed 2 MiB.

## Scope and non-goals

No database schema/migration, new task state, dependency, or networking subsystem. Existing references are not rewritten and previously lost contributors are not reconstructed.

Compatibility: adds `identifiers` to the metadata review field enum and optional `reviewVersion: 1` to persisted review results. Unversioned reviews remain readable but require a fresh search before application. Older binaries may reject new checkpoints; no downgrade conversion is provided. Task journal versions remain unchanged.

Regression tests use public preview/commit and run/get boundaries, real temporary checkpoint files, controlled response streams, and the actual renderer components. The expanded test set ran twice against the original implementations: 63 cases with the same 25 expected failures and 38 passes each time. The fixed implementation passes. Separate regressions reproduce stale apply retries, lost commands after failed draft saves, and false identifier changes from reordering or equivalent DOI spelling. All fail before their fixes and pass afterward. Command persistence fault tests close and reopen the service before injection so a finishing worker cannot race the fault fixture.

Checks below ran after the final production edits; the command fault fixture was then stabilized and its suite rerun:

| --- | --- | --- |
| Metadata/catalog, reference resolver, CSL, jobs/storage, Library/batch consumers | `npx vitest run src/main/literature src/shared/literature.test.ts src/shared/literature-csl.test.ts src/shared/literature-jobs.test.ts src/renderer/src/pages/literature src/main/storage/durable-json-file.test.ts` | 457 passed (including focused reruns) |
| IPC catalog, composition, Electron and preload contracts | `npx vitest run src/shared/renderer-contract-catalog.test.ts src/main/application-command-composition.test.ts src/main/application-command-electron-adapter.test.ts src/preload/index.test.ts` | 119 passed |
| Eight translated locales | `npx vitest run src/renderer/src/i18n/resources.test.ts` | 804 passed (including isolated scan rerun) |
| Main/sandbox/renderer types | `npm run typecheck` | passed |
| Lint and patch hygiene | `npm run lint`; `git diff --check` | passed |

Focused reruns: `npx vitest run src/main/literature/batch-jobs.test.ts src/renderer/src/pages/literature/LiteratureLibraryPage.render.test.tsx --maxWorkers=1 --testTimeout=60000` (129 passed); `npx vitest run src/renderer/src/i18n/resources.test.ts -t 'no t\(\) call passes a dotted semantic path' --maxWorkers=1` (1 passed, 803 skipped).

The first broad rerun under concurrent local validation had scan/page timeouts and exposed the fault-fixture race; focused reruns are reported separately rather than treating that run as green.

Browser QA used real Library and batch components with controlled API fixtures, without the user database. Captured identifier/date review and checkpoint-error screenshots; verified the selected date field and identical apply retry payloads. The Literature module has no module-impact owner yet, so the local set follows inspected owner/consumer and IPC relationships rather than a claimed automatic selective plan. Per maintainer request, no full local test suite was run. Full portable/platform CI and independent PR review remain authoritative; live providers and packaging were not exercised locally.

## Review focus

Worker/command ordering and object identity during active searches; identifier consent and primary flags; coupled publication dates; legacy-review refusal and downgrade implications.

### DIFF
--- a/src/main/literature/batch-jobs.test.ts
+++ b/src/main/literature/batch-jobs.test.ts
@@ -27,6 +27,7 @@ const item = (id: string): LiteratureItemView => ({
 })
 const preview = (id: string): LiteratureMetadataCompletionResult => ({
   mode: 'preview',
+  reviewVersion: 1,
   provider: 'crossref',
   sourceUrl: 'https://crossref.org',
   item: item(id),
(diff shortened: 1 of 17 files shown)


ewen-poch:
Addressed both actionable findings:

- Identifier review now compares normalized scheme/value/primary tuples without treating list order as metadata. When those tuples are unchanged, the original list and spelling are preserved. Regression cases cover both reordered identifiers and an equivalent DOI URL; both reproduced false review entries before the fix.
- The pending command is retained before saving review drafts. If that save fails, Retry saves the draft and then executes the original command. Changing the selection still invalidates the stale request. The component regression now requires the original apply to run after Retry without another click; it failed with zero apply calls before the fix.

These changes stay within the existing implementation and test files and introduce no additional contract fields, persisted state, or UI controls.

The command fault-injection tests now close and reopen the real service before replacing its checkpoint directory, so the injected failure cannot race the search worker’s final write.

Validation: the focused regression suite passes all 46 cases (three failed before the fixes). The 129-test job/page recheck passes; all 457 owner/consumer cases and 923 contract/i18n cases pass across the broad run and focused reruns. Typecheck and lint pass. The initial broad run had timeouts under concurrent local validation and the fault-fixture race described above. No full local test suite was run.

PR #2337, opened 2026-09-07
fix(literature): validate PDFs and support attachment recovery

ewen-poch:
## Problem

A PDF header alone allowed corrupt local files to become formal attachments. Users could not remove an incorrect attachment, and known missing or damaged bytes were indistinguishable from healthy attachments. Removal also needs to preserve shared content, concurrent acquisition, and saved chat PDF references.

## Proposed change

- Parse local PDFs within the existing 50 MiB automatic-processing limit before attaching, record page count, and report corrupt/password/unsupported parsing failures. Retain larger files with an explicit unvalidated indication.
- Add individual attachment removal and retry verification through the shared ActionMenu. Preserve committed deletion when cleanup or detail refresh fails.
- Persist last verification failure and attempt time, project last-known availability, and disable known-unavailable previews. Missing bytes enter the existing quarantined state; successful retry restores availability.
- Reject removal while a live chat PDF context or retained message snapshot references the attachment, including inactive conversation branches. Read the existing Session authority under its global mutation barrier; fail closed on incomplete reads. Revalidate new literature bindings before persistence so deletion cannot race a stale source into a chat. Explain the remedy in all renderer locales.
- Preserve main's locked publication commit/cancellation cleanup, provenance and authoritative transfer receipts. Local import and full-text retrieval reserve published content through reference acquisition; sweeps and short agent publication commits share the existing content lifecycle lock and reference-aware cleanup.

## Scope and non-goals

Persistence: migration `0036_content_verification_observation` adds the previously reviewed nullable `ContentBlob.lastVerificationFailure` and `lastVerificationAttemptAt`. Main's released `0035_literature_pdf_provenance` remains unchanged. Existing records remain null until checked; there is no backfill or startup scan. Private development databases that ran this PR's former unreleased migration identity are not silently rewritten.

No additional persisted Session fields, lifecycle enum values, or reference tables. Availability is a last-known `unknown` / `available` / `unavailable` view, separate from PDF parsing validity. With PDF retains its record-presence meaning. Session reference checks run only on attachment removal, using existing saved bindings; they do not hash all PDFs. Very large chat histories may make this infrequent operation slower.

Independent acquisition may complete after an older attachment is removed. Removal does not change metadataRevision or implicitly cancel a separately initiated import/download. No version-history UI, automatic replacement, background scanner, cross-process reservation protocol, or redesign of existing whole-item deletion is included. Lost bytes cannot be reconstructed.

Public Catalog/importer/content boundary regressions reproduced the original corrupt import, removal and availability failures before repair. New regressions also failed before the Session fix: the real SessionRepository retained a binding while Catalog successfully deleted its attachment, and the PDF context owner persisted a source deleted after inspection. Both pass after repair; unlinking a live context allows removal only when no retained message snapshot references the attachment. Tests cover incomplete scans, scheduler ordering, three acquisition paths, nested reservations, callback failure cleanup and retained shared references.

The latest message-history correction passed **1,768 selected cases after the final production edit**, plus complete typecheck and lint:

| --- | --- | --- |
| Removal preserves live, legacy-message and inactive-branch references; import/acquisition/authority consumers | `npx vitest run src/main/literature` | Passed |
| Session mutation ordering, facade contracts and PDF linking | Coordinator, coordinator-contract, coordinator.architecture, PDF-context-owner and operation-scheduler suites | Passed |
| Existing session serialization and message snapshot compatibility | `src/shared/session-persistence.test.ts` | Passed |
| Actionable history-reference error, preserved attachment/preview, lookup consumers and eight locales | LiteratureLibraryPage.render, LiteratureFullTextLookup.render and i18n/resources suites | Passed |
| Main, sandbox and renderer types | `npm run typecheck` | Passed |
| Lint | `npm run lint` | Passed |

Two real SessionRepository/Catalog regressions failed before the history guard: retained legacy messages and inactive-branch snapshots both allowed deletion. They now reject removal and retain resolvable content. The existing empty-history test still allows removal after unlinking. The error explicitly states that unlinking a live chat does not remove historical references. Existing history participates directly; no additional persistence, enum or migration is introduced.

Earlier validation covered 1,977 cases including storage owners, database migrations, packaged ledger and IPC adapters, plus schema generation checks. CI on preceding head 18b0de0e finished successfully, including the previously failing Notebook test shard. Those unchanged areas were not rerun locally for the narrow history predicate/copy update.

The combined module run used two workers and a 60-second CLI timeout without changing repository configuration. No full local npm test. Latest-head CI remains authoritative. The shared CodeGraph index belongs to another worktree; direct caller/composition searches establish the affected interfaces here.

Browser verification used the real Literature page with labeled controlled IPC data: attempted removal shows the history-reference explanation and preserves the attachment and preview. Filesystem/SQLite behavior is covered independently by integration tests. The screenshot is available in the task handoff.

## Review focus

Preservation of main's publication lifecycle and provenance; atomic ordering between Session binding persistence and attachment removal; reference-aware content reclamation; last-known integrity versus parser validity; and immutable released migrations.


ewen-poch:
The table preview finding is addressed: known-unavailable attachments disable the preview button, and the preview callback also rejects them. The page regression reproduces the previously enabled button and passes with the guard; browser verification confirms the recovery-details link remains usable.

The packaged migration smoke failures were corrected by registering the new immutable migration identity/checksum and removing its ledger row in historical rollback fixtures. The seven CI failures were reproduced locally before the correction and passed afterward.

The two latest verification findings are being evaluated against the approved design. Quarantining missing blobs would change the persisted state-transition policy shared by all content consumers. Rejecting an unavailable verification result also requires preserving the detail refresh on failure, so the UI can display the newly observed diagnosis. These behavior changes are awaiting maintainer confirmation before implementation; they are not claimed as resolved.

ewen-poch:
Missing-file verification now quarantines the content for both typed missing errors and filesystem ENOENT races. Explicit successful retry still restores available state. This uses the existing state values and does not backfill historical records.

The attachment verification command now rejects unavailable results. The renderer refreshes the attachment diagnosis before propagating the error through the existing action menu, keeping the failed operation visible without hiding the newly detected missing file.

All three regressions failed before the fixes for the reported reasons and pass afterward. The focused owner, migration, renderer, translation and consumer suites pass 1,043 tests; runtime typechecks pass. Full lint is still running locally. Browser verification confirms the error, updated diagnosis and disabled preview appear together.

ewen-poch:
Rebased onto main at 555f3832. Resolved the catalog constructor and IPC composition conflict by preserving the tag-assignment notification callback as the second argument and injecting the content repository as the third argument. Both tag synchronization and attachment verification/removal remain connected. Locale conflicts retain both sets of additions.

After conflict resolution, 1,011 targeted catalog, attachment, page and translation tests passed, along with all runtime typechecks and full lint. No additional schema, state values or interaction changes were introduced by this rebase.

ewen-poch:
The concurrent publication/reference-acquisition finding is reproducible through the real importer, catalog and content repository. Pausing an import immediately after publication, removing the old attachment, and then resuming the import causes it to fail. The regression uses the existing publish boundary and does not add a production test seam.

A lock only around attachment insertion does not close the preceding publication gap. A complete fix needs protection from publication through durable reference acquisition, including the local import, full-text retrieval and agent PDF acquisition consumers. This expands the previously approved lifecycle/API scope and touches additional files, so implementation is paused for maintainer approval. A grace period alone would only reduce the probability of the race.

The branch has been rebased onto main at 547541bf. Git range-diff confirms all three existing patches are unchanged. The failing reproduction is retained locally and has not been added as a failing CI test to the PR.

ewen-poch:
Publication and reference acquisition now use a scoped withPublishedContent operation. It installs a shared per-content reservation under the existing lifecycle lock before publication returns, and releases it in finally after reference insertion or failure. Sweep checks this reservation under its own claim lock, so the publication-to-reference gap is protected without holding a lock across PDF parsing or a database transaction.

All three production Literature publishers use the scoped operation: local import, selected full-text retrieval, and agent PDF acquisition. Reservations are shared across repository instances within the application main process. No database fields, enum values, relationships or UI changes were added.

The original race was reproduced independently on all three real acquisition paths before the fix, then passed afterward. Tests also cover overlapping publications, verification inside a callback, release on success/failure, and subsequent reclamation. The final targeted Literature/content/upload/artifact/command-consumer suite passed 422 tests. Runtime typechecks and full lint passed; the final test-only return annotation was rechecked with the content repository suite.

ewen-poch:
The saved-session source concern is reproducible at the existing source-resolution boundary: a literature version resolves before attachment removal and returns undefined afterward. Session PDF bindings retain that version ID, so preserving only the content blob would not repair resolution after the version row is deleted. A policy for deleting session-referenced attachments needs maintainer confirmation before expanding into session ownership or changing deletion behavior.

The in-flight acquisition concern also requires a product decision. The current regression coverage intentionally allows a separately initiated acquisition to complete while an older attachment is removed. Unconditionally advancing metadataRevision would change that behavior and conflate metadata edits with attachment cancellation. Cancellation of automatic full-text acquisition should be decided separately from an explicit local import.

Rebasing onto main at 0d3e2e88 exposed a further lifecycle design overlap: main now supports publish.commit inside the content lifecycle lock, while this branch uses a protected acquisition callback outside the lock. These must be reconciled deliberately, including verification re-entry and cleanup on callback failure. I aborted the incomplete rebase and preserved the clean branch pending the requested design confirmation. The migration sequence also needs renumbering because main now owns 0035_literature_pdf_provenance; its released migration must remain unchanged.

For the currently pushed head e4439d9, the focused local validation passed 422 tests, typechecks, and lint. Completed checks are green, but the AI review requests changes and the branch conflicts with the newly updated main. This is not ready to merge.

ewen-poch:
Saved chat references are now protected. The public Catalog removal command runs through the existing Session coordinator barrier, reads saved Session authority, rejects a referenced attachment, and fails closed if that scan is incomplete. PDF linking rechecks the version before persistence under the same scheduler, closing the inspection-versus-removal race. The UI explains that the PDF must first be removed from the chat, and retains its preview. Regression tests failed before the fix for successful deletion/stale binding persistence and now pass; a real SessionRepository test also confirms removal succeeds after unlinking. No Session fields, reference tables, or new persisted states were added.

I am retaining the independent-acquisition behavior: removing an older attachment does not cancel an explicitly started import/download or mutate metadataRevision. Cancellation here would be a separate product feature, not a necessary integrity fix. Existing main cancellation signals and authoritative transfer receipts remain intact.

Rebased onto main 0d3e2e88 and reconciled its publish.commit API and cancellation cleanup with protected parsing/reference acquisition. Both use the existing content lifecycle lock and the same reference-aware cleanup. Main's provenance migration is unchanged; this PR's unreleased observation migration is now 0036, including its checksum and packaged/historical fixture expectations.

The final production changes passed 1,961 selected test cases across the module run and focused fixture rechecks, all typechecks, lint and schema generation checks. Browser verification confirms the actionable error and preserved attachment. Updated PR scope and validation mapping describe the remaining limits. Latest pushed head: f604afc7.

ewen-poch:
The newly merged full-text lookup renderer test still supplied the old publish-only fixture. Updated it to execute the scoped acquisition callback, preserving the existing real-finder reconnect test. Its 16 cases pass, the complete main/sandbox/renderer typecheck now exits successfully, and lint passes. The final evidence total is 1,977 selected cases. This corrects the earlier typecheck status: the previous run's delayed renderer error had not yet been included when I reported completion. No production or interaction changes in this follow-up.

ewen-poch:
Rebased onto main 41fdab24 without conflicts. Range-diff preserves all six existing commits; the only contextual change is the surrounding inbox undo UI introduced by main. No architecture, data model, persisted state, or attachment interaction changes were added.

The preceding AI review was mergeable with no actionable findings. The failed Ubuntu shard report identifies one test outside this PR: NotebookRuntimeService "cancels and drains a Shell Run before application disposal completes" timed out while waiting for its execution signal (expected AbortSignal, received undefined). Neither that test nor its production module differs from main. A focused run using the portable-CI coverage options passes; this does not establish the cause of the original failure, so I have not changed its timeout or implementation.

After the rebase, 1,341 Literature/renderer/i18n cases pass, along with the focused Notebook case, complete typecheck and lint. The existing broader attachment/storage/session/migration evidence remains in the PR description. I will check the review and CI for the new head rather than treating the earlier result as current-head approval.

ewen-poch:
Confirmed the persisted-message snapshot finding using the real SessionRepository, Catalog removal command, and attachment authority. A saved user message retains its PDF version binding after the live runtime context has been unlinked. The version resolves before removal and resolves to undefined afterward; the regression fails on that exact post-removal assertion. The document reader consumes these persisted message snapshots, so checking only runtimeContext is insufficient.

This changes the previously agreed removal policy: unlinking the current chat context would no longer be sufficient when retained message history still references the PDF. A minimal guard could include retained message snapshots (including non-active conversation branches) and update the user-facing explanation without adding fields or persistence. Preserving independently readable snapshots while allowing attachment removal would require a broader ownership design.

I am pausing this policy/interaction change for maintainer confirmation, as requested. No production changes were made for this new finding. The rebase is already pushed at 18b0de0e; current static checks pass, and the portable test rerun is still in progress.

ewen-poch:
Attachment removal now checks retained user-message PDF snapshots as well as the live runtime context. It scans canonical conversationGraph.messages, including inactive branches, with legacy session.messages as the fallback. The existing Session mutation barrier and attachment identity check remain unchanged.

Two regression cases using the real SessionRepository, Catalog and content authority failed before the fix because deletion succeeded despite retained legacy or inactive-branch snapshots. Both now reject deletion and retain resolvable content. The existing empty-history case still allows removal after unlinking the live context. Updated all locale copies to explain that unlinking the current chat does not remove historical references, and verified the real rendered page with controlled IPC data.

This follows the confirmed policy. No new fields, persisted states, tables or migrations; existing history participates without backfill. After the final production edit, 1,768 affected tests, complete typecheck and lint passed. The previous head's complete CI also passed, including the Notebook shard. Pushed fix: 7ac4a97b.

PR #2340, opened 2026-09-07
fix(literature): preserve note drafts and synchronize tag updates

ewen-poch:
## Problem

Inline note failures replace the user's draft, and a concurrent rating refresh remounts the note editor. Successful inline updates also leave stale cached pages behind. Tag assignment snapshots do not refresh filtered literature results, while catalog merge/deletion transactions change tag relationships without publishing the tag service's revision event.

## Proposed change

- Preserve note drafts and their original item/revision, with compact two-row retry feedback and explicit Escape discard. Adopt incoming notes only when the editor is clean. Distinguish a successful write from a failed readback; retry acknowledged writes by reloading only, retaining any further typing and preserving concurrency checks.
- Reuse the page-cache owner for inline updates, reject stale reads, and invalidate obsolete sort/filter snapshots. Background tag-filter refresh keeps surviving editors mounted and reconciles selection/counts/pagination; unfiltered views do not reload for badge changes.
- Notify the existing TagService queue after committed single merges, batch groups and permanent deletions that changed assignments. Keep relationship writes in their existing transaction; previews and rollbacks remain silent.

## Scope and non-goals

No database schema, persisted enum, IPC payload, historical migration or new dependency. Additional state is renderer memory only. Existing data remains compatible; previously lost unsaved drafts cannot be reconstructed. No persistent draft storage, automatic conflict merge or replacement cache/event framework.

The tests exercise mounted production React controls through the existing IPC boundary and real SQLite catalog/tag services. Two isolated runs against unchanged production source each produced the same 9 expected failures and 3 passing controls. Tests reset one-shot IPC mocks so intentional failures cannot contaminate subsequent scenarios.

Rebased onto `a7350344` and verified the final two-row note feedback. The module run passed 1,430 of 1,434 tests under concurrent local load; the two affected test files were then rerun with one worker and all 930 tests passed, including all four earlier failures. No production changes or relaxed assertions were needed for that rerun. Both typechecks and lint passed. The previous commit’s complete PR Gate also passed.

| Behavior / boundary | Check | Result |
| --- | --- | --- |
| Draft failure/retry, captured revision, rating cache, tag-filter membership and mounted-editor lifecycle | Literature renderer owner tests and useLiteratureEntries tests | Passed |
| Transactional tag migration/deletion, batch/preview/rollback events and tag queue ordering | Literature catalog and tag service/repository tests | Passed |
| Cache consumers, manifest/content access, application events, shared command/renderer contracts, translations | Owner and consumer set below | 1,434 tests covered; all passed across the module run and isolated recheck |
| Main and renderer contracts | `npm run typecheck:node` (including sandbox types), `npm run typecheck:web` | Passed |
| Repository lint | `npm run lint` | Passed |
| Visible note recovery and successful retry | ego-browser with the production page/CSS and a controlled local persistence fixture | Draft retained; retry removed the error |

<details><summary>Owner/consumer test command</summary>

```sh
npx vitest run src/main/literature src/renderer/src/pages/literature src/main/tags src/renderer/src/stores/tag-store.test.ts src/renderer/src/i18n/resources.test.ts src/main/application-events.test.ts src/shared/literature.test.ts src/shared/tags.test.ts src/main/storage/content-repository.test.ts src/main/artifacts/literature-manifest.test.ts src/main/runtime-state-ownership.architecture.test.ts src/shared/application-command-contract.test.ts src/shared/renderer-contract-catalog.test.ts --maxWorkers=2 --testTimeout=60000
```

</details>

The explain plan selects the full CI fallback because `src/main/ipc.ts` has no declared module owner. Per maintainer instruction, the complete suite was not run locally; PR CI remains authoritative for full/platform checks. Local coverage includes the mutating Catalog's composition boundary and the read-only manifest consumer. No path, native packaging or migration changes require new local platform fixtures. Cross-window delivery is covered through the shared event/service/store boundaries; the browser screenshot uses controlled persistence, not a live multi-window database session.

The acknowledged-write recovery follow-up passed 142 mounted-page/cache tests, renderer typecheck and lint. Before the fix, its mounted-page regression failed because Retry issued a second write. Repeated readback failures and editing during recovery are covered; browser verification confirmed successful recovery with the existing two-row feedback.

## Review focus

Preserving the draft's original revision, background-refresh stale-response handling, and publishing tag changes only after commit through the shared service owner.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -1771,7 +1771,10 @@ const createApplicationModules = async (
     new MemoryRepository(() => getProjectDbClient(configRoot)),
     applicationEvents
   )
-  const literatureCatalog = new LiteratureCatalog(() => getProjectDbClient(configRoot))
+  const literatureCatalog = new LiteratureCatalog(
+    () => getProjectDbClient(configRoot),
+    () => tagService.notifyAssignmentsChanged()
+  )
   const literatureCitationStyles = new LiteratureCitationStyleLibrary(
     join(resolveDataRoot(), 'literature', 'citation-styles')
   )
--- a/src/main/literature/catalog.test.ts
+++ b/src/main/literature/catalog.test.ts
@@ -17,6 +17,9 @@ import { migrateApplicationDatabase } from '../database/migration-service'
 import { createProjectDbClient } from '../projects/prisma-client'
 import { LiteratureCatalog, normalizeIdentifier } from './catalog'
 import { LiteratureCitationFormatter } from './citation-formatter'
+import { TagRepository } from '../tags/repository'
+import { TagResourceCatalog } from '../tags/resource-catalog'
+import { TagService } from '../tags/service'
 
 describe('Literature identifier normalization', () => {
   it('removes text joined to the end of a DOI before provider lookup', () => {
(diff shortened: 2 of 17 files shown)


ewen-poch:
The committed-write/readback-failure finding is valid: the existing test checks retained text but stops before Retry, which currently resubmits the captured pre-write revision. I propose separating recovery of an acknowledged write from retrying a rejected write, so recovery only reloads the authoritative item and does not replay the stale update. This requires explicit maintainer confirmation because it changes retry behavior and may add renderer-only recovery state. Implementation is paused pending that confirmation. No database schema, persisted fields, or data relationships need to change.

The latest completed PR Gate and AI review runs were inspected. PR Gate passed; the cancelled CI Integrity run was superseded by a successful run. The branch has also been rebased onto main at 547e1896 without conflicts.

ewen-poch:
Fixed the acknowledged-write retry path. Once the update transaction succeeds, the note editor retains a readback recovery callback; Retry invokes that callback without replaying the update or its old metadata revision. Repeated read failures keep the draft. Text entered during recovery is retained, and its base revision advances only when the fetched note still matches the acknowledged write, preserving conflict protection against intervening note changes.

The mounted-page regression failed before the fix because Retry issued a second write. After the fix, the page/cache suites pass all 142 tests, including repeated readback failures and further editing during recovery. Renderer typecheck and lint pass. A browser check also confirmed that Retry clears the error while preserving the committed note and the two-row layout. No persistent data or schema changes.

ewen-poch:
Fixed the failed tag-refresh Retry path. The page error action now forwards the existing preserve-page option through loadEntries, keeping surviving keyed note editors mounted during recovery. Initial-load failures still load normally when there is no retained page; other refresh callers keep their existing behavior.

The mounted-page regression reproduced the issue before the fix: clicking Retry removed the note textbox immediately. After the fix, the same textbox and its unsaved text survive both the retry and successful refresh. All 143 page/cache tests pass. This reuses existing cache state and adds no persistence or new state.

PR #2343, opened 2026-09-08
fix(literature): preserve full-text acquisition lifecycle and provenance

ewen-poch:
## Problem

Canceling an `acquire_pdf` MCP request during download could still stage the PDF afterward. Closing a single-reference lookup lost its progress while the download continued. Saving a chosen PDF discarded its manuscript version and license, and batch lookup treated supplementary PDFs as full text.

## Proposed change

- Carry the MCP AbortSignal through runtime, acquisition and download, combining it with the existing timeout. Check cancellation before publication and transaction admission; cancellation does not try another source. An admitted transaction settles atomically, and temporary files still use finally cleanup. Publication can hold its existing content lock through Inbox commit; failed admission reclaims only newly created, still-unreferenced content, preserving reused bytes and concurrent publishers.
- Give the existing finder an item-addressable transfer identity. Reopening restores progress or the committed result; the dialog explains that closing continues the download in the background. A failed post-commit item read returns the successful transfer receipt; the dialog retries the read without reporting a failed download, and batch completion also accepts that receipt.
- Persist the chosen provider, source link, acquisition time, manuscript version and provider-reported license on the immutable attachment version, including PDFs accepted from Inbox.
- Require the existing `fullText` attachment role in both batch skip guards.

## Scope and non-goals

**Persistence:** add nullable `provenanceJson` to `LiteratureAttachmentVersion` and `LiteratureInboxPdf`, with an additive migration and packaged-ledger certification. Existing version numbers continue to mean content revisions.

**Historical data:** old attachments and pending Inbox PDFs remain usable with unknown provenance. No inferred backfill or source re-fetch. Duplicate content preserves the existing version and its original provenance.

**New states/API:** process-local transfer statuses are `running`, `succeeded`, and `failed`; the existing full-text command gains an item-addressable `transfer` mode and exact-ID acknowledgment. Terminal snapshots expire after 15 minutes and are capped at 128; running tasks are retained. No durable job system, restart recovery, cancel button, or attachment role editor.

Cancellation coverage is specifically the MCP request path. This change does not claim that every Agent UI Stop action sends that notification. Session history, branch selection and subscriptions are unaffected.

The initial public-boundary regression set ran twice against unchanged production code: both runs had the same six failures (67 cases, 61 passing), matching canceled Inbox writes, lost remount progress, missing durable/Inbox provenance and supplementary-PDF skips. No production test seam was added. The review regressions also ran twice before their fixes: both runs reproduced an orphaned content row, a rejected committed result, and stopped recovery in two renderer entry paths (59 existing cases still passed).

Original implementation evidence covered **1905 passing cases across 70 files**. Following the approved lifecycle review changes, **1472 cases passed after the latest rebase** across literature, storage, upload/artifact consumers, renderer literature, shared contracts and translation guards. The table identifies the broader owner/consumer mapping; untouched database and ACP paths retain their earlier evidence:

| --- | --- | --- |
| Cancellation, source retention, task recovery and batch role policy | Vitest `src/main/literature` | pass |
| Renderer remount, completion and existing detail/batch consumers | `LiteratureFullTextLookup.render`, `LiteratureLibraryPage.render`, `LiteratureBatchLookupDialog` | pass |
| Shared command validation and translations | `src/shared/literature.test.ts`, i18n resources guard | pass |
| Historical upgrades, Inbox acceptance, schema and packaged ledger | `src/main/database`, `scripts/database-migration-ledger-smoke.test.ts`, content repository | pass |
| Runtime/capability consumers | ACP runtime/base composition, session capability, provider create/resume/adopt, backend generation, Codex adapter | pass |
| MCP transport and IPC boundaries | MCP HTTP host, application command composition, Electron/preload and data-content adapters | pass |

Commands use `node node_modules/vitest/vitest.mjs run <listed paths>`; contention-sensitive reruns use `--maxWorkers=1 --testTimeout=60000 --hookTimeout=60000`. Default-timeout failures under local contention were rerun; two Codex process tests required scoped access after sandbox `ps EPERM`. The 1472-case follow-up run, typechecks and lint ran after the last material edit. Concurrency checks cover canceled publication beside another publisher, preservation of reused/unreferenced bytes and committed references, and sweep exclusion during publication-to-commit. Main/sandbox/renderer typechecks, repository lint (excluding local `docs/internal` artifacts), `db:schema:check`, and diff whitespace checks pass.

Actual Literature page UI was checked with controlled IPC: start, close, reopen with the same progress, then finish with one attachment. No external paper was fetched. Screenshots were inspected locally. The post-commit refresh fixture additionally verified the success/refresh message and automatic attachment display after read recovery. No new persisted fields or status literals were introduced by these review follow-ups.

The impact classifier requests full coverage for the schema change. At the maintainer's request, no complete local `npm test` was run; full and cross-platform authority remains PR CI. Local UI verification uses controlled IPC, not a packaged Electron end-to-end download.

## Review focus

Cancellation admission versus committed receipts; exact task-ID acknowledgment and terminal retention; nullable provenance through Inbox acceptance; additive migration/checksum compatibility; final owner/contract/consumer coverage. Independent review and PR CI remain required before calling the overall change verified.


ewen-poch:
Rebased onto current main and resolved the conflicts by retaining both the upstream metadata-integrity tests/translations and the full-text regression tests/background-download copy. The design, persisted fields, and transfer states are unchanged.

The failed AI review was a checkout failure (`couldn't find remote ref refs/pull/2343/merge`), caused by the unresolved merge conflicts; it produced no review findings or inline comments. Updating the rebased branch restores the merge ref so the review can run. Focused literature/renderer checks, typechecks, and lint were rerun on the rebased commit; full CI remains the authority for platform coverage.

ewen-poch:
The post-publication cancellation window is real: the Inbox write is prevented, but unreferenced content can remain until existing recovery sweeps it. I will not adopt the suggestion to continue staging after cancellation, because it contradicts the agreed cancellation boundary.

Immediate deletion also needs more than calling `sweep` on the returned ID: content is shared by checksum, and another acquisition may have published the same bytes but not yet committed its reference. `publish()` does not return exclusive ownership, and its lifecycle lock ends before attachment/staging.

A safe correction would extend the content repository's ownership lifecycle so publication, reference commitment, and cancellation cleanup cannot race with another consumer. That touches a file/interface outside this PR's approved change scope. Per the maintainer's instruction, implementation is paused for design/scope confirmation; no cleanup or persistence semantics have been changed in response to this finding.

ewen-poch:
The coverage failure used an incomplete input set. Ubuntu shard 3/3 was cancelled and uploaded no blob report; the merge job downloaded only two artifacts. Those reports contained 18,307 passing tests, but their partial coverage (76.4% lines) cannot establish the complete suite's coverage. The shard log reports only “The operation was canceled,” so I cannot attribute the cancellation to a timeout or code defect.

Rebased again onto current main, retaining all upstream translations and the background-download copy. No coverage thresholds, workflow behavior, data fields, or interaction semantics were changed. The new head will run fresh CI with all three shards; complete coverage must be assessed only after all reports are present.

The separate AI-review proposal concerning shared-content cleanup remains awaiting the maintainer's scope/design confirmation.

ewen-poch:
The new review repeats the post-publication cleanup issue, which remains paused for the already-described shared-content ownership scope decision.

The post-commit refresh issue is also actionable: the attachment receipt is already authoritative, but a failed catalog read can still reject the request and hide that success. The proposed correction is to retain/return the succeeded transfer independently of refresh failure and keep retrying the detail read without reporting a failed download. This can stay within the existing finder, lookup UI, and regression-test files, with no new persisted fields or status literals. It does change the failure/recovery interaction, so implementation is paused for the maintainer's requested confirmation. No review-driven behavior change has been applied yet.

ewen-poch:
The two lifecycle findings are fixed after the maintainer approved their scope.

- Publication now holds the existing per-content lock through Inbox commit. On failed admission it uses the existing transactional reference checks to reclaim only newly created, still-unreferenced content. Reused content is retained because a prior publisher may still be committing its reference. Concurrent publication and sweep tests verify that shared bytes survive.
- Post-commit catalog reads are best-effort. The finder returns the existing succeeded transfer receipt independently, the lookup retries detail reads and displays “PDF added. Refreshing attachment details…”, and batch completion accepts the committed attachment/version identity. No new persisted fields or status literals were added.

Both unchanged-code regression runs reproduced the orphaned blob, rejected committed result, and stopped recovery through both renderer entry paths. After the fixes, 1,444 focused owner/contract/consumer tests passed, along with main/sandbox/renderer typechecks and lint. The real Literature page was checked with controlled IPC; screenshots confirm automatic recovery without a second download.

The previous head's three Ubuntu shards and aggregate coverage all passed. Coverage thresholds and workflows are unchanged; the new head's CI remains authoritative.

ewen-poch:
The AI review on the lifecycle fix returned **mergeable**, with no actionable findings. Before handoff, newer main changes introduced conflicts in the eight locale files. Rebased onto `78f378bb` and retained both the upstream comparison strings and this PR's download/recovery strings.

`git range-diff` shows only translation-context changes to the two PR patches; the lifecycle implementation is unchanged. After rebase, all 1,472 focused literature/storage/consumer/renderer/contract tests passed, together with typechecks and lint. No additional data fields, statuses, or interaction changes were introduced by conflict resolution. Fresh checks on the rebased head are authoritative.

PR #2344, opened 2026-09-08
fix(literature): preserve export round trips and complete membership

ewen-poch:
## Problem

Exporting and reimporting references changes BibTeX publication months, drops PubMed/PMC/arXiv identity, and loses RIS book editors, translators and editions. Concurrent title changes during project export can also silently omit members while the total count stays unchanged.

## Proposed change

Keep the pinned citation engine and supplement its format boundaries: normalize its zero-based BibTeX publication date components, validate positional dates, project PMID/PMCID and native arXiv eprint metadata, and preserve RIS responsibilities, edition and labeled identifier notes. App-defined literal-creator notes preserve organization names containing commas without overriding subsequently edited visible fields; year zero remains compatible with the existing model. Parse RIS supplements per source record so rejected entries cannot shift fields to another reference.

Extend the existing literature search contract with an optional complete-ID query. Resolve filters and IDs inside one database transaction, then reuse those IDs for project/collection export and all-matching selection. Existing formatting rejects unavailable IDs; failed member reads now use the existing export error surface.

## Scope and non-goals

No dependency, database schema, lifecycle enum, persistent snapshot, or historical data rewrite. Existing literature fields and save/import mechanisms remain the storage owners. The optional `allItemIds`/`itemIds` fields are transient query data. Historical bad dates, lost identifiers and existing duplicates are not automatically repaired. RIS identifier notes preserve the values and round-trip within this application; other tools may retain them only as notes. Membership is fixed, not a long-lived snapshot of every metadata value. Traditional BibTeX day precision is separate from the month correction.

- Real WASM twelve-month round trips, identifiers, RIS book metadata and real SQLite PMID reuse: `src/main/literature/export-roundtrip.test.ts`.
- Invalid date positions: `src/shared/literature-csl.test.ts`.
- Complete ID selection, filters and post-read changes: `src/main/literature/catalog.test.ts`.
- Real Library + migrated SQLite + WASM: concurrent rename still saves all 101 RIS records; member-query failure saves nothing. Existing project, collection and all-matching consumers remain covered by `LiteratureLibraryPage.render.test.tsx`.
- Focused owner/contract/consumer validation covers `src/main/literature`, `src/shared/literature.test.ts`, `src/shared/literature-csl.test.ts`, `src/shared/renderer-contract-catalog.test.ts`, `src/renderer/src/pages/literature`, and `src/main/acp/literature-reference-prompt.test.ts`.
- After rebasing onto current main, all 586 focused owner/contract/consumer tests passed, along with full typecheck and lint. The only conflict joined two independently added catalog test blocks; range-diff confirms no implementation change from conflict resolution.
- Independent review follow-up: both reported boundary cases failed in two unchanged-source runs and now pass, including personal/organization role controls and stale-marker handling. All 66 affected formatter/CSL/document/LaTeX tests, full typecheck and lint pass after this follow-up.
- Both process typechecks, repository lint, changed-file formatting and diff checks passed. The checks ran after the last material edit.
- Unchanged-source replay twice: the same 17 behavior assertion failures and 2 passing controls; restored code passes all 20 focused regressions. No setup errors or timeouts counted.
- Browser verification used the actual Library component, temporary SQLite and real WASM: 101 saved records including the renamed member; subsequent injected query failure displays the existing error and does not save another file.

Per maintainer instruction, no complete local test suite ran. The impact manifest does not declare a Literature owner, so required full/platform CI remains authoritative. Document/LaTeX, catalog and renderer consumers are included locally; provider translation, subscriptions, OS paths, packaging and database migrations are unchanged.

## Review focus

Format-specific normalization must not shift RIS/accessed dates; identifier notes must stay associated with their source RIS record; complete-ID queries must retain filter semantics and exclusions. No generic replacement citation parser, title-based merging or persistent export state is introduced.

### DIFF
--- a/src/main/literature/catalog.test.ts
+++ b/src/main/literature/catalog.test.ts
@@ -179,6 +179,52 @@ describe('LiteratureCatalog', () => {
     }
   )
 
+  it('reads all matching member IDs independently of page limits and later title changes', async () => {
+    const catalog = await setup()
+    const { itemIds } = await catalog.importItems(
+      Array.from({ length: 101 }, (_, index) =>
+        literatureItemInputSchema.parse({
+          itemType: 'book',
+          title: `Paper ${String(index).padStart(3, '0')}`
+        })
+      )
+    )
+    await catalog.transact({
+      kind: 'set-project-items',
+      projectId: 'project-1',
+      itemIds,
+      included: true,
+      source: 'library'
+    })
+    const page = await catalog.search({
+      scope: 'library',
+      projectId: 'project-1',
+      sortBy: 'title',
+      allItemIds: true,
+      limit: 1,
+      offset: 100
+    })
+    expect(page).toEqual({ entries: [], itemIds, totalCount: 101 })
+    const view = (await catalog.get(itemIds[100]!))!
+    await catalog.transact({
+      kind: 'update-item',
+      itemId: view.id,
+      expectedMetadataRevision: view.metadataRevision,
+      item: { ...view.item, title: 'A moved reference' }
+    })
+    expect(page.itemIds).toEqual(itemIds)
+    const filtered = await catalog.search({
+      scope: 'library',
+      projectId: 'project-1',
+      filter: { query: 'A moved' },
+      allItemIds: true
+    })
+    expect(filtered.itemIds).toEqual([view.id])
+    expect(
+      (await catalog.search({ scope: 'library', projectId: 'missing', allItemIds: true })).itemIds
+    ).toEqual([])
+  })
+
   it.each([
     ['deleted', 'missing'],
     ['active', 'missing'],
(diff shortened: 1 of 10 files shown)


ewen-poch:
Preserved comma-containing organization editors and translators with a validated literal-creator note in the RIS adapter. The importer applies the note only while the corresponding visible creator field still matches, so later edits are not overwritten. Regression coverage includes personal creators and stale markers.

Kept year zero valid in CSL import to match the existing literature model, without changing the schema or stored data. Both reported cases failed twice before the fixes and pass afterward.

Rebased onto the latest main and retained both independently added catalog test blocks. The Windows E2E failure was a 25-minute job timeout, not a logged assertion failure; the functional suite passed and the workspace suite completed 18 tests before cancellation. A targeted job rerun was started. No workflow timeout or unrelated source files were changed.

PR #2349, opened 2026-09-08
fix(literature): validate and recover imported citation styles

ewen-poch:
## Summary

Imported citation styles can be accepted without the formatting behavior Open Science requires, damaged stored files cannot be repaired by reimporting, and overlapping style-manager requests can leave the UI loading indefinitely or restore deleted rows.

- Validate macro references and require citation and bibliography sections before accepting a style. Capture CDATA metadata through the existing text handler. These are focused support checks, not a full CSL semantic validator.
- Verify stored styles against their content-addressed IDs using raw bytes. Preserve identical imports and repair mismatched files by publishing validated original content through a same-directory temporary file and rename.
- Show available styles even if the initial list request remains pending, and make import/delete mutually exclusive using existing component state.

No new dependencies, persistent fields, database/schema changes, or IPC revisions. The interaction changes are precise import errors, working reimport recovery, and temporarily disabled conflicting mutation controls.

## Regression evidence

Added nine regression cases through the existing library and component boundaries, without new production test seams. Backend tests use the installed citeme-engine-wasm and real temporary files; component tests render the actual view with controlled API response ordering.

The initial eight cases failed consistently in two runs before the fixes. The raw-byte corruption case then failed twice before its correction. All pass after their respective fixes. A baseline formatter probe also confirmed generic fallback output for missing macros, an empty bibliography for a citation-only style, and changed output under an old ID after file replacement.

## Validation

- Literature module, style-manager component, and i18n guards: 26 files / 1,092 tests passed before the additional byte-corruption case; the final full run includes that new case, and the final style-library suite passed all 12 tests.
- Type checking, changed-file ESLint, and `git diff --check` passed. Node type checking was repeated after the byte-level correction.
- Full `npm test -- --maxWorkers=8` completed: 27,596 passed, 18 failed, 251 skipped. This is not an all-green full-suite result:
  - Seven failures in literature-list rendering and Web download tests passed when rerun with one worker; the matching baseline checks also passed.
  - Eight Office-preview failures reproduced on the unchanged base commit (`547e1896`).
  - Three package-cache sandbox integration tests failed because the production sandbox denied executing Micromamba from the worktree dependency path.
- The earlier settings/runtime test failures caused by the agent execution sandbox disappeared outside that sandbox: all 405 tests passed.

The filesystem reproductions deliberately corrupt temporary files; they do not demonstrate normal writes causing corruption. The UI races are tested at the React/API boundary, not by scheduling full Electron IPC traffic.

### DIFF
--- a/src/main/literature/citation-formatter.test.ts
+++ b/src/main/literature/citation-formatter.test.ts
@@ -232,7 +232,7 @@ describe('LiteratureCitationFormatter', () => {
       await expect(
         formatter.parseReferences('@article{example, title={A useful paper}, year={2024}}')
       ).resolves.toMatchObject({ items: [expect.objectContaining({ title: 'A useful paper' })] })
-      expect((await styles.list()).some(({ id }) => id === `custom:${digest}`)).toBe(true)
+      expect((await styles.list()).some(({ id }) => id === `custom:${digest}`)).toBe(false)
       await styles.delete(`custom:${digest}`)
       expect((await styles.list()).some(({ id }) => id === `custom:${digest}`)).toBe(false)
     } finally {
--- a/src/main/literature/citation-style-library.test.ts
+++ b/src/main/literature/citation-style-library.test.ts
@@ -1,4 +1,4 @@
-import { mkdtemp, readFile, rm } from 'node:fs/promises'
+import { mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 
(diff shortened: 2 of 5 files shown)


ewen-poch:
Rebased onto current `main` (`0d3e2e88`) without textual conflicts. The PR still changes only the original five files. The three directly affected suites pass after the rebase: 34 tests.

On the Windows replacement finding: the reported condition of an already-existing destination does not itself establish a failure. Node's Windows filesystem implementation calls `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` ([libuv implementation](https://github.com/libuv/libuv/blob/v1.x/src/win/fs.c#L2177-L2183)). The current tests repair both truncated and valid-but-replaced files through the public import boundary and check the resulting bytes and formatted output. They are not a claim that this branch's tests were executed on Windows.

Transient sharing violations or externally locked files are a different condition and may warrant bounded retries once reproduced. No concrete locking/error-code reproduction was supplied for this finding. I am retaining atomic replacement rather than adding a speculative fallback, particularly an unlink-before-publish path that would weaken recovery safety. A reproducible failing Windows case would justify revisiting this decision.

The previous PR Gate failure came from cancelled macOS accessibility and Windows workspace E2E steps; their logs show cancellation, rather than an identified citation-style assertion failure. The rebased push starts fresh checks without weakening or changing the CI workflow.

PR #2350, opened 2026-09-08
fix(literature): isolate detail requests and recover saved metadata

ewen-poch:
## Problem

A delayed detail save or metadata lookup can update a later opening of the same reference. Link loading can also replace a reference the user selected more recently. Detail saves leave the page cache stale, background completion leaves the detail stale, and a committed save whose readback fails offers another write against the obsolete revision.

## Proposed change

Track detail opening generations and navigation intent, reject obsolete UI callbacks and older metadata snapshots, and route detail saves through the existing page cache. Publish fetched background results to the current detail, including references outside the displayed page. Keep relationship operations on their existing semantic update paths. Authoritative inline/background/save/completion data updates follow item identity and revision across openings; modes, errors and previews remain scoped to the originating operation. Successful PDF imports re-read the current item before publishing attachments. Background reads publish successful results even if another completed item cannot be read, retaining the existing error and Retry surface. Metadata revision changes clear obsolete previews and pending lookup results while retaining the identifier input for a fresh search; same-revision attachment updates preserve previews. Successful completion receipts remain visible.

Retain the editing snapshot and draft when metadata changes externally. Saving stays blocked until the user explicitly chooses to discard the draft and load the latest metadata. Distinguish a committed save from a failed readback: keep the submitted fields and offer read-only Retry instead of another transaction. Add translated recovery copy in all eight locales and use the existing ErrorNotice surface.

## Scope and non-goals

Renderer state and existing catalog API usage only. Added generation, request and recovery bookkeeping is memory-only. No database, persisted enum, data relationship, migration, historical compatibility layer, persistent draft or automatic merge. Existing optimistic revision checks remain authoritative.

- Mounted-page regressions control the existing API boundary; the original six failure scenarios failed twice on untouched production code before repair. The expanded final tests were replayed twice with base production files: the same 11 regressions failed and two controls passed. Six additional review regressions for valid late data publication and one partial-background-read regression failed twice before repair and now pass. Two preview/revision regressions also failed twice before repair; an attachment-only update control verifies preview preservation. These are renderer integration tests, not a two-window end-to-end claim.
- Detail lifecycle, draft retention, navigation, read-only recovery, background publication, cache reentry and editor/import/full-text consumers -> `npx vitest run src/renderer/src/pages/literature src/renderer/src/i18n/resources.test.ts --maxWorkers=1` -> 1,103 passed.
- Renderer interfaces -> `npm run typecheck:web` -> passed.
- Source conventions -> `npm run lint` -> passed.
- Existing real SQLite optimistic revision contract -> targeted `catalog.test.ts` test `updates citation metadata atomically with optimistic revision checks` -> passed.
- Controlled browser verification with real page components: saved-but-unloaded Retry restores the detail; background updates retain the draft; loading latest replaces the draft only after the explicit action. Screenshots were inspected, including notice bounds inside the dialog and a revision change clearing the preview while keeping the identifier and Search usable.

The final renderer/i18n suite, typecheck and lint run after the last material edit. An earlier parallel module run failed two existing pagination/reading cases; both passed in the focused rerun and the full sequential module rerun without changing their tests.

The local impact explainer reports an unknown module owner for the literature controller and therefore selects a full fallback. Per maintainer instruction, no full local suite was run: the explicit local set includes all literature renderer consumers and the i18n guard; PR CI remains authoritative for full/platform coverage. No main/preload runtime or persisted contract changed. Actual multi-window and packaged cross-platform behavior were not exercised locally.

## Review focus

Please check that stale writes still update the cache without changing a newer opening, recovery never resubmits an acknowledged transaction, background refresh preserves an editor's original draft/revision, and same-revision relationship updates remain accepted.

Rebased onto main at b5a3df4c. The PDF import conflict keeps main's specific error copy and this PR's generation guard. Local Prisma Client was regenerated to match main; no schema change is part of this PR. The first module run had a timeout in an existing project export test; its focused rerun and the final sequential module run passed without changing code or tests.

Latest rebase: main c9e66afd. Additive locale conflicts retain both main and PR entries. Final module/i18n count is 1,103 passed; web typecheck and lint passed. Functional patches are unchanged.

### DIFF
--- a/src/renderer/src/pages/literature/LiteratureDetailController.ts
+++ b/src/renderer/src/pages/literature/LiteratureDetailController.ts
@@ -2,6 +2,7 @@ import type { LiteratureItemView } from '../../../../shared/literature'
 
 export type LiteratureDetailSnapshot = Readonly<{
   item?: LiteratureItemView
+  generation: number
   open: boolean
 }>
 
@@ -14,7 +15,7 @@ export type LiteratureDetailController = Readonly<{
 }>
 
 const createLiteratureDetailController = (): LiteratureDetailController => {
-  let snapshot: LiteratureDetailSnapshot = { open: false }
+  let snapshot: LiteratureDetailSnapshot = { open: false, generation: 0 }
   const listeners = new Set<() => void>()
   const publish = (next: LiteratureDetailSnapshot): void => {
     snapshot = next
@@ -28,12 +29,12 @@ const createLiteratureDetailController = (): LiteratureDetailController => {
       return () => listeners.delete(listener)
     },
     close: () => {
-      if (!snapshot.open && !snapshot.item) return
-      publish({ open: false })
+      publish({ open: false, generation: snapshot.generation + 1 })
     },
-    open: (item) => publish({ item, open: true }),
+    open: (item) => publish({ item, open: true, generation: snapshot.generation + 1 }),
     replace: (item) => {
-      if (snapshot.item?.id !== item.id) return
+      if (snapshot.item?.id !== item.id || item.metadataRevision < snapshot.item.metadataRevision)
+        return
       publish({ ...snapshot, item })
     }
   }
--- a/src/renderer/src/pages/literature/LiteratureLibraryPage.render.test.tsx
+++ b/src/renderer/src/pages/literature/LiteratureLibraryPage.render.test.tsx
@@ -7264,5 +7264,178 @@ describe('LiteratureLibraryPage', () => {
       await act(async () => pending.resolve(reply))
       expect(screen.queryByRole('button', { name: 'Apply metadata' })).toBeNull()
     })
+    it('publishes an inline save to a detail opened during its readback', async () => {
+      await showLibrary()
+      const pending = deferred<LiteratureItemView>()
+      get.mockReturnValueOnce(pending.promise)
+      fireEvent.click(
+        screen.getByRole('combobox', { name: `Reference type: ${libraryItem.item.title}` })
+      )
+      fireEvent.click(screen.getByRole('option', { name: 'Preprint' }))
+      await waitFor(() => expect(get).toHaveBeenCalledTimes(1))
+      await openReferenceDetail(screen.getByText(libraryItem.item.title))
+      await act(async () =>
+        pending.resolve({
+          ...libraryItem,
+          metadataRevision: 2,
+          item: { ...libraryItem.item, itemType: 'preprint' }
+        })
+      )
+      expect(within(screen.getByRole('dialog')).queryAllByText('Preprint').length).toBeGreaterThan(
+        0
+      )
+      await editDetail()
+      fireEvent.click(screen.getByRole('button', { name: 'Save' }))
+      await waitFor(() =>
+        expect(transact).toHaveBeenLastCalledWith(
+          expect.objectContaining({ expectedMetadataRevision: 2 })
+        )
+      )
+    })
+
+    it.each(['view', 'edit'] as const)(
+      'publishes a pending background read to a subsequently opened %s detail',
+      async (mode) => {
+        const summary = {
+          id: 'opening-job',
+          mode: 'metadata' as const,
+          phase: 'apply' as const,
+          state: 'running' as const,
+          total: 1,
+          checked: 1,
+          ready: 0,
+          done: 0,
+          failed: 0,
+          createdAt: 1,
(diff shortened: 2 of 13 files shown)


ewen-poch:
Authoritative item updates are now independent of the detail opening that initiated them:

- Inline-save and background readbacks publish through the existing item-ID and metadata-revision guard. A detail opened during the read receives the updated data; an editor keeps its draft and original base revision and shows the existing external-update notice.
- Successful PDF imports re-read the current item before publishing attachments, with the successful receipt retained as a fallback if readback is unavailable. This avoids treating metadataRevision as an attachment revision and preserves newer metadata in a reopened detail.
- The same distinction applies to completed metadata saves and completion commits: publish valid data, but keep mode changes, errors and previews scoped to their originating UI operation.

No new state, persistence, architecture, data relationships or interaction options were introduced. Changes remain within files already in this PR. The branch was rebased onto main at 0d3e2e88.

Six new controlled-response page regressions failed twice before their respective fixes, covering the reported timing races and the equivalent save/completion paths. The original stale-response and draft-preservation regressions remain in place. Final module, i18n, typecheck and lint results are recorded in the PR description.

ewen-poch:
Background read failures no longer discard successful results from the same completion notification. Fulfilled reads update the existing cache and matching detail; rejected reads retain the existing loading error and Retry surface. The mounted-page regression failed twice before the fix (the detail retained its old title), and now verifies both successful publication and visible failure reporting.

This change uses Promise.allSettled within the existing page callback. It adds no state, persistence, data-model changes, automatic retry queue, or new interaction. Retrying failed IDs automatically is deferred because it would require additional lifecycle/state ownership beyond this repair.

The preview invalidation recommendation remains pending maintainer confirmation: clearing a visible preview and its Apply action when a newer revision arrives changes the lookup interaction. That behavior has not been changed in this commit.

Validation: 1,064 literature renderer/i18n tests passed; web typecheck and lint passed. No full local test suite was run, as requested.

ewen-poch:
Metadata completion now includes metadataRevision in its request identity. The existing detail subscription resets pending and visible completion state when the revision changes, so an obsolete review token cannot remain available through Apply. The lookup stays open with its identifier input retained for a fresh Search. Same-revision attachment updates preserve the preview, and publishing a successful completion still displays its receipt.

The maintainer approved this interaction adjustment before implementation. No new state field, persistence, data model, relationship, or architecture was introduced; changes remain in the hook and page test already included in this PR.

Both pending-preview and visible-preview page regressions failed twice before the fix because Apply remained visible. They now pass, including a fresh search/commit against revision 2. An attachment-only control and the existing successful-commit coverage also pass. Final validation: 1,067 literature renderer/i18n tests, web typecheck, and lint passed. Browser verification with real page components confirmed preview removal, retained DOI input, and successful fresh lookup; a screenshot was inspected. The browser harness uses controlled API responses, not a multi-window end-to-end setup.

ewen-poch:
Rebased onto main at b5a3df4c and resolved the PDF import error-handler conflict by retaining main's specific pdfImportErrorMessage output together with this PR's detail-generation guard. A range-diff confirms that this is the only adaptation to the existing patches; no interaction or design changes were added during conflict resolution.

The latest pre-rebase PR Gate (10789780) passed, as did the preceding two heads. The only failed PR Gate in the branch history belongs to the original 801b78eb head; its Windows E2E failure did not recur in the later runs. Fresh CI will validate the rebased merge state.

The identifier synchronization recommendation remains pending maintainer confirmation because it expands the changed-file scope and revises the approved input-retention interaction. No additional lookup-state change is included in this rebase.

Rebase validation: 1,094 literature renderer/i18n tests passed; web typecheck and lint passed. The local Prisma Client needed regeneration after main changed its schema. One existing project-export test timed out in the initial module run, then passed both a focused rerun and the final sequential module run without edits. No full local test suite was run.

ewen-poch:
Rebased onto main at c9e66afd. The conflicts were additive renderer translation entries in all eight locale files; both main's batch-capacity message and this PR's draft-recovery messages are retained. Range-diff confirms that the existing functional patches are unchanged. No new interaction, state, or persistence changes were introduced during this rebase.

Validation: 1,103 literature renderer/i18n tests passed, web typecheck passed, and lint passed. The local generated Prisma Client was refreshed to match the worktree schema before the final checks. The initial run used the stale client and also timed out in an existing export test; the final sequential run passed without changing production code or tests.

PR #2352, opened 2026-09-08
fix(literature): tolerate preference failures and validate batch capacity

ewen-poch:
Literature table preferences can currently throw during page mount when storage writes fail, and another window's column changes can be overwritten by a stale local snapshot. Bulk lookup also accepts selections that exceed the task API's 1,000-reference limit, while partial page selection has no mixed checkbox state.

This change:
- Keeps preferences usable in memory after a failed write, skips unnecessary mount writes, and retries only after another edit.
- Adopts external storage changes without echoing them and merges independent column visibility edits against the latest stored preferences. This remains lightweight preference synchronization, without atomic concurrent-write guarantees.
- Rejects oversized metadata/full-text selections before creating a task, preserves selection, and displays a translated limit message in all eight locales. Exactly 1,000 references remain supported; nothing is silently truncated.
- Sets the native page checkbox's indeterminate state from visible page members.
- Adds a real SQLite regression for moving all 26 selected references to Trash despite a concurrent metadata update. The existing membership snapshot fix already handles this; no additional production change is needed for that case.

No database migration, data-model change, or dependency is introduced.

Validation:
- Nine real-page regression/boundary cases. Reported failures were reproduced before fixing; the historical offset resolver repeatedly processed only 25 of 26 records, while the current snapshot resolver processes all 26.
- After rebasing: 20 test files / 1,072 tests passed across literature renderer tests, i18n guards, shared literature contracts, and duplicate catalog tests. Low-concurrency rerun used a 60-second test timeout; the SQLite regression uses the existing database test convention of 120 seconds after a loaded-machine timeout.
- Web typecheck, changed-file ESLint, and diff whitespace checks passed.
- Earlier full suite: 27,719 passed, 11 failed in unchanged Office preview (8) and Micromamba sandbox (3) tests. Full suite is not claimed green.

### DIFF
--- a/src/main/notebook/runtime-service.test.ts
+++ b/src/main/notebook/runtime-service.test.ts
@@ -3878,15 +3878,28 @@ describe('notebook runtime service', () => {
       }
     )
 
-    it.each(['linux', 'win32'] as const)(
-      'recovers a lost background Shell receipt and idempotently cancels running work on %s',
-      async (platform) => {
+    it.each([
+      { platform: 'linux', dispatchDelayMs: 0 },
+      { platform: 'win32', dispatchDelayMs: 0 },
+      { platform: 'linux', dispatchDelayMs: 1500 }
+    ] as const)(
+      'recovers a lost background Shell receipt and idempotently cancels running work on $platform after $dispatchDelayMs ms dispatch delay',
+      async ({ platform, dispatchDelayMs }) => {
         const root = await createStorageRoot()
+        const executionStarted = createDeferred<void>()
+        const dispatchAllowed = createDeferred<void>()
+        const repository = new NotebookRunRepository(root)
+        const transitionRun = repository.transitionRun.bind(repository)
+        vi.spyOn(repository, 'transitionRun').mockImplementation(async (request) => {
+          if (dispatchDelayMs && request.run.status === 'running') await dispatchAllowed.promise
+          return transitionRun(request)
+        })
         let executions = 0
         const execute = vi.fn<NotebookShellProcess['execute']>(
           (request) =>
             new Promise((resolve) => {
               executions += 1
+              executionStarted.resolve()
               request.signal?.addEventListener(
                 'abort',
                 () =>
@@ -3904,7 +3917,7 @@ describe('notebook runtime service', () => {
           configRoot: root,
           dataRoot: root,
           projectId: 'default-project',
-          repository: new NotebookRunRepository(root),
+          repository,
           shellProcess: { execute },
           backgroundExecutionEnabled: true,
           platform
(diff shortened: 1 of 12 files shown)


ewen-poch:
Rebased onto `41fdab24` from `main`. The only conflict was overlapping test imports; both schemas are retained. The preference fallback, column synchronization, batch-capacity guard, and page-selection behavior are unchanged.

The Codex review found no actionable defects, so no additional behavior, state, persistence, or architecture changes were introduced. After rebasing, all 975 real-page and i18n guard tests pass; web typecheck, changed-file ESLint, and whitespace checks also pass.

ewen-poch:
Rebased onto `b5a3df4c` from `main` and resolved the import-only conflict by retaining the attachment imports and the shared batch-capacity constant. The PR still changes the same 11 files; no additional state, persistence, data model, or interaction change was introduced.

The failed gate was caused by Windows E2E shard 2 exceeding its 25-minute job limit. GitHub's check annotation explicitly confirms the timeout; the log reports 11 functional tests and 17 workspace tests passed before one test was interrupted and one remained unrun. The rebase incorporates the already-merged CI fix from #2355: three Windows shards and a 35-minute job limit, rather than adding a separate workflow change here.

Validation after rebasing: 1,018 tests pass across the real literature page, i18n guards, CI workflow tests, and Playwright configuration tests. Web typecheck, changed-file ESLint, and whitespace checks pass. The new Windows jobs will verify the timeout fix; the new commit also receives a fresh AI review.

ewen-poch:
All three Windows E2E shards now pass on `063ebf8a`, confirming that the previous 25-minute timeout blocker is cleared after rebasing. The AI review for this exact commit is also mergeable with no actionable findings.

The remaining CI failure is `src/main/notebook/runtime-service.test.ts:3924`: “recovers a lost background Shell receipt and idempotently cancels running work on linux” expected the execution count to become 1 but observed 0. The merged report contains the same single failing test (26,996 passed); the coverage/module job failure propagates that result. This evidence does not yet distinguish a runtime scheduling defect from test synchronization instability.

No Notebook code or tests have been changed. Further changes to that area are outside this PR's current file scope and are paused pending scope approval. CI is not yet fully green.

ewen-poch:
The Notebook execution-count failure is reproducible when the existing queued-to-running repository transition is delayed by 1.5 seconds after the receipt is recovered: the original `vi.waitFor` assertion fails with the same expected-1/observed-0 result. The no-delay Linux and Windows controls pass.

The test now awaits the fake Shell's actual execution-entry signal instead of treating the default one-second polling window as a dispatch guarantee. It still checks receipt identity and exactly one execution, explicitly checks repeated cancellation, and settles background work in cleanup. The delayed case remains as a regression. Only `runtime-service.test.ts` changed; runtime scheduling, state, persistence, and user behavior are unchanged.

The old assertion failed twice with the delayed fixture. With the fix, all 374 runtime-service tests pass when native process execution is permitted; Node and sandbox typechecks, ESLint, and whitespace checks pass. The first restricted local run could not execute seven real-Shell cases, which all pass in the native rerun.

ewen-poch:
Validated on `aeb499ae`: all three Ubuntu full portable test shards now pass, including the previously failing shard 3/3. `Module tests and coverage` also passes, so the propagated report failure is cleared. The AI review for this exact commit is mergeable with no actionable findings.

The fix remains confined to test synchronization and cleanup; no production runtime, state, persistence, or user interaction changes were required.

PR #2359, opened 2026-09-08
fix(updater): explain Windows installer handoff failures

ewen-poch:
## Problem

On Windows, Restart hands the update to NSIS and exits before the installer's result
is observable by the app. If the existing uninstaller is not writable, NSIS silently
exits 2 and leaves the old version launchable. Users receive no explanation or
recovery action. A real NSIS fixture reproduces both access denial and file sharing
violations; this does not prove the original reporter's ACL or antivirus state.

## Proposed change

For the existing interactive `/S --updated --force-run` handoff, show the affected
file and Windows error. Access denial explains manual administrator installation;
file occupation offers Retry/Cancel. Keep the write preflight, old-app/data
preservation, and unattended silent behavior. Do not elevate automatically.

## Scope and non-goals

Only the native installer failure interaction changes. No renderer, architecture,
data model, fields, enum states, persisted format or migration changes. Existing
0.25.1 handoff flags work with a future target installer; this does not modify the
already released 0.26.0 installer. Antivirus attribution, full Electron relaunch,
and automatic elevation are outside this fix.

- Native permission/lock/cancel/retry/unattended behavior ->
  `OPEN_SCIENCE_UPDATE_NOTICE_FIXTURE=<built fixture> node node_modules/vitest/vitest.mjs run scripts/windows-update-notice.test.ts --maxWorkers=1`
  -> 5 passed. Actual native dialogs were inspected and captured locally.
- Updater handoff, Windows installer/certification and packaging contracts ->
  `node node_modules/vitest/vitest.mjs run src/main/update/electron-updater-strategy.test.ts scripts/windows-updater-certification.test.ts scripts/windows-installer-smoke.test.ts scripts/electron-builder-config.test.ts --maxWorkers=1`
  -> 107 passed; combined final run: 112 passed in 5 files.
- Regression provenance: before the hook change, 3 interactive cases failed
  specifically because NSIS exited 2 without a notice and retained the old version;
  2 unattended cases passed. The access-denied failure was repeated before fixing.
- `rtk npm run lint` -> passed (final cached rerun after the last material edit).

The native test uses isolated product identity/install/profile paths and cleans up
its fixture installation; no production test seam. See the fixture README for
prerequisites and reproduction. Ordinary portable CI skips this opt-in desktop
suite, which is not native certification. Existing portable contract tests run in
CI. No runtime TypeScript/configuration changed, so runtime process typechecks and
renderer tests are not in the impact set. No full local npm test was run, as
requested. Direct Vitest was used because the shared dependency tree's unrelated
Prisma-client pretest guard is stale; this was not counted as regression evidence.

Uncovered risks: actual reporter ACL/security-product state, full Electron relaunch,
uncommon Windows errors and elevated manual recovery on a complete installation.
Independent reviews of the final diff: Standards — no findings; Spec — no findings. The final check mapping covers the changed installer boundary and existing updater/packaging consumers.

## Review focus

Confirm interactive Restart can explain failure without blocking unattended update
jobs, and the existing write-access/data-preservation safeguards remain intact.

### DIFF
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -245,7 +245,8 @@ FunctionEnd
   # Checking only after uninstalling can reject the update with the old executable already gone.
   # A stale locked uninstaller must not turn a silent install into a false success or offer the unsafe
   # Ignore choice. OPEN_EXISTING keeps this check non-destructive; Retry lets the user release a
-  # transient lock, while silent installs take the Cancel path and return a failure code.
+  # transient lock. Unattended installs stay silent; a user-requested restart must explain failure
+  # even though electron-updater starts NSIS with /S and exits before observing the result.
   Function ensureExistingUninstallerIsWritable
     IfFileExists "$INSTDIR\${UNINSTALL_FILENAME}" ensureExistingUninstallerIsWritable_retry ensureExistingUninstallerIsWritable_done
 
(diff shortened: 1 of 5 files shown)


ewen-poch:
The proposed change from control ID 2 to ID 1 does not match the native dialog observed on the Windows test host.

- The real access-denied NSIS dialog exposes its single OK button as control ID 2. The earlier ID 1 implementation failed with `Expected notice button 1 was absent` during local validation.
- I reran the existing regression against this PR commit with `OPEN_SCIENCE_UPDATE_NOTICE_FIXTURE=<built fixture> node node_modules/vitest/vitest.mjs run scripts/windows-update-notice.test.ts -t 'denied:' --maxWorkers=1`: **1 passed, 4 skipped**. It dismisses the actual dialog, observes installer exit 2, and verifies that the previous executable remains launchable.
- `GetDlgItem` looks up an actual child control identifier. The documented `MessageBoxW` return value `IDOK` is a different API observation; it is not evidence that this native child has ID 1. References: [GetDlgItem](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdlgitem), [MessageBoxW return values](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messageboxw#return-value).

I am retaining the tested button ID rather than introducing the locally reproduced failure suggested by this review. The installer behavior and its write-access safeguard remain as reviewed.

The CI failure is separate: `src/main/notebook/runtime-service.test.ts:4087` fails while waiting for the first queued shell command to start (`expected ['first'], received []`). The portable report contains one failed test; Windows and macOS E2E passed. That Notebook test is outside this PR's current file scope, so a scope confirmation has been requested before changing it.

ewen-poch:
Rebased onto `af6d20094278ac76524e1bb46dd787fd9d1e313e` (latest `origin/main` fetched for this update). The new head is `d932d20c9fc7b0368335aff490553ad6843aad4a`.

The rebase was conflict-free, and `git range-diff` confirms that the patch is unchanged. The PR still changes only the original five installer/test files. No Notebook runtime or test changes were added.

Post-rebase validation: the native installer and updater/packaging checks passed (112 tests), the previously failing Notebook queue test passed in a focused run (1 passed, 372 skipped), and lint passed. The Notebook test still contains the same initial polling assertion in the new base, so these local results do not establish that its intermittent CI failure is fixed. The new-head CI run remains authoritative.

ewen-poch:
I checked the normalized review findings for head `d932d20c9fc7b0368335aff490553ad6843aad4a` in the completed review job. GitHub suppressed the `review_body` job output and skipped publishing the feedback comment, so the workflow's successful execution is not being treated as an approval.

**NSIS `isForceRun`:** The packaging-blocker finding is not supported by the locked dependency or by a real compile. `app-builder-lib` 26.15.3 calls `scriptGenerator.flags(["updated", "force-run", ...])` in `out/targets/nsis/NsisTarget.js:579`, before adding the custom installer include. `NsisScriptGenerator.flags()` converts the hyphenated flag to `isForceRun` and emits its macro and `!define`. The dependency's own `templates/nsis/installSection.nsh` also uses this condition. I rebuilt both versioned NSIS fixture installers from the current head with electron-builder 26.15.3 and cached makensis; both compiled successfully. Adding a second parser or project-local definition would duplicate the dependency's existing behavior.

**Access-denied OK control:** This repeats the previously answered finding. The actual single-button dialog on the Windows test host exposes control ID 2, and ID 1 was absent during native inspection. The rebuilt real-installer regression suite passed all 5 scenarios, including access denial, cancellation, and releasing the lock before Retry. There is no reproduced failure supporting the proposed ID substitution.

Both findings are therefore left unchanged, with the native evidence retained. No file-scope expansion, installer behavior change, or CI workflow change was introduced to address these unsupported findings.

ewen-poch:
The remaining Windows E2E shard failure occurred before the settings test body: the startup fixture was still in `migrating` after its 60-second readiness deadline. The retained screenshot shows the database update screen. The same test passed on its automatic retry in 39.4 seconds; the functional report records 6 expected tests, 1 flaky test, and 0 unexpected tests. The job remains red because `--fail-on-flaky-tests` is enabled.

I have rerun only the failed Windows E2E shard on the unchanged head `d932d20c9fc7b0368335aff490553ad6843aad4a`. No readiness timeout, assertion, retry policy, or production code has been changed. The existing artifacts do not establish why the first migration exceeded the deadline, so they do not justify a database or installer change.

The rerun's functional journeys step passed from 11:25:15 to 11:28:04 UTC with `--fail-on-flaky-tests` still enabled. The previously failing step is now successful on the unchanged head; this does not establish a fix for the underlying intermittent migration delay. The same shard is continuing its workspace journeys, so the overall PR Gate is still pending.

PR #2360, opened 2026-09-08
fix(compute-host): save edited skeleton details against the empty stored document

### LINKED ISSUE #2358: [Bug]: Unable to save SSH compute details: `old_text does not match the current details document`
### Preflight checklist

## Bug Description

When editing the custom **Details** of an SSH compute host, saving the changes fails with the following error:

```text
Error invoking remote method 'compute:details:save':
Error: replaceDetails: old_text does not match the current details document for "ssh:open-science-server".

### Steps to reproduce

### Steps to Reproduce

1. Open **Open Science** on Windows.

2. Go to:

   `Settings → Compute`

3. Add an SSH compute host using an existing SSH configuration.

   Example configuration:

   ```ssh
   Host open-science-server
       HostName 10.143.0.121
       User user
       ServerAliveInterval 30
       ServerAliveCountMax 3
````

4. Select **SSH Config** as the authentication method.

5. Test and save the SSH compute host.

6. The SSH connection works correctly and Open Science successfully detects the remote compute resources.

   For example, the automatically generated Details section contains:

   ```text
   ## Resources

   The CPU, memory and GPU values below describe the SSH login host, not a scheduler allocation.
   Inspect scheduler partitions and provider guidance before requesting compute resources.

   cpus: 128
   mem: 251 GB
   gpus: 4x NVIDIA GeForce RTX 4090 D
   scheduler: slurm
   ```

7. Open the successfully detected SSH compute host:

   `ssh:open-science-server`

8. In the **Details** section, click **Edit**.

9. Append or replace the Details text with custom compute instructions.

   For example:

   ```text
   ## Usage Policy

   Although Slurm is detected on this server, do NOT use Slurm or any scheduler unless explicitly requested.

   Use direct SSH terminal execution for normal compute tasks.

   Preferred project root:

   /home/user/ysy

   Use one GPU by default for debugging and testing.
   Check GPU availability with nvidia-smi before starting GPU-intensive work.
   ```

10. Click **Save**.

11. The save operation fails and the following error is displayed:

```text
Error invoking remote method 'compute:details:save':
Error: replaceDetails: old_text does not match the current details document for "ssh:open-science-server".
```

12. Reopening the compute host shows that the custom Details text was not saved.

### Expected Behavior

The edited Details document should be saved successfully.

If the Details document has been modified in the background after the editor was opened, Open Science should either:

* automatically reload or merge the latest version;
* notify the user that the document has changed and ask them to reload;
* or provide a conflict-resolution mechanism.

The user-edited Details should not fail with a raw `old_text does not match` error.

### Actual Behavior

Saving the edited Details fails with:

```text
replaceDetails: old_text does not match the current details document
```

The custom compute instructions are therefore not saved.

### Additional Notes

* SSH itself works normally outside Open Science.
* The SSH compute host can be successfully detected by Open Science.
* CPU, memory, GPU and scheduler information are detected correctly.
* The failure occurs specifically when editing and saving the **Details** document.
* The issue appears after automatic compute detection has already generated content in the Details section.
```


### Operating system

Windows

### App version

0.26.0

### Provider / model

_No response_

```shell

```

chiliec:
Closes #2358

## What

Editing the **Details** of a compute host right after a successful probe failed with:

```
replaceDetails: old_text does not match the current details document for "ssh:..."
```

When the host has no stored details, `detailsGet` returns a probe-generated skeleton (`isSkeleton: true`) and the editor uses it as both the draft and `originalDoc`. On save, that skeleton text was sent as `old_text`, but the stored `detailsDoc` is still `''`, so the compare-and-set guard in `replaceDetails` rejected every first edit of auto-generated details.

## Fix

`handleDetailsSave` now sends `''` as `old_text` when the loaded document is a skeleton — the same thing `reloadDetailsForMerge` already does (`setOriginalDoc(skeleton ? '' : doc)`). One-line change in `ComputeHostDetail.tsx`.

## Tests

Added a render test that loads a skeleton details doc, edits it, saves, and asserts `saveDetails` is called with `oldText === ''`.

- Without the fix: `expected "" ... received "## Resources\ncpus: 128"` (test fails)
- With the fix: passes

```
npx vitest run src/renderer/src/pages/settings/ComputeHostDetail.render.test.tsx src/renderer/src/pages/settings/compute-i18n.render.test.tsx
 Test Files  2 passed (2)
      Tests  57 passed (57)
npx eslint <both files>      → clean
npx prettier --check <both>  → clean
```

I could not run the full `npm run typecheck:web` locally (tsc runs out of memory on my box); the change only reuses the existing `isSkeleton` state so I expect CI to cover it.

### DIFF
--- a/src/renderer/src/pages/settings/ComputeHostDetail.render.test.tsx
+++ b/src/renderer/src/pages/settings/ComputeHostDetail.render.test.tsx
@@ -1391,6 +1391,38 @@ it('keeps the draft and reloads a merge base after a details conflict', async ()
   expect(saveDetails).toHaveBeenLastCalledWith('ssh:biowulf', 'my draft', 'other writer')
 })
 
+it('saves edits to auto-generated details against the empty stored document', async () => {
+  stubDetailsGet('## Resources\ncpus: 128', true)
+  const saveDetails = vi.fn().mockResolvedValue(undefined)
+  useComputeStore.setState({ hosts: [host({ detailsDoc: '' })], saveDetails })
+  await act(async () => root.render(<ComputeHostDetail providerId="ssh:biowulf" />))
+  const section = Array.from(
+    container.querySelectorAll<HTMLElement>('[data-slot="settings-section"]')
+  ).find((section) => section.querySelector('h3')?.textContent === 'Details')!
+  const click = async (text: string): Promise<void> => {
+    await act(async () =>
+      Array.from(section.querySelectorAll('button'))
+        .find((button) => button.textContent?.trim() === text)!
+        .click()
+    )
+  }
+  await click('Edit')
+  const draft = section.querySelector('textarea')!
+  act(() => {
+    Object.getOwnPropertyDescriptor(HTMLTextAreaElement.prototype, 'value')!.set!.call(
+      draft,
+      '## Resources\ncpus: 128\n\n## Usage Policy\nUse direct SSH.'
+    )
+    draft.dispatchEvent(new Event('input', { bubbles: true }))
+  })
+  await click('Save')
+  expect(saveDetails).toHaveBeenCalledWith(
+    'ssh:biowulf',
+    '## Resources\ncpus: 128\n\n## Usage Policy\nUse direct SSH.',
+    ''
+  )
+})
+
 // Exercise live app-language changes while Intl retains the host's default locale.
 it('updates metadata dates with the interface language on an unchanged host', async () => {
   const timestamp = '2026-09-02T12:00:00.000Z'
--- a/src/renderer/src/pages/settings/ComputeHostDetail.tsx
+++ b/src/renderer/src/pages/settings/ComputeHostDetail.tsx
@@ -303,7 +303,7 @@ export function ComputeHostDetail({
     setDetailsSaving(true)
     setDetailsError(undefined)
     try {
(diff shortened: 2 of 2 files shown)


ewen-poch:
Thank you for tracking this down and adding a focused regression test. The one-line fix correctly addresses the first-save failure in Settings and is a reasonable minimal patch for that UI path.

After comparing both approaches, we've decided to proceed with #2361 as the final solution for #2358. The limitation here is that the shared read contract still returns a generated resource summary as `doc` while the persisted document is empty. The Settings-specific conversion fixes that caller, but agents following the documented read → replace workflow with `oldText: info.doc` can still encounter the same mismatch.

#2361 makes the underlying contract simpler: `doc` always contains the exact persisted text, and resource observations remain separate in `probe` and the existing resource panel. This removes the need for skeleton-specific handling across callers while preserving guarded writes and existing saved documents. It also improves the Settings message for genuine edit conflicts.

To be precise, #2361 is a larger change and includes a documented API compatibility change; its advantage is a simpler, consistent shared contract rather than the smallest diff. It still needs its outstanding validation completed before merging.

We appreciate the clear diagnosis and test in this PR—they capture the failure well. Linking #2361 here so the chosen approach and the relationship between the two PRs are explicit.

PR #2365, opened 2026-09-08
fix(literature): preserve save outcomes and search contracts

ewen-poch:
## Problem

A batch save could commit an initial candidate and then return only an error, hiding the completed work. Cancelling during metadata lookup could still lead to a write. Save cards also counted duplicate receipts and existing library items as Inbox saves. Search previews mixed contributor roles, truncated exact identifiers, searched serialized field names in selected-item scope, and omitted presentation text from their response budget. File-read failures were incorrectly reported as malformed candidate JSON.

## Proposed change

- Keep sequential candidate transactions and return the completed receipt prefix with optional `failure { inputIndex, code, message }` or `cancelled`. Stop before the next write; allow an already-started transaction to finish.
- Propagate cancellation through the MCP handler, trusted runtime wrapper, metadata lookup and file read. The stateless HTTP host forwards Library cancellation notifications to the active request on that same route. Cancellation arriving while a POST body is still being read is retained only for already-arriving same-route requests, with a byte budget, and checked before tool dispatch. These in-memory entries expire with the response; duplicate active IDs fail closed, and route removal/host shutdown closes active responses.
- Derive save counts from unique receipt identities and existing states. Show separate pending-review and existing-library results, repeated inputs, and partial failure. Reuse Inbox and exact-item navigation. Historical raw receipts override inaccurate presentation counts; summary-only history stays neutral and is not rewritten.
- Return only authors in `authors`, retain full identifier values, apply selected IDs through Catalog search while retaining merged aliases and filtering survivor metadata, and budget both emitted search text blocks with advancing pagination. Distinguish unreadable/oversized files from invalid content.

## Scope and non-goals

No database columns, tables, migrations, backfills, entity relationships, or lifecycle enum values change. Added fields are result metadata, an optional read-request ID filter, and cancellation parameters. New tool-result metadata can use the existing conversation persistence path. No durable operation journal, background task system, batch transaction, or historical conversation rewrite is introduced.

The resolver still fetches each normalized identifier once, but preserves duplicate input positions so receipts and failure indices align with the caller's input. A cancelled client may discard the final response: this change prevents further writes but does not promise durable receipt delivery after cancellation or a crash. The full-abstract tool and candidate dismissal lifecycle are unchanged.

The original behavior regressions failed twice on unchanged production sources with matching assertion causes, then passed after the fixes. Additional real HTTP cancellation tests also failed twice against the original HTTP host. Test-only mocks intercept existing runtime composition boundaries; save/search implementations are not copied and no production test seam was added.

Checks below ran on the final code after the last material production edit:

| --- | --- | --- |
| SQLite partial save, cancellation, duplicate/existing receipts, search projection, query membership, file failures | `npx vitest run src/main/literature` | Pass |
| Trusted-origin/signal forwarding for Claude Code, OpenCode, Codex Responses and Codex bridge; HTTP cancellation/isolation/cleanup | `npx vitest run src/main/acp/runtime-composition.test.ts src/main/acp/runtime-base-composition.test.ts src/main/acp/mcp-http-host.test.ts src/main/acp/session-capability-owner.test.ts` | Pass |
| Shared request/receipt contract | `npx vitest run src/shared/literature.test.ts` | Pass |
| Cards, historical output, navigation and actual Library detail destination | `npx vitest run src/renderer/src/pages/workspace/WorkspaceToolDetailsRow.render.test.tsx src/renderer/src/pages/workspace/PermissionApprovalControls.render.test.tsx src/renderer/src/pages/workspace/literature-save-contract.test.tsx src/renderer/src/pages/workspace/workspace-tool-activity-details.test.ts src/renderer/src/stores/navigation-store.test.ts src/renderer/src/pages/literature/LiteratureLibraryPage.render.test.tsx` | Pass |
| Eight locale catalogs | `npx vitest run src/renderer/src/i18n/resources.test.ts` | Pass |
| Types and lint | `npm run typecheck:node`, `npm run typecheck:web`, `npm run lint` | Pass |

The final combined targets cover **1,682 tests**. The integration run passed 1,673; eight translation guards loaded before the conflict cleanup and passed on the corrected files, and the 50,000-record concurrency test passed on its standalone rerun. That capacity test has intermittently failed in combined local runs; its passing rerun is not proof of stability. Both process typechecks and lint pass after integration. A later test-only correction updates the existing details-row integration case to exercise the explicit Inbox button with real receipts, and neutral summary-only history. Its renderer consumer set passes **159 tests**, with targeted lint passing; production sources are unchanged. HTTP checks require a loopback listener. Real production cards were rendered with isolated fixtures and checked in ego-browser; the Inbox and exact-item actions were exercised, and screenshots were delivered locally to the maintainer.

The committed `test:affected:explain -- --base b28b9e11 --head HEAD` plan selects the full fallback because the HTTP host is not yet an owned module in the manifest. Per the maintainer's explicit instruction, no full local suite was run and the manifest was not broadened merely for this PR. The complete/platform checks belong to PR CI. Local tests do not claim live model-provider or cross-platform execution. CodeGraph's shared index belongs to another checkout; its structural findings were checked against the current worktree references and consumer tests.

## Review focus

The branch was rebased only to resolve actual merge conflicts. Catalog retains main’s normalized search, task recovery and metadata integrity work, adding parameterized selected-ID/alias filtering. Translation conflicts preserve both additions and main’s removal of obsolete author labels. The module verification was rerun after this integration.

Independent review identified request-registration cancellation and merged-alias regressions. Both were reproduced twice through HTTP/SQLite before correction. CI also exposed an existing approval-card count test; it failed twice locally before restoring its approval-only display. One local 50,000-record concurrency test failed during the first final run, then passed standalone and in the repeated module set; that failure is retained in local evidence. The subsequent macOS journey lane passed. The next portable run exposed an outdated details-row test that expected a whole-card button from summary-only output; it was reproduced locally, then updated to verify both receipt-backed navigation and neutral history. The latest exact-head run remains authoritative.

Review the completed-prefix contract, cancellation forwarding across separate HTTP POSTs and cleanup, exact identifier preservation when shrinking a search page, and neutral handling of incomplete historical tool output. Please assess the final CI plan and independent review before treating the change as fully verified.

### Review clarification: HTTP keep-alive cleanup

The latest static review inferred that `ServerResponse.close` waits for the TCP connection to close. This is inconsistent with [Node's documented response lifecycle](https://nodejs.org/api/http.html#event-close_2): response completion itself emits `close`.

The unchanged production host was independently exercised twice with a real Node HTTP Agent (`keepAlive: true`, `maxSockets: 1`): 20 sequential `tools/call` POSTs reused the same route and JSON-RPC ID. Every response was HTTP 200, calls 2–20 asserted `request.reusedSocket === true`, all 20 asserted one local socket port, and the save handler ran exactly 20 times. No duplicate 409 or retained ID was observed. The temporary verification test is preserved locally; production code and committed tests were restored byte-for-byte. Existing committed HTTP coverage also tests ID reuse after completion/cancellation. No speculative production fix was applied for this unconfirmed finding.

### Final review disposition and CI

All portable shards, Windows core/E2E, macOS build/E2E, Linux isolation, coverage, static checks, CodeQL and PR Gate passed for `95d365b90c800734311ed7bcc25629840125b317`.

A second static review inferred an unhandled write after destroying a response whose body is still arriving. Without production changes, real incomplete HTTP POSTs were tested under route unregister, host clear and cancellation-buffer overflow. All three cases passed twice: the client received the expected ECONNRESET, Vitest reported no unhandled exception/rejection, no save occurred, and re-registering the route served a successful request. The temporary verification is retained locally and the committed test file was restored byte-for-byte. This alleged secondary write failure could not be reproduced; no speculative guard was added.

The concrete cancellation-registration and merged-alias findings were reproduced and corrected. The unchanged production code was independently reviewed as mergeable at `cdaecd72`; the only subsequent commit changes details-row integration tests. Later keep-alive and destroyed-response reports are not treated as confirmed defects because both were tested directly and contradicted their claimed observable outcomes. Local capacity-test intermittency remains disclosed above despite its successful final CI run.

### DIFF
--- a/src/main/acp/mcp-http-host.test.ts
+++ b/src/main/acp/mcp-http-host.test.ts
@@ -1,5 +1,5 @@
 import { mkdtemp, rm, writeFile } from 'node:fs/promises'
-import { createServer, type Server } from 'node:http'
+import { createServer, request as httpRequest, type Server } from 'node:http'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 
@@ -411,6 +411,79 @@ describe('AgentMcpHttpHost', () => {
       await client.close()
     }
     expect(saveToInbox).not.toHaveBeenCalled()
+  })
+
+  it('honors cancellation while the original Library request body is still arriving', async () => {
+    host = new AgentMcpHttpHost()
+    const { token } = await host.ensureStarted()
+    const saveToInbox = vi.fn(async () => ({ results: [] }))
+    host.registerLiteratureLibrary('slow-body', {
+      searchLibrary: vi.fn(),
+      readAbstract: vi.fn(),
+      readPdf: vi.fn(),
+      saveToInbox
+    })
+    const url = host.urlFor('library', 'slow-body')
+    const headers = {
+      authorization: `Bearer ${token}`,
+      'content-type': 'application/json',
+      accept: 'application/json, text/event-stream'
+    }
+    const body = JSON.stringify({
+      jsonrpc: '2.0',
+      id: 71,
+      method: 'tools/call',
+      params: {
+        name: 'save_to_inbox',
+        arguments: {
+          candidates: [
+            {
+              item: { itemType: 'journalArticle', title: 'Cancelled before body completion' },
+              source: { provider: 'test', rawMetadata: {} }
+            }
+          ]
+        }
+      }
+    })
+    let finish!: () => void
+    let accepted!: () => void
+    const bodyAccepted = new Promise<void>((resolve) => {
+      accepted = resolve
+    })
+    const pending = new Promise<void>((resolve, reject) => {
+      const request = httpRequest(
+        url,
+        { method: 'POST', headers: { ...headers, expect: '100-continue' } },
+        (response) => {
+          response.resume()
+          response.on('end', resolve)
+        }
+      )
+      request.on('error', reject)
+      request.on('continue', accepted)
+      request.write(body.slice(0, -1))
+      finish = () => request.end(body.slice(-1))
+    })
+    await bodyAccepted
+    // This notification can complete while the earlier POST is waiting for its final byte.
+    const cancelled = await fetch(url, {
+      method: 'POST',
+      headers,
+      body: JSON.stringify({
+        jsonrpc: '2.0',
+        method: 'notifications/cancelled',
+        params: { requestId: 71 }
+      })
+    })
+    expect(cancelled.status).toBe(202)
+    finish()
+    await pending
+    expect(saveToInbox).not.toHaveBeenCalled()
+    // The same ID is reusable after this POST closes; cancellation must not poison a future call.
+    const repeated = await fetch(url, { method: 'POST', headers, body })
+    expect(repeated.status).toBe(200)
+    await repeated.json()
+    expect(saveToInbox).toHaveBeenCalledOnce()
   })
 
   it('isolates concurrent Library request IDs by route and releases completed or cancelled entries', async () => {
--- a/src/main/acp/mcp-http-host.test.ts
+++ b/src/main/acp/mcp-http-host.test.ts
@@ -8,6 +8,7 @@ import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/
 import { afterEach, describe, expect, it, vi } from 'vitest'
 
 import { AgentMcpHttpHost } from './mcp-http-host'
+import { literatureItemInputSchema } from '../../shared/literature'
 import { ArtifactRepository } from '../artifacts/repository'
 
 describe('AgentMcpHttpHost', () => {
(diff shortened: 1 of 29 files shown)


ewen-poch:
Review follow-up for `95d365b90c800734311ed7bcc25629840125b317`:

- **Cancellation before request registration — addressed.** A real HTTP test leaves the original POST body incomplete, delivers cancellation, then finishes the body. It reproduced an unwanted save before the fix. The host now retains bounded cancellation information only for already-arriving requests on the same route and checks it before tool dispatch. Entries are removed on response completion or route/host cleanup; future requests are not poisoned by old cancellation IDs.
- **Selected merged aliases — addressed.** A real SQLite test creates references through Catalog, merges one into another, and searches using the old selected ID. It reproduced a missing result before the fix. Catalog search now joins the selected ID to its survivor, filters survivor metadata, and returns the requested ID. Coverage includes selecting both IDs and paginating correctly. Both original behavior tests failed twice before correction and passed afterward.
- **Keep-alive request cleanup — no production change.** The finding assumes `ServerResponse.close` only follows TCP connection closure. [Node documents response completion itself as a close-event condition](https://nodejs.org/api/http.html#event-close_2). With the production host unchanged, two independent executions each sent 20 sequential calls with the same route and JSON-RPC ID through a single keep-alive socket. All returned HTTP 200; calls after the first asserted `reusedSocket === true`, all calls used the same local port, and the handler ran exactly 20 times. The reported duplicate 409 was not reproduced, so an additional cleanup path is not justified by that report.
- **Writing after response destruction — no production change.** Real incomplete POSTs were tested under `unregister`, `clear`, and cancellation-buffer overflow, twice per path. Each produced the expected client ECONNRESET, no save, no unhandled exception/rejection reported by Vitest, and successful service after route re-registration. The reported secondary write failure was not reproduced. An extra guard could be defensive hardening, but the evidence does not establish a required fix for the reported symptom.

The supplemental verification used existing public HTTP boundaries without changing production code; temporary verification sources and results are retained locally, and the committed test file was restored byte-for-byte. Current production sources match the previously reviewed `cdaecd72` state; the subsequent commit only updates details-row integration tests to cover the explicit Inbox action with real receipts and neutral summary-only history.

All selected PR checks passed for the current head. No additional architecture, data-model, relationship, lifecycle-enum, persistence, or interaction changes were introduced by this review follow-up.

PR #2367, opened 2026-09-08
fix(literature): bound large reads and checkpoint writes

ewen-poch:
## Problem

Valid large Library pages and completed review tasks can exceed the Web RPC 16 MiB response budget. Conflict-heavy duplicate keys perform excessive pairwise comparisons, while batch checkpoints repeatedly rewrite previous previews. Task controls also load and clone complete payloads unnecessarily, and applying the display limit before MCP compaction prevents agent searches from returning valid oversized records.

## Proposed change

- Bound Library transport pages by serialized bytes and assemble the requested display page, preserving complete editable records and numbered navigation. Selection consumers recover exceptional records through the existing digest-checked chunk transfer. Count-only queries avoid loading metadata.
- Return bounded, version-checked task review pages. Save all continuation pages before replacing dialog rows or clearing pending selections. Replayed create requests restore review data after restart.
- Offer lossless JSON downloads for oversized retained records, including Trash, without restoring or mutating records. Join chunks before Unicode encoding; exports contain attachment descriptors, not attachment binaries.
- Partition duplicate keys by common identity schemes before comparing representatives, preserving compatible cross-key joins and the existing ambiguous-key fallback.
- Store large row payloads separately from compact task controls. Commands copy only controls; selected candidate validation and workers read rows as needed. Successful row checkpoints release payloads. Retry rewrites one affected payload at a time, retaining item snapshots and completed-row digests, then atomically publishes the new header.
- Route the main-process agent Library query through the same filtered, alias-aware catalog query before the existing MCP projection/output limit. Renderer/Web commands keep their byte limit; no unbounded search command is exposed over IPC.

## Scope and non-goals

Task journals change to **v3**, with epochs and row payload references. Legacy **v1/v2** journals remain authoritative until the complete v3 index is durable; interrupted migration is retryable. New binaries read old tasks; downgrade compatibility is not promised. No database tables, columns, relationships or task status enums change. The lazy-control follow-up adds no further persisted fields or migration.

Transport segmentation adds no manual paging interaction. No metadata is truncated before editing or applying a review. This does not promise linear duplicate detection for every ambiguous dataset or add a separate worker process. Main-process scope binding, evidence ownership and cancellation behavior remain shared across Claude Code, OpenCode, Codex Responses and Codex Bridge.

After the latest rebase and material edits, the targeted command below passed **878 tests across 5 files**. The latest test-only repair covers relocation, durable header recovery, retained row payloads, migration admission, task persistence and locale completeness. Earlier module validation covered catalog transport, renderer/IPC consumers and all four ACP paths. `npm run typecheck`, `npm run lint`, and `npm run check:web-api-map` also passed, as did the changed-file formatting check. The three control/worker regression cases each failed on the unrepaired implementation in two consecutive runs before passing after repair.

```sh
npx vitest run src/main/storage/literature-migration.integration.test.ts src/main/literature/batch-job-journal.test.ts src/main/literature/batch-jobs.test.ts src/main/literature/migration-writers.test.ts src/renderer/src/i18n/resources.test.ts
```

The latest CI failure was reproduced locally: the relocation integration still expected a v2 index and flat checkpoint. Its assertions now match the existing v3 layout and verify retained payload recovery after relocation. The previous Windows language-selection timeout passed its automatic retry but remained red under fail-on-flaky; no timeout or policy was relaxed.

The earlier broader run passed 2,052 tests but found two old checkpoint-path assertions and two timeout failures under concurrent local workloads. The checkpoint assertions now read the existing v3 header and retain their migration pause/drain checks. All seven migration tests pass. The unchanged catalog capacity test and the Library page duplicate-navigation test both pass when rerun separately; no timeout or assertion was relaxed.

Initial reproductions failed before repair with actual HTTP response-budget errors, 499,500 identity probes for 1,000 conflicting candidates, and approximately 3.94x checkpoint bytes when doubling task size. Logical write bytes are measured at the existing durable-writer boundary, not as SSD wear. The scale fixture now writes readable temporary checkpoints for lazy reads; journal tests independently cover atomic publication and recovery.

Follow-up regressions cover missing/corrupt task removal, incremental reads, complete saved selections, replayed creation, oversized selection/Trash export, and a real production ACP handler returning an MCP error for an 11 MiB abstract before repair. Control/worker regressions establish unrelated payload reads and retained row snapshots before repair; they do not claim a measured process OOM. Retry failure keeps the old header intact and successful retry preserves completed payload digests. All four agent capability paths exercise delayed-result ownership after prompt replacement.

Browser checks mounted real components with labeled controlled fixtures: a 1,000-row review retained later-page selections and applied exactly 999 checked rows; global search displayed an 11 MiB record after 45 bounded chunks. No personal catalog was used. The latest backend changes add no new interaction.

The affected-test planner requests a full fallback because `src/main/acp/runtime-composition.ts` has no registered module owner. The maintainer explicitly requested module tests locally; full portable and platform coverage remains with PR CI. Capacity checks have shown sensitivity to concurrent local verification workloads, so final verification runs them without concurrent typechecking. The latest pushed head still requires independent AI review and CI.

## Review focus

Atomic header publication and retry, unchanged payload references, bounded row ownership and response assembly, preserved review/alias semantics, complete record exports, and separation of Web display budgets from MCP compaction.


ewen-poch:
Fixed removal of tasks whose checkpoint payloads are missing or corrupt. Removal now skips payload hydration and retains the existing index-first deletion order. Two service-boundary regressions reproduced the original failure and now verify deletion, restart recovery, and preservation of another indexed task. The task journal and service suites pass all 37 tests; Node typechecking and lint also pass. No persisted format, state enum, or UI behavior was added.

The failed Ubuntu shard has one failing assertion in `src/preload/index.test.ts`: its public API inventory is missing the already-added `literature.exportRecord` method. This is reproduced locally. The required change is one inventory entry; because that test file is outside the approved changed-file set, it is pending the maintainer's scope confirmation.

ewen-poch:
Added `literature.exportRecord` to the preload public API inventory after scope approval. This fixes the exact assertion reported by the failed Ubuntu shard without changing the runtime API or weakening the inventory guard. The failure reproduced locally before the update; all 131 preload and renderer-contract tests now pass, and lint passes.

ewen-poch:
Task detail reads now hydrate one row payload at a time, retain only its display projection, and stop at the existing response budget. Read-only requests do not cache full payloads on the task's control rows. Worker and mutation paths retain their existing hydration behavior, and provider progress remains outside the command queue to preserve progress-query liveness.

A service-boundary regression failed before the change because requesting the first page read a missing payload belonging to a later page. It now verifies the first page succeeds, continuation metadata is intact, accessing the damaged row still fails, and a read does not retain a full payload cache. Existing write-failure tests now block only the header write target so they continue to test failed publication without also blocking payload reads. All 164 targeted HTTP, task/recovery, paging-consumer, duplicate and command-boundary tests pass; typechecking and lint pass. No persisted format, enum, protocol field, or UI interaction changed.

ewen-poch:
The idempotent-create finding is reproduced for both metadata and full-text tasks after restart. A local fix routes the existing-task response through the bounded read path; both regressions and all 40 task/journal tests pass, with Node typechecking and lint also passing. This fix is being held for a combined push with the related draft-save correction.

The draft-save correction will assemble all review response pages before replacing rows or clearing pending drafts. Its component regression belongs in `LiteratureBatchLookupDialog.test.tsx`, which requires approval under the maintainer's changed-file scope constraint. Two CI failures are also outside the literature changes: repository-wide CodeQL now enables C# while this head has no C# source, and the merge result contains the new README product name while the i18n guard expects the old name. Updating the branch to main and adjusting the guard are pending maintainer confirmation; no scanner configuration has been changed.

ewen-poch:
Draft saves now assemble the complete version-checked review snapshot before replacing dialog rows or clearing pending selections. Component regressions reproduce both lost later-page selections and a failed continuation; they now verify that later rows remain selected, incomplete reads retain retryable drafts, and Apply submits the correct selections.

Repeated accepted create requests use the bounded read path after restart, restoring metadata and full-text review data without eagerly caching all payloads. Both service regressions failed before the fix and pass afterward.

The branch is updated to main with maintainer approval, including the C# fixture required by the repository's newly enabled CodeQL language. The German README guard now expects the product name already published on main. All 1,699 targeted tests pass after these changes; typechecking, lint and the Web API inventory check pass. A browser check of the real dialog with controlled two-page data retained all 1,000 rows after deselecting the first row and submitted exactly 999 selections, including the last row and excluding the deselected one. No persisted format, enum or new interaction was introduced.

ewen-poch:
The oversized-record consumer finding is confirmed through the real catalog and workspace mention helper with a valid reference whose abstract exceeds the 10 MiB page budget. The same regression passes with the base catalog and fails on the current head with ApplicationCommandError from boundedLiteraturePage. Temporary reproduction changes have been restored; no implementation change is included yet.

Consumer changes require maintainer confirmation under the approved file-scope constraint. A shared partial projection also needs an explicit design decision because workspace mentions retain the complete item snapshot. I am pausing that change pending scope confirmation rather than changing the search contract implicitly.

The previously failing C# CodeQL and static checks now pass. The latest check snapshot has no failed jobs; two test jobs are still running.

ewen-poch:
Search selection consumers now assemble bounded transport pages and recover individually oversized records through the existing digest-checked chunk transfer. Global search, workspace mentions/PDF choices, restore preview and batch reading retain complete metadata; no partial item snapshot is substituted. The Library display keeps its existing oversized-record export notice.

A library-only countOnly request returns the filtered total without loading item metadata. It is a transient query option, with no additional persisted fields, task states, migration or user action. Real-catalog regressions reproduced oversized mention/count failures before the fix; component regressions reproduced missing later-page and oversized search options. They pass after the fix. A browser check displayed an 11 MiB record after 45 bounded chunks and kept the count at one.

The merge with current main resolves the reported conflicts while preserving selected-ID alias semantics, deletion cleanup/refresh notices, translations and byte limits. All 1,830 targeted tests across 55 files pass after the final edits; typechecking, lint and the Web API map check pass. The impact planner requests a full fallback because of an unowned adapter test; the maintainer requested module tests locally, so the complete portable/platform suites remain delegated to CI.

The previous Windows E2E failure was a renderer-ready timeout during fixture launch; the retry passed, and fail-on-flaky-tests correctly failed the job. Its artifact contains no renderer-level root-cause evidence. No timeout or CI check has been weakened; the same lane will validate this updated head.

ewen-poch:
The Trash export regression is fixed by reading the retained record for the requested ID, including deleted metadata and merge provenance, rather than using the active-reference resolver. Export does not restore records or change alias relationships. The real-catalog regression failed with Reference unavailable before the fix and now verifies the complete abstract and retained deletion timestamp; normal get still hides the deleted record.

The failed static job was the changed-file Prettier check, not lint or typechecking. The affected test formatting is corrected, and the exact changed-file format command passes locally. After the final edits, 358 affected catalog, Web RPC, export-reader and Library component tests pass; Node typechecking and lint pass. The earlier broader 1,830-test result applies to the preceding consumer repair.

The task-control finding remains open. A public-service/filesystem reproduction confirms that changing a first-page selection reads an unrelated later payload; that read fails when the later payload is absent, while bounded get still succeeds. This establishes eager reads, not a measured OOM. A complete correction needs lazy control drafts, preservation of unchanged payload digests during retry, per-active-row worker loading/release, and bounded command responses. That changes the in-memory ownership and journal-write workflow, so it is paused for maintainer design confirmation. The proposed approach retains the existing v3 persisted format and task states. No partial fix that risks losing unloaded payloads is included.

The latest push requires fresh CI. The earlier Windows startup flake has not yet been demonstrated resolved; its replacement run was still in progress at the last check.

ewen-poch:
I could not reproduce the reported DOI/PMID bridge omission on this head. A probe containing a DOI-only record and a PMID-only record with matching metadata, plus a DOI+PMID record with a different title, forms one three-record group in all six input permutations. All 18 tests, including the existing cross-key case, pass. The current join checks only schemes present on both components; it does not require the other component to contain the accumulated root's entire scheme set. No algorithm change is justified by this reproduction. Please provide the exact records/order if the finding depends on another case.

The ACP integration concern is supported by the current call path: runtime-composition calls catalog.search before the MCP layer's compactSearchResult, so the catalog display budget can reject data before MCP compaction. The catalog oversized-record failure is already reproduced, but this inspection is not a claim of an end-to-end ACP reproduction. Correcting the ACP read boundary expands the approved scope and needs maintainer confirmation; the existing Web response limit should remain intact. All four supported agent execution paths must be included in validation if that change is approved.

The previously identified control-operation eager-read concern also remains open pending design confirmation. This review not repeating it does not establish that it is resolved. No additional code change has been made. All CI jobs on 5d619550 have completed without failures, including static checks and the previously flaky Windows E2E shard; the AI verdict remains needs changes.

ewen-poch:
Task controls now copy compact controls only. Candidate validation reads selected rows, and the worker reads one eligible row and releases it after its durable checkpoint. Retry preserves completed payload references and item snapshots while clearing the previous review one row at a time; atomic publication is unchanged. Public task regressions failed twice against the unrepaired implementation and pass after the change. No additional persisted fields, journal version, database schema, status enum, or interaction changes are introduced.

Agent Library search now uses the same catalog filtering and alias resolution before the existing MCP compaction/output limit, without applying the renderer/Web display budget first. A real production ACP handler connected to an MCP client reproduces the oversized-record error before repair and returns a compact successful result afterward. Delayed-result ownership is covered for Claude Code, OpenCode, Codex Responses, and Codex Bridge.

The suggested cross-key duplicate grouping failure could not be reproduced across the supplied scenario permutations and existing grouping tests; the algorithm is unchanged for that finding.

Final local validation: 1,942 tests in 63 files passed, with typecheck, lint, Web API mapping, and changed-file formatting checks passing after the material edits. The affected planner requests the full fallback for an unregistered ACP composition owner; local checks remain scoped as requested by the maintainer, with complete portable/platform coverage delegated to PR CI.

ewen-poch:
Rebased onto main at 840101431c777d6f5e93666facfcac6c10f983c5. Conflicts preserve upstream attachment-deletion authority, recovery notices and accessibility behavior alongside bounded reads and complete selection recovery. The complete-selection changes originally recorded in a merge commit are retained in the rewritten history. No new storage format, schema, enum, or interaction was added.

After the final conflict resolution, 2,015 tests across 65 files passed, including the Electron command adapter alongside literature, renderer, IPC and four ACP execution paths. Typecheck, lint, Web API mapping, and changed-file formatting also passed. The prior head received a mergeable AI review with no actionable findings; the rebased head will receive its own CI and review.

ewen-poch:
Rebased onto main at 684bdaa32a260eeb1dbf610183e62980343b6611. Conflict resolution preserves upstream data-root migration admission/drain protection, including the export command, while retaining lazy journal checkpoints and complete paged selections. Upstream task button labels are preserved. Two upstream migration test reads now target the existing v3 task.json header instead of the old checkpoint path; their pause/drain and remaining-row assertions are unchanged.

Final focused validation passed 1,635 tests across 50 files. The broader run exposed two outdated checkpoint-path assertions (corrected) and two timeout failures under concurrent local workloads. The unchanged capacity and duplicate-navigation tests pass separately, and all seven migration tests pass. Typecheck, lint, Web API mapping and changed-file formatting pass. No new persistence format, schema, enum or interaction was introduced by this rebase.

ewen-poch:
Updated the legacy migration integration to the existing v3 index and nested task.json recovery layout. The CI failure was reproduced locally with the same v2/v3 assertion mismatch. The test now also retains a review message, verifying that immutable row payloads survive relocation and header recovery; review updates remain asserted after reopening. This changes test expectations only, not the persisted format or production behavior.

Rebased onto main at f4a82d4ae7ecc826f9216866ef3653e4de4da3a6, retaining upstream attachment translations. After the final edit, all 878 targeted migration, journal, task and i18n tests passed, along with typecheck, lint and changed-file formatting.

The previous Windows journey failure was a Traditional Chinese option click timeout that passed its automatic retry; fail-on-flaky correctly kept the job red. No assertion, timeout or flaky-test policy has been weakened. The new-head Windows lane will determine whether that failure repeats.

PR #2368, opened 2026-09-08
fix(literature)!: synchronize clients and reject stale collection edits

ewen-poch:
## Problem

Ordinary Library changes remained invisible in other clients, stale collection forms could overwrite previously saved fields, and late reads with equal metadata revisions could replace newer attachment or relationship snapshots.

## Proposed change

- Publish committed catalog invalidations through the existing Electron/Web event contracts. Refresh active Library views, counts, collection navigation, live details, search and mention consumers on changes, focus and Web recovery. Keep unsaved drafts.
- Use atomic collection revision comparisons and show a retained-draft conflict with the latest saved values. Loading the latest version is an explicit action; different-field edits also conflict.
- Reject obsolete per-item reads in the existing list and detail owners, including reverse completion and independent-item batches.

## Scope and compatibility

Migration `0040_literature_collection_revision` adds one persisted integer column, backfilling existing collections to 1 without changing their IDs, metadata or hierarchy. Update requests now require `expectedRevision`; old clients must refresh/upgrade. Previously overwritten values cannot be recovered by this migration. No new persisted state enum, relationship entity, event journal or draft storage is introduced.

Events contain resource IDs and a process-local monotonic revision, not full reference contents. Focus/recovery rereads authoritative data; this does not introduce a durable global version protocol. Transaction rollback and no-op writes emit no mutation event, and failed delivery cannot undo a commit. Attachment verification invalidates its persisted observation even when the observation reports unavailable bytes.

Before production edits, the same regressions failed twice at the existing SQLite catalog, React page and hook boundaries: 8 matching failures and 4 passing controls. The fixed regressions pass. Additional tests cover atomic conflict/retry/deletion, migration backfill, event delivery failure/rollback/no-ops, subscriber cleanup, read ordering and retained drafts.

Behavior-to-check mapping (the original implementation checks preceded the latest base update; the final-head rechecks are identified below):

| Behavior | Repository check | Result |
| --- | --- | --- |
| Catalog writes, cached reads, collection conflicts, drafts and all translated UI strings | `npx vitest run src/main/literature src/renderer/src/pages/literature src/shared/literature.test.ts src/shared/literature-csl.test.ts src/shared/renderer-contract.test.ts src/shared/renderer-contract-catalog.test.ts src/preload/index.test.ts src/main/application-events.test.ts src/main/renderer-broadcast.test.ts src/main/lifecycle-broadcast.test.ts src/main/data-content-application-commands.test.ts src/renderer/src/i18n/resources.test.ts` | 1,773 passed; one capacity test failed during concurrent build/test load and passed standalone without code changes |
| Migration history, upgrades and retained values | `npx vitest run src/main/database` | 305 passed |
| Packaged migration certification and historical upgrade fixtures | `npx vitest run scripts/database-migration-ledger-smoke.test.ts src/main/database/literature-collection-revision.test.ts` plus the Linux/macOS/Windows package smoke test files | 16 migration checks and 45 consumer checks passed after fixing an omitted certification ledger entry |
| Event projection, Web bootstrap/recovery, search and mention consumers | Explicit application event flow, renderer surface matrix, Web API installer/bootstrap, Web event projection/controller/stream, GlobalSearchDialog and ArtifactMentionPopup test files | 290 passed |
| Two real Electron renderers plus one real Web client: create/edit/delete/accept/classify and offline recovery | `npx playwright test e2e/literature-client-sync.spec.ts --workers=1` against rebuilt Electron/Web outputs | Passed |
| Types, formatting and generated persistence/transport contracts | `npm run typecheck`, `npm run lint`, `npm run db:schema:check`, `npm run check:web-api-map` | Passed |

The impact explain command selects the full CI fallback because these paths are not completely module-owned. Per maintainer instruction, no full local `npm test` was run; exact-head PR CI is authoritative. The new three-client scenario is a standalone macOS check; the existing Windows/Linux CI lanes do not establish that exact scenario on those platforms. The live scenario exercises all three transports but not every operation-by-client permutation. The conflict screenshot was captured from the real dialog with a controlled conflict API fixture; its interaction is also covered by the dialog test.

Latest head `63cbb337` is rebased onto `684bdaa3`, preserving upstream relocation, deletion protection, import progress and cancellation. After the final edits, 1,384 Literature/catalog/i18n tests passed; two page timing failures in the concurrent run passed unchanged in isolation. All 374 Notebook runtime tests passed on serial rerun after two unrelated startup-wait failures in the concurrent run. Process typechecks, lint, formatting, generated schema/API-map checks and PR policy pass. Earlier transport, persistence and live multi-client evidence above remains a prior checkpoint; exact-head CI is required.

Review fixes propagate the committed collection revision to the local directory. Removed or merged references close read-only details; editors retain their input and disable Save. A separate transient removed-reference ID now keeps that protection intact across locale-triggered list reloads, and the notice translates at render time. It clears on detail open/close or confirmed restoration. No new persistence or enum is added. Nine focused real-page regressions pass; browser verification confirms the retained draft, Chinese notice and disabled Save.

CI fixes await existing shell admission or explicit installer-start signals and use public batch import to prepare the 201-item collection test. The original ordering, repair-gate, atomicity and count assertions remain. Installer-start failures were reproduced with a temporary 1.3-second mock delay and passed after the fix with the same delay; the delay is removed. The previous Windows startup failure passed on retry; no timeout or flaky-test gate was relaxed. New-head CI and review remain pending.

Latest rebase checkpoint `f30de451` is based on `f4a82d4a`. Upstream attachment confirmation/version-history behavior is retained. The new component fixtures now provide this PR's required `readItem` callback. Rebase checks passed 1,320 tests and exposed one missing-callback failure; all 13 attachment component tests pass after correction, along with renderer typecheck, lint, formatting and PR policy. Node/sandbox checks also passed. The remote-deleted collection scope and non-edit detail panel findings are reproduced and awaiting interaction confirmation; they remain unresolved. Windows startup latency is under a focused `windows-e2e` dry-run; new-head CI remains required.

## Review focus

Please check committed-event boundaries and Web recovery, revision comparison/backfill, preservation of unsaved input, and obsolete-read rejection. Also verify that the retained upstream metadata-operation receipts and manual-entry recovery behavior remain intact. Independent review and required CI must complete before merge.

### DIFF
--- a/e2e/fixtures/electron-app.ts
+++ b/e2e/fixtures/electron-app.ts
@@ -1,4 +1,4 @@
-import { expect, test as base } from '@playwright/test'
+import { expect, test as base, type TestInfo } from '@playwright/test'
 import { spawn } from 'node:child_process'
 import { chmod, copyFile, mkdir, mkdtemp, readFile, readdir, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
@@ -340,7 +340,7 @@ class ElectronAppHarness implements ElectronApp {
     private readonly windowMode: E2eWindowMode
   ) {}
 
-  static async create(windowMode: E2eWindowMode): Promise<ElectronAppHarness> {
+  static async create(windowMode: E2eWindowMode, testInfo: TestInfo): Promise<ElectronAppHarness> {
     const testRoot = await mkdtemp(join(tmpdir(), 'open-science-electron-e2e-'))
     const harness = new ElectronAppHarness(
       testRoot,
@@ -361,6 +361,10 @@ class ElectronAppHarness implements ElectronApp {
       await harness.launch()
       return harness
     } catch (error) {
+      await harness
+        .captureMainLog('startup-failure.log')
+        .then((path) => testInfo.attach('main-process-log', { path, contentType: 'text/plain' }))
+        .catch(() => undefined)
       await harness.dispose().catch(() => undefined)
       throw error
     }
@@ -936,12 +940,12 @@ class ElectronAppHarness implements ElectronApp {
       this.resourceProfiler !== undefined
     )
     await this.resourceProfiler?.attach(this.application)
+    this.mainLogDirectory = await this.application.evaluate(({ app }) => app.getPath('logs'))
     this.currentPage = await openMainWindow(
       this.application,
       this.rendererFailures,
       this.windowMode
     )
-    this.mainLogDirectory = await this.application.evaluate(({ app }) => app.getPath('logs'))
   }
 
   private get runningApplication(): ElectronApplication {
@@ -1028,7 +1032,7 @@ const test = base.extend<{ app: ElectronApp; windowMode: E2eWindowMode }>({
   windowMode: ['hidden', { option: true }],
   // Playwright fixture callbacks require an object pattern even when no base fixture is needed.
   app: async ({ windowMode }, install, testInfo) => {
-    const app = await ElectronAppHarness.create(windowMode)
+    const app = await ElectronAppHarness.create(windowMode, testInfo)
 
     try {
       await install(app)
--- a/e2e/fixtures/electron-app.ts
+++ b/e2e/fixtures/electron-app.ts
@@ -1,4 +1,4 @@
-import { expect, test as base, type TestInfo } from '@playwright/test'
+import { expect, test as base } from '@playwright/test'
 import { spawn } from 'node:child_process'
 import { chmod, copyFile, mkdir, mkdtemp, readFile, readdir, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
@@ -340,7 +340,7 @@ class ElectronAppHarness implements ElectronApp {
     private readonly windowMode: E2eWindowMode
   ) {}
 
-  static async create(windowMode: E2eWindowMode, testInfo: TestInfo): Promise<ElectronAppHarness> {
+  static async create(windowMode: E2eWindowMode): Promise<ElectronAppHarness> {
     const testRoot = await mkdtemp(join(tmpdir(), 'open-science-electron-e2e-'))
     const harness = new ElectronAppHarness(
       testRoot,
@@ -361,10 +361,6 @@ class ElectronAppHarness implements ElectronApp {
       await harness.launch()
       return harness
     } catch (error) {
-      await harness
-        .captureMainLog('startup-failure.log')
-        .then((path) => testInfo.attach('main-process-log', { path, contentType: 'text/plain' }))
-        .catch(() => undefined)
       await harness.dispose().catch(() => undefined)
       throw error
     }
@@ -940,12 +936,12 @@ class ElectronAppHarness implements ElectronApp {
       this.resourceProfiler !== undefined
     )
     await this.resourceProfiler?.attach(this.application)
-    this.mainLogDirectory = await this.application.evaluate(({ app }) => app.getPath('logs'))
     this.currentPage = await openMainWindow(
       this.application,
       this.rendererFailures,
       this.windowMode
     )
+    this.mainLogDirectory = await this.application.evaluate(({ app }) => app.getPath('logs'))
   }
 
   private get runningApplication(): ElectronApplication {
@@ -1032,7 +1028,7 @@ const test = base.extend<{ app: ElectronApp; windowMode: E2eWindowMode }>({
   windowMode: ['hidden', { option: true }],
   // Playwright fixture callbacks require an object pattern even when no base fixture is needed.
   app: async ({ windowMode }, install, testInfo) => {
-    const app = await ElectronAppHarness.create(windowMode, testInfo)
+    const app = await ElectronAppHarness.create(windowMode)
 
     try {
       await install(app)
--- a/e2e/fixtures/electron-app.ts
+++ b/e2e/fixtures/electron-app.ts
@@ -2,6 +2,7 @@ import { expect, test as base } from '@playwright/test'
 import { spawn } from 'node:child_process'
 import { chmod, copyFile, mkdir, mkdtemp, readFile, readdir, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
+import { fileURLToPath } from 'node:url'
 import { delimiter, isAbsolute, join, relative, resolve, sep } from 'node:path'
 import { _electron as electron, type ElectronApplication, type Page } from 'playwright'
 import {
@@ -115,6 +116,8 @@ const closeElectronApplicationForCleanup = async (
 
 type ElectronApp = {
   readonly page: Page
+  openAdditionalRenderer: () => Promise<Page>
+  authenticatedWebUrl: () => Promise<string>
   allowRendererConsoleError: (text: string) => void
   captureMainLog: (name: string) => Promise<string>
   armDelegatedHandoffCleanupSabotage: (childName: string) => Promise<void>
@@ -547,6 +550,62 @@ class ElectronAppHarness implements ElectronApp {
       })
       await bridge.api.compute.bookmarksSet('ssh:a11y-fixture', ['/scratch/fixture/pinned'])
     })
+  }
+
+  async openAdditionalRenderer(): Promise<Page> {
+    const next = this.runningApplication.waitForEvent('window')
+    await this.runningApplication.evaluate(
+      ({ BrowserWindow }, preload) => {
+        const source = BrowserWindow.getAllWindows().find((window) =>
+          window.webContents.getURL().includes('index.html')
+        )!
+        const window = new BrowserWindow({
+          show: false,
+          webPreferences: {
+            preload,
+            contextIsolation: true,
+            nodeIntegration: false,
+            sandbox: true
+          }
+        })
+        void window.loadURL(source.webContents.getURL())
+      },
+      fileURLToPath(new URL('../preload/index.js', this.page.url()))
+    )
+    const page = await next
+    await page.waitForFunction(() => Boolean(window.api?.databaseStartup))
+    await waitForRendererReady(page)
+    return page
+  }
+
+  async authenticatedWebUrl(): Promise<string> {
+    const target = electronLaunchTarget(this.roots.userDataRoot)
+    const child = spawn(
+      target.executablePath ?? ((await import('electron')).default as unknown as string),
+      [...target.args, '--serve=0'],
+      { env: launchEnvironment(this.roots.storageRoot), stdio: 'ignore' }
+    )
+    await new Promise<void>((resolve, reject) => {
+      child.once('error', reject)
+      child.once('exit', () => resolve())
+    })
+    let port: number | undefined
+    await expect
+      .poll(async () => {
+        try {
+          port = (
+            JSON.parse(
+              await readFile(join(this.roots.storageRoot, 'web-service.json'), 'utf8')
+            ) as { port: number }
+          ).port
+        } catch {
+          return false
+        }
+        return Boolean(port)
+      })
+      .toBe(true)
+    const token = (await readFile(join(this.roots.storageRoot, 'web-token'), 'utf8')).trim()
+    return `http://127.0.0.1:${port}/?token=${encodeURIComponent(token)}`
   }
 
   async completeOnboarding(): Promise<Page> {
--- a/e2e/literature-client-sync.spec.ts
+++ b/e2e/literature-client-sync.spec.ts
@@ -1,0 +1,104 @@
+import { expect } from '@playwright/test'
+import type { Page } from 'playwright'
+import { test } from './fixtures/electron-app'
+import { literatureCandidateInputSchema, literatureItemInputSchema } from '../src/shared/literature'
+
+const library = async (page: Page): Promise<void> => {
+  await page.getByRole('button', { name: 'Library', exact: true }).click()
+  await page.getByRole('button', { name: 'All references', exact: true }).click()
+}
+
+test('synchronizes ordinary writes across two Electron renderers and a Web client', async ({
+  app,
+  browser
+}) => {
+  test.setTimeout(180_000)
+  const first = await app.completeOnboarding()
+  const second = await app.openAdditionalRenderer()
+  const web = await browser.newPage()
+  try {
+    await web.goto(await app.authenticatedWebUrl())
+    for (const page of [first, second, web]) await library(page)
+    const item = literatureItemInputSchema.parse({
+      itemType: 'journalArticle',
+      title: 'Created in desktop'
+    })
+    const created = await first.evaluate(
+      (item) => window.api.literature.transact({ kind: 'create-item', item }),
+      item
+    )
+    for (const page of [second, web])
+      await expect(page.getByText(item.title, { exact: true })).toBeVisible()
+    const saved = await web.evaluate(async (id) => {
+      const current = (await window.api.literature.get(id))!
+      await window.api.literature.transact({
+        kind: 'update-item',
+        itemId: id,
+        expectedMetadataRevision: current.metadataRevision,
+        item: { ...current.item, title: 'Edited in Web' }
+      })
+      return id
+    }, created.id)
+    for (const page of [first, second])
+      await expect(page.getByText('Edited in Web', { exact: true })).toBeVisible()
+    const collection = await second.evaluate(() =>
+      window.api.literature.transact({ kind: 'create-collection', name: 'Shared collection' })
+    )
+    for (const page of [first, web])
+      await expect(
+        page.getByRole('button', { name: 'Shared collection', exact: true })
+      ).toBeVisible()
+    await web.evaluate(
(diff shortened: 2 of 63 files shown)


ewen-poch:
The stale collection revision finding is valid and fixed. A successful atomic collection update increments the submitted revision exactly once; the editor now forwards that committed revision to the page together with the saved fields. The page keeps any higher revision it has already observed.

Two regressions exercise the real Library page with a mutable API source: immediately reopen and save again while the directory reload remains pending, and repeat when the reload rejects. Both reproduced the false conflict before the fix and pass afterward. The queueing of an authoritative reload no longer determines whether a second edit succeeds. No new persistent fields, transport contract, or user interaction was added.

The merge conflict with the attachment reliability tests was resolved by retaining both the cleanup-failure coverage from main and the attachment-observation invalidation coverage from this PR. Upstream deletion and focus-recovery behavior is preserved.

Validation on the rebased source: 283 Library page, collection editor, attachment reliability and application-command wiring tests passed. The change is in commit 99f0abe8.

ewen-poch:
The deleted-reference finding is reproducible: a real Library page test keeps the stale read-only detail open after a change notification and an authoritative missing-item response. A paired control confirms that the current path preserves an unsaved metadata draft.

Calling the existing close routine unconditionally would unmount that draft, which conflicts with this PR's draft-preservation requirement. The proposed resolution is to close only a read-only detail, retain editing input with the existing removal notice, and guard against the user opening another item or starting an edit while the read is pending. Production changes for this point are paused for the maintainer's requested design confirmation. No additional persistent data or enum is proposed.

ewen-poch:
The current Windows failures occur before their named assertions: the layout test times out waiting for the initial deterministic agent reply and passes on retry; the Russian persistence test times out during initial renderer/database readiness, with startup still reporting `migrating`, before any locale persistence checks run.

I inspected the failure logs and traces. These artifacts do not establish a layout, localization, or Literature regression, so no production change or timeout relaxation is justified yet. I started a targeted rerun of the failed Windows jobs with unchanged source and assertions. GitHub accepts one job rerun at a time here; the remaining failed shard will be rerun after the current attempt completes. If the failure repeats, further work will use focused diagnostics rather than repeatedly rerunning the full workflow.

The deleted-detail review finding remains paused for the requested confirmation of draft-preserving behavior.

ewen-poch:
Resolved the merge conflict against main at c01c0830. Both the upstream input-composition regressions and the cross-client refresh regressions are retained; the upstream accessibility and focus fixes remain intact. No change was made to the pending deleted-detail interaction decision.

Validation after conflict resolution: 1,270 Literature page, editor, search and i18n tests passed, along with the renderer typecheck, generated-contract checks and PR policy. The updated head is d254539f, and GitHub reports the branch as mergeable. The new head's Windows jobs are running; their results, rather than the superseded run, will determine CI status. No timeout or assertion has been relaxed.

ewen-poch:
The removed/merged-detail finding is fixed in 9f8696a following confirmation of the draft-preserving behavior. Focus, visibility recovery, reconnect and change notifications now use the same existing refresh subscription. A missing reference, a different survivor ID, or a deleted item closes a read-only detail and clears its preview. An open editor retains its input, displays the removal notice inside the form, and disables saving the removed reference.

The read checks the current opening generation and the latest editing mode after awaiting the API, so it cannot close a different reference or discard an edit started while the response was pending. The duplicate focus-only listener was removed. This adds only an in-memory mode reference; no persisted field, enum, transport contract or data-model change is involved.

Eight focused real-page regressions pass, including notification/focus/reconnect, retained drafts, starting an edit during a delayed read, and opening another reference before an old read finishes. The pre-fix cases reproduced the stale dialog, discarded draft and stale error. A further red/green check verifies that the deletion notice is inside the editor, rather than hidden behind it. The full Literature page module passes all 426 tests; renderer typecheck, lint and PR policy pass. Lint excludes only uncommitted downloaded reports under docs/internal. Browser screenshots of the real Library with a controlled mutable API source verify both the closed read-only detail and the retained draft with disabled Save.

The previous head's Windows shards 1 and 3 passed. Shard 2 reported flaky startup and PDF-context waits that passed on retry; these assertions and timeouts remain unchanged. The new head's CI and independent review are required before merge.

ewen-poch:
Resolved the conflicts with main at 986d4221 in 38b5781b, preserving the upstream localization and input-composition behavior. Collection errors are translated at render time so switching language retains both the draft and retry target. The new upstream localization fixture now includes revision 1 and expects revision 2 after an edit. The invalid Testing Library `exact` option was also removed without weakening the Save-disabled assertion.

Correction to the previous validation comment: the earlier renderer typecheck was not clean; its failure was obscured by the subsequent successful lint command. Both type errors are now fixed, and the renderer typecheck was rerun independently with exit code 0. After the final edits, all 1,256 Literature and i18n tests passed, as did lint (excluding only uncommitted local evidence), formatting and PR policy.

The removed-detail fix remains intact: read-only details close, while editors retain their draft and show the removal notice with Save disabled. These conflict/test adjustments add no persistent fields, states or user interaction. The new head's required CI and independent review are still pending. Existing flaky Windows assertions and timeouts were not changed.

ewen-poch:
Rebased onto main at 84010143 and resolved the deletion-protection conflicts. Both protections remain: the session coordinator checks recoverable chat history before attachment/permanent deletion, and catalog invalidations publish only after a successful commit. Upstream deletion diagnostics and translations remain intact.

After conflict resolution, 1,864 Literature, renderer, i18n and session-coordinator tests passed. The capacity test hit a SQLite transaction-start timeout during concurrent checks and passed unchanged when run alone. Node, sandbox and renderer typechecks, lint, generated schema/API-map checks and PR policy passed. No assertions or timeouts were relaxed.

The locale-switch review finding is reproduced twice through the real Library page: Save becomes enabled and clicking it sends an update for the removed reference. Replacing translated error text with a stable value alone does not fix it, because a locale-triggered list reload clears the shared error. A reliable fix needs a separate transient removed-reference identity owned by the detail view. That state addition is awaiting maintainer confirmation; the proposed failing regression is retained locally and is not included in this rebase. No persistence or database migration is needed for it. This review finding remains unresolved, so the PR must not merge yet.

ewen-poch:
Resolved the conflicts against main at 684bdaa3, retaining the upstream import progress/cancellation and data-relocation safeguards.

The locale-switch deletion guard is now fixed following confirmation of the transient state addition. A removed reference ID belongs to the current detail view, independently of list-loading errors. It clears on opening/closing the detail or an authoritative read confirming restoration. The editor translates the notice at render time and keeps Save disabled across a language switch without replacing the draft. This adds one in-memory string state only; no persistence, migration, enum or relationship changes. The real-page regression reproduced a stale update before the fix; all nine related regressions pass. A browser check of the real Library with a controlled API confirms the Chinese notice, retained draft and disabled Save.

The two Notebook CI failures came from a short polling deadline before the injected installer was reached. Both were reproduced with a temporary 1.3-second delay in the existing installer mock. Waiting for an explicit installer-start promise passes with the same delay; the delay is absent from the final source. Environment locking, execution ordering and the repair-gate assertions are unchanged.

The collection batch regression keeps its 201 references, successful first 200-item command, rejected second command and original count assertions. Its setup now uses the public batch import API instead of 402 separate per-item commands; the focused test passes in under one second locally. This reduces setup work without relaxing the test timeout or changing production catalog behavior.

Windows previously failed during database startup, remained at `migrating`, and passed on retry. That evidence does not establish a production root cause, so no Windows assertion or timeout was relaxed. The new head's CI remains required before merge.

Final validation at 63cbb337: the Literature/catalog/i18n selection passed 1,384 tests; two page timing failures during concurrent checks passed unchanged in isolation. All 374 Notebook runtime tests passed when rerun serially; the initial concurrent run had two unrelated startup-wait failures. Node, sandbox and renderer typechecks, lint, formatting, schema/API-map checks and PR policy pass after the final edits. No full local portable suite was run; exact-head CI is authoritative.

ewen-poch:
Both new review findings reproduce through the real Library page with a mutable API fixture. Two unchanged-source runs each fail the same four cases: after remote collection deletion the heading changes to All references while the expected reference remains absent; after reference deletion the citation, metadata-completion and full-text dialogs remain open. The draft-protection cases from the preceding fix remain the intended behavior.

The proposed interaction is to return a deleted collection scope to All references and to close the non-edit detail modes after remote deletion, retaining only metadata-edit drafts. This changes navigation/panel behavior and requires maintainer confirmation under the agreed scope rules. No new persistent field, enum, relationship or transport is proposed. Regressions are preserved locally; no production change has been applied for these findings yet.

ewen-poch:
Rebased onto main at f4a82d4a and resolved the conflicts in f30de451, retaining attachment deletion confirmation, version history and the existing shell-admission wait. The new upstream attachment component tests now supply the required readItem callback using their existing API mock. No assertion or production reading path changed.

The rebase check initially passed 1,320 tests and failed the attachment refresh-error case because readItem was missing; renderer typechecking confirmed the same missing prop. After the fixture correction, all 13 attachment component tests, renderer typecheck, lint, formatting and PR policy pass. Node/sandbox typechecks passed during the rebase check. No full local portable suite was run.

The previous Windows failure trace now includes early migration timings and a storage probe: sequential I/O took 6 ms while the synchronous write probe took 3,763 ms. Startup spent most of its initial 15 seconds waiting on I/O and had only reached early migrations, before the collection revision migration. This supports investigating runner storage latency; it does not establish a production root cause. The existing windows-e2e dry-run is running on the previously failing head to isolate that path without rerunning the full suite. No timeout or assertion was relaxed.

The two remote-deletion interaction findings remain pending maintainer confirmation and are not included in this fixture adjustment. The PR must not merge while those findings or required CI remain unresolved.

ewen-poch:
Rebased onto main at 04b683e2. The conflict resolution retains the upstream bounded selection reads, oversized-record notice and export mock together with this PR's change subscription and contract. All eight commits are preserved; range-diff shows only the expected integration differences. No additional interaction, persistence or model change was introduced.

Checks after the rebase: 30 test files / 1,594 tests passed across the Library renderer (including pagination, stale reads, drafts and attachments), global search, catalog, shared literature contracts, preload and i18n guards. Node, sandbox and Web typechecks, lint, conflict-file formatting, database schema generation check, Web API map check and commit/migration policy passed. No full local portable suite was run; exact-head CI remains authoritative.

The preceding head f30de451 passed PR Gate, including all Windows E2E shards, and its latest AI review is mergeable with no new findings. This does not resolve the two previously reproduced remote-deletion interaction findings, which still await maintainer confirmation. They remain outside this rebase and the PR should not merge until resolved.

ewen-poch:
Rebased onto main e69dbc20, retaining its attachment operation store, bounded waits and cleanup barrier. The store now uses the existing detail-controller read callback, so a delayed operation read cannot overwrite newer attachment history at an unchanged metadata revision. The real attachment component regression failed twice before this connection (two versions replaced by one) and passes afterward. No additional persistent field, enum or UI behavior was introduced.

After the change, 1,454 tests passed across 26 Library, catalog, PDF reliability and i18n files; the component/controller subset also passed all 15 tests. Node/sandbox/Web types, lint, schema and API map checks passed. Playwright successfully collected the two affected diagnostic suites (27 tests); this is collection validation, not a local Windows execution. No full local portable suite was run.

The last Windows failures occurred while startup was still migrating and while the chat was Thinking before its deterministic response. The uploaded traces contained neither main-process logs nor stdout, so they do not establish a root cause. The E2E fixture now attaches its existing main log when startup fails before fixture creation; normal test failures retain the upstream log capture. No timeout or assertion was relaxed. These CI failures remain under investigation pending the new exact-head results.

The Web event-stream compatibility recommendation requires a protocol decision and remains pending, as do the two previously reproduced remote-deletion interactions. The agreed resumed review limit has been reached; no further review-driven redesign or merge is being performed automatically.

ewen-poch:
The latest widespread E2E failure was introduced by my diagnostic change: resolving the main log directory before window initialization caused `electronApplication.evaluate: Resulting promise was garbage collected`. The unchanged real Electron CSL test reproduced the same failure at that line locally. I reverted that diagnostic change, restoring the upstream lookup after window initialization. No timeout or assertion changed.

Rebased onto main d3d6e03d, preserving the reviewed-source deduplication guard, source API and translations together with the existing commit notifications. The attachment store ordering fix remains intact.

After rebuilding the current source, the real Electron CSL test passed. The two previously failing scenarios (French locale after restart and Memory/history replay) also passed locally without retries. These macOS results do not establish that the prior intermittent Windows failures are resolved; exact-head Windows CI remains required.

Final local checks after the rebase: 29 files / 1,490 tests passed across Library renderer, catalog, PDF reliability, migration writers, renderer contracts and i18n. Node/sandbox/Web types, lint, formatting, schema, Web API map and commit/migration policy passed. E2E build passed; three real Electron tests passed without retries. No full local portable suite was run. The pending interaction and Web protocol decisions remain unchanged; no automatic merge is authorized while those findings remain unresolved.

ewen-poch:
Rebased onto main a8e9c688. Both sets of PDF tests are retained: verified content identity/permission recovery from main and cross-client invalidation/read ordering from this branch. The new upstream component fixture supplies the existing read callback. No runtime behavior or data model was changed by this conflict resolution.

The active-collection deletion finding is valid and matches the previously reproduced public-page failures. Returning to All references and clearing selection after authoritative collection removal is a worthwhile fix, consistent with local deletion. It changes navigation behavior, so implementation is still awaiting the maintainer's explicit confirmation under the agreed scope rules.

The earlier Tags panel recommendation also identifies a consumer outside this PR's file scope. Source inspection confirms its literature reads depend on resource IDs and manual retry; adding this consumer and its regression test requires scope confirmation first. It has not yet been dynamically reproduced. Neither this recommendation nor the pending protocol compatibility decision is silently being treated as resolved.

After the rebase, all 53 targeted PDF/attachment/controller tests passed, along with Node/sandbox/Web types, lint, conflict-file formatting, schema/API-map checks and commit/migration policy. Range-diff confirms only the expected test integration differences. The preceding head passed the complete required CI; the rebased head still requires its own CI. No full local portable suite was run.

ewen-poch:
After authoritative collection refresh, the page now checks the current selection and uses the existing All references navigation path if that collection no longer exists. This clears row selection and reloads the unfiltered Library. A late refresh checks the latest selected ID, so switching to a surviving collection while the request is pending is preserved. Failed or superseded collection reads do not trigger navigation.

The real-page regression failed before the fix because the reference remained absent after remote deletion, despite the All references heading. It passes after the fix, including cleared row selection. A second test covers switching collections during the delayed refresh. All 304 Library page/change-hook tests passed. The real Library page was also verified in Chromium with a mutable API fixture: the deleted collection disappears, All references shows the remaining reference, and its checkbox is cleared. This is not a two-Electron-window E2E claim.

The approved navigation change adds no persistent field, enum or database migration. It retains only an in-memory reference to the current selection for asynchronous reconciliation. Other pending panel, Tags and protocol findings are outside this change.

Renderer typecheck, lint, formatting and commit/migration policy also passed after the final edit. No full local portable suite was run; exact-head CI remains authoritative.

ewen-poch:
Permission-denied verification observations now notify other literature clients before the original OS error is rethrown. This follows the content repository's existing contract: it commits the permission-denied observation before rejecting with EACCES/EPERM. The event invalidates committed data; it does not report verification success. Unrelated errors are not broadcast from this rejection handler.

The existing real-file/SQLite permission test reproduced the missing notification before the fix, while confirming that the unavailable observation had already persisted. It passes afterward. A separate public-command control confirms that a failed observation write rejects with the original failure and emits no notification. No schema, persistent field, enum or UI change was added. The real permission test is POSIX-only, as before; it is skipped on Windows and when running as root.

Rebased onto main 9ae0f766, preserving batch PDF import copy and all existing fixes. The only textual conflicts were independent additions to the eight locale catalogs.

The current Windows failure is a separate session-persistence timeout: the page reports Prisma could not start a transaction in time, and main logs show sessions:save-session rejected after about 22 seconds. The agent had connected successfully. This narrows the failure but does not establish why the transaction could not start. The existing Windows-only dry-run 34304340912 is checking the failing head; no session timeout or test assertion was changed.

After the final rebase, all 1,490 tests passed across 28 literature catalog, PDF, renderer and i18n files. No full local portable suite was run. The Windows-only dry-run is still in progress; the session transaction failure is not being marked fixed by this literature change.

Node/sandbox/Web typechecks, lint, formatting, database schema and Web API map checks, and commit/migration policy passed on the final state.

ewen-poch:
The focused Windows run 34304340912 reproduces the same workspace-files failure as PR Gate 34301231451. Both captured pages show `sessions:save-session` failing with Prisma's transaction-start timeout before the first agent reply. The first main log reports about 22 seconds in that IPC call; the focused run reports about 11 seconds. The failed condition is session persistence, not the expanded-file assertion itself. The cause of transaction contention is still unproven.

Investigating a code fix now requires entering the session-persistence/SQLite transaction owner and its tests, outside this PR's current file scope. No timeout increase, assertion relaxation or speculative session change has been applied. Maintainer scope confirmation is required before extending the patch. The literature verification-notification fix is independent and remains in the current head.

PR #2371, opened 2026-09-08
fix(literature): restore focus and expose accessible control states

ewen-poch:
## Problem

Closing reference details loses the initiating title's focus. Metadata overwrite choices, library navigation, two popovers, and the background-task indicator omit accessible state or names even though their visible UI changes.

## Proposed change

Restore detail focus to the usable initiator, falling back to All references when it disappears, while leaving child-preview focus ownership intact. Derive overwrite pressed state and field/candidate descriptions from existing values. Expose the current navigation destination and name Filters/Customize columns from their visible headings. Describe background-task progress and announce status changes politely without announcing numeric polling updates. Explain existing arrow-key column ordering in all eight translated locales.

## Scope and non-goals

Renderer accessibility only. Also align the German README banner assertion with the product rename already merged into main; its stale old-name expectation failed identically on main. No new dependencies, database fields, enum values, API payload changes, persistence changes, historical migration, or shared Dialog redesign. Metadata overwrite and column-preference persistence retain their existing behavior.

The ten core regression cases fail twice against unchanged production code and pass with this patch. Tests exercise real components through existing native API boundaries; no production test seam was added. Repeated baseline failures confirm body focus, missing pressed/current state, unnamed popovers, and empty task descriptions.

The initial integrated implementation passed all 1,174 literature/i18n tests. The approved review follow-up adds one transient component-local completion-announcement flag; it suppresses historical completions and stale completion replay after errors without changing persistence or visible layout.

Final checks ran after resolving conflicts with main. The resolution preserves main’s deleted-reference guard and disabled Trash buttons alongside initiating-element focus restoration:

| Behavior | Command / check | Result |
| --- | --- | --- |
| Background-task owner, library-page consumer, metadata submission, focus fallback, preview focus, historical hydration and error-recovery announcements | `npx vitest run src/renderer/src/pages/literature src/renderer/src/i18n/resources.test.ts --maxWorkers=2` | 18 files / 1,198 tests passed |
| Original core regressions and focus/announcement boundaries | Included in the final owner/consumer suites above | Passed |
| Lint | `npm run lint` | Passed |
| Patch hygiene | `git diff --check` | Passed |
| Visible layout and browser semantics | ego-browser with production components/CSS and isolated native API fixtures | Filters/customizer names, selected metadata field description, task description, Close/Escape focus return verified; screenshots retained locally |

No shared component interface or main/preload logic changed. Browser and native API boundaries are exercised by the literature render suites; the i18n guard covers all changed locale entries. `npm run test:affected:explain -- --base origin/main --head HEAD` reports full fallback because literature files are not registered module owners. Full local `npm test` was intentionally omitted as requested; exact-head PR CI remains authoritative for broader and platform coverage. Manual screen-reader speech and cross-platform Electron runs were not performed locally.

## Review focus

Check usable-initiator fallback and preview focus ownership, exactly one current navigation destination, and the mounted polite region remaining quiet for unchanged/numeric task updates. The large background-task text diff is primarily JSX indentation after preserving the live region when the button disappears.


ewen-poch:
Confirmed: loading only historical completed jobs changes the polite live region from empty to `Completed`, although no completion occurred during this component's lifetime. A regression test against the current component fails twice for exactly this reason; it uses the existing API fixture and adds no production test seam.

The proposed correction is to compare incoming jobs with the existing previous snapshot and retain a component-local completion announcement flag. Initial hydration and newly discovered historical completions would remain silent; a previously observed pending job completing successfully would remain announceable. Unchanged polling must not repeat announcements, and existing pause/failure descriptions and task recovery behavior must remain intact.

This introduces transient renderer state, but no task enum, API, database, persistence, or architectural change. Product-code changes are paused for maintainer confirmation, as requested for any added state. The regression test is currently local and has not been pushed.

ewen-poch:
Fixed in 47c50b9. Completion announcements now require an observed pending-to-successful-completion transition for the same job. Initial hydration and newly discovered historical completions remain silent. Repeated identical polls preserve the live-region text, while list-read and task-removal errors clear the transient completion flag so recovery cannot replay an old completion. Existing pending-task visibility and library refresh behavior remain unchanged.

The added component-local boolean was approved by the maintainer. It is not a task enum, API field, or persisted value. Changes are limited to LiteratureBackgroundTasks.tsx and its existing test file.

The initial-hydration regression failed twice before the fix. Focused coverage also reproduced the task-removal error recovery case before correcting it. The final background-task and library-page suites passed all 244 tests; lint passed. No full local test suite was run, and there are no visible layout changes.

ewen-poch:
Resolved the integration conflicts with main by preserving both behaviors: deleted references remain blocked from opening details, while valid reference buttons pass their actual initiating element for focus restoration. The title and unavailable-attachment buttons retain main's disabled state in Trash. No additional state, persistence, schema, or interaction design changes were needed.

Validation after the final resolution: literature renderer tests and the i18n guard passed (18 files, 1,198 tests), including detail focus restoration and the updated lifecycle behavior. Web typechecking, lint, and diff whitespace checks passed. Full portable and platform verification remains with CI.

Issue #2382, opened 2026-09-09
[Bug]: Quarantined sessions in unrelated projects block all new Compute dispatch after startup


bobo-xxx:
### Preflight checklist

- [x] I searched existing issues and this has not been reported yet.
- [x] I'm running a recent version of Open Science.

### What happened?

Eight quarantined session files belonging to two older projects prevent Open Science from starting its Compute queue dispatcher. A new Slurm job in a different project remains `queued` and never reaches submission.

The affected files are under the application's session storage and have names ending in `.json.invalid-<timestamp>-<sequence>`:

| Project | Quarantined files | Matching original `.json` files present | Files that parse as JSON |
| --- | ---: | ---: | ---: |
| single | 2 | 0 | 2 |
| Ap | 6 | 0 | 6 |

Read-only inspection confirmed that all eight have a version-2 session envelope. Their stored session IDs match the original filenames, and their project IDs match the owning directories. An isolated replay against local source commit `3ae1e1dfa0f9dd3b0683a2e19c92b5d0aecb0f5c` now confirms that all eight return `invalid` from `decodeSessionFile`. Direct conversation-graph validation throws `Message response target is invalid.` for every file.

Across the eight graphs, 41 `responseToMessageId` links point to IDs absent from the graph's message list (per-file counts: 4, 6, 5, 5, 3, 2, 4, 12). This establishes a present-day validation failure, but not which historical writer, migration, or operation introduced the dangling references. Omitting `conversationGraph` only in in-memory copies makes all eight decode successfully. No modified copies were saved, and dropping the graph is **not** proposed as a recovery procedure: it could discard authoritative conversation structure.

**Observed impact:** an unrelated project, `Compute workspace verification`, initially had a Slurm job with `status=queued`, `submittedAt=null`, `startedAt=null`, and `remoteHandle=null`. Startup logs on two launches report that Session concurrency limits could not be restored and Compute queue reconciliation remains stopped. The latest `glm-5.3-flash` retry also returned `queued` with no scheduler job ID. Its agent then explicitly cancelled it about 17 seconds after creation; the cancellation returned `cancellation_status=cancelled` and `status=failed`. This terminal state is not a Slurm execution failure. The original job is also now `failed`, but its terminal transition was not traced in this retry inspection.

[...]

PR #2384, opened 2026-09-09
fix(compute): explain and retry blocked queue restoration

ewen-poch:
Unresolved quarantined Sessions can prevent startup from restoring authoritative concurrency limits, leaving unrelated Compute jobs queued. The existing storage warning only explains the archive restriction, and rereading repaired Session data does not restart dispatch.

This change explains the Compute impact in the existing recovery alert, adds a translated “Recheck saved conversations” action, and returns `queue_blocked_reason: session_limits_unavailable` from concurrency status while restoration is blocked. Session list/load requests retry only a previously failed restoration. Successful recovery restores the durable limits before dispatching; incomplete recovery keeps dispatch stopped. Shutdown invalidates in-flight recovery and prevents later catalog reads from restarting the queue.

No database migration, Session format change, quarantine deletion, or automatic conversation-graph repair is included. The historical source of dangling response links and lossless recovery of the eight reported files remain unresolved. This addresses the visibility and retry portions of #2382; it does not remove the cross-project fail-closed safeguard.

Validation:

- Four regression assertions failed twice on unfixed code for the reported missing explanation, missing blocker status, and absent dispatch after recovery; they pass after the fix.
- Added coverage for restored concurrency limits and shutdown/retry races.
- Related Compute, Session persistence, graph, recovery UI, and i18n suites: 2,606 passed and 18 skipped across the broad run and failed-file reruns. SSH socket tests passed outside the sandbox after the initial EPERM failures; two translation glossary violations were corrected and the full i18n guard passed.
- Full typecheck, final Node typecheck, ESLint for changed TypeScript/TSX files, web API map check, and git diff whitespace check passed.

Related: #2382


ewen-poch:
Confirmed the concurrent Session-limit update issue with a deferred-load regression through the public admission boundary. The retry captures a limit of 5, `setSessionLimit` successfully persists 1, and releasing the stale snapshot then admits another job as `submitted` despite one active job. The regression failed twice on the current PR code.

The proposed fix is an in-memory Session-limit mutation revision checked before installing a restored snapshot, with rereading when the snapshot is stale. Holding the admission lock across catalog loading is unsafe because catalog recovery can re-enter limit projection/cleanup. This proposal adds no persisted data or user-facing behavior, but implementation is paused for the maintainer's requested approval of additional state.

CI failures were also traced to expectations for the previous UI: `src/renderer/src/App.test.tsx:1595` still requires the recovery retry button to be absent, and the macOS `session-recovery-warning` visual baseline does not include the new explanation/button. Updating those two files is outside the original changed-file set, so those updates are also awaiting scope approval. The aggregate Module tests and PR Gate failures follow the failing checks; they are not separate confirmed product defects.

ewen-poch:
Rebased onto `2722da2a` and resolved the eight locale-catalog conflicts by preserving both the new main-branch entries and this PR's recovery copy.

Fixed the CI expectations in `6cab6b7e`: the application test now checks that “Recheck saved conversations” invokes the existing reload callback, and the recovery-warning baseline includes the approved explanation and button. The old application assertion was reproduced locally before updating it. The CI screenshot was visually inspected, and a fresh Electron visual-regression run passed against the updated baseline, including the existing narrow-viewport layout checks.

Validation: 960 related tests passed, Electron E2E build passed, the representative conversation/project/recovery visual test passed, and ESLint, formatting, and web API map checks passed.

The concurrent Session-limit snapshot issue remains pending approval for the proposed additional in-memory mutation revision. Its failing reproduction is retained locally and is not included in this CI-only follow-up; no new runtime state, persistence, or interaction change was introduced here.

ewen-poch:
Confirmed the overlapping-retry finding with a deferred-load regression through `admit()`: two catalog reads request recovery, one snapshot succeeds and the duplicate read fails, and admission remains `queued` despite available capacity. The regression failed twice on the current head. A later explicit recheck can still recover, so the reproduction establishes a discarded successful attempt, not an irrecoverable queue.

Coalescing in-flight retries is worth addressing alongside the previously confirmed stale-limit snapshot race. It would require additional in-memory coordination (an in-flight promise and a mutation revision), with no persistence or UI change. Those additions remain paused under the maintainer's explicit approval boundary. Both reproductions are retained locally and have not been pushed as failing CI tests.

PR #2398, opened 2026-09-09
fix(lifecycle): recover interrupted data cleanup independently

ewen-poch:
## Problem

Interrupted content publication can leave temporary bytes after its staging row is swept. A retained remote cleanup failure can also block an unrelated Session deletion, and a new Project deletion that fails before metadata commit does not wake idle recovery.

## Changes

- Keep Compute deletion plans per owner, preserving parent/child exclusion while allowing unrelated owners to progress. Recover independent owners before reporting aggregate failures.
- Wake existing recovery after a durable foreground deletion fails. Preserve Session projection-before-Project-removal ordering and per-Project retry accounting when independent remote recovery fails.
- Recover recognized publication temporaries during full ContentRepository sweeps, including historical files without database rows. Protect active publishers and existing shared references; targeted sweeps keep their existing scope.
- On POSIX, atomically capture a candidate into a private, locked per-attempt quarantine and verify its identity after capture. Delete only the verified private payload. Recover receipts before scanning source names so conflicts remain protected across retries. Windows verifies identity on the opened file handle and deletes through that same handle.

No database columns, tables, migrations, public enum values, UI or interaction changes. Existing deletion intents/job rows remain authority; Compute plans and active publication tracking remain in memory.

The approved persistence addition is limited to POSIX publication recovery: `content/blobs/<prefix>/.publication-recovery-<UUID>/receipt` and an optional `payload`. The versioned receipt records the original temporary basename, parent and file device/inode, size, and modification time in nanoseconds. The receipt and directories are synced before capture; capture, payload deletion and receipt retirement are synced in order. Recovery covers interruption before receipt creation, before capture, after capture and after unlink. On Linux without `RENAME_NOREPLACE`, capture uses `renameat` into the locked, previously empty private destination. Restore creates a non-replacing hard link, syncs the public parent, then unlinks only the private payload. An interrupted restore preserves both aliases behind the existing receipt barrier; no new persisted format is needed.

A reused source name retains both files and the receipt. A captured file with a mismatched identity is restored without overwriting another file, while its receipt remains a conflict barrier. Unknown names, malformed receipts, symlinked directories and references to published content are not deleted. Errors retain evidence and identify the quarantine requiring attention; nothing is recursively erased.

Historical rowless temporaries retain the approved bounded filename/root admission. Receipts are tied to the original filesystem identity: copied/rebound identities fail closed, and POSIX receipts transferred to Windows are retained rather than guessed or migrated. There is no automatic identity-rebinding migration or new recovery UI. The private quarantine follows the existing application single-writer contract; this is not a promise against deliberate same-account rewriting of private receipts or secure media erasure.

## Verification

The public source-replacement regression failed twice before the fix by observing missing replacement bytes. The extended crash-state and conflict cases also failed twice on the earlier implementation; fixed storage/native/integration checks pass 54 tests on macOS; three Linux-only syscall-failure cases are skipped there and exercised separately on real Linux. Recovery fixtures model persisted interruption boundaries; no actual power-loss or process-kill experiment is claimed.

Final local Test Impact Set: the repository-declared `project_lifecycle`, `compute_service` and `upload_repository` plans plus ContentRepository, native publication, user-file publisher, Notebook save, provenance, PDF acquisition/attachment and wiring/manifest consumers. Across 102 selected files: 3446 passed, 9 conditional skips, zero failures. Skips cover five real-SSH opt-in scenarios, a Windows-only SSH helper and the three Linux-only syscall-failure cases.

The Linux compatibility regression runs the real native removal/recovery API in a seccomp-restricted child, forcing ENOSYS, EOPNOTSUPP and EINVAL from renameat2. Each case failed twice before the fix and passed twice afterward on ARM64 Linux with Node 20.19.4, using offline cached containers. It also covers existing receipt recovery and repeated preservation of captured replacement bytes. CI supplies Node 22 platform validation. No production test hook was added.

The native macOS rebuild, Node/sandbox typechecks, global lint and module/consumer suite run after the Linux fallback change. Renderer typechecks, the application build and three no-retry repetitions of the formerly flaky macOS startup/exit scenario passed before this Linux-only follow-up; renderer behavior and the macOS mutation primitive are unchanged. Full local `npm test` was explicitly excluded by the maintainer; PR CI remains the complete/platform validation authority. Native tests retain the existing availability guard, and skips are never treated as platform evidence. Required checks and independent review must pass before merge.

### DIFF
--- a/packages/safe-file-publisher-native/index.d.ts
+++ b/packages/safe-file-publisher-native/index.d.ts
@@ -11,3 +11,11 @@ export type StoragePathCapabilities = {
 }
 
 export function inspectPath(path: string): StoragePathCapabilities
+
+export function removeAnchoredFile(
+  rootPath: string,
+  relativeParentPath: string,
+  filename: string,
+  parentDev: bigint,
+  parentIno: bigint
+): void
--- a/packages/safe-file-publisher-native/index.d.ts
+++ b/packages/safe-file-publisher-native/index.d.ts
@@ -17,5 +17,18 @@ export function removeAnchoredFile(
   relativeParentPath: string,
   filename: string,
   parentDev: bigint,
+  parentIno: bigint,
+  fileDev: bigint,
+  fileIno: bigint,
+  fileSize: bigint,
+  fileMtimeNs: bigint,
+  quarantineName: string
+): void
+
+export function recoverAnchoredRemoval(
+  rootPath: string,
+  relativeParentPath: string,
+  quarantineName: string,
+  parentDev: bigint,
   parentIno: bigint
 ): void
--- a/packages/safe-file-publisher-native/src/safe_file_publisher_native.cc
+++ b/packages/safe-file-publisher-native/src/safe_file_publisher_native.cc
@@ -639,12 +639,21 @@ bool MatchesRemovalFile(const struct stat& info, const RemovalReceipt& receipt) 
       modified.tv_sec * INT64_C(1000000000) + modified.tv_nsec == receipt.mtime;
 }
 
-// Both names are relative to held directories. Never overwrite a concurrently installed leaf.
-int MoveRemovalFile(int from, const char* source, int to, const char* destination) {
+// Both names are relative to held directories. Only capture targets the private, locked quarantine.
+int MoveRemovalFile(int from, const char* source, int to, const char* destination,
+                    bool private_destination = false) {
 #ifdef __APPLE__
   return renameatx_np(from, source, to, destination, RENAME_EXCL);
 #else
-  return static_cast<int>(syscall(SYS_renameat2, from, source, to, destination, RENAME_NOREPLACE));
+  if (syscall(SYS_renameat2, from, source, to, destination, RENAME_NOREPLACE) == 0) return 0;
+  if (errno != ENOSYS && errno != EOPNOTSUPP && errno != EINVAL) return -1;
+  // FinishRemoval already observed an absent payload while holding the quarantine lock. Its
+  // single writer can use ordinary rename here; no public source name is ever unlinked.
+  if (private_destination) return renameat(from, source, to, destination);
+  // Restore without replacing a public name. Make its link durable before unlinking the private
+  // payload. An interruption leaves both copies and the existing receipt's conflict barrier.
+  if (linkat(from, source, to, destination, 0) != 0 || fsync(to) != 0) return -1;
+  return unlinkat(from, source, 0);
 #endif
 }
 
@@ -713,7 +722,7 @@ int FinishRemoval(int parent, int quarantine, const std::string& directory, cons
     if (errno != ENOENT) return -1;
     if (fstatat(parent, receipt.name.c_str(), &source, AT_SYMLINK_NOFOLLOW) == 0) {
       if (!MatchesRemovalFile(source, receipt)) { errno = ESTALE; return -1; }
-      if (MoveRemovalFile(parent, receipt.name.c_str(), quarantine, "payload") != 0) return -1;
+      if (MoveRemovalFile(parent, receipt.name.c_str(), quarantine, "payload", true) != 0) return -1;
       // The receipt is already durable. Make capture durable before destroying its only copy.
       if (fsync(quarantine) != 0 || fsync(parent) != 0) return -1;
       if (fstatat(quarantine, "payload", &payload, AT_SYMLINK_NOFOLLOW) != 0) return -1;
--- a/packages/safe-file-publisher-native/src/safe_file_publisher_native.cc
+++ b/packages/safe-file-publisher-native/src/safe_file_publisher_native.cc
@@ -1,6 +1,8 @@
 #include <node_api.h>
 
 #include <cerrno>
+#include <charconv>
+#include <sstream>
 #include <cstddef>
 #include <cstring>
 #include <limits>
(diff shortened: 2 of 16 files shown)


ewen-poch:
The native availability feedback is addressed in 64f56da0. I reproduced the missing-addon failure twice before changing the suite to use the existing availability guard. With the addon built, all 8 native tests pass; with it deliberately unavailable, all 8 are skipped and the command succeeds. Skips are not counted as platform validation.

The POSIX leaf-replacement finding is valid and remains blocking. A public ContentRepository sweep regression replaces the leaf after the final path check and fails twice on the current implementation because the replacement bytes are removed. Passing another inode check into native code would narrow, but not eliminate, the check-to-unlink window. A complete quarantine/recovery protocol introduces durable filesystem receipts beyond the approved scope, so that design is awaiting maintainer confirmation. The PR will not be merged with this unresolved preservation failure.

The failed PR Gate jobs were both flaky E2E failures: macOS timed out waiting for Electron to exit, and Windows timed out waiting for database startup to reach ready. Both passed their built-in retry but correctly failed the fail-on-flaky gate. The macOS main-process attachment records application shutdown and backend teardown completing in 21 ms before the process-exit timeout; the Windows failure occurred during fixture startup, before the Memory conflict scenario ran. I have retained the logs and traces and have not relaxed timeouts or disabled the flaky check. The new head reruns these checks; current evidence does not establish a lifecycle-cleanup root cause.

ewen-poch:
The latest review of 64f56da0 now reports only the POSIX leaf-replacement issue. It is the already reproduced blocker awaiting approval for a durable quarantine/recovery protocol; no extra persistence, state values, UI, or out-of-scope file changes have been introduced.

For the macOS CI failure, I built the current application and ran the exact unavailable-saved-model Electron scenario three times with retries disabled. All three passed, with no skipped or flaky result. This does not establish a root-cause fix for the original exit timeout. The Windows startup timeout has not been locally reproduced on macOS. The new PR Gate run re-executes both platform checks, with the existing timeouts and fail-on-flaky policy unchanged.

Further cleanup implementation is paused at the maintainer's design boundary. The failing replacement-preservation regression remains available locally; the PR stays unmerged.

ewen-poch:
The leaf-replacement finding is addressed in 703a7a6 after approval of the additional filesystem persistence. POSIX no longer unlinks the candidate's source name: it first records and syncs a bounded recovery receipt, atomically captures the candidate without replacement into a private locked quarantine, and verifies the captured identity before deleting the private payload. A reused source name preserves both files; a captured replacement is restored without overwriting anything, and the retained receipt prevents later sweeps from treating it as a fresh orphan. Windows verifies device/inode, size and modification time on the same open handle used for deletion.

The public replacement-preservation test failed twice before the fix and now passes. Persisted interruption fixtures cover pre-receipt, pre-capture, post-capture and post-unlink recovery, plus repeated conflicts, captured replacement, modified payload, malformed receipts, linked quarantine and rejection of published-content names. Storage/native/integration checks pass 54/54 on macOS. The final module/consumer set passes 3446 tests with 6 existing conditional skips; typechecks, lint, build and three no-retry repetitions of the previously flaky macOS startup/exit scenario also pass.

The only added persistent format is the approved POSIX quarantine receipt under the existing managed blob parent. There are no database/schema, public-enum or UI changes. Receipts copied to a different filesystem identity, or transferred to Windows, fail closed rather than being rebound automatically. This uses the existing application single-writer contract for private quarantine contents.

The previous head's PR checks are now complete without failures. New exact-head CI and review must still validate this implementation; neither cancelled Windows tests nor availability-guard skips are counted as platform proof.

ewen-poch:
The latest review identifies a remaining Linux compatibility gap in 703a7a63: removal has no fallback when the filesystem rejects RENAME_NOREPLACE. Inspection confirms that these errors preserve the candidate and its receipt and propagate out of recovery; this is a cleanup-availability concern, not evidence that the replacement-preservation fix deletes data. I have not yet reproduced the unsupported-filesystem case in a Linux runtime, so I am not claiming a validated fix.

The existing publication linkat/unlinkat fallback cannot simply be copied into removal: unlinking the public source name would reintroduce the replacement race this change addresses. A fallback needs its own externally observable failure reproduction and interruption/preservation coverage before implementation. No additional persistence format, schema, enum, UI, or file-scope change has been made for this finding. Implementation is paused for maintainer direction, and the PR remains unmerged.

The current head's static checks, Windows core and Linux Notebook isolation have passed. The remaining PR Gate and supplemental Windows full tests are still running; pending checks are not counted as successful validation.

ewen-poch:
The Linux compatibility finding is addressed in dadab52c. I reproduced ENOSYS, EOPNOTSUPP and EINVAL from renameat2 in a real ARM64 Linux child using a seccomp filter. Each case failed twice against the previous implementation at the native removal boundary, then passed twice after the fix. The regression exercises the public native removal/recovery operations, including existing receipts and repeated replacement preservation; no production test hook was added.

The fallback is deliberately directional. Capture uses renameat only into the previously absent payload of the held, private, exclusively locked quarantine. Restore uses linkat without replacement, syncs the public parent, and unlinks only the private payload. It never performs a fallback unlink of the public source name. An interrupted restore leaves both aliases protected by the existing receipt; unsupported hard links preserve the payload and receipt. This reuses the approved single-writer contract and receipt format, with no new schema, enum, UI, persistence fields or changed-file scope.

After the final edit, the macOS native build and 54 focused checks pass; three Linux-only tests are correctly skipped on macOS and separately validated on Linux. The complete selected module/consumer set passes 3446 tests with 9 conditional skips. Node/sandbox typechecks, global lint and diff checks pass. The offline Linux reproduction uses Node 20.19.4; exact-head CI remains responsible for Node 22 validation. The Linux syscall behavior is documented in https://man7.org/linux/man-pages/man2/renameat2.2.html.

PR #2400, opened 2026-09-09
feat(cli): add JSON readiness projection for headless environments

LiuZhaoyan:
## Summary

Add JSON-only `open-science doctor --json`, `OpenScienceClient.doctor()`, and authenticated `GET /api/v1/doctor` so headless callers can inspect selected runtime/provider readiness and enabled, available Skills.

Distinguish absent resources (`missing`) from configured but unusable resources (`not_ready`) and return stable next-action codes. Codex inspection requires nonempty locally stored API-key or access-token material: an empty `auth.json` object no longer passes because a provider validated previously. Inspection does not launch/repair Codex or normalize its credential configuration; remote token expiry/revocation validation remains outside this local check. Authorization is revalidated after asynchronous inspection before returning results.

## Architecture and scope

CLI → OpenScienceClient → Task HTTP API → HeadlessTaskApi → existing Settings preflight and Skill catalog owners.

No new runtime installation, onboarding mutation, automatic fix command, credential storage, or database migration. Related to #1987; this is a diagnostic slice, not complete first-run bootstrap.

The follow-up merges current main while preserving the original contributor commits and both Settings test changes.

- Public doctor states: `ready`, `missing`, `not_ready`; next codes: `runtime_missing`, `runtime_not_ready`, `provider_missing`, `provider_not_ready`. Provider failures can include a reason.
- Existing renderer preflight boolean contract remains available.
- No historical data migration, stored-field addition, or credential rewrite. Local inspection recognizes API-key/access-token material with or without an auth-mode field.
- No new states were introduced by the final credential-validation fix.

## Validation

After the final source edit:

- Credential inspection, shared readiness, and doctor CLI/SDK/Task composition → `npm test -- packages/open-science/cli.test.ts packages/open-science/client.test.ts src/main/web-service/task-api.test.ts src/main/application-command-composition.test.ts src/main/settings/preflight.test.ts src/main/settings/provider-auth-lifecycle.test.ts src/main/settings/agent-runtime-manager.test.ts src/main/settings/service.test.ts src/main/settings/codex-auth.test.ts` → **576 passed**.
- Authenticated HTTP doctor and revocation during inspection → `npm test -- src/main/web-service/http-server.test.ts -t 'rejects a delayed doctor result|serves the versioned task API'` → **2 passed**, 63 unrelated tests skipped.
- `npm run lint` and `git diff --cached --check` → **passed**.
- `npm run typecheck:node` and `npm run typecheck:web` → **blocked by missing local `reflect-metadata` and `@peculiar/x509` dependencies** in unchanged Notebook network gateway code; not reported as passed.

The new credential-free fixtures failed before the fix and passed afterward. Independent review found no blocking issue in the final fix or conflict resolution. The change affects the common Settings preflight consumed by doctor and the renderer; owner and consumer checks are included. It adds no ACP wire behavior, platform path semantics, or persisted format, so no extra local protocol/platform migration matrix was introduced. Full portable and platform verification is delegated to the required CI on the updated head; prior-head green results are not reused.

### DIFF
--- a/packages/open-science/CLI.md
+++ b/packages/open-science/CLI.md
@@ -88,9 +88,11 @@ open-science doctor --json
 ```
 
 The command prints one JSON object. `ready` requires both the selected Agent runtime and active
-provider to be ready. `checks.skills.enabled` contains the sorted IDs of enabled, available Skills;
+provider to be ready. A check is `missing` when its resource is absent and `not_ready` when it is
+configured but unusable; provider failures may include a stable `reason`, such as
+`credential_invalid`. `checks.skills.enabled` contains the sorted IDs of enabled, available Skills;
 Skills do not block overall readiness. `next` contains stable action codes such as
-`runtime_missing` and `provider_missing`, never commands that may not exist. A completed inspection
+`runtime_missing` and `provider_not_ready`, never commands that may not exist. A completed inspection
 exits with code `0` even when `ready` is false; connection and command failures keep the existing
 nonzero CLI error codes.
 
--- a/packages/open-science/CLI.md
+++ b/packages/open-science/CLI.md
@@ -90,8 +90,13 @@ open-science doctor --json
 The command prints one JSON object. `ready` requires both the selected Agent runtime and active
 provider to be ready. A check is `missing` when its resource is absent and `not_ready` when it is
 configured but unusable; provider failures may include a stable `reason`, such as
-`credential_invalid`. `checks.skills.enabled` contains the sorted IDs of enabled, available Skills;
-Skills do not block overall readiness. `next` contains stable action codes such as
+`credential_invalid`.
+
+Codex credential inspection checks supported material in the app-owned `auth.json` without launching
+Codex or validating expiry/revocation remotely; historical provider validation remains a separate gate.
+
+`checks.skills.enabled` contains the sorted IDs of enabled, available Skills; Skills do not block
+overall readiness. `next` contains stable action codes such as
 `runtime_missing` and `provider_not_ready`, never commands that may not exist. A completed inspection
 exits with code `0` even when `ready` is false; connection and command failures keep the existing
 nonzero CLI error codes.
--- a/packages/open-science/CLI.md
+++ b/packages/open-science/CLI.md
@@ -78,6 +78,22 @@ Among lifecycle commands, `status`, `stop`, and `update` support `--json`. `star
 it; run `start --no-open` followed by `status --json` for machine-readable startup status. Errors
 with `--json` are reported on stderr as a JSON error object with a nonzero exit code.
 
+## Readiness checks
+
+Inspect the running service without changing runtime, provider, credential, onboarding, or Skill
+settings:
+
+```bash
+open-science doctor --json
+```
+
+The command prints one JSON object. `ready` requires both the selected Agent runtime and active
+provider to be ready. `checks.skills.enabled` contains the sorted IDs of enabled, available Skills;
+Skills do not block overall readiness. `next` contains stable action codes such as
+`runtime_missing` and `provider_missing`, never commands that may not exist. A completed inspection
+exits with code `0` even when `ready` is false; connection and command failures keep the existing
+nonzero CLI error codes.
+
 ## Application updates
 
 Check, download, and apply an Open Science application update without opening the browser or desktop
--- a/packages/open-science/cli.mjs
+++ b/packages/open-science/cli.mjs
@@ -34,6 +34,7 @@ Commands:
   status      Show backend status
   url         Print the authenticated web URL
   update      Check, download, and apply an application update
+  doctor --json  Inspect headless readiness
   codex login [--force]
   connector list | show <id> | enable <id> | disable <id>
   connector add | update <id>      Read configuration JSON from stdin
@@ -146,6 +147,7 @@ const VALUE_OPTIONS = {
 }
 
 const TASK_COMMANDS = new Set([
+  'doctor',
   'project',
   'run',
   'session',
@@ -168,6 +170,7 @@ const GROUP_COMMANDS = new Set([
 // Project create, update, and session-defaults intentionally remain unbounded because their
 // positional Project names may contain multiple unquoted words.
 const POSITIONAL_LIMITS = new Map([
+  ['doctor', 0],
   ['connector list', 0],
   ['connector show', 1],
   ['connector enable', 1],
@@ -378,6 +381,9 @@ export const parseCliArgs = (argv) => {
   }
   if (options.json && options.jsonl) {
     throw new CliUsageError('Use only one of --json or --jsonl.')
+  }
+  if (command === 'doctor' && !options.json) {
+    throw new CliUsageError('doctor requires --json.')
   }
   if (options.json && (command === 'start' || command === 'url')) {
     throw new CliUsageError(`--json is not supported for ${command}.`)
@@ -1438,6 +1444,11 @@ export const runTaskCommand = async (parsed, dependencies = {}) => {
   const deps = { ...TASK_DEPS, ...dependencies }
   const { command, subcommand, positionals = [], options } = parsed
   const client = await deps.connect({ configRoot: options.configRoot })
+
+  if (command === 'doctor') {
+    deps.log(JSON.stringify(await client.doctor()))
+    return
+  }
 
   if (command === 'connector' || command === 'credential') {
     const id = positionals[0]
--- a/packages/open-science/cli.test.ts
+++ b/packages/open-science/cli.test.ts
@@ -39,10 +39,10 @@ describe('task CLI', () => {
       checks: {
         daemon: { status: 'ready' },
         runtime: { status: 'missing', framework: 'codex' },
-        provider: { status: 'missing' },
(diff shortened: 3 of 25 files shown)


ewen-poch:
Thanks for keeping this focused and reusing the existing Settings owners. The underlying gap is real: the CLI can inspect service status, but previously had no unified machine-readable runtime/provider readiness report. This is a useful incremental slice of #1987; I would not require the rest of bootstrap or Settings parity in this PR.

Before merging, could we tighten the diagnostic contract in two places?

1. **A previously validated Codex provider can still produce `ready: true` after its app-owned authentication is lost.** `doctor()` directly promotes `preflight.activeProviderReady` into readiness (`src/main/web-service/task-api.ts:244–251`), but `ProviderAuthLifecycleOwner.isProviderKeyUsable()` unconditionally returns `true` for Codex subscription providers (`src/main/settings/provider-auth-lifecycle.ts:526–527`). This path never reads the current Codex auth status. A valid runtime plus historical validation can therefore yield `ready: true` and an empty `next`, even though authentication is now missing. Could we cover this case through the existing auth owner, or explicitly narrow the contract to stored configuration/preflight readiness? The relevant case is missing app-owned credentials, not merely logging out of the external `~/.codex` profile, since shared credentials are imported.

2. **`missing` is not equivalent to a failed preflight.** Lines 241–251 report `provider_missing` for an existing provider with failed validation, unusable credentials, or an incompatible model/framework. Likewise, an installed runtime can fail its executable/version check. This loses the distinction an automation client needs to select a recovery action. A neutral `not_ready` status/code would be more accurate if detailed reasons are outside this slice; otherwise, preserve the reason through the existing owner. Please add cases for an existing incompatible provider and an installed-but-unusable runtime before freezing these public values.

One smaller HTTP consistency issue: the new route checks external authorization only before `await tasks.doctor()` (`http-server.ts:992–995`). If authorization is revoked during inspection, the report is still returned. The adjacent connector/credential route rechecks after the awaited operation. Please apply the same post-await check here, ideally with a delayed-inspection revocation test. The response is limited to readiness and Skill IDs, so this has a smaller impact than exposing credentials, but it should honor revocation consistently.

Validation on `117e642fa87d77176de915a2c3b8bba088f69742`: 174 focused CLI/client/Task/composition/preflight/auth tests passed, plus the authenticated versioned HTTP API test. I also checked the exact-commit CI results; PR Gate and selected checks passed. No source files were changed during this review.

ewen-poch:
Thanks for the follow-up changes. I re-reviewed `2d7f3c1ccfbdbbe8d57bac4f0bc37b98084fd153`. The new `not_ready` distinction addresses the misleading missing-resource diagnosis, and checking current Codex authentication through its owner is the right direction. There are still two integration issues in that new call path, plus the earlier HTTP follow-up:

1. **A broken/missing Codex runtime now prevents the diagnostic report itself.** At `provider-auth-lifecycle.ts:529`, `getStatus()` opens an auth session through `resolveCodexExecutable()`. That resolver throws when the native path is absent, the managed adapter is missing, or the adapter version is unsupported. `AgentRuntimeManager.getPreflight()` awaits this without converting expected probe failures into diagnostics. With a retained Codex provider and a removed/broken runtime, `doctor` therefore fails instead of returning its runtime `missing`/`not_ready` result and `next` action. This also affects the existing renderer preflight. I reproduced the exception propagation using the actual controller and auth-owner method with a failing session opener. Could we preserve a structured diagnostic outcome when the authentication probe cannot run, rather than let the dependency being diagnosed abort the report?

2. **The new status path is not read-only.** Its executable resolution calls `ensureManagedCodexContextUsage()`, which can rewrite the installed adapter. Opening the auth session also calls `ensureCodexAuthHome()` / `ensureCodexAuthConfigHome()`, which can create the profile and write `config.toml` to normalize credential-store and transport configuration. These are existing setup helpers, but they are newly reachable from `doctor` and conflict with the documented promise not to change runtime or credential settings. Please keep diagnosis separate from repair/normalization and cover unchanged adapter/config files in the regression tests.

3. **The previously mentioned authorization-revocation check remains unchanged.** `http-server.ts:992–995` checks only before awaiting `tasks.doctor()`. Please recheck before writing the response, as the neighboring connector/credential route does, and cover revocation during a delayed inspection.

Compatibility notes: `not_ready`, `runtime_not_ready`, `provider_not_ready`, and the provider reason values expand the new public API contract. No database migration or new stored field was added. However, the filesystem writes in item 2 are a real persistence effect despite the absence of a schema change.

Validation: the seven focused CLI/client/Task/composition/preflight/auth/runtime-manager suites passed (218 tests). Settings service and Codex auth suites passed on rerun with local-port access (321 tests); the first sandboxed Settings run hit port restrictions. Exact-commit PR Gate and selected CI checks are successful. No source files were changed. I would address the remaining integration issues before merging.

ewen-poch:
Thanks for addressing the previous integration issues. I re-reviewed `69dd36240c6b1c166ee930d6e5a513416c3125bf`: credential inspection no longer launches/repairs Codex or normalizes its configuration, and the HTTP route now correctly rechecks authorization after inspection.

One readiness issue remains before I would recommend merging:

**`inspectAppOwnedCodexAuthentication()` treats arbitrary JSON objects as usable credentials** (`src/main/settings/codex-auth.ts:642–644`). I executed the actual inspection function against `{}`, `{"tokens":{}}`, `{"tokens":null}`, and `{"tokens":{"access_token":""}}`; every case returned `{ state: 'present' }`. The auth owner maps `present` directly to `isProviderKeyUsable: true`, so a previously validated provider with one of these credential-free files can still be reported as ready. This checks JSON syntax/object presence, rather than credential presence.

Could we keep the new read-only approach but recognize supported stored credential shapes, rejecting objects with no usable credential material? Please cover empty objects and null/empty credential fields alongside supported credential variants. This does not require a live network validation or a return to the runtime-launching path.

GitHub currently also reports `CONFLICTING` / `DIRTY`. The latest head has administrative/integrity checks but no successful PR Gate/test/build result, so the earlier green CI is not evidence for this revision. Please resolve the base-branch conflicts and let the required checks complete on the resulting head.

No source files were changed during review. The focused Codex auth/provider/preflight/Task suites passed (112 tests).

ewen-poch:
Re-reviewed `e6505254d39a3cc3122ed929706aba8822cd90fe`. The previous empty-credential, runtime side-effect, and authorization-revocation issues are addressed. I found no additional confirmed blocking production-code regression in the supported path, but the current head is not ready to merge because a required test is reproducibly failing.

**Please update the remaining positive Settings integration fixture** (`src/main/settings/service.test.ts:3056`). The test `resolves a validated codex-isolated subscription without API routing` still writes only `{"tokens":{"access_token":"app-owned"}}`, while the new validator intentionally requires the fuller ChatGPT token structure and `last_refresh`. It therefore receives `activeProviderReady: false` at line 3065 instead of the expected `true`.

This is the actual failing assertion in the Ubuntu shard-1 blob report for run [34572081079](https://github.com/aipoch/open-science/actions/runs/34572081079), and I reproduced it locally with:

`npm test -- src/main/settings/service.test.ts -t 'resolves a validated codex-isolated subscription without API routing'`

Please replace this positive fixture with the same realistic supported ChatGPT credential shape used in the new inspection tests, preserve the successful readiness assertion, and rerun the Settings suite and required CI. The Module tests/coverage and PR Gate failures propagate from the failed unit shard; they are not three independent reported defects.

The four focused auth/provider/preflight/Task suites passed locally (140 tests), and current-head static checks and platform E2E checks passed in CI. GitHub reports no merge conflict. This review changed no files.

Compatibility note: the new checks are stricter readers of existing auth files, not a data migration or a persisted-format change. No new public readiness state was added in this revision. Keep the documented distinction between locally supported credential structure and live remote credential validity.

ewen-poch:
Fixed the remaining positive subscription fixture in `cc54971d7`. It now uses the same complete ChatGPT credential shape as the other readiness tests. The `activeProviderReady: true` and backend assertions are unchanged; no production behavior, public states, persisted formats, or migrations changed.

After the edit, `npm test -- src/main/settings/service.test.ts src/main/settings/codex-auth.test.ts src/main/settings/provider-auth-lifecycle.test.ts src/main/settings/preflight.test.ts` passed **391 tests**. Single-file ESLint, Prettier, and diff checks passed. Independent review found no issue. The original CI failure was reproduced before the edit. Full verification remains with CI on this new head.

ewen-poch:
Fixed the remaining portable-test failure in the shared-interpreter queue test. The test assumed the second run would be durably admitted within 20 ms; on the CI runner it was still absent when the assertion ran.

The test now waits for the observable queued state before checking serialization. A deliberately delayed repository append reproduces the original failure and passes with the fix. The original concurrency, execution-order, and completion assertions remain intact. No production code, persisted formats, or states changed.

- Targeted queue regression: passed, including the delayed append.
- Entire runtime-service test file: 385 passed, one unrelated shell exit-code test failed locally. That shell test also fails with the unmodified HEAD file in the same worktree.
- Full repository lint and changed-file formatting checks: passed.
- Independent review: passed.
- Node typecheck: blocked by missing local reflect-metadata and @peculiar/x509 dependencies.

The new head still requires the PR CI results; local verification is not being presented as a completely green suite.

PR #2408, opened 2026-09-09
fix(update): preserve artifact integrity and recovery

ewen-poch:
## Problem

Cancelled downloads could become ready again when transfer finalization arrived late. A manually
downloaded installer could be replaced before opening, and a refused install gate left a downloaded
update without a direct installation retry. Linux download estimates assumed AppImage, and research
blockers remained English in translated dialogs. Historical website mirrors also overwrote newer
channel entry points.

## Proposed change

- Accept download events only for the actual uncancelled transfer, including while a retry drains
  the old request; check manual download ownership before publishing ready.
- Recheck a manual installer's file type, size and streaming SHA-256 before opening. Reuse `applying`
  for visible verification and return invalid files to download. Preserve stale-response protection
  and coalesce concurrent apply calls.
- Keep a ready artifact with actionable error feedback after pre-install gate refusal. Every retry
  runs the gate again; installer handoff failures still remain failures.
- Omit Linux pre-download estimates when package format is unknown. Actual SDK download progress
  continues to supply the size; package selection is unchanged.
- Translate research/subagent and installer-validation feedback in all eight renderer catalogs.
- Default website mirrors to versioned backfill. Explicit promotion serializes channel writers,
  compares every current feed and manifest version, requires complete platform feeds, and permits
  same-version recovery after a partial upload.

## Scope and non-goals

No application database, settings, public status field or enum is added. Installer validation reads
existing files. Publication preserves existing client paths/formats and adds a versioned copy of the
existing `version.json` for recovery; no new release ledger or automatic rollback is introduced.

Windows installer name-only process termination remains deferred: no native Windows reproduction
was available, so this PR does not modify or claim to validate that recovery path. SideChat process
reaping is a separate unresolved gate dependency, not fixed by the mocked gate tests here. This PR
also does not claim a native installer run, adversarial same-user TOCTOU protection or atomic
multi-object S3 publication. No real release or storage publication was performed.

Desired-behavior regression assertions failed repeatedly on unchanged production code before repair.
The failures matched the reported cancellation, changed-file open, gate retry, untranslated alert,
unsupported size estimate and old-version overwrite behaviors. Positive controls passed. Publication
reproduction executes the actual workflow command against local fake storage; it does not simply
assert workflow text. Transfer timing uses explicit Promise barriers and existing public boundaries.

Final local coverage: 26 selected files / 1,151 passing tests, plus the final added publication
concurrency contract (17 publication tests passing). After return-type/style-only cleanup, the four
affected test files passed again (41 tests). All listed behavior checks ran after the last production
logic edit. The test-impact classifier requests full CI because publication scripts/tests are global
or unowned inputs; full local testing was intentionally omitted at the maintainer's request.

| Behavior | Check | Result |
| --- | --- | --- |
| Cancellation, drain, integrity, gate retry and creation/IPC contracts | `npx vitest run src/main/update` | Passed |
| Shared transfer and version selection | `npx vitest run src/main/net/resilient-download.test.ts src/shared/update.test.ts` | Passed |
| Dialog/capsule/store and host command consumers | UpdateDialog render/lifecycle/blockers, UpdateCapsule render, update-store, host-application-commands and application-command-wiring tests | Passed |
| Locale completeness and actual localized rendering | `npx vitest run src/renderer/src/i18n/resources.test.ts src/renderer/src/components/UpdateDialog.blockers.test.tsx` | Passed |
| Publication, transforms and workflow contracts | mirror-channel-publication, release-workflows, windows-release-workflows, generate-version-manifest and inject-feed-notes tests | Passed |
| Runtime types | `npm run typecheck:node`; `npm run typecheck:web` | Passed |
| Style | `npm run lint`; changed-file Prettier; `git diff --check` | Passed |
| Visual behavior | Real UpdateDialog, repository CSS and i18n in ego-browser: Chinese retry, verification and invalid installer states | Screenshots inspected locally |

macOS local public-boundary tests cover file access using platform path helpers. The Bash publication
harness intentionally skips Windows because production publication runs on Ubuntu. Native packaging
and complete platform validation remain the responsibility of PR CI.

## Review focus

Check transfer-event ownership while a newer request waits, gate/admission recovery, stale manual
verification results, and same-version publication recovery without downgrade. A channel concurrency
group does not protect against independent external storage writers. Missing/unreadable current
metadata fails closed; initial channel bootstrap and explicit rollback remain separate operations.

### Current verification and review boundaries

Current-head PR Gate, all Ubuntu portable-suite shards, merged coverage, Windows core/E2E, both macOS
E2E lanes and CodeQL passed. The three host-dependent size assertions exposed by CI now explicitly
exercise macOS, Windows and Linux; the updated owner suite passed all 178 tests. The strategy creation
suite also passed on the final head.

- Linux estimates are intentionally omitted because the adapter lacks the selected package format.
  The installed SDK can choose DebUpdater from resources/package-type, so restoring a platform-only
  AppImage assumption would reintroduce the mixed-feed error. Actual transfer totals remain available.
- The reported Windows manual same-path retry is outside the production Windows update path:
  createUpdateStrategy selects ElectronUpdaterStrategy for win32. A local manual cancel/drain/retry
  probe also passed without changes (macOS, not native Windows). The downloader uses fs.rename, and
  [libuv's Windows implementation](https://github.com/libuv/libuv/blob/v1.51.0/src/win/fs.c#L2109-L2115)
  supports replacing an existing destination. No speculative deletion of the user's chosen path was added.
- Revalidation does not eliminate concurrent pathname replacement before shell.openPath. The bounded
  repair detects stale or accidentally changed downloads; an immutable native handoff is outside this
  change. This remaining race is acknowledged, not claimed fixed.

The latest automated review still reports needs changes on these points. The evidence above records
the disposition and limits; it does not claim that review approved the final head.


ewen-poch:
Response to the update review findings:

- **Linux download-size estimate:** The omission is intentional when the selected package format is unknown. electron-updater can select DebUpdater from `resources/package-type`; a platform-only AppImage mapping gives the wrong estimate for mixed feeds. This accepts the loss of the pre-download estimate, including for AppImage users, while retaining actual SDK transfer totals. The host-dependent tests have been corrected to explicitly cover macOS, Windows and Linux. Package-format-aware estimation is not being added to this bounded change.

- **Cancellation and same-path retries on Windows:** `createUpdateStrategy` selects `ElectronUpdaterStrategy` for Windows; the reported `UpdateService` path is not the production Windows update path. A public-service cancel/drain/same-path-retry probe passes on macOS without a production change. This is not native Windows validation, but destination existence alone is also insufficient to establish the claim: [libuv uses `MOVEFILE_REPLACE_EXISTING` on Windows](https://github.com/libuv/libuv/blob/v1.51.0/src/win/fs.c#L2109-L2115). The strategy-selection suite passes all 13 tests. No speculative deletion of the user's chosen destination is being added.

- **Replacement between verification and opening:** This residual race is real and remains explicitly acknowledged. The approved repair revalidates downloads changed while awaiting installation; it does not promise protection against a concurrent same-user writer. The PR adds verification to a path that previously opened the installer without revalidation, rather than introducing the replacement window. Eliminating the window would require a separately designed and validated native handoff; an extra hash or another pathname check would not establish that guarantee. That architectural expansion is not being made in this PR.

No further code changes are proposed for these findings. Current-head full CI, including portable tests, coverage, Windows/macOS E2E and CodeQL, passed. These dispositions do not change the automated review's `needs changes` verdict or claim that the remaining race has been fixed.

PR #2418, opened 2026-09-09
fix(skills): restore scoped Codex document loading

ewen-poch:
## Problem

Codex advertises Connector Skill paths while native shell execution is disabled. Explicitly selected
Skills already load through native Skill inputs, but an unselected Connector document has no
supported progressive reader. The reported Agent tried Notebook Shell/REPL filesystem reads, then
mistook the authored-Skill composer (`host.skills`) for the generated Connector catalog and continued
retrying after failures.

## Proposed change

- Reuse the bounded `skills/load_skill` MCP loader for Codex's app-owned Skill directory. Keep the
  current Specialist allowlist and omit the loader from restricted roles or empty scopes.
- Wire API and subscription Skill roots separately. Preserve native selected Skill inputs.
- Refresh a live Codex loader with close → compatible resume of the same native thread on Specialist
  switches. Reject stale resume/adoption before publication, prevent attached-resume scope changes,
  and keep failed or superseded reconfiguration from retaining the prior loader. Existing history,
  binding revisions and pending-switch handling are reused.
  Check negotiated close and resume support before detaching/provisioning; an unsupported runtime gets an
  explicit update/retry error rather than an unsupported request, stale scope, or silent context reset.
- Expose only `mcp__skills` directly using Codex's existing
  [`direct_only_tool_namespaces` configuration](https://github.com/openai/codex/blob/main/codex-rs/core/config.schema.json).
  Preserve its current request schema through the Chat bridge; restricted sessions still drop it.
  Real Codex 0.144.6 checks confirmed this is needed because deferred `tool_search` is not forwarded.
- Auto-approve this read-only loader only when its app-owned server and trusted Codex tool identity
  match. Keep shell/REPL restrictions intact.
- Explain the composer/catalog boundary and provide stopping/recovery guidance through existing
  Notebook instructions, failure hints and authored-Skill read errors.

## Scope and non-goals

No database, persisted-format, historical-data migration, state-enum, page or human interaction
changes. `skillsDirectory` and its environment variable are optional internal process configuration;
the existing `CODEX_CONFIG` gains one tool-exposure setting. Existing ACP child lifecycle owns the
loader. The default Claude projection layout remains supported.

This does not diagnose the original PowerShell FileSystem initialization failure or classify arbitrary
REPL data such as `{ok:false}` as runtime failure. It does not merge the authored and generated Skill
catalogs, add a retry state machine, or permit raw access to Skill directories.

The initial 12-test regression set ran twice against unchanged production files from `b50946ae7`: **6 failed,
6 passed each time**. With the fix, all **12 passed**. The failures were missing unselected document
bodies/loader schemas and missing guidance, rather than missing dependencies or failed test startup.
The supplied trace does not identify selected Skill inputs; the original multi-turn model loop and
Windows sandbox startup failure have not been reproduced end to end.

Checks below ran after the last material production edit; the final test-only edits were rerun too.
Commands use the existing Vitest directly because `npm test` pretest rejects the shared stale Prisma
client. No full local suite or dependency installation was run.

| Behavior | Project-owned check (`npx --no-install vitest run <paths>`) | Result |
| --- | --- | --- |
| Bounded catalog, traversal/link rejection, composer boundary | `src/main/skills/{runtime-mcp-server,host-skills-service}.test.ts` | Passed |
| Framework config and existing loading paths | `src/main/agent-framework/{codex,claude-code,opencode,codebuddy}.test.ts` | Passed; 2 optional native tests skipped |
| Notebook public result and captured Windows error guidance | `src/main/notebook/{mcp-server,skill-loading-guidance}.test.ts` | Passed |
| Trusted permission identity, scope, create/resume and selected inputs | `src/main/acp/{permission-policy,permission-context,session-presentation-policy,provider-session-creator,provider-session-resumer,turn-skill-owner,session-capability-owner}.test.ts` | Passed |
| Provider roots, request/reply translation and restricted roles | `src/main/settings/{backend-resolver,responses-bridge,responses-bridge.skill-loading,native-responses-compatibility}.test.ts` | Passed |
| Actual Codex + ACP document delivery over Responses and Chat, selected/unselected | `src/main/settings/codex-skill-loading.integration.test.ts`, with `CODEX_ACP_PATH` and `CODEX_NATIVE_PATH` | 4 passed; local synthetic upstream, no paid model call |
| Source lint | `npm run lint` | Passed |
| Sandbox package types | `npm run typecheck:sandbox` | Passed |
| Main/renderer types | `npm run typecheck:node`; `npm run typecheck:web` | Blocked by shared generated Prisma client mismatch; all 558 main diagnostics reproduce on baseline, no added production diagnostics |

The final impact run passed **764 tests across 27 files**, with 4 optional tests skipped. Lifecycle
coverage additionally includes `provider-session-adopter.test.ts`, `session-replacement-workflow.test.ts`
and `runtime-provider-session-composition.test.ts`. Targeted Runtime consumers passed **16 tests** via
`npx --no-install vitest run src/main/acp/runtime.test.ts -t 'specialist|Specialist|[Ll]iterature.*[Mm][Cc][Pp]|reconfigur'`.
Every lifecycle finding was reproduced before its fix: unchanged MCP on switching, stale resume and
adoption, attached-resume bypass, late switch after another prompt/binding, and restored attachment
generation. Real Codex testing proved plain resume retained old MCP processes; close → resume
replaced scope, allowed the new Skill, rejected the prior Skill, removed empty-scope tools and restored
them later on both routes, including an already-closed retry. The actual Codex integration is opt-in and
skips when native paths are absent in CI. Cross-platform directory-link fixtures use Windows
Junctions or POSIX directory symlinks. PR CI with lockfile-installed/generated dependencies is
authoritative for full typing and platform coverage. No screenshot is applicable because no renderer
surface changed; subscription authentication was not exercised against a live account.

## Review focus

Current Skill scope through create/resume/switch/adoption and restricted roles, exact trusted tool
identity, directory containment, and request-local bridge declarations. Independent Standards and
Spec reviews identified the lifecycle gaps above; those were reproduced and fixed. Both final reviews
found no remaining blockers and confirmed the final evidence mapping.

PR review reproduced four unsupported-close failures on both Codex routes (attached and detached);
the negotiated-capability guard now passes them. CI exposed three further consumer failures, all
reproduced locally before changes: the added Notebook instructions exceeded the static context budget,
and two old fake providers omitted their advertised resume protocol/modes. Guidance is now ~35 tokens
with a documented 40-token budget increase; fixtures implement the advertised protocol and retain
their assertions. Final coverage adds `connection-lifecycle-workflow.test.ts`,
`specialist/specialist-identity-injection.test.ts`, `context-usage-static-context.test.ts`, and
`codex-completion-handoff.integration.test.ts`. The final impact set and Runtime consumers ran after the last material edit. Both independent reviews passed.

Prior-head CI passed lint, Node/Web types, Windows core/E2E and macOS presentation. Its first macOS
functional attempt hit `database_validation_failed` at `0030_literature_foundation`; the same test's
retry passed. No database code changed. That failure remains recorded separately pending current-head
CI; a green earlier lane is not substituted for current-head verification.

A full Runtime/ACP regression additionally reproduced loss of the old attachment after an unsupported
Specialist switch (`ACP session not found` on the next prompt). The same capability preflight now
runs before binding mutation; the original runtime identity and attachment remain usable. Failures
after a real scope change still discard the stale loader. This does not change the UI's existing
persisted pending-switch send guard. No new persisted state or history compatibility layer is added.

The preflight also requires negotiated resume support. The public Runtime regression fails on the
pre-fix code with the unsupported request and now preserves the original session for either missing
capability. The automated review's detached-close hypothesis was checked against the actual adapter:
on both Responses and Chat, terminate the adapter, start a fresh process/connection, close the saved
but locally unloaded session ID, and resume it successfully. This test passed without changing the
close behavior; no exception swallowing or speculative legacy fallback was introduced.

### Final review adjudication

The final automated comment alleged that closing a missing session bypasses resume recovery. A
focused local experiment with actual Codex 0.144.6 tested a never-existing valid UUID: close succeeds,
and resume produces exactly the same error before and after close on both routes. Independent
Standards and Spec reviewers agree this is not a reproduced regression and does not block this PR.
The experiment patch and logs are retained locally; no speculative error swallowing was added.
It did expose a separate pre-existing policy gap for `Internal error` with
`data.details: no rollout found for thread id ...`, which occurs without close and is outside this fix.
The native compatibility conclusion is bounded to the tested binary/adapter.

Current-head complete portable tests and macOS lanes passed. Windows core hit a 60-second timeout in
unchanged `normalize-legacy-paths.test.ts` (both tests pass locally in about 1.5 seconds), and Windows
E2E shard 3 timed out while startup remained `migrating`; its retry passed. Only failed Windows jobs
and their dependent gate are being rerun. No migration code or timeout threshold was changed.

### Final CI result

All required checks passed for `2255377c4b6611212dc3195478b13712cf3d6254`.
[PR Gate, attempt 2](https://github.com/aipoch/open-science/actions/runs/34423545583)
succeeded after rerunning only the failed Windows jobs and dependent gate. Complete portable tests,
Windows core/E2E, Linux isolation, macOS journeys/presentation, Node/Web types, lint and CodeQL passed.
The two runner timing failures did not require source or timeout changes.


ewen-poch:
Review findings at head `2255377c4b6611212dc3195478b13712cf3d6254`:

- **Negotiated close support:** Addressed in `4585f5f0c`. The scoped Codex refresh checks the connection's advertised `session/close` capability and reports an explicit unsupported-runtime error. It does not silently discard persisted history.
- **Preserve the session when a Specialist switch cannot start:** Addressed in `04a2c62b8`. Capability preflight runs before the Specialist binding is mutated, so an unsupported refresh preserves the original identity and live attachment. The regression exercises the public Runtime/ACP behavior, including sending another prompt after the rejected switch; it failed before the fix.
- **Negotiated resume support:** Addressed in `2255377c4`. The same preflight now requires both close and resume. The missing-resume regression failed with an ACP `Internal error` before the fix and passes with the explicit capability error and original usable session afterward.
- **Detached close / lost-session recovery:** The reported close-induced failure did not reproduce with actual Codex 0.144.6 and the installed ACP adapter on either Responses or Chat. Committed native integration coverage terminates the adapter, starts a new process/connection, closes the persisted but locally unloaded session, and successfully resumes it; repeated close also succeeds. A separate exploratory probe closed a never-existing valid session UUID successfully and observed the same resume error before and after close. Therefore no speculative close-error suppression or fresh-context fallback was added. This conclusion is bounded to the tested runtime, rather than a claim about all future adapters.

The exploratory probe did identify a separate pre-existing error-classification gap: `Internal error` with `data.details` containing `no rollout found for thread id ...` is not classified as adoptable. It occurs without the new close step, and the classifier is unchanged by this PR. Expanding that recovery policy is outside this PR's approved scope and is left for a separately scoped fix.

The final local impact set passed 780 targeted tests, including the four native integration cases. PR Gate and static/platform checks passed at the current head. No architecture, persisted-data format, state enum, or user interaction changes were added in response to the remaining speculative finding.

PR #2420, opened 2026-09-10
fix(privacy)!: require explicit media loads and secure model input

ewen-poch:
## Problem

Viewing model-generated Markdown can start remote media requests without activation, small model images can include private metadata, and custom remote HTTP endpoints can receive credentials and content in plaintext.

## Proposed change

- Defer remote image/audio/video sources and posters until the user activates a control showing the destination hosts. Approval is scoped to the rendered element and its complete URL set. Block Mermaid image nodes before Mermaid can issue requests; preserve ordinary diagrams. After activation, reuse Streamdown’s native image renderer to preserve downloads and the existing Electron save bridge.
- Normalize images at prompt assembly for attachments, references, current messages, and history. Strip ancillary metadata from model-input derivatives, preserve original files, and retain existing image budgets and animation handling.
- Require HTTPS for remote model URLs at validation, request sending, and runtime configuration projection. Preserve localhost and IPv4/IPv6 loopback HTTP.
- Document data recipients, cancellation/deletion limits, and the remaining third-party traffic verification work. Exclude frozen `docs/internal/` evidence from lint, matching the existing Vitest boundary.

## Scope and non-goals

**Historical compatibility:** existing remote HTTP configurations remain stored and editable but cannot run until changed to HTTPS. Previously saved messages use the new media activation and image-derivative rules when viewed or replayed. Because vision-cache keys include the emitted image bytes, normalization can cause a one-time cache miss and recomputation for old images. No attachment, cache, or database migration is performed.

**State:** add only per-element in-memory media approval and transport/render error branches; no new persisted state enum.

**Persistence:** no schema, database fields, persisted-format changes, or original-file rewrites. No new durable external resource.

This does not provide an offline mode, a final-origin media allowlist after activation, third-party telemetry guarantees, or provider-side deletion. Subscription lifecycle and active-branch selection remain owned by their existing runtime/persistence boundaries.

The image-download review regression was reproduced against the first PR commit (missing download button), then covered by the real save bridge and a Chromium download test.

Original failures were reproduced before the production changes: premature browser requests, EXIF canaries in outgoing image blocks, and credential-bearing HTTP requests reaching the injected transport. Regression tests now assert the safe behavior alongside legitimate controls.

Checks ran in the dedicated worktree after the final material edits:

| --- | --- | --- |
| Renderer, prompt/image preparation, provider resolution, uploads, and translations | Deduplicated `createModuleTestPlan`/`executeModuleTestPlan` plans plus new/direct regression files | 128 files: 127 passed, 1 skipped; 5,297 tests passed, 4 skipped |
| Media activation, reopened history, Mermaid rejection and ordinary diagrams | `playwright test ... e2e/browser/media-privacy.spec.ts` | 5 passed |
| Main, sandbox and renderer types | `npm run typecheck` | Passed |
| Maintained source and configuration | `npm run lint` | Passed |

Module plans: `session_renderer`, `workspace_runtime`, `image_input_compatibility`, `settings_backend_resolution`, `settings_provider_accounts`, `settings_service_facade`, `upload_repository`, `i18n_catalog`, and `i18n_renderer_adapter`. Their curated checks are reproducible with `npm run test:module -- <module-id>`; new regressions are checked in beside their public boundaries. Runtime consumer coverage includes Claude Code, OpenCode, Codex Responses, and Codex Bridge. Existing branch replay and lifecycle tests remain included through the runtime modules.

Real Chromium and isolated Electron renderer fixtures exercise the browser network boundary; no real provider credentials were sent. The exact-commit impact explain command selects the full CI plan because the new browser fixtures and lint configuration are not owned by its manifest. Local validation uses the explicit module set above at the maintainer’s request; Windows/Linux execution and complete validation are delegated to PR CI. Third-party runtime login/idle/telemetry and remote-Web traffic measurements remain unverified.

## Review focus

Check metadata minimization without modifying scientific originals, preserved prompt budgets/history selection, transport enforcement for saved configurations and subprocess projections, and media approval invalidation when URLs change. The deliberate compatibility change for remote HTTP also applies to private-network endpoints.

The latest automated review's Mermaid metadata concern was checked against the locked and installed Mermaid **11.17.2**. Its `FlowVertex` type declares `img?: string`, and `FlowDb.addVertex` assigns `vertex.img = doc?.img`. Running `mermaidAPI.getDiagramFromText` with both `img:` and `"img":` produced `{ img: "https://privacy-canary.invalid/image.png", props: {} }`; `vertex.props?.img` was undefined. The checked-in real-browser tests also reject those image nodes before any request with media both enabled and disabled. This finding was not reproduced, so the production guard was not changed to the incorrect property. The earlier download finding is covered by the subsequent fix and regression tests.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -61,6 +61,63 @@ practices govern how it processes that content.
 Treat project content, model output, downloaded files, Skills, Specialist packages,
 custom Connectors, MCP servers, and remote compute hosts as untrusted until reviewed.
 
+### Data recipients
+
+Local storage and local processes do not imply offline operation. The selected provider endpoint
+receives the message, relevant history, and attachments prepared for the task. Scenario models may
+use another provider: a fixed title model receives the first message text; a vision model receives
+an image and returns evidence that can then be sent to the main model. Review both the provider name
+and endpoint in Settings when choosing these models.
+
+Read-only Connectors still send query terms, identifiers, filters, or other tool arguments to their
+sources. Literature lookup may query several services with DOI/PMID identifiers; a metadata lookup
+is not an automatic upload of the local PDF. Remote jobs transfer the scripts and selected inputs
+needed for the job. In the Web UI, uploaded files first go to the machine running Open Science,
+which can be different from the machine running the browser.
+
+Custom remote model endpoints require HTTPS. HTTP is supported for `localhost`, IPv4 loopback, and
+IPv6 loopback; private-network addresses alone do not establish an encrypted connection. Existing
+remote HTTP configurations remain editable but cannot supply a running model until corrected.
+
+Messages and Markdown previews defer external media until the user activates it. The control shows
+the destination hosts and exposes the full URLs on hover. This permission is local to the displayed
+media and its URL set, not a permanent domain grant. After activation, media uses the browser's
+network stack and normal redirect behavior; it is not routed through the Notebook domain policy.
+Mermaid image nodes are blocked before diagram rendering; use a separate Markdown image to load
+them explicitly. Ordinary diagrams remain available. Managed artifact previews retain their existing
+authorization. Import previews that disable media continue to suppress it entirely.
+
+Image understanding uses metadata-minimized derivatives. Original attachments remain intact for
+scientific provenance and explicit file/metadata analysis. Small still images are losslessly encoded
+from their decoded, oriented pixels when they fit the payload budget; larger derivatives retain the
+existing resizing/compression limits. Removing metadata does not remove identifying information
+visible in the picture or prevent an explicitly authorized file tool from reading the original.
+
+### Cancellation, deletion, and retention
+
+Stopping a request can stop subsequent work or stop waiting for a response. It cannot retract bytes
+already received by a service. Pausing a batch may wait for the current item to settle. Deleting a
+local session, attachment, temporary profile, or vision cache removes that local resource; it does
+not instruct the model provider to erase its copies. Remote job file cleanup and provider-side
+deletion are separate operations and must be confirmed by the owning service.
+
+Retention, training use, and deletion guarantees depend on the selected service and account policy.
+Open Science does not infer a zero-retention guarantee from HTTPS, local models, or cancellation.
+
+### Runtime network verification
+
+The app proxy settings affect Electron sessions and the environment of subsequently started
+processes. They do not change the OS proxy, reconfigure already running agent processes, or control
+a remote browser's network stack. Third-party runtime telemetry cannot be established from this
+repository's environment flags alone.
+
+Before making a runtime-specific privacy claim, test the exact runtime version in an isolated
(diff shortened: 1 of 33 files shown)


ewen-poch:
Regarding [preserving the Streamdown image wrapper and download action](https://github.com/aipoch/open-science/pull/2420#issuecomment-5611235732): fixed in 361df05b91c63b7717ed2ac47844d0c78536d0c3.

The regression was reproduced before the fix through the existing save-bridge boundary: after approving an image, the test failed because the `Download image` button was absent. Approved images now use Streamdown's public renderer, retaining its native wrapper, download action, and blob-download attribution behind the explicit-load gate.

The save-bridge regression passes after the fix, and a real Chromium test verifies that the approved image downloads successfully as `image.png`. All five media privacy browser tests pass. The final local module test set passed with 5,297 tests passing and four skipped; typechecks and lint also passed after the last material edit.

ewen-poch:
Regarding [the Mermaid image metadata property](https://github.com/aipoch/open-science/pull/2420#issuecomment-5611415956): I could not reproduce the reported bypass with the locked Mermaid 11.17.2 dependency, and the installed parser contradicts the suggested property path.

- `dist/diagrams/flowchart/types.d.ts` declares `img?: string` directly on `FlowVertex`.
- The installed flowchart parser assigns `vertex.img = doc?.img` in `dist/chunks/mermaid.core/chunk-SHT3W25Y.mjs`.
- Calling the actual `mermaidAPI.getDiagramFromText` with both `img: "https://privacy-canary.invalid/image.png"` and `"img": "https://privacy-canary.invalid/image.png"` produces a vertex with that URL in `vertex.img` and an empty `vertex.props` object.
- The checked-in Chromium regressions in `e2e/browser/media-privacy.spec.ts` exercise Mermaid image nodes with media both enabled and disabled and assert that no image request is issued. Both pass; the ordinary shape-metadata diagram also renders successfully.

The existing `vertex.img` guard therefore matches this dependency's actual runtime shape. Replacing it with `vertex.props?.img` would remove the effective check for the tested inputs, so no code change was made for this suggestion. The review workflow has completed, but its published verdict remains `needs changes`; this comment records the runtime evidence for declining the finding rather than treating workflow success as review approval.

ewen-poch:
The [additional AI review execution](https://github.com/aipoch/open-science/actions/runs/34430181781) completed and reported only the same Mermaid `vertex.props.img` suggestion. It did not identify another issue. The runtime evidence in the [existing response](https://github.com/aipoch/open-science/pull/2420#issuecomment-5611646502) still applies to the unchanged head `361df05b91c63b7717ed2ac47844d0c78536d0c3`: Mermaid 11.17.2 stores the image URL directly in `vertex.img`, and the real-browser request regressions pass.

No code change is warranted for that suggestion. GitHub Actions also emitted `Skip output 'review_body' since it may contain secret`, so the workflow skipped publishing its review comment. This is a publication limitation, not a mergeable verdict. The review workflow was not modified, and the previously published finding and its evidence-based response remain available.

PR #2424, opened 2026-09-10
fix(web): guard Electron-only handoff lifecycle APIs in web mode

### LINKED ISSUE #2423: fix(web): white screen crash in v0.27.0 — IpcHandoffLifecycleClient calls Electron-only APIs without web guard
## Description

v0.27.0 introduces a white screen crash when accessing Open Science via web browser (Remote Access). The desktop Electron app works correctly. This is a regression — v0.26.0 was not affected.

## Version

- **Open Science:** v0.27.0 (AppImage on Linux)
- **Regression introduced in:** v0.27.0 (v0.26.0 works correctly)
- **Access method:** Web browser via Remote Access (localhost web UI)

## Error Messages

```
Uncaught TypeError: t.onHandoffLifecycleEvent is not a function
    at h_t.ensureChangedListener (mermaid-GHXKKRXX-0oBxQ010.js:643:64464)
    at h_t.subscribe (mermaid-GHXKKRXX-0oBxQ010.js:643:63772)
```

```
Uncaught (in promise) TypeError: n.getHandoffEvents is not a function
    at h_t.load (mermaid-GHXKKRXX-0oBxQ010.js:643:63980)
```

## Root Cause

`IpcHandoffLifecycleClient` (introduced in v0.27.0 in `src/renderer/src/pages/workspace/handoff-lifecycle-source.ts`) calls `api.onHandoffLifecycleEvent()` and `api.getHandoffEvents()` without checking whether these methods exist on the web platform.

In the Electron desktop app, the preload script (`src/preload/index.ts` lines 333-336) hardcodes these methods on `api.specialist`, so they are always available. In the web version, however:

1. The contract catalog (`src/shared/renderer-contract-catalog.ts` lines 337-338) marks `onHandoffLifecycleEvent` and `getHandoffEvents` as `ELECTRON` profile
2. The web API installer (`src/renderer/web/api-installer.ts`) only installs methods with `surfaceInstallation.localWeb` of `"web-rpc"`, `"web-event"`, or `"browser-native"` — ELECTRON-only methods get `"unavailable"` and are skipped
3. The generated web API map (`src/shared/web-api-map.generated.ts`) does not include `specialist:handoff-lifecycle-changed` or `specialist:get-handoff-events` in `WEB_EVENT_CHANNELS` or `WEB_INVOKE_CHANNELS`

**Contrast with `WorkspacePage.tsx`** which correctly guards these calls:

```typescript
// WorkspacePage.tsx line 2303 — correct guard
if (!specialistApi?.onHandoffLifecycleEvent) return
```

But `handoff-lifecycle-source.ts` does not:

```typescript
// handoff-lifecycle-source.ts line 117-128 — missing guard
private ensureChangedListener(): void {
  if (this.stopChangedListener) return
  const api = this.getApi()
  if (!api) return  // checks api exists, but NOT whether the method exists
  this.stopChangedListener = api.onHandoffLifecycleEvent(...)  // crashes on web
}
```

Same issue in `load()` (line 99-105):

```typescript
async load(sessionId: string): Promise<void> {
  const api = this.getApi()
  if (!api) return
  const retained = await api.getHandoffEvents(sessionId)  // crashes on web
}
```

## Suggested Fix

Add method existence checks in `handoff-lifecycle-source.ts`:

```typescript
// ensureChangedListener()
const api = this.getApi()
if (!api || typeof api.onHandoffLifecycleEvent !== "function") return

// load()
const api = this.getApi()
if (!api || typeof api.getHandoffEvents !== "function") return

// retry()
const api = this.getApi()
if (!api || typeof api.retryHandoff !== "function")
  throw new Error("Handoff lifecycle API is unavailable")
```

Alternatively, consider adding these channels to the web API map if handoff lifecycle support is intended for web mode.

## Environment

- OS: Ubuntu 24.04
- Browser: Edge, Chrome
- Deployment: AppImage installed on remote machine, accessed via browser

justemu:
## Summary

Adds `typeof` function guards to `IpcHandoffLifecycleClient` methods (`load`, `retry`, `ensureChangedListener`) so they gracefully no-op when the underlying Electron-only APIs (`onHandoffLifecycleEvent`, `getHandoffEvents`, `retryHandoff`) are not available in the web Remote Access UI.

## Problem

v0.27.0 introduced `IpcHandoffLifecycleClient` which calls `api.onHandoffLifecycleEvent()`, `api.getHandoffEvents()`, and `api.retryHandoff()` without checking whether these methods exist. In the Electron desktop app these methods are registered via the preload script, but in the web version the contract catalog marks them as `ELECTRON` profile and the web API installer skips them.

This causes a white screen crash (`TypeError: t.onHandoffLifecycleEvent is not a function`) when accessing Open Science via web browser.

## Fix

```typescript
// Before
if (!api) return
this.stopChangedListener = api.onHandoffLifecycleEvent(...)

// After
if (!api || typeof api.onHandoffLifecycleEvent !== "function") return
this.stopChangedListener = api.onHandoffLifecycleEvent(...)
```

This matches the existing guard pattern in `WorkspacePage.tsx` (lines 2303, 2309).

## Testing

- Verified: web Remote Access no longer crashes on load
- Verified: Electron desktop app continues to work (methods are present, typeof check passes)
- Verified: `retry()` now throws a clear error message instead of `TypeError` when called on web

Fixes #2423

### DIFF
--- a/src/renderer/src/pages/workspace/handoff-lifecycle-source.test.ts
+++ b/src/renderer/src/pages/workspace/handoff-lifecycle-source.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it, vi } from 'vitest'
 
 import type { HandoffLifecycleEvent } from '../../../../shared/handoff-lifecycle'
 import type { CompletionHandoffLifecycleEvent } from '../../../../shared/specialist'
+import { installWebRendererContracts } from '../../../web/api-installer'
 
 import { IpcHandoffLifecycleClient } from './handoff-lifecycle-source'
 
@@ -25,6 +26,42 @@ const event = (
 })
 
 describe('IPC handoff lifecycle client', () => {
+  it('handles the web-installed specialist API without Electron-only lifecycle methods', async () => {
+    const api: Record<string, unknown> = {}
+    const invoke = vi.fn()
+    const subscribe = vi.fn()
+    installWebRendererContracts(api, {
+      availableRpcChannels: new Set(['specialist:list']),
+      restrictedRpcChannels: new Set(),
+      invoke,
+      subscribe,
+      nativeAdapters: {}
+    })
+
+    expect(api.specialist).toMatchObject({ list: expect.any(Function) })
+    for (const method of ['getHandoffEvents', 'onHandoffLifecycleEvent', 'retryHandoff']) {
+      expect(api.specialist).not.toHaveProperty(method)
+    }
+    // The shared API type describes Electron; the web installer intentionally omits these methods.
+    const client = new IpcHandoffLifecycleClient(
+      () => api.specialist as typeof window.api.specialist
+    )
+    const listener = vi.fn()
+    const snapshot = client.getEvents('session-1')
+    expect(snapshot).toEqual([])
+    const unsubscribe = client.subscribe(listener)
+
+    await expect(client.load('session-1')).resolves.toBeUndefined()
+    expect(client.getEvents('session-1')).toBe(snapshot)
+    expect(unsubscribe).not.toThrow()
+    await expect(
+      client.retry({ sessionId: 'session-1', originatingTurnId: 'turn-1' })
+    ).rejects.toThrow('Handoff lifecycle API is unavailable')
+    expect(listener).not.toHaveBeenCalled()
+    expect(invoke).not.toHaveBeenCalled()
+    expect(subscribe).not.toHaveBeenCalled()
+  })
+
   it('keeps the empty event snapshot stable before a session has handoff events', () => {
     const api = {
       getHandoffEvents: vi.fn(async () => []),
--- a/src/renderer/src/pages/workspace/handoff-lifecycle-source.ts
+++ b/src/renderer/src/pages/workspace/handoff-lifecycle-source.ts
@@ -98,7 +98,7 @@ class IpcHandoffLifecycleClient implements HandoffLifecycleEventSource {
 
   async load(sessionId: string): Promise<void> {
     const api = this.getApi()
-    if (!api) return
+    if (!api || typeof api.getHandoffEvents !== 'function') return
 
     const retained = await api.getHandoffEvents(sessionId)
     this.merge(sessionId, retained.map(toHandoffEvent))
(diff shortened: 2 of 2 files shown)


ewen-poch:
Thanks for the focused fix! I checked the problem against the renderer contract catalog, the web API installer, and the conversation subscription path. The issue is real: `specialist` exists on web, while these three Electron-only methods are omitted. I reproduced the subscription TypeError and the rejected `load()` on the base revision with an existing but incomplete API object; the PR handles both correctly and gives an explicit error for unavailable retry.

The guards look appropriate for restoring web startup while preserving the existing desktop-only handoff boundary. This does not add web handoff status/retry support, and I do not think broadening the web RPC contract is necessary for this crash fix.

Before merging, could you please add a regression case to `handoff-lifecycle-source.test.ts` where the API object exists but the lifecycle methods are missing? It should check that subscribe/unsubscribe and load are safe, the empty snapshot stays stable, and retry rejects with the explicit unavailable error. The current three tests all provide the complete API, so they pass even without this fix. An integration case using the web installer would also help protect the platform boundary.

I ran the existing lifecycle client, web installer, and message scroller render/interaction suites at `3807ea1f`: 4 files / 166 tests passed. A separate in-memory base-versus-head check also confirmed the missing-method behavior without modifying repository files.

The current PR Gate is still red: Windows E2E shards 1/3 and 3/3 time out in `waitForRendererReady`, observing database startup phase `migrating` rather than `ready` (the theme-persistence and Ctrl+F scenarios). I have not established that these failures are caused by this patch; please investigate or rerun the affected lanes before merging. Static checks, all three portable test shards, and module tests passed.

No historical-data migration, new state/enum value, or persistence-format/write-path change is introduced by this patch.

ewen-poch:
I've added the regression coverage directly to this PR in 53cb0ec5. It uses the real web API installer to produce a specialist object without the Electron-only lifecycle methods, then verifies safe subscribe/unsubscribe and load, stable empty snapshots, explicit retry rejection, and no transport calls.

Validation after the final test edit:
- Lifecycle owner, web installer contract, and message scroller consumers: `vitest run src/renderer/src/pages/workspace/handoff-lifecycle-source.test.ts src/renderer/web/api-installer.test.ts src/renderer/src/pages/workspace/WorkspaceMessageScroller.render.test.tsx src/renderer/src/pages/workspace/WorkspaceMessageScroller.interaction.test.tsx` — 4 files, 167 tests passed.
- `tsc --noEmit -p tsconfig.web.json --composite false` — passed.
- `npm run lint` and the changed-file Prettier check — passed.
- Temporarily removing each of the three guards independently makes the new regression test fail; all guards were restored and the four suites passed again before committing.

This follow-up changes only the test. No new state, persistence changes, or historical-data compatibility requirements. The earlier Windows startup failures remain unproven to be related; the new-head CI results are still needed before merge.

PR #2438, opened 2026-09-10
feat(reproducibility): verify artifact versions with captured execution and lineage

ewen-poch:
This adds replayable verification for captured Artifact Versions. Users can inspect the input files, upstream Notebook runs, and environment required for a check, restore a sealed recipe in isolation, and compare the reproduced files with the original version.

- Capture version-bound execution evidence, file generations, environment locks, lineage graphs, and replay recipes. Support checks from captured inputs or available intermediate-file checkpoints, plus batches of Artifact Versions in a session.
- Show comparison outcomes and per-file details instead of treating execution completion as proof of reproduction. Include exact byte checks, bounded image/TIFF and table comparisons, optional scientific comparison rules, and exportable verification records with output retention, size, and deletion controls.
- Strengthen Python and R dependency analysis for cross-cell variables and package setup, callbacks, pipes, scientific plotting, paths, and intermediate files. Keep kernel epochs separate and retain conservative results for unsupported operations or missing evidence.
- Version derived analysis separately from immutable captured evidence, and fix startup cleanup when a timed-out window has already been destroyed.

Validation covers a broad range of real-world Python and R scientific workflows, supplemented by synthetic regression cases. Scenarios include volcano plots, heatmaps, Venn and chord diagrams, tabular and single-cell data workflows, cross-cell and cross-turn dependencies, intermediate-file checkpoints, and failure recovery. Coverage spans dependency and lineage analysis, recipe capture, replay execution, and output comparison, with automated regression tests and live end-to-end checks. The Web client currently requires the desktop interface for Check. Large RDS/H5AD files are not loaded for content comparison.

Validation after rebasing onto `32b4ef255`:

- `npm run typecheck`: passed (main, sandbox package, and renderer).
- Focused `npm test` for Artifact, Notebook, shared reproducibility contracts, startup, and i18n: 250 files passed, 7,675 tests passed, 141 skipped.
- Reproducibility start-point UI and preview target regression: 12 tests passed.
- Remaining changed-file tests: 1,376 tests passed across 32 files; ACP runtime: 604 tests passed.
- `git diff --check`: passed.
- The full local suite was attempted, then stopped after broad local process/network and installed-dependency failures. It is not reported as passing.
- CI Integrity requires an explicit maintainer ruleset bypass for the existing `scripts/ci/change-impact.json` expansion. The protected check itself is unchanged.

The three newly added Markdown documents under `docs/` have been removed from this PR as requested.

Additional regression validation:

- Preview divider names remain distinct from workspace divider names in all three layout modes; translations are present in all eight locales. The new assertion fails against the previous shared name.
- UI and i18n guard checks: 819 tests passed. Type checking, targeted ESLint, formatting, and Electron build passed.
- Real Electron: all three comparison-preview scenarios and both original workspace divider/modal-isolation scenarios passed without retries.
- The German 414px/125% scenario now reserves scrollbar space and retains the strict overflow assertion. A bounded, wrapping submit button fixes the reproducible 3px overflow without changing form behavior.


ewen-poch:
Post-rebase validation passed for the main and renderer type checks, 7,675 focused Artifact/Notebook/protocol/startup/i18n tests, and 11 reproducibility start-point UI tests. The 141 conditional skips and the incomplete full-suite attempt are documented in the PR description.

CI Integrity is blocked by the existing expansion of `scripts/ci/change-impact.json`: the repository treats this as a protected gate control-plane change and requires an explicit maintainer ruleset bypass. The added paths extend runtime-test selection for reproduction execution, environment restoration, and Python analysis; the guard has not been changed or bypassed.

ewen-poch:
Fixed the preview context-menu E2E target: `react-resizable-panels` overwrites the `data-testid` passed to its Group, so the existing `preview-file-content-region` locator disappeared. The marker now lives on a native wrapper around the same panel group. This preserves the existing preview interaction and adds no application state or persistence.

The regression test uses the real resizable component: it fails against the previous implementation with the missing locator and passes with the fix. All 12 tests in that UI suite pass. Additional checks passed: 1,376 tests across 32 changed-file suites, plus all 604 ACP runtime tests. The history remains consolidated into four commits.

On the updated head `77fd7ede2`, the macOS regression E2E job now passes, including the real preview context-menu test. The delegation E2E job also passes; its earlier failure was a first-attempt timeout followed by a passing retry, and no delegation behavior was changed. The protected CI impact-map gate still requires maintainer review.

ewen-poch:
Registered `e2e/reproducibility-preview.spec.ts` in the existing `test:e2e:regressions` command. The CI execution-ownership guard correctly found that the new spec was not reachable from any scheduled test command. No workflow or integrity rule was weakened. All four ownership checks now pass locally. The actual standalone Electron preview E2E also passed all three scenarios: 414px English, 1280px English, and 414px German at 125% zoom, exercising the real comparison controls above the Preview dialog.

The previous portable CI summary reported 32,423 passing tests and this one ownership failure. The updated command addresses that failure; the next CI result must confirm it. The AI review for `77fd7ede2` reached the existing 30-minute job timeout before publishing a verdict, so it does not supersede the earlier published review. The review workflow and timeout have not been changed.

ewen-poch:
The latest macOS run rules out font/layout settling as the cause of the narrow German form failure: the overflow remains 3px after the full wait. Reserving a 12px scrollbar gutter reproduces the same failure locally. The form is 186.875px wide, while the non-wrapping “Apply comparison rules” button is 189.7625px wide.

A scoped fix is prepared locally: constrain that button to the available width and allow its label and height to wrap. The existing overflow threshold is unchanged. All three Electron preview scenarios, including the reserved-scrollbar case, and all 12 related unit tests pass. This fix has not yet been pushed.

The macOS journeys and Windows shard 3 failures share a separate cause: the workspace divider and the preview provenance divider both use the accessible name “Resize right panel,” making the existing locator ambiguous. Changing the preview divider name and its translations is awaiting owner confirmation under the requested interaction-change boundary.

All three portable-test shards, module tests/coverage, static checks, and Windows core passed on the current PR head. CI Integrity still requires maintainer handling for the protected change-impact configuration; no guard has been weakened.

The AI review for `d824693e6` has now finished with a 30-minute timeout and did not publish a verdict. Earlier completed reviews reported mergeable, but those verdicts apply to their respective earlier heads. Further changes are paused pending the divider-name confirmation, preserving the final permitted review attempt.

ewen-poch:
The preview provenance divider now has the distinct accessible name “Resize provenance panel,” with all eight translations. This fixes the duplicate-name locator failure shared by the macOS journeys and Windows shard 3 jobs without changing pointer or keyboard behavior. Regression assertions cover content, split, and provenance modes and fail against the former shared name.

The narrow German comparison form now constrains the submit button to its available width and allows wrapping. Reserving scrollbar space reproduces the CI runner’s exact 3px overflow before the fix; the original strict overflow assertion passes afterward. Waiting for fonts alone did not fix this issue.

Validation: 819 UI/i18n checks passed, all three real Electron comparison-preview scenarios passed, and both original workspace divider/modal-isolation scenarios passed without retries. Type checking, targeted ESLint, formatting, and the Electron build also passed.

The three newly added Markdown documents under `docs/` have been removed. These changes are folded into the existing four-commit branch. The protected change-impact configuration still requires maintainer handling through CI Integrity; its guard is unchanged.

ewen-poch:
The empty per-command host list does not replace the Notebook network policy. `NotebookNetworkSandboxOwner.createSandbox()` passes `buildNotebookNetworkPolicy(settings)` into the sandbox, whose locked `packageRegistries` group includes `pypi.org`, `files.pythonhosted.org`, `cran.r-project.org`, `cloud.r-project.org`, and the catalogued mirror hosts. `allowedNetworkHosts` is an additional command grant checked by the fallback network-access handler.

The reported blanket failure for uncached native packages therefore does not follow from an empty supplemental list. The existing default-policy, package-process sandbox, and network-owner suites pass (61 tests), including preservation of the required package-registry group. This is policy/call-chain validation, not a claim that every private registry has been tested live.

A non-catalogued private registry still needs explicit authorization. Automatically trusting arbitrary URLs embedded in a captured lock would change the network boundary, so that is not being added as a response to this finding. Any configured-private-mirror handoff issue should be validated separately against its exact configuration before changing that boundary.

ewen-poch:
Sanitized the reported regression fixtures and their matching assertions: personal paths, session identifiers, uploaded-file hashes, source-language prose, and original dataset/cohort labels now use synthetic English examples. The fixtures still exercise the same dependency, file-access, and replay cases. A regression guard rejects original host paths, upload-style identifiers, accessions, and Chinese prose in these fixtures.

The four commits were rewritten so these originals are also absent from the current branch's commit trees, and the three newly added docs files are removed throughout this branch history. Previously published Git objects are not erased by this rewrite.

Fixed the portable-test failure in `PreviewFileSurface.test.tsx`: its selector still used the old separator name after the accessibility fix renamed it to `Resize provenance panel`.

Validation: all 36 affected test files passed (1,607 tests passed, 34 skipped); formatting, ESLint, and diff checks passed. The previously failing macOS regression/journey and Windows shard 3 E2E jobs passed on the preceding CI run. This push requests one additional CI/AI review for the new changes.

CI Integrity still requires the documented maintainer ruleset bypass for the protected change-impact manifest update; the guard has not been weakened or bypassed.

ewen-poch:
Confirmed the R lock snapshot finding. The lock-capture executor was the remaining probe path using raw `process.env`; the inventory and fingerprint paths already activate the bound R Conda environment.

A local fix now passes `environmentCaptureProcessEnv(target, process.env, this.platform)` to that executor. A new regression test exercises native R lock capture with Windows platform settings and checks the snapshot command's activated PATH. The test failed against the old executor and passes with the fix; the two environment/lock suites pass (138 tests, 16 skipped). This reuses the existing helper and does not change the persistence model or network policy.

This fix is currently local, not included in the reviewed commit. The agreed additional review allowance is exhausted, so no further push/review cycle is being started automatically.

The latest CI passed all three portable-test shards and the module/coverage aggregate, confirming the stale separator selector fix. Windows core hit a separate existing timing assertion (10,637 ms against a 10,000 ms bound). A targeted rerun is currently refused while the remaining workflow job is still running; the timing assertion has not been relaxed.

ewen-poch:
The native R lock snapshot fix is now included in the branch. Lock capture uses the same Conda-activated environment as the existing R inventory and fingerprint probes, including the Windows DLL directories. The regression test covers the snapshot command itself and was confirmed failing before the fix.

Validation: environment tracker/lock suites passed (138 tests, 16 skipped), full typecheck and ESLint passed. The change is folded into the existing notebook commit, preserving four commits and the sanitized fixtures.

ewen-poch:
Confirmed the Compute ordering issue. Priority job IDs should decide which activities survive the 100-activity bound, but should not become execution ordinals ahead of earlier selected jobs.

The fix retains the existing inclusion bound, then sorts the selected activities by `createdAt` with an `id` tie-breaker. Both database queries also use that deterministic ordering. The existing harvested-output regression now inserts same-timestamp jobs in reverse ID order and a later prioritized output owner beyond the normal bound; it verifies that the owner remains included while all emitted activities are chronological. This assertion failed before the fix.

The portable CI failure was separately reproduced by delaying the mocked Shell executor by 1.5 seconds: the FIFO test's default polling deadline expired even though queue state was correct. The test now waits for an explicit first-execution signal before checking running/queued state and cancellation, retaining all behavioral assertions. The delayed case passes; the temporary delay is not included in the patch. Production Shell scheduling and timeout limits are unchanged.

Both fixes are now included in the branch. Validation: all 517 tests in the affected provenance/write-contract/graph/runtime suites passed with real subprocess permissions; full typecheck, ESLint, formatting, and diff checks passed. The branch still contains four commits.

ewen-poch:
The helper-trimming P2 is deferred from this PR at the maintainer's request. It concerns oversized execution snapshots that require repeated helper removal; the follow-up should verify recipe/snapshot consistency across successive trims, including evidence already marked incomplete before trimming.

No P2 patch is included in the current head. The release candidate remains `0cf35f536845f1996b7b77d409ae28758cbb7c03`, whose PR Gate, portable tests, platform tests, E2E suites, and CodeQL checks all passed. CI Integrity still reports the protected change-impact manifest update and requires the documented maintainer approval path.

PR #2445, opened 2026-09-11
perf(notebook): skip installs for satisfied package requests

ewen-poch:
## Problem

Repeated `manage_packages` requests invoke mirror discovery and installer transactions even when the target managed Python or R environment already satisfies every requested package. Presence-only `installed` metadata cannot by itself establish version satisfaction.

## Proposed change

Check fresh interpreter-native metadata under the existing environment mutation lease before mirror resolution, cache retention, journal/dirty markers or imported-lock invalidation. Managed Python probes use the canonical managed environment and isolated mode; R probes use the canonical managed library environment and activated PATH.

Return the existing successful result with unchanged package versions and no installer attempts only when all requirements are proven satisfied. Cover bare Python/R names and strictly equal numeric Python pins. Keep complete original requests for partial/missing/uncertain matches, explicit installers (including `usePip: true`), sources/channels, complex specs, external environments, uninstalls and repair-required targets. Do not create a new restart recommendation or clear an existing repair marker on a no-op. Fresh probes receive cancellation signals. The renderer no longer guesses Conda for a confirmed no-installer result, including results compacted by MCP.

## Scope and non-goals

No new result fields, enum values, persistent formats, database migrations or historical-data rewrites. The fresh path does not write inventory state. Imported environment locks remain intact. No new dependency/version solver, partial-request filtering or implicit upgrade workflow. Ordinary cached inspection is unchanged.

Checks after the final material edit at `80bd94a1d`:

- Fresh package metadata, fallback and explicit-installer intent, Python/R full and compacted result rendering -> `npx vitest run src/main/notebook/package-mutation.test.ts src/main/notebook/package-operations.test.ts src/main/notebook/environment-state-tracker.test.ts src/renderer/src/pages/workspace/WorkspaceActivityGroup.i18n.render.test.tsx src/renderer/src/pages/workspace/workspace-tool-activity-details.test.ts` -> 239 passed, 13 existing opt-in skips. Includes real subprocess cancellation tests for both probe paths.
- Runtime binding and repair consumers -> `npx vitest run src/main/notebook/runtime-service.test.ts -t 'managePackages|repair aliases|runtime bindings'` -> 132 passed, 254 outside selection.
- Main/renderer/sandbox types -> `npm run typecheck` -> passed; `npm run lint` and `git diff --check` -> passed.

Additional baseline contract/translation coverage at `d6cae0b60` (not rerun after the final edit):

- Current metadata, Python/R identity and versions, source/repair/cancellation fallback, no mutation side effects -> `npx vitest run src/main/notebook/package-mutation.test.ts src/main/notebook/package-operations.test.ts src/main/notebook/package-admission.test.ts src/main/notebook/environment-state-tracker.test.ts src/main/notebook/mcp-server.test.ts src/renderer/src/pages/workspace/WorkspaceActivityGroup.i18n.render.test.tsx src/renderer/src/pages/workspace/workspace-tool-activity-details.test.ts src/renderer/src/i18n/resources.test.ts` -> 1,165 passed, 13 existing opt-in cases skipped.
- Runtime binding and repair consumers -> `npx vitest run src/main/notebook/runtime-service.test.ts -t 'managePackages|repair aliases|runtime bindings'` -> 132 passed, 254 outside selection.
- Main/renderer/sandbox types -> `npm run typecheck` -> passed.
- Source standards -> `npm run lint` and `git diff --check` -> passed.
- Real renderer component fixture in Ego Browser -> Python/R version and unchanged rows shown, with no guessed Conda or restart prompt. Screenshot retained locally; this is component presentation evidence, not a live installation E2E.
- Independent Standards review confirmed the impact map; independent Spec review caught host-library contamination, then confirmed the final fix resolved it.

The impact explainer selects full fallback because the tracker test has no registered module owner. Per maintainer instruction, no full local suite was run; PR CI remains the complete/platform authority. Explicit local ownership is tracker/mutation/operations -> runtime-service/MCP -> renderer result consumers. No ACP wire contract changed, so four-provider protocol matrices are outside this diff.

Limitations: Windows/macOS/Linux process-environment cases are simulated locally; live Windows/R probe timing and runtime behavior remain CI/runtime checks. An expanded local runtime-service run had an unrelated Shell exit-code failure in the linked worktree; the isolated corresponding test passed in primary main. No unrelated Shell changes were made. Package-specific final tests pass.

## Review focus

Fail-safe fallback on uncertain metadata; isolation from host Python/R libraries; preservation of repair and imported-lock evidence; both language paths and legacy renderer results.

### DIFF
--- a/src/main/notebook/environment-state-tracker.test.ts
+++ b/src/main/notebook/environment-state-tracker.test.ts
@@ -53,6 +53,94 @@ const readBinding = async (
 }
 
 describe('EnvironmentStateTracker', () => {
+  it.each(['python', 'r'] as const)(
+    'fresh %s inspection bypasses cache and does not publish state',
+    async (language) => {
+      dataRoot = await mkdtemp(join(tmpdir(), 'fresh-package-inspection-'))
+      const name = language === 'python' ? 'scikit-learn' : 'R.utils'
+      const requested = language === 'python' ? 'scikit_learn' : name
+      const inspectInstalled = vi.fn().mockResolvedValue({
+        packages: [
+          {
+            name,
+            ecosystem: language,
+            version: '1.0',
+            versionStatus: 'known',
+            evidenceSources: []
+          }
+        ]
+      })
+      const captureFingerprint = vi.fn()
+      const resolveMicromamba = vi.fn()
+      const tracker = new EnvironmentStateTracker({
+        dataRoot,
+        inspectInstalled,
+        captureFingerprint,
+        resolveMicromamba
+      })
+      const capture = { ...target, language, runtimeSource: 'managed' as const }
+      expect(
+        (await tracker.inspectPackages(capture, [requested], { fresh: true })).packages[0]
+      ).toMatchObject({ status: 'installed', version: '1.0' })
+      inspectInstalled.mockResolvedValue({ packages: [] })
+      expect(
+        (await tracker.inspectPackages(capture, [requested], { fresh: true })).packages[0].status
+      ).toBe('missing')
+      expect(inspectInstalled).toHaveBeenCalledTimes(2)
+      expect(captureFingerprint).not.toHaveBeenCalled()
+      expect(resolveMicromamba).not.toHaveBeenCalled()
+      expect(await readdir(dataRoot)).toEqual([])
+    }
+  )
+
+  it('does not certify conflicting Python distribution versions', async () => {
+    dataRoot = await mkdtemp(join(tmpdir(), 'ambiguous-python-inspection-'))
+    const tracker = new EnvironmentStateTracker({
+      dataRoot,
+      inspectInstalled: async () => ({
+        packages: ['1.0', '2.0'].map((version) => ({
+          name: 'numpy',
+          version,
+          versionStatus: 'known' as const,
+          ecosystem: 'python' as const,
+          evidenceSources: []
+        }))
+      })
+    })
+    expect(
+      (await tracker.inspectPackages(target, ['numpy==2.0'], { fresh: true })).packages[0].status
+    ).toBe('unknown')
+  })
+
+  it('uses exact R names and the first library rank during fresh inspection', async () => {
+    dataRoot = await mkdtemp(join(tmpdir(), 'ranked-r-inspection-'))
+    const tracker = new EnvironmentStateTracker({
+      dataRoot,
+      inspectInstalled: async () => ({
+        packages: [2, 1].map((libraryRank) => ({
+          name: 'R.utils',
+          version: `${libraryRank}.0`,
+          libraryRank,
+          versionStatus: 'known' as const,
+          ecosystem: 'r' as const,
+          evidenceSources: []
+        }))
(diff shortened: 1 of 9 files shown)


ewen-poch:
The R library-scope finding is addressed in 53a6ac1. The no-op path now requires the selected R package to have `libraryScope === 'environment'`; user, system, unknown, and missing scope all fall through to the original installer request. Added regression coverage for each case while retaining the managed-library success case.

Explicit R Conda IDs were already excluded by the plain R-name allowlist (`^[A-Za-z][A-Za-z0-9.]*$`), which rejects the hyphen in both `r-*` and `bioconductor-*`. Added explicit tests for `r-ggplot2` and `bioconductor-deseq2`, confirming neither invokes metadata preflight and both reach the installer. No Conda provenance mechanism is needed for this deliberately excluded path.

Validation after the change: 138 package mutation/operations/inventory tests passed (13 existing opt-in cases skipped), Node/sandbox typechecks passed, and lint passed. These changes remain within the existing files and reuse the existing library-scope field; no architecture, persistence, state-enum, or UI change.

ewen-poch:
Cancellation is addressed in 86f1f20. The existing caller AbortSignal now reaches the fresh inventory reader and Node's execFile call. A cancelled request is checked again before any installer or mutation resource starts. Added subprocess-backed tests for both Python and R probe paths that wait until the child is running, abort it, and assert prompt AbortError completion without inventory writes; the mutation test also verifies signal propagation and no installer invocation.

The exact-version suggestion is intentionally not adopted. The approved scope uses strict numeric exact pins and leaves normalized-equivalent spellings, suffixes, epochs and richer syntax to the installer. This can miss an optimization (e.g. 2.0 versus 2.0.0), but cannot turn an uncertain version match into a false successful install. Expanding the fast path's accepted version semantics offers limited benefit relative to its additional correctness surface, and is not necessary for the agreed Python/R ensure-present path. Added a code comment documenting this deliberate fallback; no behavior was broadened.

Validation after the change: 140 mutation/operations/inventory tests passed (13 existing opt-in skips), 132 selected runtime-service package/binding/repair tests passed, Node/sandbox typechecks passed, lint and diff whitespace checks passed. No new files, persisted fields, enum states, architecture, or UI were introduced.

ewen-poch:
Addressed the explicit-installer and compacted-result findings in 80bd94a:

- Explicit `usePip: true` requests now retain the installer path, even when installed metadata reports a matching version. This preserves installer intent without introducing package-provenance state. A regression test verifies that the preflight is bypassed.
- The activity row recognizes an unchanged, successful result without an installer method using fields retained by MCP compaction. It no longer depends on `attempts`, which compaction removes. Python and R rendering tests cover both full and compacted results; actual installer results retain their installer label.

Validation after the final edit: 239 focused owner/renderer tests passed (13 existing opt-in skips), 132 selected runtime consumer tests passed, full process typechecks and lint passed. The changes remain within the existing PR files and add no persisted fields, enum values, data-model changes, or interaction changes.

PR #2446, opened 2026-09-11
fix(notebook): repair Windows R startup and admission

ewen-poch:
## Problem

Windows R startup failed before user code executed when the command TEMP directory included the product name (`Open Science (DEV)` or `Open Science`). A first R admission probe after a live Python kernel also repeatedly rebuilt directory ACLs and could exceed its deadline before showing administrator authorization. Short-lived directory locks then caused cleanup to report EBUSY. Inherited POSIX `C.UTF-8` variables produced four startup warnings even after a successful run.

## Proposed change

- Put command temporary directories under the existing config root, removing spaces introduced by the product name. Retry Windows directory cleanup after confirmed process termination.
- Skip ACL restoration/regrant when recovery has exactly the same validated live leases and receipts, and for lease operations with no filesystem grants. Preserve recovery snapshots when an empty lease completes after interrupted cleanup. Rebuild and pin both Windows helper architectures.
- Allow the native R admission host 60 seconds for ACL preparation, R execution and cleanup. Keep the full elevated verifier's existing execution deadline.
- Remove only inherited `C.UTF-8` / `C.utf8` locale overrides for Windows R, allowing R to choose the system locale. Keep valid explicit locale settings and other platforms unchanged.

## Scope and non-goals

This fixes product-name spaces; a Windows home or explicitly configured root that itself contains spaces remains an existing limitation. This PR does not claim arbitrary spaced-root support (the automated review's spaced-home example remains reproducible).

No new enum or database/receipt schema, migration, or renderer layout. Existing ACL receipts keep their format and recovery authority. `notebook-command-temp` was introduced after v0.27.0, so no unreleased-directory migration is added, as requested by the maintainer. Existing user data and isolation directories are preserved.

Checks below ran after the corresponding final material edits; only targeted local checks were run. Complete portable/platform validation is delegated to this PR's current-head CI.

| Behavior | Command / evidence | Result |
| --- | --- | --- |
| Kernel, admission owner and environment consumers | `npx vitest run src/main/notebook/process-environment.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/network-sandbox-owner.test.ts` | 94 passed, 80 skipped |
| Real Windows R locale startup | `OPEN_SCIENCE_TEST_RSCRIPT` opt-in `process-environment.test.ts` using the provisioned R and DLL PATH | 11 passed; four warnings reproduced before fix, absent after fix |
| Real AppContainer R execution | `OPEN_SCIENCE_TEST_RSCRIPT` / `OPEN_SCIENCE_TEST_R_DLL_DIR` opt-in `windows-external-r.integration.test.ts` | 1 passed; kernel output has no locale startup warnings |
| Live/no-grant ACL leases and interrupted recovery | `cargo test --offline --locked --manifest-path packages/notebook-network-sandbox/vendor/windows-src/Cargo.toml` | 28 passed; interrupted-cleanup regression confirmed red before snapshot guard |
| Sandbox contracts and pinned resources | `npm run test:module -- notebook_network_sandbox` | 107 passed, 13 skipped |
| Main-process types | `npm run typecheck:node` | passed |
| Lint and formatting | `npm run lint -- --ignore-pattern '.dev-isolate/**'`, focused ESLint after final test edits, Prettier check, `git diff --check` | passed; exclusion is generated local runtime/cache data only |
| Packaged helper consistency | Offline release builds for x86_64-pc-windows-msvc and aarch64-pc-windows-msvc; SHA-256 resource tests | passed; ARM64 not executed locally |

Maintainer verified first authorization followed by successful R plotting (`sin_plot_r.png`) in isolated Electron dev; a subsequent R run completed in about 1.1 seconds. The fresh-R-after-Python admission path was also exercised by a temporary real-owner harness before R was authorized. The retained `windows-first-r-access.integration.test.ts` is opt-in and requires an unapproved R installation with the desktop stopped; its default skip is not counted as a pass. Initial authorization still includes Windows permission checks and user consent, so it is slower than a warm run.

## Review focus

Independent Standards and Spec reviews found and then verified the interrupted-cleanup snapshot fix. Final Standards review confirmed the affected-test mapping and both packaged binary hashes; no remaining blocker in the final changes. Please review conservative recovery fallbacks and the stated spaced-home limitation. No renderer layout changes require screenshots.

### Follow-up to current-head review

The installer WSL2 restart smoke now creates both valid and malformed command receipts under the config root returned by the actual packaged-app launch. It no longer derives the owner directory from APPDATA. `npx vitest run scripts/windows-installer-smoke.test.ts`: 44 passed after the change; focused ESLint, Prettier and diff checks passed. Independent Standards review found no blocker; actual packaged WSL2 restart certification remains a CI/platform check.

No old-directory migration is added: commit `1b152084c` introduced this command-temp mechanism after tag `v0.27.0` (`git merge-base --is-ancestor 1b152084c v0.27.0` returns 1), and the maintainer explicitly excluded compatibility for unreleased layouts.


ewen-poch:
Regarding command-temp cleanup compatibility and the installer smoke path:

- The installer smoke mismatch is fixed in aad8586258fe915df8e1e1746d43f44260fc55ea. Both valid and malformed WSL2 ownership fixtures now use the configuration root returned by the actual packaged-app launch, matching the production owner directory. All 44 `scripts/windows-installer-smoke.test.ts` tests, focused ESLint and Prettier checks passed.
- No legacy-root migration is being added. The command-temp receipt mechanism was introduced in 1b152084c240125fcfa26eb07d2dd7a67be07630, after the v0.27.0 release tag (f0c0e08125c2c2acc091e5c7ae10e3e8c88a5341). The maintainer explicitly requires compatibility for released versions through v0.27.0 and excludes unreleased layouts. Supporting orphan receipts created by development builds would add another cleanup authority/path without serving that supported upgrade contract. The PR description records this scope; old development directories are not scanned or deleted.

Regarding the earlier spaced-home finding:

This fix removes spaces introduced by the product-named command directory; it does not guarantee that an arbitrary Windows home or explicit configuration root is space-free. That pre-existing limitation is explicitly documented in the PR. General spaced-root support is outside this focused startup/admission fix and is not being claimed as resolved.

The current-head review's remaining legacy-layout suggestion is therefore intentionally not adopted. The substantive installer-test mismatch has been addressed, and the Windows R startup, ACL recovery and locale changes retain their targeted regression coverage.

PR #2452, opened 2026-09-11
fix(recovery): harden recovery and interaction quality

ewen-poch:
## Summary
Interrupted operations and stale settings could leave users without a clear recovery path, lose unsent input, or incorrectly finalize retryable work. This change preserves recoverable state and exposes explicit retry and conflict handling across workspace, settings, Compute, Notebook, Literature, and Review flows.

## Architecture and behavior
- Keep recovery effects in the owning main-process services and expose the required operations through application commands, IPC/Web RPC contracts, preload inventories, and SDK types.
- Treat transient remote collection failures as retryable, expose an explicit harvest retry command, and preserve completed job state while collection recovers.
- Scope Notebook rechecks to interrupted startup operations and surface retained operations and recovery blockers.
- Add provider configuration revision checks, Skill replacement previews, Connector diagnostics, defensive Review JSON parsing, and update manifest validation.

- No new database tables, columns, or migration files. The Prisma schema change only clarifies the existing harvestError/harvestedAt semantics.
- Add optional provider configRevision in stored settings; reject stale configuration edits instead of overwriting newer settings.
- Preserve Literature failure details in the existing batch journal and reset only the rows selected for retry.
- Keep Review update timestamps monotonic and handle malformed persisted JSON defensively.
- Store Web composer drafts in scoped sessionStorage for same-tab reload recovery. This stores draft text in browser session storage; it is not an application-restart backup.
- Recover completed uploads through signed, caller-bound receipts with file identity checks. Receipts expire after 24 hours and become invalid when the host restarts.
- Keep transient harvest errors retryable by leaving harvestedAt unset until collection is finalized.

## User interaction
- Preserve composer drafts, unsaved form edits, and preview edit contents across supported failure/recovery paths.
- Provide actionable notification, Web event, storage, runtime, job, and Literature error/retry states.
- Show configuration conflicts and Skill replacement differences before users retry or replace content.
- Improve keyboard/focus behavior, accessible status feedback, long-form/zoom handling, and translations across all eight locales.
- Add browser recovery fixtures/specs and an Electron UX regression spec.

## Validation
- Passed after squash: 279 tests across eight focused recovery/persistence suites.
- Passed after squash: 893 tests across the i18n guard and preload inventory suites.
- Passed: git diff --check.
- Full npm test attempt after squash reported 1,646 files passed, 83 failed, 35 skipped; 32,849 tests passed, 804 failed, 535 skipped; 7 unhandled errors (958.75s).
- Confirmed environmental blockers include EPERM when binding loopback TCP/Unix sockets and missing reflect-metadata / @peculiar/x509 dependencies in this worktree. These packages are already declared and locked; installation attempts did not complete.
- Type checking is also blocked by those missing modules. Remaining full-suite failures have not all been individually triaged, so this PR does not claim a green full regression run.
- Browser/Electron coverage is included, but no new browser/Electron execution is claimed for this PR preparation.

## Before marking ready
- Restore dependencies from the committed lockfile in a working network environment.
- Rerun typecheck and the full test suite with the local socket/process permissions required by integration tests; investigate any remaining failures.
- Run the relevant browser/Electron recovery regressions.


ewen-poch:
Accepted the notification recovery finding. `retrying` is now reset in the promise cleanup before the boundary is reset, so a refresh that completes without repairing the notification surface leaves Retry enabled for another attempt. Added a regression test covering a failed refresh followed by a second retry.

The PR title was also changed to `fix(recovery): harden recovery and interaction quality` to satisfy the repository Conventional Commit title rule.

ewen-poch:
The PR policy failure was caused by a comment-only change in `prisma/schema.prisma`; this repository requires every schema file change to include a migration. The comment carried no runtime or persistence behavior, so I removed it instead of adding a migration for documentation-only text. No data model, relationship, or persistence behavior changed.

ewen-poch:
Follow-up CI fixes are now pushed in `61d7224c`:

- corrected the formatter findings in the touched renderer files;
- removed the test double type mismatch in the notification recovery regression;
- extended the Notebook runtime serialization readiness wait to avoid a startup race under CI scheduling.

These changes are test/static-check stability fixes only. They do not change architecture, data models, persistence, or user-facing interaction behavior.

ewen-poch:
Accepted the P1 draft-loss finding. Authorization expiry now preserves the current composer envelope in a separate, temporary sessionStorage recovery key before pausing the Web session. A subsequent bootstrap consumes it only when the returned draft scope exactly matches the original host/principal scope; mismatched scopes discard it. The envelope is removed after consumption, and the existing explicit revoke path clears both keys.

Added coverage for same-scope recovery and cross-scope rejection. This changes only the existing Web session draft-recovery lifecycle; it adds no database fields, migrations, or durable persistence.

ewen-poch:
Rebased the PR onto the latest `main` at `c1327034`. During the rebase I retained both sides of the locale catalogs, then removed the stale `Running jobs in this session` entries exposed by the i18n orphan guard. The focused i18n and draft-recovery checks now pass: 875 tests across 3 files.

No architecture, data model, relationship, durable persistence, or user interaction design changed in this rebase follow-up.

ewen-poch:
Fixed the settings test's invalid `any` suppression with the existing repository type, normalized native CRLF clipboard line endings in the draft-copy assertion, and increased notification metadata and section-label contrast using the existing text token.

The accessibility test now waits for opening animations to finish. This reproduces the contrast failure reliably instead of intermittently missing content while the panel is moving. No accessibility rules were disabled.

Validation: targeted ESLint and formatting checks passed; 333 unit tests passed; 10 of 11 browser tests passed locally, including draft copy. The remaining browser failure is the “Needs approval” badge from `notification-event-visual.ts`. That shared palette is outside the current PR file scope and also affects live notifications, so it remains unchanged pending the requested scope approval. CI is not yet fully resolved. No state, persistence, or interaction-flow changes were made.

ewen-poch:
The remaining accessibility failure was in the shared notification badge palette. I updated success, danger, waiting, and neutral badge labels to use the existing `text-text-100` contrast token while preserving their semantic backgrounds and borders.

The affected browser suite now passes all 11 tests, including the settled-animation accessibility check and Windows clipboard normalization. No state, persistence, or interaction-flow changes were introduced.

ewen-poch:
The Ubuntu shard failure was a stale assertion expecting the previous `text-text-300` badge token. I updated the existing regression test to assert the approved high-contrast `text-text-100` token. The focused suite passes all 39 tests; the fix is pushed in `4ff7f811` and CI has restarted.

PR #2454, opened 2026-09-11
fix(update): allow force update after install gate

ewen-poch:
## Summary

- add an explicit force-update action after the install gate reports background processes that could not be fully stopped
- preserve the safe default update behavior and pass force intent through the transient update IPC options
- localize the new action label across supported renderer locales

## Validation

- `npx vitest run src/main/update/electron-updater-strategy.test.ts src/renderer/src/i18n/resources.test.ts`

- No persisted data changes or new status enum values.
- Force mode may interrupt running work and can still fail if the operating system retains file locks.

### DIFF
--- a/scripts/ci/change-impact.json
+++ b/scripts/ci/change-impact.json
@@ -125,6 +125,7 @@
         "src/main/**/*powershell*.ts",
         "src/main/**/*path*.ts",
         "src/main/**/*acl*.ts",
+        "src/main/window-shortcuts.ts",
         "src/main/acp/agent-process.ts",
         "src/main/acp/claude-executable.ts",
         "src/main/acp/permission-broker.ts",
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -3815,9 +3815,9 @@ const createApplicationModules = async (
   )
   const updateStrategy = createUpdateStrategy(process.platform, {
     translate,
-    installGate: async () => {
+    installGate: async (options) => {
       releaseSettingsInstallAdmission = settingsService.holdInstallAdmission()
-      return updateInstallGate()
+      return updateInstallGate(options)
     },
     releaseInstallHandoff: abortUpdateHandoff
   })
--- a/src/main/update/electron-updater-strategy.test.ts
+++ b/src/main/update/electron-updater-strategy.test.ts
@@ -900,6 +900,63 @@ describe('ElectronUpdaterStrategy', () => {
     })
   })
 
+  it('forces active work through teardown and durability before installing', async () => {
+    const updater = new FakeUpdater()
+    let active = true
+    const teardown = vi.fn(async () => {
+      active = false
+      return { completed: true, reaped: true }
+    })
+    const durability = vi.fn(async () => true)
+    const strategy = createStrategy({
+      updater,
+      currentVersion: '0.2.0',
+      broadcast: vi.fn(),
+      installGate: createActiveResearchSafeInstallGate(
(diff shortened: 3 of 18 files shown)


ewen-poch:
Fixed the Static checks failures by aligning the renderer store apply signature with the transient force option and adding an explicit return type for the force-update handler. Pushed in commit 14ed6c84.

ewen-poch:
Addressed the Codex review: the renderer contract now accepts UpdateApplyOptions, and force mode only bypasses degraded process-reaping blockers while preserving durability and migration safety gates.

ewen-poch:
The force-update behavior is intentionally user-authorized: after the warning, force=true may proceed even when active tasks remain, and the user accepts interruption risk. Corrected the inverted gate condition so force bypasses any degraded install-gate refusal, matching the requested interaction.

ewen-poch:
Handled the actionable P2: the force action is now hidden when the install gate throws, since retrying the same exception cannot recover. The P1 safety recommendation is intentionally not adopted per the confirmed product decision: users may force through active work after the explicit warning and accept interruption risk.

ewen-poch:
Added a second confirmation step before force update. The confirmation explicitly warns that active tasks will be interrupted and unsaved work may be lost; force proceeds only after the user confirms. Added translations for all renderer locales.

ewen-poch:
Refined force-update scope: explicit confirmation may bypass active-task/process-reaping blockers, while migration and renderer-durability refusals remain non-forceable. This preserves the confirmed interruption-risk behavior without weakening data-integrity gates.

ewen-poch:
Addressed the audit finding: UpdateService.apply now accepts and carries UpdateApplyOptions through its admitted apply lifecycle, so the force intent is preserved consistently across update strategies.

ewen-poch:
Fixed the latest CI/review issues: removed the unused-options lint failure and aligned force-action eligibility with the backend by showing the second-confirmation action for degraded teardown, research, delegated, and settings-install blockers. Migration/durability refusals remain non-forceable.

ewen-poch:
Addressed the P2: generic teardown errors no longer expose the force action when the backend cannot force through them. The action remains available for explicit active-task blocker errors and keeps the second risk confirmation.

ewen-poch:
Fixed the CI classification failure by adding src/main/window-shortcuts.ts to the Windows-sensitive impact overlay. The classifier regression now covers this production Windows runtime signal.

ewen-poch:
Resolved the remaining P1: forced apply now passes force intent into the install gate, which bypasses only the early active-work admission check and still runs the full teardown/reaping plus durability and migration checks before quitAndInstall.

ewen-poch:
Fixed two remaining production-path gaps: the IPC-owned install-gate wrapper now forwards the force options, and quitAndInstall no longer bypasses an unsuccessful gate result. Force skips the initial active-work admission check, but teardown, renderer durability, migration checks, and the final active-work check must succeed.

Added regression coverage for ordinary refusal followed by confirmed force, teardown/durability/install ordering, and forced refusal for migration, failed durability, unreaped processes, or remaining work. Focused update tests: 78 passed. Targeted ESLint passed. Local node typechecking is blocked by missing reflect-metadata and @peculiar/x509 dependencies in the shared notebook-network-sandbox runtime.

CI Integrity separately rejects the change-impact.json classification fix under the protected-gate-control-plane rule. This requires an explicit maintainer ruleset bypass; no CI protection has been disabled or bypassed.

PR #2456, opened 2026-09-11
fix(settings): allow deleting selected provider

ewen-poch:
## Problem

The Settings provider list disables deletion for the provider that supplies the currently selected model. When that provider is still referenced by the current conversation or default selection, users must first switch the response model before they can delete the provider.

## Proposed change

Allow deleting the selected provider whenever another provider remains. Keep the existing protection against deleting the last provider.

Existing conversations retain their session-scoped provider/model configuration, while the existing settings repository clears the default provider/model pointer after deletion.

## Scope and non-goals

- No database fields, migrations, enum values, or persisted-format changes.
- No change to provider deletion semantics beyond the Settings UI guard.

- Regression test fails before the fix because the selected provider delete button is disabled.
- Regression test passes after the fix.
- `npx vitest run src/renderer/src/pages/settings/ProviderList.render.test.tsx src/renderer/src/pages/settings/ProvidersPanel.race.test.tsx src/main/settings/repository.test.ts`
- `npm run typecheck:web`
- `npm run lint -- --quiet`
- `git diff --check`

All checks ran after the final material edit.

## Review focus

Verify that the last remaining provider remains undeletable and that deleting a selected provider still clears the default settings pointer through the existing repository behavior.

### DIFF
--- a/src/main/settings/repository.test.ts
+++ b/src/main/settings/repository.test.ts
@@ -1441,6 +1441,20 @@ describe('settings repository: v2 official providers & activeModel migration', (
     expect((await repository.getSettings()).activeModel).toBeUndefined()
   })
 
+  it('selects the first remaining provider when the active provider is deleted', async () => {
+    const repository = new SettingsRepository(await createStorageRoot())
+
+    await repository.upsertProvider(provider({ id: 'p1', model: 'm1' }))
+    await repository.upsertProvider(provider({ id: 'p2', model: 'm2' }))
+    await repository.setActiveProvider('p1', 'm1')
+
+    const settings = await repository.deleteProvider('p1')
+
+    expect(settings.providers.map((item) => item.id)).toEqual(['p2'])
+    expect(settings.activeProviderId).toBe('p2')
+    expect(settings.activeModel).toBe('m2')
+  })
+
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed in 61bcb652.

The delete path now selects the first remaining provider atomically when the deleted provider is active, carrying its stored model when present. If no provider remains, the existing clear behavior is preserved. This keeps new sessions routable while leaving existing session-scoped configurations unchanged.

Added a repository regression test covering active-provider deletion with a remaining provider. The last-provider guard remains in the renderer.

ewen-poch:
After the follow-up review, the active-provider deletion change has been reverted in 74ce7a09.

The review correctly identified that persisted historical sessions can still reference the deleted provider after restart, and that a raw storage-order fallback can select an incompatible provider. This PR now leaves the existing deletion guard in place.

The requested follow-up will require a project-wide persisted-session reference check before deletion, plus a separate unavailable-session flow that lets the user select a replacement provider/model without rewriting historical messages or attribution.

ewen-poch:
Closing this no-op PR. The implementation was reverted after review identified historical session references and fallback compatibility concerns. Follow-up work will continue on a separate branch.

PR #2461, opened 2026-09-11
fix(settings): allow deleting providers with unavailable sessions

ewen-poch:
## Problem

The provider delete action was disabled for the selected or last provider, forcing users to keep a provider they explicitly wanted removed. A deleted provider can still be referenced by saved conversation history, but preserving that provider only to keep the old history selectable is the wrong trade-off for deletion.

## Proposed change

Allow the Settings delete action for the selected provider and for the last provider. The existing settings mutation removes the provider credentials/configuration. Existing Session records keep their provider/model identity unchanged; the existing Session resolver then presents those conversations as unavailable. The user can choose an available provider/model from the existing composer selector before sending a new turn.

## Scope and non-goals

- No soft deletion, new persisted fields, enum values, database schema, migration, or automatic history rewrite.
- Historical messages, branches, runtime attribution, and provider/model snapshots remain intact.
- The unavailable/reselection interaction is reused; this PR only removes the UI guard that prevented deletion and adds regression coverage for the public Settings boundary.
- Existing active-provider reconnect behavior remains the same. Deleting a provider may leave the persisted default unavailable until the user selects a replacement, which is intentional.

- Selected and only remaining providers expose an enabled Delete action.
- Deleting a provider still uses the existing settings workflow and clears its settings record.
- A saved Session that references the deleted provider remains readable, preserves its saved configuration, reports unavailable, and accepts an explicit provider/model replacement.

Targeted checks after the final edit:

- `src/renderer/src/pages/settings/ProviderList.render.test.tsx`
- `src/renderer/src/pages/workspace/ComposerModelPicker.render.test.tsx`
- `src/renderer/src/pages/workspace/workspace-session-agent-configuration-controller.availability.test.ts`
- `src/renderer/src/lib/acp/session-agent-configuration.test.ts`
- `src/main/settings/workflows.test.ts`
- `src/main/settings/ipc.test.ts`

Result: 6 files, 183 tests passed. The existing CI Static checks for the preceding implementation passed; this follow-up commit will rerun the affected CI checks. A browser component verification previously confirmed the unavailable selector and explicit replacement path; no new persistent UI state or new visual surface was added.

Per maintainer instruction, no complete local `npm test` run. CI remains authoritative for the full test plan.

## Review focus

Deletion of selected/last providers, preservation of historical provider/model attribution, and the existing unavailable-to-explicit-reselection transition.


ewen-poch:
The current implementation preserves saved conversation references before provider settings or authentication cleanup. The scan is read-only, includes archived/cross-project history and inactive branches, and fails closed when the saved-session catalog cannot be read completely. Existing unavailable sessions require an explicit provider/model selection; historical messages and attribution remain unchanged. No soft-delete state, schema migration, or additional persisted fields were introduced.

Validation on `08a1d947`: 13 targeted test files / 1,395 tests passed locally. Restoring the original deletion call makes the eight real-file reference regressions fail by allowing deletion. CI Static checks passed, including Node/web typechecks, lint, formatting, and i18n. The browser component checks confirmed the localized rejection and manual model reselection.

The AI review is **mergeable** with **no actionable findings**, so no additional code changes are needed for this review. Remaining CI checks are left running; this PR has not been merged.

ewen-poch:
Follow-up based on the maintainer clarification: provider deletion is intentionally allowed even when the provider is selected or is the last configured provider. The reference-blocking scan and its persistence guard have been removed.

Saved Sessions still retain their original provider/model attribution. After deletion they resolve to the existing `unavailable` state, and the user can explicitly choose a replacement provider/model before continuing. No history rewrite, soft-delete state, schema change, migration, or new persisted field was added.

The regression now covers the public Settings UI boundary: the selected and only provider has an enabled Delete action, while the existing unavailable/reselection tests remain green. Targeted validation after this change: 6 files, 183 tests passed.

PR #2467, opened 2026-09-11
feat(runtime): allow external R installs and conditional package restore

ewen-poch:
## Problem

External R runtimes could execute notebooks but could not install missing packages. Their dependency evidence also could not be exported for conditional restoration, even when a native package lock was available.

## Proposed change

- Allow explicit installation consent for an existing personal R library. Validate the interpreter and canonical writable library, reject system/site libraries, scope sandbox writes to the selected library, and recheck consent after waiting for the mutation lock. Preserve the existing external Python installation path.
- Capture external R `renv.lock` or existing hash-pinned Python requirements as schema-v2 conditional evidence, with exact interpreter/platform/architecture/package-manager prerequisites. Keep managed schema-v1 locks unchanged.
- Export a user-invoked restore script that checks prerequisites and checksums before creating a new destination, restores packages, and verifies effective package versions/paths. Keep full-environment reproduction and output-verification gates unchanged. Restore destinations remain user-owned; the app never adopts or deletes them.
- Replace the disabled R installation explanation with library-scoped consent controls. Detect eligible personal libraries automatically, select the sole candidate by default, and offer optional folder selection under Advanced options. Revalidate the selected directory before granting consent; preserve revocation of historical authorization. Translate the controls into all eight locales.

## Scope and non-goals

R captures require an available renv installation; Python requires a supported existing hash-pinned requirements lock. Conditional restoration is an exported bundle workflow, not in-app full runtime recreation. R repository/version records do not cryptographically pin archive contents. Repository availability, interpreter native dependencies, original inputs and result comparison remain prerequisites. No automatic uninstall of personal packages, package-manager migration, or OS provisioning.

Historical compatibility: retain existing v1 locks without rewriting prior runs. Historical external R authorization booleans do not grant library access and require renewed consent. Older applications cannot interpret new v2 locks.

Discovery: add a transient personalRLibraries candidate array to the existing runtime response; it is not persisted as authorization. No automatic directory creation.

Persistence: add optional `installLibraries` to per-runtime settings; new immutable v2 lock records include `externalRuntime.version` and `installerVersion`. No database schema migration. Revocation and app uninstall preserve personal packages.

New state/format: lock schema version 2 and partial reason `external-interpreter-required`; reuse the existing top-level partial/unavailable/reproduction states.

The final targeted impact set covers the changed owners and their consumers; the complete suite is left to PR CI as requested by the maintainer.

| --- | --- | --- |
| Consent, mutation ordering, installation and sandbox scope | Targeted Vitest admission, mutation, operations, package-manager, package-process-sandbox, external-r-library suites | Passed |
| Capture, legacy locks and full-reproduction gates | Targeted environment-lock, external-environment-lock, environment-state-tracker, native-lock-restoration, restored-environment-verification, artifact recipe/execution/projection suites | Passed except one Windows symlink fixture creation blocked by EPERM |
| Persisted consent and application interfaces | Settings codec/repository/service/capabilities, runtime workflows/IPC, Electron/Web adapters, shared contract/Web RPC suites | Passed except nine unrelated SettingsService symlink fixture creations blocked by EPERM |
| Export and visible behavior | Export, both Artifact panel suites, RuntimesPanel render suite, eight-locale guard; authenticated Web screenshot check | Passed |
| Process/build checks | `npm run typecheck:node`, `npm run typecheck:web`, `npm run lint`, `npm run build:web`, generated Web API map check | Passed |

The combined 31-file run reported 2,052 passed, 29 skipped and 10 failures, all at Windows symlink fixture creation before business assertions. These are unverified cases, not passing checks. Export/capture checks were rerun after the final script version-comparison change (32 passed). Other listed checks cover the unchanged final implementation; independent review confirmed the impact mapping. Exact-head PR CI must pass before merge.

Additional local spike evidence: R 4.4.3/renv 1.2.4 restored a recursive dependency set including compiled DLLs; Python restored hash-pinned pandas/NumPy wheels and reproduced a numerical result. The production restore script rejected an R version mismatch and modified native lock before creating a destination. It restored indented/continued requirements and equivalent numeric release versions. Earlier probes showed renv accepts a mismatched R version and same-version changed package content, motivating the explicit conditional claim.

Uncovered locally: standalone CRAN R distribution, macOS/Linux native execution and OS sandbox enforcement. The available Windows R is Conda-distributed; sandbox path policy is covered by unit tests, not a full protected installer E2E run. No claim of complete reproducibility is made.

PR follow-up: extracted the kernel's existing Windows R-home resolver and reused it for package installation, preserving read access for both `bin/Rscript.exe` and `bin/x64/Rscript.exe` while keeping only the authorized library writable. The new read-root fixtures plus kernel/install checks passed (62 passed, 80 skipped). Registered the new Windows-sensitive source files in the CI impact manifest; classifier/workflow checks passed (106 passed, 3 skipped). Node/sandbox typechecks and lint passed after these fixes. The earlier Windows CI failure was in an unchanged Rust test's temporary-directory cleanup (file-in-use error 32); its focused local rerun passed. Require the updated head's full CI result before merging.

## Review focus

Library-scoped authorization and revocation, immutable v1/v2 decoding, preservation of strict full-reproduction gates, and portable restore-script prerequisite/identity checks.


ewen-poch:
Addressed the Windows external R read-root issue in a87121991dccd8a6d7a49daba3495550fd07864d. Package installation now shares the kernel's Windows-aware R-home resolver, covering both `bin/Rscript.exe` and `bin/x64/Rscript.exe`. The R home is readable; package writes remain scoped to the explicitly authorized personal library. Focused installer/kernel checks passed (62 passed, 80 platform skips), and Windows core passed on the current commit.

The CI impact manifest now registers the three Windows-sensitive source files. Its formatting was corrected in 6e849c5a66abe2fa030dfc0b37bb15554c20e364, and the current Static checks job passed. CI Integrity still requires explicit maintainer authorization for this protected manifest change.

The latest automated review did not produce a code-review verdict: subscription authentication failed, and the API fallback exited before reading code. The workflow retained no error artifact, so the underlying fallback failure remains undiagnosed. No review approval or complete reproducibility claim is inferred from that failure.

ewen-poch:
Addressed the restart-flag finding in 3521973f4. External R installation now returns needsRestart based on the installer exit status, matching the managed CRAN path. The regression test covers successful and failed exits and verifies that failed attempts retain possible mutation risk.

Partial installation safety remains in the existing mutation layer: confirmed package changes or a possibly mutating failed attempt can still recommend restarting a live R session. An unsuccessful batch does not prove that its earlier package installations were rolled back. No new states, persistence fields, or interaction flows were introduced.

Validation after the change: 185 tests passed (2 skipped) across external-r-library, package-manager, package-mutation, and package-operations; Node/sandbox typechecks, lint, and formatting checks passed.

ewen-poch:
Addressed legacy R consent revocation in 85b93dc61. The Settings toggle now reflects persisted consent, so an older installAuthorized=true entry can be revoked without a library path. Once revoked, enabling installation again requires a personal library path. Backend package admission still refuses R installation without library consent.

The render regression test covers revocation with no library and the subsequent re-enable gate. Focused renderer, settings, admission, and i18n suites passed (884 tests); Web typecheck, lint, formatting, and Web build passed. A real-browser check against the rebuilt Web renderer used an in-memory legacy-settings fixture, confirmed the revoke request carried authorized=false with no library, and captured the resulting UI. No saved user authorization was changed by that browser test.

Historical settings require no migration. This changes only the existing consent toggle behavior; no enum values or persisted fields were added.

ewen-poch:
The two R patch-version findings share a premise that does not hold for R.version. On the real R 4.3.3 runtime, R.version$major is "4" and R.version$minor is "3.3". Executing the production expression paste(R.version$major, R.version$minor, sep=".") returns "4.3.3", exactly matching as.character(getRversion()). It does not return "4.3".

Both the inventory and live-manifest scripts in environment-state-tracker.ts use that expression. Their RUNTIME parser preserves the complete version string without truncation. The capture check therefore compares full versions, and the conditional restore probe uses the same full version. These findings do not demonstrate the reported rejection of normal patch-version locks.

No version normalization change is applied: relaxing the comparison to major.minor would allow a different patch release and weaken the approved matching-interpreter contract. The existing exact-version checks are retained.

ewen-poch:
Implemented automatic personal R library discovery and optional advanced folder selection in a9a4a198b. A single eligible library is selected automatically; multiple candidates require a choice. Users can optionally choose another existing folder through the native folder picker under Advanced options. Cancellation preserves the selection, and historical consent remains revocable.

Discovery is read-only and excludes system/site libraries. Enabling consent re-probes the selected R and validates the canonical writable directory; installation retains its existing validation and sandbox write boundary. The new personalRLibraries field exists only in the discovery response. No enum values, persistent fields, historical migrations, or automatic directory creation were added.

Validation: 917 focused tests passed across discovery, workflows, renderer, admission, settings and i18n; an additional IPC/renderer run passed 67 tests. Node/sandbox and Web typechecks, lint, formatting, and Web build passed. The production probe was also exercised against standalone R 4.3.3. Browser verification covered the default, cancellation, and advanced selection using in-memory fixtures; the native dialog itself was not exercised by that browser test.

ewen-poch:
Rebased this PR onto origin/main at d637a650e. Resolved the eight locale-file conflicts by merging translation keys, retaining the current main-branch entries and the external R consent/restore entries, including removal of the obsolete manual-path placeholder.

Range-diff verification confirms that the functional patches are preserved; the rebase differences are locale context/order only. No new behavior, data model, or persistence changes were introduced by conflict resolution.

Post-rebase validation: all 1,025 tests passed across nine targeted runtime, consent, artifact export, renderer, and i18n suites. Lint passed. Node/sandbox and Web typechecks passed. The complete test suite remains delegated to PR CI as requested.

ewen-poch:
Fixed the portable-suite failure in 708a6d1e4 by registering external-r-library.ts and runtime-workflows.ts in the Windows-sensitive source list. The exact classifier assertion failed locally before the change and passed afterward. Classifier/integrity tests passed (111 tests), and PR-gate/integrity workflow contracts passed (38 passed, 3 skipped). Formatting and diff checks passed. The coverage aggregation failure reported the same classifier assertion, not a separate coverage threshold failure.

The macOS presentation failure is an unchanged main-branch visual assertion: e2e/visual-regression.spec.ts requires white-space: nowrap, while the compact ErrorNotice introduced on main uses whitespace-normal and the design specification allows wrapping. Updating that test is pending approval to extend the current file scope; the product wrapping behavior is retained.

The two Windows browser failures (Japanese notification recovery and German settings layout) passed twice each in a local Edge diagnostic run. The CI Chromium browser was not installed locally, so these results do not establish that the original CI timing failures are fixed. The protected change-impact manifest still requires the repository's explicit maintainer ruleset bypass; no integrity check was weakened.

The repeated R version finding remains inapplicable after rebase: the production manifest expression paste(R.version$major, R.version$minor, sep=".") equals as.character(getRversion()) on the actual R 4.3.3 runtime (minor is "3.3"). The parser preserves that full version. Exact interpreter-version validation remains unchanged.

ewen-poch:
Updated the recovery-action assertions in c3b2557a7 to match the existing compact ErrorNotice wrapping behavior. At 320, 375, 414, and 768 px, the test now requires the action to be fully in the viewport, below the message, and free of horizontal or vertical text clipping. Product styles, screenshot baselines, and pixel comparison thresholds are unchanged.

Validation after the edit: lint, formatting, diff checks, and the E2E build passed. An isolated Electron run of the recovery section passed all four widths. The complete representative-states test stopped earlier on this Windows host because the Bash provenance fixture executed printf through PowerShell; it did not reach the recovery assertions. The complete macOS visual result remains subject to the new CI run.

This change adds no product interaction, state, persistence, or compatibility changes. CI Integrity still requires the repository's explicit maintainer handling for the protected change-impact manifest; the guard remains intact.

ewen-poch:
The new macOS run reached and passed all four recovery-action layout checks. It then failed the unchanged session-recovery-warning.png snapshot (8,623 differing pixels on the first attempt). Inspection of the expected and actual artifacts shows that the expected image predates the shared ErrorNotice conversion on main in 41d10062a: the current notice has the warning icon, revised typography, and the existing shared action layout. This is a baseline mismatch, not a failure of the new visibility or clipping assertions.

The renderer layout suite also passed all 46 tests on macOS. Updating the single recovery screenshot baseline is pending approval to extend the file scope. No baseline or pixel threshold has been changed. The latest CI Integrity diagnostic still reports protected-gate-control-plane for scripts/ci/change-impact.json and requires an explicit maintainer ruleset bypass.

ewen-poch:
Addressed the architecture-casing finding in 53e25dc47. External-lock validation now uses the existing normalizeRuntimeArchitecture helper, accepting AMD64/ARM64 and the established aliases while continuing to reject unsupported architectures. The original architecture evidence remains unchanged in the lock, so this does not introduce a persisted-format change or migration. Regression tests reproduced the rejection before the fix and now pass, including a Windows Python pip-lock case (8 focused tests passed). Node/sandbox typechecks, lint, formatting, and diff checks passed. The broader capture/reproduction run passed 101 tests, skipped 3, and failed 3 on local Windows constraints: file-symlink permission and two platform-specific path expectations. These failures are not reported as successful validation.

Updated only the approved recovery screenshot baseline in 51fcf7a (see the commit in this PR). The first macOS attempt and retry produced byte-identical actual screenshots. The baseline now reflects the shared ErrorNotice already on main; pixel tolerances and product UI remain unchanged. A fresh CI result is still required.

The R library-context finding is valid for libraries supplied solely through filtered R_LIBS/R_PROFILE_USER settings. Implementing the suggested library injection would change the existing installation-time .libPaths visibility contract. That design adjustment is pending maintainer confirmation; it has not been silently applied. A bounded option is to revalidate the selected library using the chosen interpreter before installation, then pass only that authorized library into the controlled install context, retaining system-library exclusion and the existing write boundary.

ewen-poch:
Rebased this PR onto origin/main at be5aeb8b6 and resolved the overlap with #2475 in the recovery visual test. Main's wrapping/bounds checks and this PR's full-visibility/text-clipping checks are both retained. The final test file is byte-for-byte unchanged from the pre-rebase feature head; range-diff confirms all other patches are preserved. The branch was updated with an explicit force-with-lease.

Addressed the approved R library-context finding in 57495efd1. Before installing, the selected R re-probes its profile-visible personal libraries and must still report the canonical authorized destination. Probe failure or a missing destination stops before installer launch. Only that destination is propagated through R_LIBS_USER; R_LIBS and R_PROFILE_USER remain filtered, and the existing system-library rejection and filesystem write boundary remain in place.

Validation after rebase: 978 tests passed and 3 skipped across package-manager, external library, package sandbox, runtime workflows, external locks, and i18n. Node/sandbox and Web typechecks, lint, formatting, and diff checks passed. A real R 4.3.3 spike exercised the production installer and environment projection with both R_LIBS-only and R_PROFILE_USER-only library configuration: the old context rejected the library, the fixed context installed cached praise 1.0.0 into the authorized library, and removing the configuration prevented another installer launch. The spike used a local package mirror and did not re-test native isolation or create new runtime grants.

No new persisted fields, enum values, historical migrations, or UI changes were introduced. Fresh CI remains authoritative for full cross-platform verification; CI Integrity still requires explicit maintainer handling for the protected change-impact manifest.

ewen-poch:
Investigated all three findings against real CRAN R 4.3.3 and renv 1.1.4. The version finding does not apply: R.version$minor is "3.3", so the production expression returns "4.3.3", matching renv.lock. A real snapshot explicitly requesting stats, utils, and praise recorded only praise. This agrees with renv_snapshot_library, which excludes base packages: https://github.com/rstudio/renv/blob/main/R/snapshot.R#L630-L631 . No patch-version relaxation or host-package containment exemption was introduced.

The investigation did uncover a different concrete capture defect: renv::snapshot returns an S3 renv_lockfile object, and jsonlite::toJSON rejects that class. Fixed in 7ed800cee by serializing unclass(lock), preserving the native records. Before the fix, the production capture returned unavailable with "No method asJSON S3 class: renv_lockfile"; afterward, it captured a conditional bundle with the full R version while excluding observed base packages. An opt-in real-R regression test now covers that path. All 9 tests passed with the existing R/renv installation; Node/sandbox typechecks, lint, formatting, and diff checks passed. No persisted-format, state, UI, or restore-policy changes were made.

Other CI failures require separate file-scope approval: SkillsPanel.render.test.tsx still expects the old two-argument deletion call, and user-skill-repository.architecture.test.ts counts overload declarations as extra public operations. Both failures reproduce locally and those files are unchanged by this PR. The Windows native test failed at temporary-directory cleanup with sharing violation 32; no native source change has been made. The prior macOS presentation run now passes. CI Integrity remains subject to the existing explicit maintainer bypass requirement.

ewen-poch:
Addressed both actionable R findings. In fe0cc5d33, external R installers receive write access only to the authorized personal library and the workload cache, while the app-managed runtime root is read-only. Inherited Conda/Mamba cache roots cannot reintroduce the broad grant. Managed runtime behavior is unchanged. The boundary regression failed before the fix and passes afterward.

In 9a6281d99, conditional R restore retains the library search paths probed from the supplied interpreter, with the new destination first. This preserves separately installed renv/jsonlite tools without accepting library paths from lock data. Locked package versions and destination containment are still verified. A private real-R fixture with renv and jsonlite in separate libraries reproduced failure after package installation; the fixed production restore script completed successfully using a cached local package mirror.

In f17bc0a13, the skill renderer test expects the current optional directory argument, and the architecture inventory counts method implementations rather than overload declarations. Both CI failures reproduced locally before the change. The Windows timeout test now checks its termination proof and timeout before cleanup, closes its own handles, and retries only sharing violation 32 for up to one second within its unique test directory. The original native failure was not reproduced in 100 local repetitions; this is bounded cleanup hardening, not a claimed diagnosis of the external lock holder.

Final validation: 129 targeted tests passed, 1 skipped; all 28 Windows native tests passed. Node/sandbox and Web typechecks, lint, changed-file formatting, and diff checks passed. The changed Rust file passes rustfmt; whole-crate formatting reports an existing difference in unchanged wfp.rs, which was left untouched. No new persisted fields, enum states, migrations, or UI interactions were introduced. Fresh CI remains pending, and the protected change-impact manifest still requires explicit maintainer handling.

ewen-poch:
Rebased onto main at daaa7fd5b and resolved the overlapping changes. Retained upstream host-aware path assertions and the network-readiness gate, kept the bounded native cleanup after termination assertions, and omitted fixes already present upstream. Removed duplicate test declarations exposed by integration.

Validation: Node/sandbox and renderer typechecks passed; lint passed; 313 focused tests passed (4 skipped), with the sandbox-owner suite additionally passing all 58 tests; all 28 native Windows tests passed. No full npm test run. No new runtime behavior, persistent fields, or states were introduced by this rebase.

The separate review finding about propagating a selected external R library into notebook kernels remains pending design confirmation; it is not included in this rebase.

ewen-poch:
The macOS temporary-directory finding does not apply to the composed installer path. NotebookNetworkSandboxOwner.wrap creates an owned per-command directory, sets TMPDIR/TEMP/TMP, and adds that exact directory to readWriteRoots before calling the platform backend. Cleanup remains gated on backend/process teardown. Adding another installer-owned temp lifecycle would duplicate the existing owner. The sandbox-owner and package-sandbox suites passed: 85 tests passed, 1 skipped. This verifies the composition contract, not a new live macOS R installation claim.

Current CI failures are separate: Ubuntu kernel-executor.test.ts:713 received the stdin-pipe diagnostic before the expected exit-code/stderr diagnostic; the Windows layout test never reached its assertions because renderer startup remained starting past the 90-second fixture budget. Its diagnostics show database migration consuming roughly 88 seconds before subsequent startup work. These require focused investigation and file-scope approval for the kernel regression tests and E2E startup fixture; no timeout increase or assertion weakening has been applied.

CI Integrity was reproduced locally: scripts/ci/change-impact.json is a protected gate control-plane file and requires an explicit maintainer ruleset bypass. The guard and coverage registrations remain intact.

ewen-poch:
The kernel exit assertion reproduced locally with the real Python interpreter: the immediate-exit fixture can trigger EPIPE before exit handling. The local fix makes both exit-diagnostic fixtures consume the request before exiting, preserving exit-code, stderr, and sandbox-annotation assertions. Both scenarios passed ten consecutive runs. This stabilizes the intended post-dispatch exit test; it does not claim to change production pre-dispatch EPIPE behavior.

The Windows startup draft uses a bounded 180-second readiness budget and a separate 240-second fixture budget, leaving the test-body timeout unchanged. Validation found that scripts/electron-fixture-teardown.test.ts must also understand Playwright's tuple fixture declaration and the platform-specific deadline. That additional test file is awaiting file-scope confirmation; the incomplete Windows change has not been committed or pushed.

ewen-poch:
Implemented the approved CI fixes in 72d2f4824. Exit-diagnostic fixtures now consume the request before exiting, so EPIPE cannot race the intended exit assertions. The exact CI failure reproduced locally before the change; both real-Python scenarios passed ten consecutive runs afterward.

Windows renderer readiness now has a shared 180-second deadline and a separate 240-second Playwright fixture budget. Test-body timeouts and readiness assertions remain unchanged, as do non-Windows budgets. This accommodates the observed 89-second schema initialization without disabling migrations or changing database behavior. Updated lifecycle tests cover startup after the old deadline, bounded failure with diagnostics, shared settings deadlines, and teardown.

Validation: all 24 fixture tests passed; Node/sandbox typechecks, changed-file lint/formatting, and diff checks passed. Playwright successfully registered the affected layout tests. Hosted Windows timing will be verified by CI. No production code, persistent fields, enum states, or user interactions changed.

ewen-poch:
The R-version finding is not applicable. Both r_loop.R and the inventory probes use paste(R.version$major, R.version$minor, sep="."). R.version$minor already includes the patch component: the real R 4.3.3 probe returned minor="3.3", runtime="4.3.3", and renv.lock R.Version="4.3.3". Re-ran this against renv 1.1.4 successfully. Exact version matching remains intact.

The restart-recommendation finding is valid. Package operations record the recommendation through the environment process key, and session read models consult that same key. External R bindings sharing r:default-r can therefore share a recommendation incorrectly. Correcting this requires changing in-memory recommendation ownership and its session/runtime lookup and clearing semantics, with additional owner/read-model tests. This design/file-scope expansion is awaiting explicit confirmation; no runtime identity or persistent schema change has been made.

ewen-poch:
Fixed restart recommendation isolation in d3e788da4. External R recommendations now include runtime identity and aggregate session identity in the existing in-memory set. Package operations mark only sessions bound to the affected external runtime; read models resolve the same scope. Targeted restart, whole-session restart, and binding teardown clear only the corresponding external recommendation. Managed environment behavior and kernel process keys remain unchanged.

The new service regression failed against the original production code because another external R session incorrectly reported restartRecommended=true. After the fix, both restart paths preserve the recommendation in a second session sharing the runtime, exclude unrelated external/managed runtimes, and clear stale recommendations on switch-away/rebind. Validation: 54 focused tests passed; Node/sandbox typechecks, changed-file lint, formatting, and diff checks passed. No persisted fields, migration, enum values, or renderer interaction changes.

The preceding CI run now passes all Ubuntu shards, coverage, and Windows E2E shard 1, confirming the earlier fixes. A separate macOS journey failed its tool-layout assertion with an 80px transient excursion (limit 2px). This is distinct from startup timing and requires separate file/UI scope approval; no assertion threshold was relaxed. CI Integrity still requires the documented explicit maintainer handling for the protected impact manifest.

ewen-poch:
Fixed the anchored transcript excursion in 2b86afa8f. A minimal browser reproduction using the production scroller matched the CI trace: shrinking trailing content by 80px clamped scrollTop before the spacer caught up, moving the tool row from 172px to 252px and back. The cumulative @shadcn scroller patch now re-runs its content synchronization layout effect when React children change, updating the anchor spacer before paint. ResizeObserver scheduling, user-scroll intent, and the 2px assertion threshold are unchanged.

The browser regression fails against the original installed source with an 80px excursion and passes against the exact patched source with 0px excursion. Five repeated runs passed all 10 checks for shrink/grow anchoring and free user scrolling. The 12 dependency-patch lifecycle tests and 7 existing scroller tests passed, including historical patch recovery and refusal to overwrite unknown edits. Lint, formatting, and diff checks passed. Local browser verification used installed Edge with an isolated patched-module alias; shared node_modules was not modified. Hosted macOS verification remains pending.

The completed review on d3e788da4 correctly reiterates the separate external-library-to-kernel integration gap. That runtime-binding/execution change remains pending design confirmation. The latest Windows failure is also separate: settings-layout.spec.ts:50 could not find the initial .github-star-cta element. Neither that unrelated test nor the library propagation path was changed in this scroller fix.

ewen-poch:
The external-R library-to-kernel finding is addressed in 6a1cda805. Execution admission reads the current Settings consent, validates the canonical personal library, and rechecks it after the environment queue. The transient interpreter context carries that library into R_LIBS_USER and an explicit read-only sandbox root. Kernel identity includes the library, so revocation or replacement cannot reuse the old process on the next execution. The path is checked again before sandbox preparation. Durable binding and frozen-run schemas are unchanged; historical consent without a selected library grants no additional root.

Validation: the service regression first failed because the library was absent from dispatch. All 140 targeted binding/session/queue tests and 6 kernel checks now pass. A real R 4.3.3 kernel loaded glue from the personal library and replaced the process after revocation. That integration uses a recording pass-through sandbox: it validates the requested read-only grant, not native enforcement. Node/sandbox typechecks, ESLint, and formatting passed. No full npm test was run.

The Windows settings-layout failure on d3e788da4 is a resource-loading failure: its trace records net::ERR_NO_BUFFER_SPACE for /assets/error-notice-C8hDLhaN.js before the page mounts. The exact Russian 768px scenario passed 10 repetitions with two workers against the built fixture in local headless Edge. No assertion, retry, or timeout was weakened. This does not prove the hosted Chromium resource failure is resolved.

The new review about allowing app-managed directories as external personal libraries is valid: profile visibility and canonicalization do not establish that the directory is user-owned. Expanding authorization and pre-install validation to reject protected application roots changes the accepted-path policy and may invalidate existing consent. That change is pending maintainer confirmation.

A related follow-up is required before claiming end-to-end capture/reproduction for this new execution context: environmentCaptureTarget currently omits the selected personal library, and the independent inventory/lock subprocesses use their existing environment. A read-only real R probe confirmed the mismatch: find.package("glue") returned the selected personal-library package path with the kernel context, but an empty result with environmentCaptureProcessEnv. Extending that propagation and checking inventory-cache identity is pending scope confirmation; successful kernel loading alone is not proof of aligned lock capture.

PR #2471, opened 2026-09-11
fix(skills): allow deleting identity conflicts

ewen-poch:
## Problem

A user Skill copied or retained under the writable `personal/` or `imported/` roots can reuse the name of a bundled Featured Skill. The catalog correctly marks it as `identity-conflict`, but the Settings page hides its action menu, so the user cannot remove the orphaned package.

## Proposed change

- Keep runtime actions unavailable for identity-conflicting Skills.
- Expose the existing delete action for conflicting user-owned Skills.
- Reject built-in Skill deletion explicitly in the main-process catalog service.

## Scope and non-goals

No new data fields, enum values, database migration, or persisted-format changes. Built-in Skills remain non-deletable. Export, edit, toggle, and detail actions remain unavailable for conflicting entries.

- The focused regression test failed before the production change and passes afterward.
- `npx vitest run src/renderer/src/pages/settings/SkillsPanel.render.test.tsx src/main/settings/skill-catalog.test.ts` — 112 passed.
- `npm run lint` — passed.
- `npm run typecheck:web` — blocked by pre-existing missing dependencies in `packages/notebook-network-sandbox/runtime/src/gateway/local-ca.ts` (`reflect-metadata`, `@peculiar/x509`), unrelated to this change.

## Review focus

Verify that an identity-conflicting personal/imported Skill can be deleted, while built-in Skills cannot be deleted through the main-process service.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -2155,8 +2155,8 @@ const createApplicationModules = async (
     )
   }
   composition.phase('marketplace-recover')
-  settingsService.setSkillDeletionGuard((skillId) =>
-    specialistPackageService.assertSkillDeletionAllowed(skillId)
+  settingsService.setSkillDeletionGuard((request) =>
+    specialistPackageService.assertSkillDeletionAllowed(request.id, request.source)
   )
   // Per-session specialist binding store. Shared between the SET_SESSION_SPECIALIST barrier
   // (validate + record) and the runtime switch so a hot-switch lands on the same source of truth.
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -2156,7 +2156,11 @@ const createApplicationModules = async (
   }
   composition.phase('marketplace-recover')
   settingsService.setSkillDeletionGuard((request) =>
-    specialistPackageService.assertSkillDeletionAllowed(request.id, request.source)
+    specialistPackageService.assertSkillDeletionAllowed(
+      request.id,
+      request.source,
+      request.directoryName
+    )
   )
   // Per-session specialist binding store. Shared between the SET_SESSION_SPECIALIST barrier
   // (validate + record) and the runtime switch so a hot-switch lands on the same source of truth.
--- a/src/main/settings/service.test.ts
+++ b/src/main/settings/service.test.ts
@@ -4380,7 +4380,7 @@ describe('SettingsService: skills', () => {
     await expect(service.deleteSkill({ id: 'personal-my-skill' })).rejects.toMatchObject({
       code: 'protected-skill'
     })
-    expect(guard).toHaveBeenCalledWith('personal-my-skill')
+    expect(guard).toHaveBeenCalledWith({ id: 'personal-my-skill' })
     await expect(service.getSkillDetail('personal-my-skill')).resolves.toBeDefined()
   })
 
--- a/src/main/settings/service.ts
+++ b/src/main/settings/service.ts
@@ -287,7 +287,7 @@ class SettingsService {
   private deviceCredentialAuthenticator?: (credentialId: string) => Promise<void>
   private deviceCredentialAuthenticationCanceller?: (credentialId: string) => Promise<void>
   private deviceCredentialDisconnector?: (credentialId: string) => Promise<void>
-  private skillDeletionGuard?: (skillId: string) => Promise<void>
+  private skillDeletionGuard?: (request: DeleteSkillRequest) => Promise<void>
 
   hasActiveInstall(): boolean {
     return this.installCoordinator.getActiveId() !== undefined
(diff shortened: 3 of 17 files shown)


ewen-poch:
Thanks for the review. Both findings are valid.

The current delete request carries only the catalog id, so a duplicate sidecar id can resolve to the wrong user source, and the built-in guard can reject a user package whose sidecar id matches a bundled Skill. I will address this by carrying the concrete user source in the delete request and resolving that exact `personal` or `imported` package under the existing mutation lock; built-in deletion will remain rejected based on the actual catalog entry. This requires a shared IPC request-shape change, so I am pausing implementation for maintainer confirmation before expanding the approved scope.

ewen-poch:
Implemented both findings.

- The renderer now passes the Skill source with every delete request.
- The shared delete request, preload-facing store action, repository, and store resolve the exact `personal` or `imported` source under the existing mutation lock, preventing cross-source sidecar-ID deletion.
- Built-in protection applies only when the request is not explicitly targeting a user source, so a user package whose sidecar ID matches a bundled Skill remains removable while the actual built-in entry stays protected.

Validation: 423 focused tests passed; lint and `git diff --check` passed. Node typecheck has no change-related errors and remains blocked only by the existing missing `reflect-metadata` and `@peculiar/x509` sandbox dependencies.

ewen-poch:
Addressed the source-validation concern.

`UserSkillStore` now validates the source against the exact `imported` / `personal` allowlist before constructing any user-skill path or scanning directories. Invalid runtime IPC values fail closed with `Invalid user Skill source.` A regression test covers a traversal-like source value.

Validation: 231 focused tests for the updated catalog/repository/UI paths passed; lint and `git diff --check` passed. The existing CI typecheck failure remains limited to missing notebook sandbox dependencies.

ewen-poch:
The latest findings are valid: source selection currently stops at file deletion. The production Specialist guard still resolves only the id, and enablement/tag cleanup can affect the surviving built-in entry with that id. I am holding those changes for maintainer approval because they extend into the Specialist guard and Settings tag workflow outside the approved file set. The proposed fix will retain package ownership checks and preserve settings/tags for surviving identities; it requires no new persisted fields or enum values.

Correction to my earlier CI report: full static-job logs show change-related TS2775 errors in the validation helper and TS2345 errors in both renderer deletion calls. These reproduced locally before the fix. The compiler fixes are committed locally but not pushed yet, to avoid another review of a partially addressed head. The four change-related diagnostics are gone; 231 focused tests, lint, and diff checks pass. The local typecheck still reports missing notebook sandbox dependencies, which is distinct from the actual CI failures.

The latest E2E logs also show a flaky Windows conversation reply check and a repeated macOS recovery-alert bounding-box timeout. Their relationship to this PR has not been established; neither test nor timeout has been modified.

ewen-poch:
Implemented the remaining identity propagation and relationship-safety fixes.

- The Specialist deletion guard now receives the requested user source, so a user package whose sidecar ID matches a built-in Skill is not misclassified as the built-in target. Existing ownership and reference protections remain enforced for the selected user source.
- After deletion, enablement reset and tag cleanup are skipped when another catalog entry with the same ID survives. This preserves the built-in Skill's state and tags.
- Added regression coverage for source-aware protection and surviving same-ID relationships.

Validation: affected focused tests pass, lint passes, and the only remaining local typecheck errors are the pre-existing missing `reflect-metadata` and `@peculiar/x509` notebook sandbox dependencies.

ewen-poch:
Addressed the duplicate same-source package identity issue. Settings rows now carry the stable user package directory name through `DeleteSkillRequest`; the main-process resolver validates and targets that exact source directory, and the Specialist deletion guard receives the same qualifier. This prevents deleting the wrong package when two Personal or Imported packages share a sidecar ID. Focused Skill catalog/workflow/store-facing tests and lint pass.

ewen-poch:
Fixed the regression for ordinary Personal and Imported Skills without specialist sidecar metadata. Source-qualified deletion now accepts the conventional `source-directoryName` identity before falling back to sidecar metadata, so passing the stable directory qualifier works for both ordinary and Specialist-managed packages. Added a repository regression test covering deletion with source and directory name; focused tests and lint pass.

ewen-poch:
Fixed the TypeScript compatibility regression in the repository deletion facade. `UserSkillRepository.delete` now has explicit overloads for the legacy `(id, guard)` form and the source-qualified `(id, source, directoryName, guard)` form, with a narrowed implementation before delegating to `UserSkillStore`. Node typechecking passes; the sandbox typecheck remains blocked only by the repository's pre-existing missing `reflect-metadata` and `@peculiar/x509` dependencies. Focused deletion tests and lint pass.

ewen-poch:
Fixed bulk Skill deletion to pass each row's `directoryName` through the existing delete action. This keeps bulk deletion aligned with single-row deletion for packages whose sidecar ID differs from their directory name, and the bulk renderer regression tests now assert the qualified call.

PR #2474, opened 2026-09-12
feat(cli): add terminal profile initialization

ewen-poch:
## Problem
The existing headless CLI can control Open Science without opening the desktop window, but it had no explicit terminal initialization command. Future standalone CLI work also needs a stable profile spelling without changing the current Electron-backed runtime.

## Proposed change
- Add `open-science init` to create the local configuration directory without starting the app.
- Add `--profile` as a forward-compatible alias for `--config-root`.
- Document the terminal-first initialization and headless workflow.

## Scope and non-goals
This PR does not publish an npm package, split the daemon, remove Electron, add a TUI, or change persistence formats. `start` continues to use the existing Electron headless runtime.

- `open-science init --profile <path>` creates the profile directory and is repeatable.
- Existing lifecycle and task commands remain unchanged.
- CLI module tests pass: `npm run test:cli` (152 tests).
- ESLint passes for changed CLI files.
- Checks ran after the last material edit.

## Review focus
Please review the terminal workflow and whether `--profile` is an appropriate stable spelling to preserve if a standalone CLI is added later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,10 +363,11 @@ A: Enable the **Remote Compute (SSH)** skill under **Settings → Skills**, regi
 
 ### Is there a command-line interface?
 
-A: Yes. Install it in one click from **Settings → General → Command line tool → Install command** (adds `open-science` to your PATH; no separate Node.js needed). The CLI controls the local service and submits research tasks without opening a browser:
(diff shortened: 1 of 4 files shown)


ewen-poch:
Addressed both P2 findings:

- `init` now resolves packaged vs development mode through the same app discovery path as `start`, so source checkouts use `.open-science-project` and packaged installs use `.open-science`.
- Packaged initialization now rejects `--config-root`/`--profile` with the same restriction as `start`.

Added regression coverage for profile resolution and packaged override rejection. Focused CLI tests pass (153 tests), and ESLint passes.

PR #2489, opened 2026-09-12
feat(runtime): expose runtime list command

LiuZhaoyan:
## Problem

Issue #1987 requires an agent-operable first-run bootstrap that can eventually reach a runnable first task.

Open Science already has internal runtime detection and managed installation support for Claude Code, OpenCode, Codex, and CodeBuddy, but those capabilities are not available through the public Task API, SDK, or CLI. Runtime setup therefore remains one of the Settings-only steps in the larger bootstrap flow.

## Proposed change

This PR exposes that existing Agent runtime functionality as a focused slice of #1987:

* `GET /api/v1/runtimes`
* `POST /api/v1/runtimes/:framework/install`
* `OpenScienceClient.listRuntimes()`
* `OpenScienceClient.installRuntime()`
* `open-science runtime list [--json]`
* `open-science runtime install <framework> [--json]`

Supported framework IDs are `claude-code`, `opencode`, `codex`, and `codebuddy`.

Runtime listing exposes only `framework`, readiness `status`, `version`, and managed/external `source`, without exposing executable paths or the full Settings snapshot.

Runtime installation reuses the existing Settings install commands and remains local-only. It waits for the existing final install result rather than introducing a new job or progress API. The SDK uses a one-hour method-specific timeout by default while allowing `timeoutMs` overrides; timing out the client request does not cancel an installation already accepted by the service.

## Scope

This PR only makes the existing Agent runtime list/install functionality operable through the public headless interfaces.

It does **not** complete #1987. Provider setup, connector/OpenAlex configuration, skill enablement, PATH launcher repair, onboarding finalization, and overall bootstrap orchestration remain follow-up work.

It also does not add a progress/event API, persistent install jobs, separate repair/update commands, or automatic framework selection.

Part of #1987.

## Validation

* Focused Task API / HTTP / SDK / CLI tests pass.
* ESLint passes.
* Prettier passes.
* `git diff --check` passes.
* The Node typecheck issue identified in review was fixed by widening the test fixture parameter type.

### DIFF
--- a/e2e/fixtures/electron-app.ts
+++ b/e2e/fixtures/electron-app.ts
@@ -1074,7 +1074,14 @@ class ElectronAppHarness implements ElectronApp {
             throw new Error('Electron E2E forced close did not reap the process tree.')
         }
       },
-      { gracefulTimeoutMs: 10_000, forcedTimeoutMs: 10_000, requireGraceful }
+      // Windows CI occasionally needs more than 10s to flush the Electron/SQLite shutdown path
+      // after a session restart. Keep the forced-close budget bounded, but avoid classifying a
+      // successful graceful shutdown as a test failure solely because of Windows teardown latency.
+      {
+        gracefulTimeoutMs: process.platform === 'win32' ? 30_000 : 10_000,
+        forcedTimeoutMs: 10_000,
+        requireGraceful
+      }
     )
   }
 }
--- a/packages/open-science/CLI.md
+++ b/packages/open-science/CLI.md
@@ -187,6 +187,15 @@ Repeat the same absolute invocation with `cli install --json`. Respect the retur
 Windows may require a new terminal and Unix shells may need the existing launcher directory on PATH.
 The launcher uses the existing ownership receipts and does not claim unrelated executables.
 
+## Agent runtimes
+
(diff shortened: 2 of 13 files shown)


ewen-poch:
Request changes

The PR currently fails the required Node typecheck at src/main/web-service/task-api.test.ts:274. commandsFrom() infers its settings parameter from the minimal taskSettings fixture, so the richer settings object used by the new runtime projection test is rejected by TypeScript. I reproduced this from the PR head and fixed it in commit e6bbdf62 by widening the test fixture parameter type; the affected task-api, CLI, and SDK tests then pass (155 tests).

There is also a scope/issue-link mismatch. Issue #1987 describes an agent-operable first-run bootstrap that can reach a runnable first task. This PR adds runtime listing and local runtime installation, but it does not provide provider setup, connector/OpenAlex setup, skill enablement, PATH launcher repair, or onboarding finalization. Those flows remain blocked behind Settings, so the issue’s stated end-to-end outcome is still not achieved.

Please either remove Closes #1987 and track this as the runtime slice of the bootstrap work, or extend the implementation and acceptance criteria to cover the remaining initialization steps.

PR #2493, opened 2026-09-12
fix(ci): pack Windows full-test dependencies through bash

ewen-poch:
## Problem

Scheduled Windows Full Test has failed every run since #2482 (`c2e5ee552`). The last four batches never reached the shards:

```
Prepare Windows test dependencies / Upload dependencies
No files were found with the provided path: D:\a\_temp/windows-dependencies/
```

Pack logged `E2E setup pack-dependencies: 687.8 MiB` then upload-artifact looked at `${{ runner.temp }}`. The pack step runs on `windows-latest`, which defaults to pwsh, so `"$RUNNER_TEMP/windows-dependencies"` expands to `/windows-dependencies` instead of `D:\a\_temp/windows-dependencies`. PR Gate's Windows E2E setup already sets `shell: bash` for the same `$RUNNER_TEMP` pattern; this workflow copied the commands without the shell.

The last batch that still ran shards (https://github.com/aipoch/open-science/actions/runs/34686212713, before the pack job) also failed:

- `managed-preview-protocol.test.ts` — in-place overwrite on NTFS kept file ID and timestamp quantum, so the open-handle fstat did not reject
- `literature/catalog.test.ts` — 60s timeout on `accept-candidate` after deleting the source project (not changed here; it has not run since #2482)

## Proposed change

- Set `shell: bash` on pack and both restore steps, matching PR Gate Windows E2E.
- Guard those steps in the workflow topology tests.
- Mutate preview stream size so Windows fstat detects replacement.

## Scope and non-goals

CI pack/restore path and one preview protocol test. No production preview or catalog change. Catalog timeout left for a follow-up if it recurs after shards run again.

After the last material edit, from `.worktree/windows-full-test-runner-temp`:

| --- | --- | --- |
| Workflow topology | `npx vitest run scripts/ci/release-workflows.test.ts scripts/windows-release-workflows.test.ts scripts/ci/e2e-execution-ownership.test.ts src/main/managed-preview-protocol.test.ts` | 61 passed |

Windows Full Test `regressions` dry-run will be dispatched on this branch (reuses pack + restore + the existing subset). Full scheduled coverage remains authoritative.

## Historical compatibility

None.

## New state / enum values

None.

## Data persistence

None.

## UI / interaction

None.

Refs:
- https://github.com/aipoch/open-science/actions/runs/34718660753/job/103620347799
- https://github.com/aipoch/open-science/actions/runs/34686212713
- https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

### DIFF
--- a/.github/workflows/windows-full-test.yml
+++ b/.github/workflows/windows-full-test.yml
@@ -80,6 +80,9 @@ jobs:
       - name: Install dependencies
         run: node scripts/ci/npm-ci.mjs
       - name: Pack dependencies
+        # Windows defaults to pwsh, which does not expand $RUNNER_TEMP. Pack into the same
+        # runner.temp directory that upload-artifact reads, matching PR Gate's Windows E2E setup.
+        shell: bash
         run: node scripts/ci/e2e-setup-snapshot.mjs pack-dependencies "$RUNNER_TEMP/windows-dependencies"
       - name: Upload dependencies
         id: upload
@@ -121,6 +124,7 @@ jobs:
           merge-multiple: true
 
       - name: Restore dependencies
+        shell: bash
         run: node scripts/ci/e2e-setup-snapshot.mjs restore-dependencies "$RUNNER_TEMP/windows-dependencies"
 
       - name: Test complete suite shard
@@ -243,6 +247,7 @@ jobs:
           path: ${{ runner.temp }}/windows-dependencies/
           merge-multiple: true
       - name: Restore dependencies
+        shell: bash
         run: node scripts/ci/e2e-setup-snapshot.mjs restore-dependencies "$RUNNER_TEMP/windows-dependencies"
       - name: Build Electron application
         run: npm run build:e2e
--- a/scripts/ci/release-workflows.test.ts
+++ b/scripts/ci/release-workflows.test.ts
@@ -10,6 +10,7 @@ type Step = {
   if?: string
   name?: string
   run?: string
+  shell?: string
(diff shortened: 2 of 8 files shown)


ewen-poch:
Follow-up for the macOS journeys flake on `e47befb92`: `deletes a project through confirmation and keeps it absent after relaunch` hit `No handler registered for 'notifications:peek-pending-open-session'` during `app.restart()`, then retried green.

The startup window is created before `registerIpcHandlers` (same reason network and database-startup IPC are pre-registered). After relaunch the renderer peeks pending notification sessions as soon as App mounts, which can beat adapter installation. Pushed `319f8f5cf`: treat that missing handler as not-ready and retry, matching `storage:get-status`.

ewen-poch:
Regressions dry-run https://github.com/aipoch/open-science/actions/runs/34723255731 got past pack/upload/restore (the original empty-artifact failure). Shard 1 then failed:

`network-sandbox-owner.test.ts` / `waits for a transient Windows directory lock after probe cleanup`
`wrapped.setExecutionActive is not a function`

#2467 removed that field from the Windows-only wrap mock. Restored it in `b548f1123` to match the sibling mocks. Dispatching `regressions` again on this SHA.

ewen-poch:
Addressed the Codex P2: a late peek rejection can no longer schedule a retry after hook cleanup. `ab381d759` bumps a retry epoch in the effect cleanup and ignores stale timers; App.test covers unmount.

PR #2500, opened 2026-09-13
feat(literature): add local PDF extraction and table exports

ewen-poch:
PDF previews previously exposed only the original document. This adds an optional local figures-and-tables workflow: install verified model resources, analyze pages in an isolated Node process, and reopen cached extraction results without starting over.

## Changes
- Add model installation, integrity checks, cancellation, download telemetry, cache management, and application command integration.
- Reconstruct figure crops, captions, cross-page content, table groups, merged headers, source text, and footnotes using shared geometry and evidence rules. Preserve unresolved text and review warnings instead of silently exporting it.
- Add the figures-and-tables preview, image copy/download, compact progress with a scan animation and remaining-time estimate, research table styling, and a format selector with copy/download actions for HTML, TSV, and Markdown.

The branch is organized into three commits: model and preview infrastructure; extraction repairs and performance; progress and export UI. 124 regression fixtures use compact JSONL records (one field or source-array item per line), reducing fixture lines from 202,721 to 38,347 (81% fewer). A shared test loader reconstructs the original nested data. Local PDF audit reports, temporary files, and planning Markdown are excluded.

## Validation
- JSONL conversion: all 124 reconstructed fixtures matched the original committed JSON by deep equality; 101 PDF regression test files / 1,115 tests passed after conversion. Node/sandbox type checks, ESLint, and formatting checks passed again.
- 118 changed test files: 1,498 tests passed.
- Translation guards and mainline IPC registry regression checks: 815 tests passed.
- Full Node, sandbox, and renderer type checks passed.
- ESLint and Prettier checks passed for all changed files they cover; generated web API mapping and git whitespace checks passed.
- Browser verification covered narrow layouts at 320, 375, 414, and 768 px, dark-mode tables, reduced-motion scanning, and format-selection/download requests.

Extraction remains a candidate workflow requiring comparison with the source PDF. Existing unsupported-page and unassigned-text states remain visible. Full desktop end-to-end and packaging suites were not run for this PR preparation.


ewen-poch:
Addressed the production table diagnostics finding in `fbd082b`: raw grids and unassigned text are now logged only outside production mode. The worker stdout limit remains intact. Re-running page 6 of `36701846.pdf` reduced production stdout from 1,279 bytes to zero, with identical extracted content, geometry, and detection scores. All 39 related tests, ESLint, and formatting checks passed.

The total-page budget finding is valid. It changes behavior for very long PDFs, so implementation is awaiting the requested product confirmation. The proposed change is to apply the same 500-page bound to analysis and cache restoration, using the existing incomplete/limited result rather than adding persistent state.

The static-check failure is a Node heap exhaustion, not a TypeScript diagnostic: CI reached its approximately 4 GiB heap limit. The same Node 22 typecheck passed with `--max-old-space-size=6144` (reported TypeScript memory: approximately 4.1 GiB). Updating `typecheck:node` in `package.json` would extend the current PR file scope and is awaiting confirmation. The separate CI Integrity failure concerns the existing change to protected `scripts/ci/change-impact.json`; no guard has been bypassed or weakened.

ewen-poch:
Updated `typecheck:node` to invoke TypeScript with a 6 GiB heap in `24523c0a6`. This addresses the static-check process abort at the previous approximately 4 GiB limit without skipping checks. The exact npm command, including the sandbox typecheck, passed locally on Node 22; 39 CI workflow tests and formatting checks also passed.

Added a “Model source” link to the local parsing model card in `5ea28f7`. It opens the Microsoft Table Transformer repository through the same external-link component used by Agent settings. The link remains informational and does not alter installation, parsing, or persisted data. All eight translated locales are included; 810 settings/i18n tests and the renderer typecheck passed.

The total-page budget proposal remains pending product confirmation. The separate protected CI configuration finding has not been bypassed.

Remote verification: [Static checks](https://github.com/aipoch/open-science/actions/runs/34739860332/job/103677799669) passed on `5ea28f7d3274871cafa5cf310a54fce6fbeec3ca`, including Node/Web typechecks, lint, formatting, CLI/SDK checks, i18n, and runtime-bundle verification.

ewen-poch:
Fixed the initial-install concurrency finding in `50f961a`. When an initial model installation is already active, acquiring the model now returns the existing retryable “not installed” code. The existing PDF install-and-retry path joins the in-flight download, waits for readiness, and retries the page. Removal and same-revision repair retain their existing exclusion behavior; no new state, persistence, or IPC contract is introduced.

The regression test failed before the fix with “Local model management is in progress” and now passes. Coverage includes joining an existing install without duplicate downloads, acquiring the completed model, and continuing PDF extraction from both `notInstalled` and `installing`. All 181 selected tests, full Node 22 typechecks, ESLint, and formatting checks passed.

CI diagnosis:
- Static checks passed on the previous head after the heap adjustment.
- Windows E2E shard 1 reported a flaky `session-package` test: its first execution timed out during Electron graceful restart, while the automatic retry passed. The attached main-process log ends at a 2.5-second `renderer-session-preflight` timeout, before PDF/model owner teardown. This evidence does not establish a PDF regression. The new head will rerun the unchanged test; no timeout or flake policy has been weakened.
- CI Integrity rejects the protected `scripts/ci/change-impact.json` change registering the PDF process engine for Windows coverage. Removing that registration drops all Windows lanes for engine-only changes and violates the existing coverage guard. This requires explicit maintainer handling of the protected configuration; the guard and coverage registration remain intact.

The previously discussed total-page cap remains pending product confirmation.

ewen-poch:
Partially adopted the interrupted-publication recommendation: added four recovery regression cases in `111e76c`, without changing the publication protocol.

The stated Windows failure premise is incorrect for ordinary existing files. Node's `fs.rename` overwrites an existing destination file, and libuv's Windows implementation explicitly calls `MoveFileExW(..., MOVEFILE_REPLACE_EXISTING)`: [Node documentation](https://nodejs.org/docs/latest-v22.x/api/fs.html#fsrenameoldpath-newpath-callback), [libuv Windows implementation](https://github.com/libuv/libuv/blob/v1.51.0/src/win/fs.c#L2109-L2115). This code renames individual validated regular files, not directories. Merely having a destination file left by an interrupted attempt does not make retries permanently fail. File locks or permissions can still cause errors, but a delete/backup step would not establish a fix for those separate conditions.

The new cases recreate interruption after one asset has moved, and after all assets have moved but before `active.json` is committed, for both initial installation and upgrade. They exercise real filesystem renames, verify the complete destination and receipt, preserve the previous revision during upgrades, and acquire the installed model after another owner restart. All 29 owner tests, Node 22 typechecks (including sandbox), ESLint and formatting checks passed locally on macOS. This is not a claim that these new cases have already run natively on Windows.

A broader replacement/recovery protocol is therefore not justified by this finding; runtime behavior, persistence format, and UI are unchanged.

PR #2502, opened 2026-09-13
fix(session-persistence): roll back failed first-save allocations

ewen-poch:
## Problem

A first Session save can allocate its SQLite projection and pending save before JSON publication fails. Later catalog recovery sees no JSON and turns that allocation into a deletion tombstone. Retrying the original, never-published draft then fails with `Cannot save a deleted Session`.

## Proposed change

Return an internal allocation receipt from `prepareSave`. On failed publication, compensate only a newly created projection whose primary JSON was absent at admission and whose directory contains no primary, temporary, backup or quarantine evidence. Remove the unchanged allocation and pending save in one SQLite transaction. Preserve tombstones, published child facts, independently recorded auxiliary usage and the number sequence; retain both errors if compensation fails. JSON-success/projection-failure recovery is unchanged.

The public Session Repository/IPC return contract remains unchanged. The three projection test consumers unwrap the internal receipt.

## Scope and non-goals

- Persistence behavior changes; no database columns, JSON fields, migrations, business enum values, UI or provider protocol changes. The receipt is transient.
- No bulk repair of historical tombstones. No generic transaction framework or append-log migration.
- Crash before compensation, failed compensation, and temp-only pending replay remain separate recovery concerns. A probe found that existing pending replay can tombstone a temp-only first save; this patch retains that file/pending evidence at the failed-save boundary and does not claim to fix the subsequent replay.

The real-filesystem/real-SQLite regression failed repeatedly on unchanged production code, including immediately before implementation, at the retry promise with the reported deleted-Session error. It now passes. New controls cover trusted writes, existing authority, temporary retention, errors after rename, projection commit failures, failed compensation, tombstones and published facts.

An additional real-SQLite regression recorded auxiliary usage through the existing rename boundary before failing first publication. Before the guard, it failed twice because the usage projection became empty. It now verifies that usage stays visible, the owner and number remain intact, and the draft retries successfully.

Local checks after the last material edit:

| --- | --- | --- |
| Repository, projection, durable JSON and usage contracts | `npm test -- src/main/session-persistence/projection.test.ts src/main/session-persistence/repository.test.ts src/main/session-persistence/repository-queue.test.ts src/main/session-persistence/auxiliary-turn-usage.test.ts src/main/session-persistence/usage-regressions.test.ts src/main/storage/durable-json-file.test.ts` | 174 passed |
| Main ownership, hydration, deletion and IPC | `npm run test:module -- session_persistence` | 653 passed |
| Project deletion, renderer flush and merge consumers | `npm test -- src/main/projects/deletion-coordinator.test.ts src/main/session-persistence/deletion-integration.test.ts src/main/session-persistence/renderer-flush.test.ts src/renderer/src/lib/session-persistence/session-persistence.test.ts src/renderer/src/stores/session-store.test.ts` | 434 passed |
| Main types | `npm run typecheck:node` | Passed, including sandbox package |
| Style | `npm run lint`; Prettier and `git diff --check` | Passed |

CodeGraph traces the root-checkout Repository/projection relationship; worktree references confirm Repository is the sole production `prepareSave` consumer. Its private receipt remains inside this boundary. The affected-plan classifier requests full CI because changed projection files/test consumers lack module ownership entries (the exact-head plan first flags auxiliary-turn-usage.test.ts). Per maintainer instruction, no full local suite was run; exact-head CI and independent review remain required. Tests use portable temporary paths but do not simulate OS power loss or exercise live providers/multiple windows. No renderer/interface change requires a new UI screenshot.

## Review focus

Please check the evidence required before compensating and the conditional transaction's non-resurrection guarantees. In particular, missing authority after a crash is not treated as proof of first publication.

Prior art: VS Code's [chat write confirmation](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/common/model/objectMutationLog.ts) distinguishes prepared from confirmed state for retries. Its storage model differs; this patch does not adopt its log format.

### DIFF
--- a/src/main/session-persistence/auxiliary-turn-usage.test.ts
+++ b/src/main/session-persistence/auxiliary-turn-usage.test.ts
@@ -34,7 +34,7 @@ beforeEach(async () => {
   await migrateApplicationDatabase(client)
   await client.project.create({ data: { id: 'project', name: 'Project' } })
   projection = new SessionProjectionRepository(async () => client)
-  const session = await projection.prepareSave({
+  const { session: session } = await projection.prepareSave({
     id: 'session',
     projectId: 'project',
     title: 'Session',
--- a/src/main/session-persistence/projection.test.ts
+++ b/src/main/session-persistence/projection.test.ts
@@ -1,4 +1,4 @@
-import { lstat, mkdtemp, readFile, rename, rm, truncate, writeFile } from 'node:fs/promises'
+import { lstat, mkdir, mkdtemp, readFile, rename, rm, truncate, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 
@@ -398,8 +398,8 @@ describe('Session projection', () => {
     })
     const repository = new SessionProjectionRepository(async () => client!)
 
-    const first = await repository.prepareSave(session('session-1', 100))
-    const second = await repository.prepareSave(session('session-2', 200))
+    const { session: first } = await repository.prepareSave(session('session-1', 100))
+    const { session: second } = await repository.prepareSave(session('session-2', 200))
     expect([first.number, second.number]).toEqual([1, 2])
 
     await repository.commitSave(first)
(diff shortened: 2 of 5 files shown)


ewen-poch:
Addressed the auxiliary-usage ownership finding. The new regression uses the existing Repository rename boundary and the public auxiliary-usage recorder with real SQLite. It failed twice before the fix: usage was visible before the publication failure and disappeared afterward.

The rollback now checks for auxiliary usage inside its existing transaction and refuses to discard the owner when such evidence exists. The original Session number and pending save are preserved; the test also verifies a successful same-draft retry without losing or duplicating usage. This uses the existing compensation-error behavior and changes no schema, persisted fields, states, relationships or UI. Crash/pending-replay recovery remains outside this bounded fix.

Validation after the last material edit: storage/usage contracts 174 passed; session_persistence module 653 passed; deletion/renderer-flush consumers 434 passed; main typecheck, lint, formatting and diff checks passed. Full and platform validation remains with exact-head CI.

PR #2508, opened 2026-09-13
fix(session-persistence): retain unresolved temporary authority

ewen-poch:
## Problem

After a first Session publication and its temporary-file cleanup both fail, a live creator PID keeps that temp ineligible for recovery. Pending replay treated the missing primary as deletion and permanently rejected the original ID. A subsequent projection rebuild could also discard the unresolved allocation when its catalog callback omitted diagnostics.

## Proposed change

Include recognized temporary-file primary identities in Session scans. Report a missing primary with unresolved evidence as unreadable/incomplete, retaining the file, Session owner and pending save. Independently prevent rebuilding an initialized projection while pending recovery remains unresolved. Existing per-Session reads now report temp-only recovery barriers without duplicate directory warnings. Explicit deletion also removes recognized temps for its valid primary before removing that primary, leaving deletion retryable if cleanup fails; this closes a pre-existing residue recovery gap.

Keep the durable reader's live-writer exclusion and valid-primary precedence. Revisioned saves remain blocked while authority is unreadable; once the writer exits and existing recovery restores readable JSON, saving can continue against the recovered revision.

## Scope and non-goals

- Only Session Repository code and its Repository/projection tests change. No schema, persisted fields, enum values, relationships, migration, shared file-writer changes or new UI surfaces/copy.
- Existing supported legacy temporary filenames remain readable. Existing incomplete-recovery diagnostics now cover this previously misclassified case.
- No historical tombstone resurrection, immediate retry of unreadable authority, or guarantee of recovery from corrupt files, ongoing disk failure, PID reuse or OS power loss.

The final retention regressions failed twice on unchanged production code: Session rows became deletion tombstones, pending evidence disappeared, and direct diagnostics incorrectly returned missing. The same cases pass after the fix. An additional real Node child holds a PID/UUID temporary file while alive, then exits; storage owners reopen SQLite, recover the file and successfully save using the recovered revision. No production test seam was added. Review regressions additionally verify that invalid current PID/UUID suffixes are ignored while supported legacy numeric-only suffixes remain recoverable; the invalid PID cases failed twice before correction. A real child-process deletion regression also failed on both the PR and pre-PR base; after cleanup ordering was corrected, deleted JSON stays absent and cleanup failures retain the primary.

Checks after the last material edit:

| Behavior | Check | Result |
| --- | --- | --- |
| Pending/catalog retention, revision guard, exited-writer recovery, legacy/oversized/quarantine and true-absence controls | `npm test -- src/main/session-persistence/projection.test.ts src/main/session-persistence/repository.test.ts src/main/storage/durable-json-file.test.ts` | 176 passed |
| Persistence owner and representative coordinator/IPC consumers | `npm run test:module -- session_persistence` | 653 passed |
| Deletion, queue serialization, replay and storage normalization consumers | `npm test -- src/main/session-persistence/deletion-integration.test.ts src/main/session-persistence/repository-queue.test.ts src/main/session-persistence/claude-replay.test.ts src/main/storage/normalize-legacy-paths.test.ts src/main/projects/deletion-coordinator.test.ts` | 85 passed |
| Main types and style | `npm run typecheck:node`; `npm run lint`; Prettier; `git diff --check` | Passed |

CodeGraph identified direct reads, catalog rebuilding and identity lookup as shared Repository consumers; current-worktree source confirms the paths. Tests use portable paths and an actual exited child process. The exact-head impact classifier requests full CI because repository.test.ts has no declared module owner; its fail-closed routing is unchanged. Full portable/platform validation remains with exact-head CI per maintainer instruction. No renderer layout or controls changed; live multiwindow/provider and power-loss behavior were not exercised locally.

## Review focus

Check that retained evidence cannot authorize deletion or rebuild, while known deletion, valid primary precedence and revision admission stay protected. This is a follow-up to #2502; that PR's immediate no-file compensation remains unchanged.


ewen-poch:
Addressed the PID/UUID recognition mismatch. The new cases failed twice before the fix: zero and unsafe-integer PIDs with UUID suffixes were incorrectly classified as unreadable Session authority. They now report true absence, matching the durable reader; recognized live writers remain protected.

One compatibility distinction matters: the existing reader deliberately recognizes legacy numeric-only suffixes, including `.0.tmp` and numeric overflow, without interpreting them as a live PID. Those formats remain supported. Two additional controls verify that their valid JSON is still recovered. No shared-reader, schema or state changes were made.

Post-edit validation: storage/diagnostic contracts 174 passed; session_persistence module 653 passed; direct consumers 85 passed; main typecheck, lint, formatting and diff checks passed. Exact-head CI remains the full/platform authority.

ewen-poch:
Confirmed the retained-temp deletion problem through public deletion and filesystem recovery with a real child writer. The same regression also fails against the pre-PR base `c5bec5ea`, so the resurrection predates this scan change; the existing durable reader already promoted the residue after the writer exited.

The fix keeps the existing primary-first ownership proof: when a valid primary authorizes explicit deletion, remove recognized temporary files for that exact primary before removing the primary itself, just as the existing path removes matching backups. A failed temporary cleanup leaves the primary intact and deletion retryable. Shared recovery still excludes live writers; only explicit deletion removes its owned candidates. Unknown filenames and other Session temporary files are preserved.

Both regressions failed before the fix and now pass: deleted JSON must not reappear after writer exit, and temporary-cleanup failure must preserve the primary. The filename recognizer is shared privately by scan and deletion so PID/UUID validation and legacy compatibility remain consistent. No new files outside the PR scope, schema, persisted fields or states were added.

Final post-edit checks: storage/projection contracts 176 passed; session_persistence module 653 passed; direct consumers 85 passed; main types, lint, formatting and diff checks passed. Full/platform CI remains authoritative for the new head.

PR #2518, opened 2026-09-13
fix(acp): preserve cancellation and quarantine unreaped delegates

ewen-poch:
## Problem

Stopping a prompt during forced Skill reload can still dispatch it after reconnect, and provider cancellation can become successful completion. Unconfirmed delegated process cleanup must also remain visible to Session/Project deletion and application teardown after runtime composition is reconstructed.

## Proposed change

- Preserve request-scoped cancellation across reload without cancelling a successor, and preserve provider cancellation using existing cancellation reasons.
- Persist a generation ownership receipt before physical provider launch. Share ownership checks across launch, workspace reuse, deletion, quit and update; retain files, capacity and port ownership until process cleanup is proven. Recover at existing lifecycle boundaries.
- Release terminal executions' model bridge/transport references even when process exit is unconfirmed. Existing reference counting retains transports for active siblings; process-owned resources remain protected separately.
- On Windows, create providers atomically inside an exclusive current-user Job with kill-on-close and Node/libuv pipes. Preserve live POSIX process-tree proof and block ambiguous recovery after a crash.
- Forward OpenCode's existing cleanup-recovery hook so proven cleanup releases quarantined capacity and workspace/runtime-home exclusions.
- Retain focused native-process validation and register the new files and consumer tests in the current module inventory.
- Let queued renderer snapshots reach the existing Main transcript owner after CLI completion, avoiding a false legacy revision conflict that blocks conversation export. Preserve captured preferences and conversation commands; legacy transcript conflict checks remain in place.

Rebased onto main `caa8fbedc539d0c3defd272f3dd768cc05300a9b`. At the rebased head `7b26e6fa`, the production tree matched the previous reviewed PR combined with current main, including upstream artifact-publication authority and cancellation handling. The additional export fix below adapts one remaining renderer check to that existing Main ownership contract.

Follow current main's split registration format: keep the metadata-only `scripts/ci/module-impact.json`, move the 39 still-needed PR registrations into nine existing `scripts/ci/module-impact/*.json` files, and use upstream's shared manifest loader in architecture tests. The maintainer approved this file-scope adaptation. All other module fields and upstream registrations are unchanged. The old inline manifest fixes already present in main are dropped.

## Persistence, states and compatibility

Version-1 JSON receipts live under `<dataRoot>/delegation-process-ownership/<projectId>/<sessionId>/<receiptId>.json`, outside removable workspaces. They contain provenance, generation identity, diagnostic timestamps and platform ownership material; no credentials or prompt contents. Resource phases are `starting`, `owned` and `cleanup-pending`. No Task/Attempt status, cancellation-reason enum, database schema or historical-record migration is added.

The approved rollout protects newly launched executions only; it does not infer ownership of historical orphans. Older versions do not enforce receipts. Missing PIDs or markers are not whole-tree exit proof. A verified Linux boot-ID change can clear prior-boot ownership; ambiguous macOS receipts can remain blocked indefinitely. Uninstall does not sweep unproven processes. Existing UI/error surfaces are reused; no renderer layout or copy changes.

Reproduction evidence is retained in the PR discussion:

- Connection-release policy: four unreaped/stop/startup variants failed twice through public durable-delegation behavior before correction, while three completion/cancellation controls passed. All seven pass with the approved policy; files and capacity remain protected.
- OpenCode recovery: the public regression failed twice with capacity rejection after proven cleanup and recovery, then passed with forwarding, path reuse and idempotent disposal. [Evidence](https://github.com/aipoch/open-science/pull/2518#issuecomment-5714045743).
- Workflow isolation: the focused Windows dry-run regression failed twice before correction. [Evidence and passing Windows run](https://github.com/aipoch/open-science/pull/2518#issuecomment-5713544483).
- Restart/deletion coverage checks actual filesystem preservation after composition reconstruction with cleanup failure injected through the execution boundary. It does not claim to reproduce an escaped process. Upstream cancellation-after-preparation and basic live unreaped cleanup remain regression controls, not new defect claims.

Checks at rebased head `7b26e6fa`, before the additional export fix:

| --- | --- | --- |
| ACP cancellation, resume, Plan context and all four provider framework paths; delegated execution/recovery/ownership/composition | Targeted `npm test --` across 23 ACP/framework/delegation suites, including agent-process | 1,600 passed |
| Split manifest loading, ownership, consumers, impact planning, workflow classification, every architecture suite using the shared manifest loader, Electron crash fixture and process snapshots | Targeted `npm test --` across 22 suites | 381 passed |
| Process-tree termination and delegated artifact evidence | `npm test -- src/main/process-tree.test.ts src/main/delegation/delegated-artifact-evidence.test.ts` | 35 passed |
| Process types and style | `npm run typecheck`, `npm run lint`, targeted Prettier and `git diff --check` | Passed |
| Trusted-base admission | `node scripts/ci/check-ci-integrity.mjs --base caa8fbedc539d0c3defd272f3dd768cc05300a9b --head 7b26e6fa452d18fc9488d29b484d560e7b3f4c20` | Passed; no violations or ownership gaps |

The previous five architecture failures and missing notebook-sandbox consumer entries are fixed upstream. Before rebasing, eight relevant architecture/consumer suites passed all 66 tests on untouched latest main. No duplicate correction was added.

The prior Windows brand-storage crash-restart failure is also addressed by upstream #2744: the fixture verifies observed process-tree exit after taskkill races. This PR retains that implementation unchanged. Upstream documents stable reproduction, a deterministic regression and real Windows repetitions; the local fixture/process-snapshot checks pass, but new-head native and branding journeys still require current-head Windows CI. This local run is not a Windows E2E result.

The impact explain plan selects full CI because the PR contains global CI inputs. Per maintainer instruction, no full local suite was run. The rebase checks above ran after the final rebase edit; the export-fix checks below run against the subsequent change. CI remains authoritative for the complete portable suite, Windows/native and E2E lanes; previous-head green results are historical. This integration adds no new runtime architecture, model, relationship, state, persistence, migration or interaction change beyond the previously approved PR design. No bypass or merge is requested.

### CLI export regression

The Windows workspace journey failed with `Session revision conflict: expected 3, actual 5` before export. A queued partial renderer reply was evaluated against Main's completed Task transcript by a legacy merge check, even though Main now accepts only captured user commands/preferences and owns the transcript.

The maintainer approved expanding the file scope to only `session-persistence.ts` and `session-persistence.test.ts`. One condition keeps the legacy merge check for renderer-owned transcripts and lets Main-owned snapshots reach the existing authoritative save path. No new fields, states, storage format, migration, historical compatibility requirement, or UI flow is introduced.

Two deterministic public-boundary cases (with and without a concurrent rename) both failed twice on the unfixed implementation with the exact CI conflict. The regression uses the actual store, ordered save queue, flush boundary and Main persistence owner; no new test seam or delay is needed. It verifies complete reply preservation, the rename, and no extra disk write/revision for a passive snapshot.

Checks after the last material edit (`089a2b5b`):

| --- | --- | --- |
| Queued saves, revision conflicts, Main transcript ownership, Branch handling, export/quit flush consumers, framework event handling | `npm run test:module -- renderer_session_persistence` | 194 files / 5,421 tests passed |
| Incremental impact selection | `npm run test:affected:explain -- --base HEAD^ --head HEAD` | Selects the same module and registered consumers; no new test seam or registration |
| Node, sandbox and renderer types; source style | `npm run typecheck`, `npm run lint`, targeted Prettier, `git diff --check` | Passed |

No full local suite was run. The two regression cases failed on the unfixed implementation twice, then passed with the fix. Windows execution remains subject to exact-head CI; the earlier flaky attempt is not considered a pass. Previous rebase results remain historical, since this commit only changes the two approved renderer persistence files.

## Review focus

Review ownership proof before destructive cleanup, failure propagation, and the separation between connection references and process-owned resources. The previously discussed preconnected Windows pipe concern has a response citing version-pinned libuv and real Windows results. An AI workflow's successful execution alone does not establish its text verdict.


ewen-poch:
**Cleanup failure propagation and workspace deletion**

Addressed within the live execution owner's lifetime in `9581e7c5`. Cleanup errors survive terminal-history writes and propagate through Stop/all-work shutdown to the coordinator's quit/update gates. Session/Project deletion stops before its independent workspace-removal pass if live cleanup fails, and retains the failed owner for retries. Six public production-composition/coordinator cases failed twice before the fix, then passed: they observe temporary evidence-file preservation and reject the previous false `reaped: true` outcome. Final local validation passed 3,525 tests across 155 files, node/sandbox typechecks, lint, formatting, and the scoped Electron Stop E2E.

**Cancellation before interaction admission**

The reported provider dispatch has not been reproduced on `9581e7c5`. I held plan preflight and Skill authorization separately, accepted cancellation through the real interaction owner and captured workflow callback, then released preparation and ran the real provider executor. Both cases returned `stopReason: 'cancelled'` without calling `provider.prompt`, on two repeated runs. The pending reservation carries its accepted-cancellation state into the active interaction. This evidence covers those composed boundaries; it does not establish every external-provider/disconnection scenario. I have not added a speculative production guard. The characterization patches and results remain local and are not presented as failing regression tests in this PR.

**Quarantine across restart**

Confirmed at the reconstruction boundary: after a cleanup error is written to durable Session history, rebuilding the production composition loses the in-memory quarantine. Session and Project deletion then remove a real temporary evidence file. Both cases fail consistently in two runs. This reconstructs the owner from durable data; it does not create an actual orphan process or simulate an OS-level process restart.

A complete fix requires durable process/resource ownership and cleanup evidence, startup recovery, and an explicit policy for historical executions that lack trustworthy ownership records. That expands the approved live-owner design and adds persistence/state beyond this PR's current scope. Implementation is paused pending maintainer confirmation. The current change does not claim cross-restart safety, and the remaining finding is not considered resolved merely because the local tests or CI pass.

ewen-poch:
**Proposed durable ownership scope — awaiting maintainer approval**

The persistence fix needs a recovery contract, not only a boolean quarantine marker. The proposed design keeps a versioned, delegation-owned receipt outside the removable workspace subtree for each provider process-tree generation. It records launch intent before spawn, then verified platform ownership; unconfirmed cleanup remains durable. Workspace reuse, Session/Project deletion, and quit/update gates read the same owner. Task statuses and cancellation reasons remain unchanged; the new resource-only phases would be `starting`, `owned`, and `cleanup-pending`.

Recovery must prove whole-tree cleanup. A root PID disappearing is insufficient. Existing POSIX ownership/immutable-identity primitives can inform a narrow recovery boundary, with ambiguous identity or incomplete observation remaining blocked. Windows needs genuine Job containment and assignment before the process runs; Microsoft's `PROC_THREAD_ATTRIBUTE_JOB_LIST` guidance avoids the suspended-process assignment crash window: https://devblogs.microsoft.com/oldnewthing/20230209-00/?p=107812 . This requires launch/native-process files beyond the current diff and is not being implemented without approval.

Historical executions lack trustworthy receipts. The bounded rollout proposal protects new executions without fabricating historical ownership; conservative protection of old workspaces would require a separate inventory/recovery policy and could block existing deletion/reuse. That compatibility choice also awaits maintainer confirmation. There is no approved schema, state, native-launch or interaction implementation yet, and this proposal does not resolve the finding by itself.

ewen-poch:
The maintainer approved the delegation-owned receipt/recovery design and the new-executions-only historical policy. The two public reconstruction regressions were rerun twice on `9581e7c5`; both still fail because deletion resolves and the real evidence file is removed. These are composition reconstruction tests, not OS restart certification.

Physical launch tracing identified an omission in the proposed file scope: `AgentSpawnInput`, the four framework spawn implementations, and Claude’s `acp/agent-process.ts`. The frameworks own command rewriting, credential/environment isolation, shell handling, and Codex install admission. The owned platform spawner must receive the prepared launch at that boundary; wrapping a returned `ChildProcess` would be too late for atomic Windows Job assignment. The proposed addition is an optional internal spawn callback, with existing preparation and default callers preserved. It adds no further persisted fields, states, or UI changes beyond the approved design.

Implementation is paused for the maintainer’s required approval of these additional source files. The failing regressions remain local, and this review finding remains unresolved in the PR.

ewen-poch:
Addressed cross-restart workspace deletion in 865b4749. The two public reconstruction regressions failed twice on 9581e7c5 by deleting a real evidence file, and now pass. One production ownership owner records each process generation before physical launch and gates workspace reuse, Session/Project removal and quit/update cleanup. Failed teardown retains files, capacity and backend claims; proven recovery releases them.

The maintainer approved versioned JSON receipts and the resource phases `starting`, `owned`, `cleanup-pending`, with historical Policy A: protect new launches only, no backfill or database migration. Every framework keeps its existing command/environment preparation and injects ownership only at physical spawn; four launch-admission regressions were red before that wiring and green after it. Windows uses atomic native Job-list creation and exact-Job recovery. POSIX receipts remain blocked after lost observation unless existing live proof or a verified Linux boot change proves cleanup; ambiguous macOS receipts can remain blocked indefinitely.

Final local evidence: 187 files, 4,732 passed, 0 failed, 31 platform/existing skips; main/sandbox types, lint, formatting, native macOS build and both scoped Electron journeys passed after the last material edit. Two Notebook process checks failed during an earlier concurrent run; both whole consumer files passed in isolation (418 tests), followed by a passing identical 187-file run without concurrent build/Electron work. The restart UI fixture is explicitly distinguished from real-process crash certification. A focused Windows process-ownership CI run is pending.

The ordinary pre-admission cancellation allegation remains unsubstantiated by the repeated composed plan/Skill cases, so no speculative production guard was added for it. The PR description now includes the final impact command and compatibility/platform limits; please review this exact head and its evidence mapping.

ewen-poch:
I read the final result from AI workflow 34760155633 for 865b4749; its Post Codex feedback job was skipped, so there is no published comment for that head. Responding to its two findings here:

- **Cancellation before normal admission:** the composed public workflow/interaction/provider-executor cases for plan preflight and Skill authorization pass again on 865b4749. Accepted cancellation returns `stopReason: cancelled` and `provider.prompt` is never called. The pending reservation preserves cancellation when promoted into the active interaction. The reported dispatch is still not reproduced, so no speculative production guard is added.
- **Named-pipe descriptors in net.Socket:** the stated premise is incorrect for Node 22. [Node's createHandle](https://github.com/nodejs/node/blob/v22.x/lib/net.js#L173-L187) detects `PIPE` and constructs `Pipe`; Socket opens the supplied descriptor through that handle. [libuv's Windows pipe implementation](https://github.com/libuv/libuv/blob/v1.x/src/win/pipe.c) supports the corresponding pipe descriptor boundary. The native launcher transfers overlapped pipe handles using the host libuv descriptor table. This source evidence rebuts the claim that Socket accepts only TCP descriptors; real Windows integration remains required.

The focused Windows run instead exposed a concrete build failure before tests: windows.h included legacy Winsock before uv.h pulled in Winsock 2, causing duplicate sockaddr/fd_set definitions. The minimal correction includes uv.h first. No architecture, persisted format, state or interface changes are needed. The same focused CI will validate compilation and then the actual Job/stream lifecycle.

ewen-poch:
Windows certification now passes on 90c42f6 in [focused run 34760763814](https://github.com/aipoch/open-science/actions/runs/34760763814): all 92 tests passed, including all four real-process tests for concurrent piped ACP IO/argv, descendants after root exit, application-owner crash with Job kill-on-close, and the cmd launch contract. The header-order fix resolved the actual compiler failure. The successful real-pipe tests also contradict the prior review's claim that net.Socket rejects these pipe descriptors.

A separate check remains blocked: CI Integrity reproduces `protected-gate-control-plane` for the PR Gate workflow edit. The trusted checker states that modifying this established control-plane file requires an explicit maintainer ruleset bypass. No bypass or policy weakening has been attempted. The maintainer must choose whether to retain the tested configuration with that explicit approval, or approve a separate automatically triggered/manual process-certification workflow and revert the PR Gate edits. The latter expands the current file list, so implementation is paused for that decision.

The fourth exact-head AI review is still running. Platform test success is not being represented as overall PR readiness.

ewen-poch:
Responding to [the review of 90c42f6](https://github.com/aipoch/open-science/pull/2518#issuecomment-5653735250):

- **Backend lease on ownership rejection:** confirmed through the existing public production framework/execution boundary. Two blocked preparations leave two outstanding provider transport leases on unchanged 90c42f6; the same test failed twice. Moving the existing ownership assertion before backend resolution fixes it without adding state, persistence or architecture. The local follow-up passes its regression and the 41-file delegation/framework/backend consumer set (916 passed, 7 existing skips). It will be pushed with the approved CI scope decision, so the current PR head must not yet be described as containing this fix.
- **ConnectNamedPipe OVERLAPPED:** [Microsoft's API contract](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-connectnamedpipe) does require a valid OVERLAPPED pointer for an overlapped pipe. This remains a documented contract concern. However, this launcher successfully opens the synchronous client handle before calling ConnectNamedPipe, and the same documentation describes the already-connected case as a valid connection. The actual Windows pipe/Job/crash/cmd tests passed on 90c42f6; the claimed disconnected launch has not been reproduced. Per the maintainer's regression-first requirement, no speculative native fix is being presented as a reproduced bug fix. This is not a claim that passing tests waive the documented API contract.

Four AI review revisions have now been examined. The remaining review allowance is reserved for the final diff after the maintainer decides how to resolve the protected PR Gate workflow edit; no ruleset bypass or control-plane policy change has been performed.

ewen-poch:
The backend-lease fix is now published in 0fface86. The public regression failed twice on unchanged 90c42f6: two ownership-blocked preparation retries retained two provider transport leases. Moving the existing ownership check before backend resolution eliminates acquisition on that path. Final follow-up validation passed 916 tests across 41 delegation/framework/backend consumer files (7 existing skips), node/sandbox typechecks, lint and formatting. No fields, phases, persistence, architecture or UI were added.

The maintainer explicitly confirmed retaining the tested PR Gate changes and permits a ruleset bypass **only for the protected-gate-control-plane policy finding**. Other checks remain required; no repository-wide rule was changed and no unrelated failed check is covered by that approval. The next completed AI review is the fifth and final revision under the requested cap.

ewen-poch:
Additional primary-source evidence for the ConnectNamedPipe finding: the exact Node version used by the passing Windows run, **Node v22.23.2**, vendors the same pipe-pair setup pattern in libuv. [uv__create_pipe_pair](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L321) creates the server with FILE_FLAG_OVERLAPPED, successfully opens the client, then calls ConnectNamedPipe(server_pipe, NULL), accepting ERROR_PIPE_CONNECTED. Its [standard-stream caller](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L368-L418) explicitly sets the server's UV_NONBLOCK_PIPE flag.

This matches the ordering here: both endpoints already exist, and failure to open the client returns before ConnectNamedPipe. It is not an outstanding asynchronous accept waiting for an unknown client. Together with the passing real Windows stream/Job tests, this gives a concrete upstream basis to retain the current paired-pipe setup. No reported disconnected launch has been reproduced, and I am not adding a different connection state machine solely from the generic asynchronous-accept warning. The backend lease finding was independently reproduced and fixed in 0fface86.

ewen-poch:
I inspected the final review output for 0fface86 in workflow 34761997518; the comment-publication job was skipped. Both findings repeat Windows startup claims contradicted by concrete upstream behavior and the successful process integration tests:

- **ConnectNamedPipe:** [Node v22.23.2's vendored libuv](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L321) uses this same already-created pipe-pair sequence, including the null OVERLAPPED argument and ERROR_PIPE_CONNECTED handling. The reported pending-client condition is not reproduced here.
- **net.Socket descriptors:** [Node v22.23.2 createHandle](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L173-L187) selects Pipe for PIPE descriptors; the [Socket constructor](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L399-L407) opens that handle. The assertion that descriptors always follow the TCP path is incorrect.

The current-head Windows core job also passed the delegated process ownership step, following the earlier 92-test focused run. These tests exercise real concurrent stdio, descendants after root exit, owner crash cleanup and cmd launch. No reported Windows startup failure has been reproduced, so no speculative transport rewrite is added. The confirmed backend lease leak is fixed in 0fface86 with a twice-failing regression followed by a passing 41-file consumer set.

The automated verdict remains needs changes; this comment records the evidence-based maintainer response rather than relabeling that verdict. Full PR Gate is still running. Only the separately authorized protected-gate-control-plane exception is eligible for bypass.

ewen-poch:
Added `src/main/delegation/process-ownership.ts` to the existing Windows-sensitive overlay so changes to this owner select Windows runtime/path checks and Windows functional/workspace consumers.

The CI source-coverage failure reproduced twice. The added case exercises the public `classifyChanges` boundary and also failed twice before the manifest fix, with `main_runtime` present but `windows_sensitive` missing. After the fix, all 200 tests in the classifier, PR Gate workflow, evaluator, module-shadow and integrity suites pass; full lint, focused formatting and diff checks pass. No runtime, data model, persistence, state or interaction changes are included in this follow-up.

The maintainer explicitly extended the per-PR `protected-gate-control-plane` exception to `scripts/ci/change-impact.json`, in addition to the existing PR Gate workflow exception. Other failures remain blocking; no bypass has been applied. Complete portable/platform verification is delegated to the new head's CI, as requested.

ewen-poch:
Rebased onto `86a379741db06dbef7d1809bc3bb175c73fd369d` as requested by the maintainer. All six commits replayed without conflicts; `git range-diff` reports identical patches. Published with an explicit lease against the prior remote head, preserving concurrent-update protection.

The Windows failure artifact shows the Session-package journey timed out waiting for the full visible reply after Send. The main log records roughly 18 seconds of pre-prompt work, then about 1 second to `end_turn`; the failure snapshot already contains a partial reply and “Response completed.” This supports assertion-budget pressure, but does not prove missing content or justify a speculative runtime change. After rebase, the exact Electron journey passed twice locally with retries disabled (26.6s and 24.2s). No timeout, retry policy or product code was changed.

Validation after rebase: Electron build, Main/Renderer/sandbox typechecks and lint passed. The 194-file owner/consumer set returned 5,092 passed, 2 failed and 31 skipped. Serial reruns of the two failing Notebook files returned 142 passed, 1 failed and 2 skipped: the R cancellation stress case passed, while the verification fixture's 1-second PID-readiness wait failed again. Its PID files appeared after 1.88, 1.19 and 1.24 seconds across three attempts. These Notebook production/test files are unchanged by the rebase; no unrelated fixture adjustment was included. This is partial local evidence, not an all-green verification claim. New-head CI must certify the portable/platform suites and determine whether the original Windows flake recurs.

ewen-poch:
Rebased onto `3f73d6008ebd2e604083118866b0a9be9b971438` without conflicts; the six existing patches remain identical under `git range-diff`. The maintainer approved extending this follow-up to `packages/notebook-network-sandbox/src/network-enforcement.integration.test.ts` only.

A controlled `EAI_AGAIN` on the second `example.com` DNS lookup reproduced the CI's identical disposal failure twice on unmodified code. The fixture's local parent proxy did not isolate DNS. After a DNS-dependent assertion failed, cleanup for an already-exited command was skipped, so disposal correctly retained the sandbox owner and masked the initial assertion. The original CI log cannot establish that DNS caused that particular run; this is a reproduced failure path, not an inferred production defect.

The live-policy fixture now pins its public DNS answer, keeps the real curl/gateway/policy/private-address checks, and restores the DNS spy afterward. Completed-command cleanup is asserted before policy assertions, preserving diagnostics and preventing owner leakage when an assertion fails. The fixture server closes even if disposal rejects. Production cleanup rules, data, state, persistence and UI are unchanged; no timeout or retry policy was relaxed.

The identical DNS-failure injection passes after the fix. After the final edit: six native/policy/lifecycle files passed 189 tests with one platform skip; the separate real REPL-through-production-sandbox CI check passed. Sandbox typecheck, full lint, focused Prettier and diff checks passed. Full portable/platform verification remains the new head's CI. The five-review iteration cap and the narrowly authorized protected-file exceptions remain unchanged; no bypass or merge has been performed.

ewen-poch:
Rebased onto `c66c61c0` and rechecked the bug evidence against the current upstream code.

- Upstream `b1e6ec8f` already fixes cancellation after framework preparation. The same public regression passes twice on unpatched latest main, so I removed this PR's redundant checkpoint and retained the test.
- Five cases still fail consistently on latest main in two runs: Stop during Skill reload at backend resolution/provider resume, provider cancellation reported as completion, and resource reuse after unreaped/throwing shutdown.
- The two restart deletion cases also fail twice with latest main plus live cleanup handling but before persistent ownership (`4994754d`): Session/Project deletion resolves and the actual evidence file becomes ENOENT. These tests inject cleanup failure at the execution boundary and reconstruct production composition; they do not create a real escaped process or establish that every persistence/native implementation detail is uniquely necessary.
- Conflict resolution preserves upstream OpenCode per-Attempt isolation and its disposal, upstream replay conditions, and all current workflow dispatch options. The previously approved ownership architecture, resource phases, persistence format, historical rollout and macOS ambiguity limitation are unchanged.

After the final edit, the explicit 363-file owner/consumer set passed **9,716 tests, 0 failed, 94 skipped**. Main/sandbox and renderer types, full lint, changed-file formatting and diff checks passed. No full local test suite was run. Current classification still selects expanded macOS and Windows runtime/E2E validation; the new head requires its own CI evidence.

The five-review iteration cap remains in effect. No merge or policy bypass has been performed; the existing protected-control-plane exception remains limited to the two approved files.

ewen-poch:
Registered the two new process-ownership implementations and their two tests in the existing `acp_runtime` module. This preserves the existing lifecycle consumers and Windows/E2E overlays without introducing another module or weakening the integrity guard.

Four public `createAffectedTestPlan` regressions failed twice before the manifest change: each new file fell back to an unowned full plan instead of resolving its owner, tests and Windows coverage. All four now pass. The six focused CI suites passed **212 tests, zero failures/skips**; focused lint, formatting and diff checks passed.

The exact committed head `1b2eb56b4d80eaec6f1222a92985f170a8ecbd3d` passes `check-ci-integrity.mjs` against `c66c61c0fb8235fdb740ab9092d41cc152e2d1ac`, with no violations. The prior historical ownership gaps remain nonblocking and were not expanded into this fix. This is test-selection metadata only: no runtime, architecture, data model, persistence, state or user interaction change. Current-head CI is running; no bypass or merge has been performed.

ewen-poch:
Integrated the intervening main commit `27ff9391` (#2698), which caused a merge conflict and prevented the new PR Gate from triggering. The conflict resolution preserves Specialist Skill preparation before spawn, passes prepared Skills into the runtime, and retains upstream cleanup ordering. The existing process-ownership guard remains before any resource disposal; the already-prepared spawn is wrapped once. No new design, state, persistence or interaction change is introduced. The other eight patches are unchanged under `git range-diff`.

Validation after this rebase: **2,475 passed, 0 failed, 4 skipped across 54 focused files**, including all delegation/framework tests, upstream Skill/Task/renderer consumers and the six CI modules. Main/sandbox and renderer typechecks, focused lint/formatting and diff checks pass. Six cancellation/cleanup regressions fail consistently twice on unpatched `27ff9391`; the preparation-cancellation control passes. The additional failing case is the existing recoverable-cleanup variant, which still releases resources before confirmation on main. All pass with this PR.

Exact head `ace4bf7dc125b3a97bf7bc84dd0a22914c7d2b1e` also passes the current CI Integrity checker against `27ff9391` without violations or bypass. The preceding metadata-only head passed remote CI Integrity. Current-head PR Gate must still complete; no merge or bypass has been performed.

ewen-poch:
The CodeBuddy construction-failure report is covered by a new production-boundary test using a real owned child process. It passed twice before any production change: forcing `createAcpRuntime` to throw still terminates the child, clears its ownership receipt, removes the Attempt runtime directory and makes capacity available again. The shared ACP disposer calls `confirmProcessCleanup()` even when runtime construction never assigns a runtime; production confirmation calls `ownership.recover(..., true)` and awaits tree termination. The adapter's immediate `kill()` is not the final cleanup boundary. I retained the test and did not change this already-working path.

For the Windows pipe finding, this function opens the client synchronously before `ConnectNamedPipe` and accepts `ERROR_PIPE_CONNECTED`. This matches the preconnected stdio-pair pattern in [Node v22.23.2's libuv](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L321), including the overlapped server selected by [its stdio caller](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L368-L418). The [Windows core job for ace4bf7d](https://github.com/aipoch/open-science/actions/runs/35169647518/job/105038519271) uses that Node version and passed all 92 process-ownership/composition/native integration checks, including actual piped IO and crash/descendant cleanup. This is evidence for the preconnected case, not a claim that a null OVERLAPPED is safe for a pending asynchronous connection. No speculative native change was made.

Separately, the macOS delegation failure was reproduced twice through the production execution boundary: a successful OpenCode run rejected with `EACCES` while unlinking `.catalog_stamp` inside a copied read-only Skill directory. Cleanup now uses the existing Skill materializer to clear only the Attempt-owned projection before removing the runtime directory. The regression checks successful completion, directory removal, and unchanged shared-source contents and permissions. Preparation-failure cleanup uses the same path. No new persisted fields, resource phases, historical migration, public interface or UI behavior is introduced.

The Settings interface guard also failed twice because upstream added `prepareDelegatedSkills` without adding it to the exact expected method set. The guard now lists the already-existing method; the production interface is unchanged.

ewen-poch:
The remaining named-pipe finding is the same preconnected-pair case addressed in [the evidence response](https://github.com/aipoch/open-science/pull/2518#issuecomment-5707269895). `CreateFileW` succeeds before `ConnectNamedPipe`, so this is not an asynchronous accept waiting for a client. The linked Node v22.23.2/libuv implementation uses the same overlapped-server, client-first, null-OVERLAPPED sequence and accepts `ERROR_PIPE_CONNECTED`.

There is no new failing reproduction in this review, and the [new head's Windows core job](https://github.com/aipoch/open-science/actions/runs/35172581035/job/105047268884) has now passed. I am retaining the implementation rather than adding an unverified native change. This confirms the native pipe/process checks on `cca245b5`, not just the prior revision. Remaining macOS/portable/E2E jobs are still monitored separately. The raw AI verdict remains `needs changes`; this response does not override or bypass any required check.

ewen-poch:
Integrated main `e12a6d8a` in merge commit `d091f803159656673cf79d63c32465dbb275ea7e`, resolving the four conflicting files without rewriting published history.

The maintainer-approved cleanup boundary is now explicit: terminal cleanup releases its backend/transport reference even when process exit is unconfirmed. Shared transports remain reference-counted; files, capacity, process receipts and potentially occupied ports remain protected until ownership recovery proves cleanup. The redundant retained-backend-claim map is removed. The existing PRD ownership paragraph reflects this distinction; no new enum, persisted format, migration or UI interaction is introduced.

Upstream's read-only OpenCode cleanup, including symlink protection, is retained in place of this branch's duplicate materializer-based cleanup. Its running-reservation and asynchronous-disposal guards are preserved. Module registration now uses upstream's exact owner inventory and includes the reachable consumer tests identified by its new guard.

Before changing production behavior, the public durable-delegation test ran twice: four unreaped/stop/startup variants failed because their transport claim was not released; three ordinary completion/cancellation controls passed. All seven now pass with file and capacity protection still asserted.

Validation after the source resolution:
- Focused cleanup, production composition, backend reference counting and test-selection checks: 237 passed.
- Eight CI contract/ownership suites plus the newly selected ACP consumer: 321 passed. Complete ownership audit: 4,605/4,605 files. Exact-commit CI Integrity against the integrated main: passed, no violations or gaps.
- Main, sandbox and renderer typechecks, lint, changed-file formatting and diff checks: passed.
- Expanded `main_delegation` module run: 10,365 passed, 20 failed, 142 skipped across 411 files, with one cascading unhandled storage error. Eighteen failures used a stale locally linked native file-cleanup addon predating main's API changes. Rebuilding that addon only inside this worktree with cached headers and rerunning the three affected files passed all 111 tests without errors. The two Notebook-import cases timed out in the broad run and both passed unchanged in isolation. The final manifest's additional ACP consumer also passed above. This is focused rerun evidence, not a claim that the first expanded run was green.

No full local test suite was run. Current-head CI must validate the merged revision; the earlier green CI is historical. No PR merge, bypass or workflow rerun has been performed.

ewen-poch:
This is the previously discussed preconnected stdio-pair finding. `windows_owned_process.cc` is unchanged by this integration. The synchronous `CreateFileW` succeeds before `ConnectNamedPipe`, and the code accepts `ERROR_PIPE_CONNECTED`; it is not waiting for a future client. The same sequence is used by the overlapped-server stdio implementation in [Node v22.23.2/libuv](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L321), selected by [its stdio caller](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L368-L418).

The [existing evidence response](https://github.com/aipoch/open-science/pull/2518#issuecomment-5707360922) links the prior head's passing real Windows ownership, pipe and process tests. The [new head's Windows core job](https://github.com/aipoch/open-science/actions/runs/35210994567/job/105168299115) is still running and remains necessary current-revision evidence. No new failing reproduction was supplied, so no speculative native change is included. The raw review verdict remains needs-changes; this response does not bypass required checks or authorize a PR merge.

ewen-poch:
Fixed the current CI test failures in `ce0b6410`:

- Updated the two exact consumer-list assertions after the delegation coverage registrations. Both failed twice before the correction with the same four missing expected entries as CI.
- Added the eight missing `variants-gnomad.test.ts` consumer registrations. The failed run checked out GitHub's synthetic merge `c5d185effd34667be60dfc9aa69ae032e23aeac8`, which includes a newer main-side `ConnectorService` import. This explains why the guard passed on the PR head alone. The exact merge revision reproduced the same eight missing edges twice; all three architecture/coverage suites pass after applying this patch to that revision (14 tests).
- Changed the CodeBuddy construction-failure test to await the existing asynchronous `ChildProcess` exit observation. Windows cleanup proves that the native Job has no active processes independently of the adapter's 25 ms exit-status poll. The test continues to require process exit, receipt removal, runtime-directory removal, and reusable capacity. Production cleanup and native pipe code are unchanged. This is a test synchronization correction; the prior failure alone does not establish a leaked process.

Validation after the final edits: the three architecture/consumer suites pass on both the PR checkout and the failed CI merge checkout (14 tests each); focused delegation ownership/runtime and manifest/impact checks pass (157 tests); Node/sandbox/web typechecks, lint, formatting, diff checks, and exact-commit CI Integrity pass. No full local suite was run, per maintainer instruction; the manifest change still selects full portable CI. The [current-head Windows core job](https://github.com/aipoch/open-science/actions/runs/35213888681/job/105177752267) has now passed: all 92 native/ownership tests and all 341 Windows behavior tests passed, including the previously failing construction-cleanup case.

The maintainer explicitly approved expanding the diff to the two architecture test files. This commit changes test expectations and additive test registrations only; it adds no runtime states, persisted fields, data-model changes, or user interaction.

ewen-poch:
The review for `ce0b6410` completed with two findings. GitHub suppressed the reusable workflow's `review_body` output as potentially sensitive, so the normal comment publisher was skipped; I read the findings from [the completed review run](https://github.com/aipoch/open-science/actions/runs/35213884414). Workflow success is not a mergeable verdict.

**Windows pipe transport:** no change is needed for the reported Winsock-only premise. [Node v22.23.2 `createHandle`](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L180-L195) distinguishes `PIPE` from `TCP`, and the [`Socket({ fd })` constructor](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L423-L437) opens that selected handle. This is already a libuv pipe transport. The [same-head Windows core run](https://github.com/aipoch/open-science/actions/runs/35213888681/job/105177752267) passed all 92 native/ownership tests, including real piped ACP IO and argument boundaries for concurrent providers, and all 341 Windows behavior tests. There is no failing reproduction supporting a replacement transport.

**Focused Windows process dry-run:** accepted and fixed in `67504889`. The wheel-evidence step now excludes only `windows-process`, matching the other unrelated runtime steps. Normal PR wheel validation remains mandatory. The existing workflow contract failed twice before the one-condition fix; all 191 workflow/classifier/platform tests pass afterward, as do lint, formatting and exact-commit CI Integrity. No production code, data model, resource states, persisted fields or UI changed. The [same-head focused manual run](https://github.com/aipoch/open-science/actions/runs/35215460748) passed: delegated Windows process ownership tests executed successfully, and the wheel-evidence step was explicitly skipped.

ewen-poch:
I read both findings for `67504889`.

**OpenCode recovery forwarding:** the adapter currently returns only `reserve` and `run`, while `protectExecution` invokes the optional `execution.recoverCleanup`. This is a plausible recovery-path gap and warrants a public-boundary regression covering failed cleanup, explicit recovery, and reusable capacity/workspace. It has not yet been behaviorally reproduced, so I am not presenting it as a confirmed runtime defect. The likely minimal change forwards the existing hook; no new lifecycle state or persistence format is expected. Both `opencode-execution.ts` and its test are outside the approved file scope. The maintainer's five-review cap has also been reached, so further edits and another review are paused pending explicit approval.

**Overlapped pipe connection:** this repeats the already answered preconnected-pair finding. The client is synchronously opened before `ConnectNamedPipe`, and `ERROR_PIPE_CONNECTED` is accepted, following the [version-pinned Node/libuv stdio implementation](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L321). The [current-head focused Windows process run](https://github.com/aipoch/open-science/actions/runs/35215460748) and Windows core in [the current PR Gate run](https://github.com/aipoch/open-science/actions/runs/35215400959) both passed. No new failing reproduction supports changing this native path.

The raw review verdict remains `needs changes`. No merge, bypass, or additional review cycle has been initiated.

ewen-poch:
Fixed the OpenCode cleanup-recovery finding in `7c8d5515` by forwarding the existing `recoverCleanup` method through the adapter.

The new public-boundary regression failed twice on the unmodified adapter: after a cleanup failure, proven cleanup and the same optional recovery call used by production still left `reserve(1)` rejecting with `code: capacity`. It now passes and also checks that disposal is withheld before proof, repeated recovery does not dispose twice, and the same workspace/runtime-home paths can execute successfully again. The test uses the existing runtime/scope contract; no production test seam or real orphan process is introduced.

Validation after the final edits:
- OpenCode recovery and framework/ownership/composition consumers: 349 tests passed across 10 files.
- Claude Code certification sibling controls: 17 tests passed.
- Node and sandbox typechecks, full lint, formatting, diff checks and exact-commit CI Integrity: passed.

The other framework adapters preserve the common execution object; the missing forwarding was specific to the OpenCode wrapper. No new API, enum, stored field, migration, or user interaction is added. This is a bounded adapter correction rather than a change in ownership policy.

The maintainer authorized these two files and one additional AI evaluation. Previous head `67504889` completed its entire PR Gate successfully; that is historical evidence, and this commit's CI/review must still be read. No full local test suite or PR merge was performed.

ewen-poch:
The additional review for `7c8d5515` is complete. Its only finding repeats the claim that `Socket({ fd })` cannot wrap a Windows named-pipe descriptor; it no longer reports the OpenCode recovery gap. GitHub again suppressed `review_body`, so I read the raw result in [the review run](https://github.com/aipoch/open-science/actions/runs/35219057814).

No transport replacement is justified by this finding. The [version-pinned Node documentation](https://nodejs.org/download/release/v22.23.2/docs/api/net.html#class-netsocket) defines `net.Socket` for TCP and streaming IPC endpoints, including Windows named pipes. More directly, [Node's `createHandle`](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L180-L195) selects a libuv `Pipe` for a `PIPE` descriptor, and [the fd constructor path](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L423-L437) opens that selected handle. The current head's [Windows ownership test step](https://github.com/aipoch/open-science/actions/runs/35219061900/job/105194596104) has passed, including real concurrent provider pipe IO. The remaining Windows core steps and full CI were still running when this response was written.

The raw verdict remains `needs changes`; this response does not alter review labels or bypass any gate. The single additional evaluation authorized by the maintainer is now complete. No further code/review cycles are initiated without renewed authorization; remaining CI monitoring is read-only.

ewen-poch:
The Ubuntu failures were caused by the synthetic merge, not by the OpenCode recovery change. CI checked out `dd81187aaebd629f60a8b94cee69b873b4c3dd79` (PR head `7c8d5515` merged with main `2a42b011`). Both branches registered the same gnomAD consumer at different positions, so Git retained duplicate entries in eight modules.

Commit `b52819b5` places those eight existing entries at the same positions as main and aligns the approved settings architecture expectation. Every evidence member and all ownership, capability and routing metadata are preserved. No production code, architecture, data model, state, persistence or UI changes are involved.

Reproduction and validation:
- The existing repository-manifest regression failed twice on the unmodified CI checkout with `settings_provider_accounts.testFiles.consumer contains duplicate entries`.
- After the fix, the same eight focused suites passed on both the PR branch and its merge with current main: 169 tests per checkout. These cover manifest validation, ownership inventory, consumer coverage, impact planning, source-regression routing, and the settings/skills architecture contracts.
- The actual `git merge-tree` result matches the two tested files byte for byte; all consumer membership sets are unchanged and no duplicates remain.
- Node and sandbox typechecks, lint, formatting, diff checks and exact-commit CI Integrity passed after the final edit.

The existing repository-manifest test already exercises this failure directly, so no duplicate test or production seam was added. The full PR impact plan selects complete CI because this PR already changes global CI inputs. Per maintainer instruction, no full local test suite was run. Current-head remote CI remains required; earlier Windows results apply only to the previous head.

ewen-poch:
I checked the only finding on `b52819b5`. It repeats the previously discussed preconnected-pipe case; the review did not report a new issue in the consumer-registration fix.

The Microsoft warning is relevant to pending overlapped connection operations. Our `Pipe` helper successfully opens the client synchronously before calling `ConnectNamedPipe`, and accepts `ERROR_PIPE_CONNECTED`. The [Microsoft return-value documentation](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-connectnamedpipe#return-value) identifies a client connected before that call as an established connection. More directly, [Node v22.23.2's libuv implementation](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L321) uses this same order and null argument for its pipe pair, including when the server has `FILE_FLAG_OVERLAPPED`. This is evidence for the specific preconnected pair, not a claim that null is generally safe for asynchronous connection setup.

The [current-head Windows core job](https://github.com/aipoch/open-science/actions/runs/35222055375/job/105204484690) passed. Its process-ownership step passed all 92 tests, including all four real Windows process integration tests and concurrent ACP pipe IO; the adapter/composition step passed another 343 tests. No failing reproduction supports the reported disconnected/hung startup on this path, so I am retaining the implementation instead of changing the native connection lifecycle speculatively.

The raw review verdict remains `needs changes`. This response does not change labels or bypass any gate. No additional review cycle is being initiated; remaining CI is monitored read-only.

ewen-poch:
Commit `2e0ad705` merges current main (`a155bbed`) into this branch to resolve the conflicts while preserving the existing PR history.

Both text conflicts were competing additions to test-evidence arrays: `scripts/ci/module-impact.json` and `src/main/skills/user-skill-repository.architecture.test.ts`. The resolution retains the Plan tests from main and the delegation/process tests from this PR. It also removes a duplicate gnomAD registration produced by automatic merging. No parent test evidence was dropped.

Validation found an additional registration gap already present on main: two renderer tests now import main-process behavior, but 91 reachable consumer edges across 66 modules were missing. The consumer-coverage guard failed twice on the merged checkout; a separate untouched main checkout reproduced exactly the same 91 edges. The fix only adds those entries to the existing manifest and synchronizes the two previously approved architecture-test expectation lists. All assertions remain intact.

Local validation of the final integration:
- ACP cancellation, resume, Plan context and framework paths: 12 targeted suites, 1,234 tests passed.
- Delegated execution, cleanup recovery, ownership and composition: 10 targeted suites, 349 tests passed.
- Final manifest/ownership/consumer/workflow guards, both architecture expectations and both newly registered renderer consumers: 9 targeted suites, 343 tests passed.
- Node, sandbox and renderer typechecks, lint, formatting and diff checks passed.

The manual resolution differs from Git's automatic merge in only the manifest and the two approved architecture tests. There are no additional production changes, new states, persisted fields, migrations or UI changes introduced by the resolution; main's existing changes are retained. Local verification was targeted as requested, not a complete local test run. Windows/native and full-suite verification remain with the new exact-head CI; the previous head's green CI is historical evidence only.

A separate base-branch blocker remains: `origin/main` itself contains duplicate gnomAD evidence in `main_permission_grants.testFiles.consumer`. The candidate manifest is deduplicated and validates successfully, but exact-commit CI Integrity rejects the invalid **base manifest** before checking the candidate. The workflow checks out trusted base code and `checkModuleOwnership` validates `baseManifest` first. Thus this is not a passing Integrity result. Fixing the candidate alone cannot repair the trusted main baseline. I have not changed protected CI policy or bypassed a gate; baseline remediation requires separate maintainer direction.

ewen-poch:
The exact-head AI review for `2e0ad7051ca6cba9765c60c8b9ea4c46d5af21e8` has completed with **mergeable** and no actionable findings: https://github.com/aipoch/open-science/pull/2518#issuecomment-5715769100.

The remaining CI failure is the invalid trusted-main manifest, independently confirmed in both CI Integrity and PR Gate preflight. Baseline repair is isolated in #2749 (`7058dea1`), changing only the manifest and two architecture expectation lists. It reproduces the failures before the fix and passes 383 targeted tests, all typechecks, lint, formatting and diff checks afterward. That repair PR encounters the same trusted-base failure, so normal admission cannot bootstrap it.

No protected-branch bypass or merge has been performed. The repair is being prepared for a separate, narrowly scoped maintainer decision; this PR retains its ordinary CI requirements afterward.

ewen-poch:
Rebased onto current main `1c4ba5c2`; the new head is `19a6ab78`.

The final production tree matches the original reviewed PR combined with main. The replay preserves the previously approved terminal connection-release policy, cleanup assertions and Windows impact routing that were recorded in historical merge resolutions. Manifest conflicts retain every test-evidence member from both revisions, with main's order first; the two approved architecture expectations match those lists.

After the final edit: 31 targeted suites / 1,926 tests passed, together with Node/sandbox/web typechecks, lint, formatting, diff checks and exact-commit CI Integrity. No complete local suite was run. Current-head full/platform CI remains required.

Main's #2748 resolves the earlier invalid trusted baseline. There is no remaining bootstrap/bypass request. This rebase introduces no additional architecture, persistence, state, data-model or UI changes. The PR description now describes the final integration and evidence rather than obsolete merge history.

ewen-poch:
The named-pipe finding does not reproduce for the preconnected pipe pair used here, so I am retaining the implementation.

`Pipe()` successfully opens and retains the client handle with `CreateFileW` before calling `ConnectNamedPipe`; it accepts `ERROR_PIPE_CONNECTED`. This is the preconnected case, not a pending asynchronous accept. Microsoft's [return-value documentation](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-connectnamedpipe#return-value) identifies that result as an established connection. The general OVERLAPPED requirement should not be used to claim that this sequence necessarily prevents every launch.

There is a version-pinned upstream example: Node v22.23.2's bundled libuv [sets FILE_FLAG_OVERLAPPED on the server, opens the client first, then calls ConnectNamedPipe(server_pipe, NULL) and accepts ERROR_PIPE_CONNECTED](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L309). Its [stdio pipe setup explicitly requests an overlapped server](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L368-L400).

More importantly, the current head `19a6ab78` has already passed the **Test delegated Windows process ownership** step in [this Windows job](https://github.com/aipoch/open-science/actions/runs/35234274947/job/105246188080). The existing real-process regression launches three providers and exchanges ACP stdin/stdout while checking argument boundaries; the suite also covers Job recovery, owner crash and cmd launch. Both the native pipe implementation and that integration test are byte-for-byte identical to the earlier [passing focused Windows run](https://github.com/aipoch/open-science/actions/runs/35215460748).

No failing public-behavior reproduction supports adding an asynchronous connection state machine or switching the transport mode. No production change is made for this finding. The AI review's raw verdict remains `needs changes`; this comment records the evidence-based disposition, not an overridden verdict. Other current-head Windows/PR checks are still running and are not represented as fully passed.

ewen-poch:
#2749 is merged. Rebased this PR onto its main commit `f6ee8875`; the new head is `ad332619`.

The intervening branding merge (#2567) and #2749 caused four additive-list conflicts. The resolution retains both sets of platform/ownership registrations and consumer expectations. The production tree matches the previous reviewed head combined with main. The subsequent main commit `7619e75d` also merges cleanly in a read-only merge preview, so no further history rewrite is needed merely because main advanced.

Validation exposed three additional baseline assertion failures, reproduced twice on both the unchanged rebased checkout and an untouched main checkout: 31 missing `notebook_network_sandbox` consumer registrations, two omitted settings-repository owner paths, and eight branding/credential consumer expectations in each of the two architecture tests. The follow-up only corrects the existing manifest and the two already-approved architecture-test files. It preserves every surviving evidence member and removes nothing beyond paths already deleted by main. No production logic, architecture, model, state, persistence, migration or UI change is introduced.

After the rebase, 23 ACP/framework/delegation suites passed 1,594 tests; no production changes followed. After the final metadata correction, all 41 affected consumer/architecture/CI suites passed 1,083 tests. Final Node/sandbox/web typechecks, lint, formatting, diff checks and exact-head CI Integrity passed. The full local suite was not run, as requested; current-head remote CI remains authoritative.

The prior Windows pipe review response remains applicable: this rebase does not alter that native implementation or its real-process integration tests. Earlier Windows success and review conclusions remain historical evidence until the new head's checks finish.

ewen-poch:
Rebased onto latest main `caa8fbed`; new head `7b26e6fa` follows the current split module-registration layout.

The original PR's 39 still-needed registrations now live in nine existing module files. The metadata-only root and shared manifest loader remain exactly as upstream defines them. The two existing settings/skills architecture assertions retain upstream expectations plus this PR's process consumers. No other module fields change, and no inline aggregate is restored. The final production tree matches the reviewed PR combined with latest main.

The five previously failing architecture expectations and missing notebook consumer entries are already fixed upstream; the untouched latest-main check passed 66 tests. The Windows brand-storage crash fixture also has an upstream fix in #2744, including documented failing reproduction and real Windows verification. Both are retained instead of introducing duplicate corrections.

After the final rebase edit, 47 targeted suites passed 2,016 tests: 1,600 ACP/framework/delegation tests, 381 manifest/architecture/CI/fixture checks, and 35 process-tree/artifact-evidence tests. Node/sandbox/web typechecks, lint, formatting, diff checks and exact-commit CI Integrity passed. No full local suite was run. Current-head Windows and full PR CI remain necessary; the local crash fixture tests are not a Windows E2E result.

This is the approved adaptation to upstream file organization, with no additional runtime architecture, data model, states, persisted fields, compatibility migration or UI change. The native pipe implementation and its real-process test remain unchanged; the earlier review response still describes that code, while new-head CI/review results must be evaluated independently.

ewen-poch:
The current review repeats the previously evaluated preconnected named-pipe finding. No source change is warranted by the available evidence.

`CreateFileW` opens and retains the client before `ConnectNamedPipe`, and the latter accepts `ERROR_PIPE_CONNECTED`. This is not a pending asynchronous accept. Node v22.23.2's bundled libuv uses the [same sequence with an overlapped server and a null OVERLAPPED pointer](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L232-L309); its [stdio setup explicitly requests the overlapped server](https://github.com/nodejs/node/blob/v22.23.2/deps/uv/src/win/pipe.c#L368-L400). See the earlier [detailed response](https://github.com/aipoch/open-science/pull/2518#issuecomment-5716290369).

For this exact head `7b26e6fa`, **Test delegated Windows process ownership** has now succeeded in [the current Windows job](https://github.com/aipoch/open-science/actions/runs/35339533989/job/105582101742). It exercises real provider creation, piped ACP IO, argument boundaries, Job recovery and crash cleanup. The native implementation and integration tests are byte-for-byte unchanged from the prior passing heads.

The claimed launch failure has not reproduced. Introducing an asynchronous connection state machine or changing pipe mode would add behavior without a demonstrated regression. The AI's raw verdict remains `needs changes`; this response records our evidence-based disposition without changing review labels. Remaining Windows/E2E and portable CI results are still pending and are not represented as passed.

ewen-poch:
Fixed the CLI conversation export failure in `089a2b5b`.

The failure artifact shows `Session revision conflict: expected 3, actual 5`; the error occurs in renderer serialization before the save reaches Main. A queued partial reply was still subjected to the legacy graph merge after Main had published Task completion. Main already owns this transcript and applies only captured user preferences and conversation commands.

The minimal fix excludes Main-owned transcripts from that legacy pre-save merge check. Renderer-owned conflict protection remains intact. No timeout increase, flaky-test exemption, new state, storage format, migration, or interaction change is introduced.

Two deterministic regressions exercise the real store, ordered queue, flush boundary and Main persistence owner, with and without a concurrent rename. Both failed twice on the unfixed implementation with the same expected-3/actual-5 conflict, then passed after the fix. They verify that the complete reply survives, the rename persists, and a passive snapshot adds neither a disk write nor a revision. No test seam was added.

After the final edit: `npm run test:module -- renderer_session_persistence` passed 194 files / 5,421 tests; the incremental impact plan selects this module and its registered consumers. All process typechecks, lint, formatting and diff checks passed. Windows E2E and AI review still require the new-head results; previous-head results are historical.

ewen-poch:
The current AI review for `089a2b5b` reports that `net.Socket({ fd })` treats Windows named-pipe descriptors as TCP. No change is warranted by the implementation or the current Windows evidence.

In the exact Node version used by this job, [`lib/net.js` lines 180–195](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L180-L195) calls `guessHandleType(fd)` and constructs a `Pipe` for `PIPE`, or a `TCP` for `TCP`. The Socket constructor [uses and opens that handle at lines 427–431](https://github.com/nodejs/node/blob/v22.23.2/lib/net.js#L427-L431). This is already a pipe-aware wrapper.

The [current-head Windows core job](https://github.com/aipoch/open-science/actions/runs/35345373486/job/105600652876) ran Node v22.23.2 and passed all four real `process-tree.windows.integration.test.ts` tests, including writing an ACP request through child stdin and reading its response through stdout. The three ownership/composition suites passed all 93 tests. Replacing this wrapper would add an alternative transport without reproducing the alleged defect.

The raw AI verdict remains `needs changes`; this response records the evidence-based disposition of its sole finding. The separate Windows conversation-export journey is still pending and is not claimed as verified by these process tests.

PR #2523, opened 2026-09-13
fix(session-persistence): preserve explicit branch selections across clients

ewen-poch:
## Problem

With two Electron windows and a Web client open on the same Session, one client can keep an older local Branch while another edits and selects a new revision. Passive runtime saves could publish the older selection again, so reloading the editing client reopened an older revision. A queued snapshot could also adopt a newer revision without containing its new Branch. During the same scenario, an incoming idle snapshot that predates a newly appended prompt could clear its local activeRun before provider dispatch.

## Proposed change

- Ignore persistence triggers consisting only of transient running/navigation indicators.
- Keep an explicit local Branch selection, including descendant frames, pending until its own save receipt. Passive saves retain the authoritative selection while still persisting runtime results. Preserve window-local selection when applying those receipts.
- Rebase a queued passive snapshot when its source authority predates a different durable Branch, and retain the submission baseline during revision-conflict recovery. Explicit navigation remains an intentional selection write.
- Keep a pending local run when a newer, non-archived idle snapshot does not contain its prompt anywhere in the graph. A matching terminal snapshot or an archive still settles it.

## Scope and non-goals

Four files: two renderer production files and their public-boundary/unit and existing Electron/Web regression coverage. No new public interface, persisted field, state enum, database migration, dependency, or Markdown documentation. The pending selection record lives only in the existing saver instance. Existing Session graph selection and run fields are written as before, with corrected ownership. No historical selection repair: existing files cannot establish the user's last intended selection. No layout or new control changes; reload now restores the intended revision.

The targeted/module checks below ran against the final production sources, including descendant navigation. The E2E waits for the Web recovery alert to close before capturing a screenshot. The local impact classifier requests full fallback because this existing E2E path is unclassified; the classifier is unchanged. Local targeted/module validation follows the maintainer's instruction not to run the complete portable suite locally; PR Gate remains the final authority for full/platform CI, which is still running. AI review of 9e0dffb4 completed with **mergeable** and no actionable findings ([review](https://github.com/aipoch/open-science/pull/2523#issuecomment-5653816615)).

- Baseline reproduction: eight public regressions fail twice on untouched cab6b38a production (both changed production files temporarily restored from HEAD, then restored to the fix). Failures are the older selected prompt replacing the remote edit, or a new prompt retaining its message while losing activeRun.
- Descendant-navigation regression: failed twice on 953111db before its follow-up fix, selecting the remote root prompt instead of the explicitly selected child prompt. It now passes through public Store actions, save acknowledgement and reload; remote hydration alone does not enqueue a save.
- Store/runtime/shared Session behavior: `npm run test:module -- session_renderer` — 952 passed.
- Persistence, lifecycle subscription, quit flush, export, runtime events/branches, skill-save, compute-host and Plan receipt consumers: `npm test -- src/renderer/src/lib/session-persistence/session-persistence.test.ts src/renderer/src/lib/session-persistence/session-persistence.render.test.tsx src/renderer/src/hooks/useLifecycleSync.test.tsx src/renderer/src/hooks/useQuitPersistenceFlush.test.ts src/renderer/src/lib/session-package-export.test.ts src/renderer/src/lib/acp/workspace-runtime-save-as-skill-owner.test.tsx src/renderer/src/lib/acp/useWorkspaceAgentRuntime.characterization.test.tsx src/renderer/src/lib/acp/workspace-events.test.ts src/renderer/src/lib/acp/workspace-runtime-session-branch-owner.test.ts src/renderer/src/pages/workspace/workspace-compute-host-access-controller.test.ts src/renderer/src/pages/workspace/previews/PreviewToolContent.plan.test.tsx` — 372 passed across 11 files.
- Renderer and E2E type contracts: `npm run typecheck:web`; `npm run typecheck:node` (including sandbox check).
- Source checks: `npm run lint`; `git diff --check`.
- Electron and Web build: `npm run build:e2e`; `npm run build:web`.
- Real client regression: `npm run test:e2e -- e2e/session-client-consistency.spec.ts --repeat-each 3` — 6 passed (all three repetitions). Exercises the delayed deletion receipt and edits across two Electron windows and an authenticated Web client, including reload and screenshots. Both scenarios also passed after the descendant-navigation follow-up (2 passed), including the reload screenshot.

Uncovered: the real client fixture uses fake OpenCode, not live Claude Code/OpenCode/Codex Responses/Codex Bridge providers. Existing runtime tests cover their shared renderer paths; live provider certification and Windows/Linux execution are not claimed. Main/preload protocols, storage schema and platform-specific implementation are unchanged.

## Review focus

Check passive versus explicit root/descendant selection, delayed/coalesced receipts, the fixed conflict baseline, and prompt identity across all graph branches. Confirm terminal/runtime data still persist and that genuine terminal or archive authority is not suppressed.


ewen-poch:
Addressed the descendant-navigation finding. The saver now recognizes active-frame and descendant-branch changes as explicit local navigation and preserves the complete selection already present in that snapshot. Remote hydration can update descendant choices, so those updates do not create a local navigation intent. Existing retained-root/context-reset behavior remains covered.

A public Store/save/reload regression failed twice on 953111db: selecting the original child revision persisted the remote root prompt instead. It now passes and also checks that remote hydration alone causes no save. The session module passes 952 tests, and the persistence/runtime consumer set passes 372 tests. The change stays within the existing saver and regression file, with no new API, persisted field, enum or migration.

PR #2554, opened 2026-09-14
fix(update): recover interrupted shell launches before restarting

ewen-poch:
## Problem

Restart to update briefly enters loading, returns to ready, and never calls the installer when old Shell launch intents cannot prove process cleanup. The supplied diagnostic report confirms two version-1 launching records without PID or start identity. Restarting Windows does not supply those missing fields. The failure reason was also below long release notes, making refusal appear to be a no-op.

## Proposed change

- Admit new Windows supervised Shell commands through the existing process host. It atomically promotes a pending receipt into a PID-addressable active receipt before starting the native supervisor. Recovery cancels pending admission or validates the recorded host nonce before stopping it. Normal cleanup retains the native termination proof. Restore the workload's original Electron environment flag so the wrapper cannot change command behavior.
- Offer in-app recovery for intact, old version-1 launch intents with no process identity. Stop identifiable backends first; bind confirmation to the exact records; move original bytes into an exclusive backup directory; repeat the normal cleanup and renderer durability gate before installation. Changed, malformed, claimed and current-instance records are excluded. Partial backup failure preserves all originals and requires fresh consent for the remaining records.
- Put failure feedback before release notes and use ErrorNotice's recovery button. Include all eight translated locales.

## Scope and non-goals

ShellProcessOwnershipRegistry owns the receipts under shell-process-ownership. New hosted receipts use version 2 with receiptId and commandIdentityMarker; PID-addressable filenames close the crash gap before JSON PID publication. Old version-1 records remain readable and fail closed unless the user explicitly confirms the bounded recovery. No new UpdateState/UpdateBlocker enum values, database schema, service, native binary or dependencies. The existing host gains an optional argument to restore only its child's environment.

Historical recovery persists the original records under shell-process-ownership-backups/<recovery-id>. It does not delete projects, notebook history, conversations or settings. This action acknowledges unknown historical work; it does not claim an unidentified process was killed. Backups are retained without automatic pruning; the owner documentation records a non-overwriting restoration route with the app closed. Recovery consent is transient status/options metadata (legacyShellRecovery and legacyShellRecoveryToken), not a database field.

Already installed versions do not gain the new recovery button until updated. An affected old installation may need one manual installation of a build containing this fix. No further computer reboot is required by this design.

Red-before-green evidence: the real Electron crash fixture failed on unreaped cleanup before the hosted launch fix; the historical updater integration failed twice at zero quitAndInstall calls before recovery was implemented. Both now pass. Browser recovery was also red before the button existed. Independent review found a malformed-shutdown-result edge and wrapper environment leakage; both were reproduced by failing tests, fixed and re-reviewed.

Final focused impact set:

| Behavior / consumers | Check | Result |
| --- | --- | --- |
| Shell ownership/execution, runtime, shared kernel host consumers, updater/provider/IPC, shutdown/app lifecycle, reproducibility | Main command below | 16 files passed; 723 tests passed, 106 conditionally skipped |
| Dialog, update store, eight locale guards | Renderer command below | 5 files, 857 tests passed |
| Visible refusal, explicit recovery consent/cancel, long/short notes and small viewport | Browser command below, repeated twice | 10 passed |
| Main + sandbox types | npm run typecheck:node | Passed |
| Renderer/bridge types | npm run typecheck:web | Passed |
| Source/style rules | npm run lint | Passed |

```powershell
npm test -- src/main/notebook/windows-shell.integration.test.ts src/main/notebook/shell-process.test.ts src/main/notebook/shell-process-ownership.test.ts src/main/notebook/runtime-service.test.ts src/main/notebook/runtime-service.architecture.test.ts src/main/notebook/kernel-process-lifecycle.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/kernel-startup-retry.integration.test.ts src/main/update/legacy-shell-recovery.integration.test.ts src/main/update/strategy.test.ts src/main/update/electron-updater-strategy.test.ts src/main/update/electron-updater-provider-regressions.test.ts src/main/update/ipc.test.ts src/main/lifecycle-shutdown.test.ts src/main/artifacts/artifact-reproducibility-lifecycle.test.ts src/main/app-lifecycle.test.ts --maxWorkers=4
npm test -- src/renderer/src/components/UpdateDialog.render.test.tsx src/renderer/src/components/UpdateDialog.lifecycle.test.tsx src/renderer/src/components/UpdateDialog.blockers.test.tsx src/renderer/src/stores/update-store.test.ts src/renderer/src/i18n/resources.test.ts --maxWorkers=4
npx vite build --config vite.browser-test.config.ts
npx playwright test -c playwright.update-local.config.ts e2e/browser/update-dialog.spec.ts --repeat-each=2
```

The local browser config only selects installed Chrome and the prebuilt localhost fixture server; it is not shipped. All verification is on Windows. The main set and browser/renderer set ran after the final behavior changes. The subsequent optional-chaining type correction was checked again with the real Windows workload/environment and crash regressions (3 passed), both typechecks and lint. No complete npm test was run, as requested.

## Review focus

Review the pending/active admission race, exact identity validation, consent-to-backup binding, partial failure preservation and normal installer gate. Standards and Spec reviews are complete with no remaining findings after fixes; both reviewers confirmed the impact map including shared-host kernel consumers. Linux/macOS-specific and unavailable-interpreter cases remain conditional skips locally. Packaged installer handoff and cross-platform packaging must be covered by PR CI; no CI run is claimed before the PR exists.


ewen-poch:
The reported TypeScript compilation failure is not reproduced on PR head `6afb724f757b4621d9ad9c13a6aecd44a3831eb4`. The [Static checks job](https://github.com/aipoch/open-science/actions/runs/34806404203/job/103859082623) passed both **Typecheck node and web** and **Verify published runtime bundle**. I am leaving the existing dispatch unchanged because the asserted compiler error is contradicted by the project checks.

The actual CI failures have both been reproduced locally with the focused command `npm test -- src/main/application-command-wiring.test.ts scripts/ci/classify-pr-changes.test.ts -t "holds side chat admission|covers every production source"`: the wiring assertion still expects the old options object, and the interrupted-launch fixture does not match the Windows-sensitive path rules. The proposed fixes are to update that assertion to include the recovery-token forwarding and rename the existing Windows-only fixture to `windows-interrupted-shell-launch.ts`, updating its test reference. These changes are pending maintainer confirmation of the additional test-file scope; no architecture, persisted format, state, or UI change is proposed.

ewen-poch:
Fixed the two reproduced CI failures in `42d9692`:

- Updated the application wiring guard to verify both side-chat admission and the recovery-token forwarding already implemented by this PR.
- Renamed the existing Windows-only crash fixture to `windows-interrupted-shell-launch.ts` and updated its test reference. It now matches the existing Windows-sensitive classification rules; the classifier and workflow are unchanged.

Validation after these edits: `npm test -- src/main/application-command-wiring.test.ts scripts/ci/classify-pr-changes.test.ts src/main/notebook/windows-shell.integration.test.ts` passed all 112 tests, including the real Windows/Electron interrupted-launch recovery. `npm run typecheck:node` (including sandbox types), `npm run lint`, and formatting checks passed. Both CI guard failures were reproduced before these edits. These are test-only changes with no additional runtime, persistence, state, or UI impact.

ewen-poch:
I checked both findings against the actual Windows runtime and the admission protocol, and am leaving the implementation unchanged:

- **Wrapper exit / surviving supervisor:** an isolated probe used the bundled Electron 39.8.10 (Node 22.22.1, libuv 1.51.0), the production admission host, the real native supervisor, and a long-running harmless workload. Both child PIDs were alive before SIGKILL of the Electron wrapper; both were gone before recovery was invoked, in three consecutive runs. `NotebookRuntimeService.shutdownAll()` then returned `{ reaped: true }`. This matches [libuv 1.51.0's non-detached Windows child ownership](https://github.com/libuv/libuv/blob/v1.51.0/src/win/process.c#L65-L115): closing the wrapper's Job Object terminates its direct native supervisor; that supervisor's kill-on-close Job Object contains the workload. The probe was kept local because it did not reproduce the reported defect. Adding another persisted process identity based on this finding is not justified by the observed behavior.
- **Pending-to-active gap:** `kernel_process_host.js` first renames the pending file directly to the final PID-addressable active path. That operation itself publishes the active receipt; the subsequent temporary-file replacement only enriches its JSON. Recovery already accepts the pre-enrichment record using the PID in the filename. There is no separate removal/publication interval matching the finding. The existing pending-cancellation and pre-PID-publication tests both passed when rerun.

The current-head Static checks job also passed, including Node/Web typechecks. No additional runtime changes, persisted fields, states, or interaction changes were introduced in response to these findings.

ewen-poch:
Rechecked the wrapper-exit concern with 20 consecutive isolated runs using the bundled Electron, the production admission host, the native Windows supervisor, and a live harmless workload. The supervisor and workload PIDs were alive before wrapper termination and absent before recovery ran in every case. The ordinary update shutdown then succeeded. The proposed P1 explanation omits libuv's non-detached Windows child Job Object: wrapper exit terminates the native supervisor, whose own kill-on-close Job Object contains the workload. See [libuv 1.51.0](https://github.com/libuv/libuv/blob/v1.51.0/src/win/process.c#L65-L115). No production change or extra persisted identity/proof is justified by the currently reproduced behavior; this is bounded experimental evidence, not a claim to have tested every environment.

The confirmation-component finding is valid: `docs/design.md` requires `AlertDialog` for destructive confirmations. The new historical-record recovery confirmation uses `window.confirm`; this was missed in the previous review. The proposed correction is to reuse `ConfirmActionDialog`, retain the existing warning and explicit consent, and test cancel/confirm/focus behavior through the browser. It would introduce only transient dialog state, not a persisted field, backend state enum, data model, or recovery-protocol change. UI implementation is pending the maintainer's requested interaction-change confirmation.

ewen-poch:
Replaced the historical Shell recovery `window.confirm` with the existing `ConfirmActionDialog` (`AlertDialog`) in `a81c095`. The warning and explicit recovery action are unchanged. Cancel receives initial focus; Cancel/Escape dismiss without applying, and focus returns to the recovery button. Confirmation is tied to the captured recovery token and the current ready-state offer.

Only `UpdateDialog.tsx` and its browser test changed. The additional state is transient UI state; there are no new persisted fields, backend enums, data-model changes, or recovery-protocol changes. Existing translations are reused.

The updated browser test failed twice on the previous UI because the app-owned alert dialog was absent, then passed after the change. Final validation: 857 tests across the update dialog/store/i18n suites passed; all 10 repeated browser update-dialog tests passed; Web typecheck, lint, and formatting passed. Chinese screenshots at 1000×720 and 560×420 were visually checked. An independent review found no additional high-value issue in this UI slice.

ewen-poch:
The reported admission release does not occur in the production update composition. `ipc.ts` passes `holdSideChatAdmission: true` for the complete durable install gate. `SideChatRuntimeOwner.suspendAll()` clears its in-flight suspension on completion but retains `handoffAdmissionHeld`; only the install-abort path calls `resumeAfterHandoff()` to release it. The consented notebook pass therefore does not reopen Side Chat admission.

The outer `createActiveResearchSafeInstallGate` also rechecks active research after both notebook passes and renderer durability, before returning to the installer. Newly active work is refused by that final check. An independent review of this finding reached the same conclusion.

Idle ACP reconnection is intentionally non-latching, and an asynchronous interval after runtime teardown already exists in the base gate. Repeating the teardown in the second pass would relocate that interval rather than establish an admission fence. No production reproduction currently establishes the reported regression at line 234, so I am not adding extra teardown or new admission state based on this finding. A broader idle-runtime admission guarantee would need a separately reviewed design and a reproducing test.

ewen-poch:
Fixed the recovery confirmation's early-Escape race in `UpdateDialog`. The macOS trace shows the second click had opened the AlertDialog, but Escape arrived before its dismissable layer finished registering; the outer dialog handled the key and both dialogs closed. The update dialog now routes Escape to cancellation of the open recovery confirmation during that interval. Ordinary Escape and IME handling remain unchanged.

The existing immediate-Escape browser regression failed in both the original macOS run and its retry. It remains unchanged: no extra readiness wait, timeout increase, or retry was added to hide the race. Local Chrome did not reproduce the baseline failure, including repeated and CPU-throttled probes, so macOS CI is still required to confirm resolution.

Local validation: all 15 repeated update-dialog browser cases passed; renderer module tests, web typecheck, and lint passed. Independent review found no additional issue. This is limited to the existing UpdateDialog file and introduces no data or persistence changes.

The other failing job was Windows workspace E2E: the paced-tool bottom-follow test timed out waiting for a deterministic reply, then passed on retry and failed the fail-on-flaky gate. That test is outside this change; no unrelated test or CI policy was modified.

PR #2572, opened 2026-09-14
fix(remote-access): observe shutdown cleanup failures immediately

ewen-poch:
## Problem

If a browser has been approved for persistent access but has not claimed its grant, shutdown removes that unclaimed authorization. When a provider detection is still pending and this existing persistence write fails, the invalidation promise rejects before serialized shutdown attaches its handler. Node emits `unhandledRejection` even when the shutdown caller immediately catches its returned promise, potentially interrupting the remaining exit cleanup.

This also occurs in the normal web-first shutdown order: the web-stop callback observes its own failed cleanup, but the retained cleanup IDs are retried by remote-access shutdown while detection still holds the queue.

## Change

Observe invalidation rejection immediately, then continue awaiting the original promise in the existing serialized shutdown. Preserve immediate access denial, queued disposal, shutdown promise identity, and propagation of the original cleanup error.

The regression uses public HTTP authorization and service APIs, real pairing and a temporary-file repository, with existing provider/persistence injection points. It covers direct shutdown and prior web stop, verifies no process-level rejection before or during shutdown, and checks that the caller still receives the original write failure after detection drains.

## Scope

Two files; one production statement plus explanation. No architecture, API, UI, state/enum, data model, persisted format, migration, or historical-compatibility change. The existing authorization deletion remains strict and can still fail; this does not repair the disk failure or introduce retries. No production test seam or dependency added.

## Validation

All final behavior checks ran after the last material edit. An independent review confirmed the following impact set and found no actionable issues.

| Behavior | Check | Result |
| --- | --- | --- |
| Reported early rejection, including web-first shutdown | `npm test -- src/main/remote-access/service.test.ts -t 'handles shutdown cleanup failure'` using unchanged base production source | Both new cases failed twice, with exactly `shutdown cleanup persistence failed` observed as an unhandled rejection |
| Pairing, repository, provider, IPC and shutdown regression | `npm test -- src/main/remote-access` after restoring the exact fix | 6 files, 117 tests passed |
| Shutdown ordering/failure handling, command composition, fatal-process policy, and Web consumers | Explicit impact command below | 16 files, 366 tests passed |
| Main and sandbox TypeScript | `npm run typecheck:node` | Passed |
| Source quality | `npm run lint`; final changed-file ESLint, Prettier check and `git diff --check` | Passed |

```sh
npm test -- src/main/remote-access \
  src/main/application-runtime.test.ts \
  src/main/lifecycle-shutdown.test.ts \
  src/main/app-lifecycle.test.ts \
  src/main/index-fatal-errors.test.ts \
  src/main/host-application-commands.test.ts \
  src/main/application-command-composition.test.ts \
  src/main/runtime-electron-wiring.test.ts \
  src/main/web-service/controller.test.ts \
  src/main/web-service/http-server.test.ts \
  src/main/web-service/index.test.ts
```

The first impact run hit sandbox `listen EPERM` in the HTTP suite; the same bounded set passed with local socket access enabled. No tests were skipped to work around it. Full local `npm test` was intentionally not run. No renderer/build/platform-specific logic changed; the final PR Gate remains authoritative for portable and OS lanes. This is a controlled service-boundary reproduction, not a real remote.it outage or disk-fault desktop E2E.

## Review focus

The immediate catch observes the private promise only: the queued shutdown must still await the original rejected promise and preserve the failure rather than reporting success.


ewen-poch:
Investigated the two failing CI lanes before retrying:

- Ubuntu shard 3 failed in `package-mutation.test.ts` teardown with `ENOTEMPTY` while removing its temporary root. The unchanged test file passes locally (46 tests).
- Windows E2E shard 1 encountered `ERR_NO_BUFFER_SPACE` during `run-marks` navigation. The notification-recovery timeout has the same loading failure: its Playwright trace records `ERR_NO_BUFFER_SPACE` fetching the shared JavaScript bundle. Both unchanged browser cases pass locally with two workers and no retries.

Neither area is modified by this PR. No code, test expectations, timeouts, or workflow configuration were changed. After the entire first attempt completed, I requested one failed-jobs-only rerun, retaining the successful lanes. Local passes do not establish the CI host root cause; any repeated failure needs further investigation rather than another automatic retry.

ewen-poch:
Final verification for head `3b24de5cd95c64c1786b8cee4d12b3655d31383b`: PR Gate run 34833303926, attempt 2, completed successfully, including the final aggregate. The Ubuntu shard 3 and Windows E2E shard 1 reruns passed; CodeQL and CI Integrity also passed. The exact-head AI review is mergeable with no actionable findings.

Only one `--failed` rerun was requested. Comparing both attempts confirms that previously successful lanes retain their original execution timestamps despite receiving new job IDs; the failed lanes and related report/aggregate jobs ran again. No source, test expectations, timeouts, or workflow changes were made during CI triage. This successful retry does not establish a permanent fix for the original runner/cleanup failures.

Issue #2602, opened 2026-09-15
[Feedback]: Avoid an unexpected second Codex login during initial setup when launched from Codex


imjszhang:
### Feedback type

Improvement

### What would you like to share?

During first-time setup, Open Science asks me to sign in to Codex again even though I have already signed in to Codex and launched Open Science from there.

From a user's perspective, this feels like a duplicate authentication step: I am already using Codex, but entering Open Science's configuration flow requires another Codex login. This adds friction to onboarding and makes it unclear whether my existing sign-in has been recognized.

### Steps to encounter this experience

1. Start with Codex already signed in.
2. Launch Open Science from Codex with Open Science in its initial, unconfigured state.
3. Enter the initial configuration flow and configure Codex.
4. Open Science asks for another Codex sign-in.

### Expected experience

Where technically supported and with the user's consent, detect the existing Codex sign-in and offer to reuse it so that setup can continue without repeating authentication.

If Open Science requires a separate authentication context, explain this before asking the user to sign in again—for example, why the existing Codex session cannot be reused and whether this additional sign-in is a one-time setup step.

### Additional context

This is feedback about the first-run experience, rather than a claim that authentication is broken. The main concern is the unexpected second login when the user has just launched Open Science from an already authenticated Codex environment.

ewen-poch:
Verified on current main (f799a608): this onboarding feedback has been addressed by the existing authentication-import flow and #2609 (merged September 15, 2026).

- Codex onboarding offers **Import existing Codex sign-in** and explains what will be copied into Open Science's app-owned profile. A separate Open Science sign-in is also available and is explicitly described as separate from the CLI profile.
- If no file-backed credential (`auth.json`) is available, the import now explains that the existing sign-in may be held in the system credential store, which Open Science cannot import from, and directs the user to the supported Open Science sign-in. This message is translated in all eight supported locales.

Validation in an isolated worktree: all 22 authentication-import tests passed, as did the onboarding, provider-form, validation-message, and locale-guard suites. I also ran the existing missing-credential regression against the pre-#2609 implementation: it failed with the old generic error, then passed after restoring current main. The merged PR's final CI checks passed.

For transparency, the broader local authentication suite had one separate runtime-install admission test failure (runtime remained reported as in use after teardown); the import-focused tests passed. No real OS-keyring authentication session was exercised in this verification.

This addresses the reported unexpected-login experience; it does not add OS-keyring credential import or guarantee that every existing Codex login can be reused. Closing as completed.

PR #2609, opened 2026-09-15
fix(settings): explain missing file-backed Codex credential imports

LiuZhaoyan:
## Problem

When a user launches Open Science for the first time from an already-authenticated Codex environment (#2602), the onboarding provider step seeds the `codex-shared` import path. That import reads only `~/.codex/auth.json`. If Codex keeps its credential in the OS credential store (macOS Keychain / Windows Credential Manager / Linux keyring), the file is absent, and the import fails with the generic `The selected Codex profile does not contain importable authentication.` — which reads like a required second sign-in even though Codex itself is signed in.

## Proposed change

Split the `ENOENT` branch of `readCodexAuthenticationSnapshot` (`src/main/settings/codex-auth.ts`) so a *required* import over a *missing* `auth.json` raises a message that:

- states that no file-backed Codex credential was found to import,
- explains the existing sign-in may live in the system credential store, which Open Science cannot import from,
- points the user to the supported Open Science Codex sign-in to continue.

A present-but-unparsable `auth.json` keeps the existing generic message. The new message is localized through the existing `localizeProviderResourceMessage` switch in `validation-message.ts` (the same path the onboarding `ProviderStep` already uses for main-process provider errors), with catalog entries added to all eight translated locales. The Settings page `Re-import Codex login` error detail benefits from the same clearer text.

## Scope and non-goals

In scope: the error copy at the import boundary, its localization, focused tests.

Non-goals (per the issue's narrow framing):

- No OS keyring/Keychain access or credential migration.
- No parent-process or "launched from Codex" detection.
- No change to the shared/isolated import architecture, provider upsert flow, or `prepareCodexRuntimeHomeAuthentication` (which already tolerates a missing file via `required: false`).

1. A required shared import with no `auth.json` raises the credential-store explanation.
2. A required shared import with an unparsable `auth.json` keeps the existing failure.
3. The new message localizes on the provider-validation surfaces.
4. All eight locale catalogs carry the new key with no placeholder drift.

Evidence mapping — all commands below ran on the final head `f7763c38`, after the last material edit:

| --- | --- | --- |
| 1 — import boundary messages | `npm test -- src/main/settings/codex-auth.test.ts` | passed |
| 2 — retained generic failure | same file (`keeps the generic import failure for a present-but-unreadable auth.json`) | passed |
| 3 — localization on validation surfaces | `npm test -- src/renderer/src/pages/settings/validation-message.test.ts` | passed |
| 4 — i18n catalog guard | `npm test -- src/renderer/src/i18n/resources.test.ts` | passed |
| Type safety, both processes | `npm run typecheck:node`, `npm run typecheck:web` | passed |
| Lint | `npm run lint` | passed |
| Formatting of changed files | `npx prettier --check <changed files>` | passed |

Environment note: typechecks ran with an increased Node heap (`NODE_OPTIONS=--max-old-space-size`) because the default invocation crashes on this 7 GB machine on a clean `main` tree as well; the crash is a local memory limitation, not diff-related.

Consumers: the changed interface is the thrown error message consumed by `ProviderStep` (onboarding) and `ProvidersPanel` (re-import detail); both already display main-process provider messages verbatim or localized, so no consumer contract changed.

Uncovered risks: no real keychain-stored Codex login was exercised end to end (requires a Codex account configured for OS credential storage); platform lanes (Windows/macOS) and the full portable suite are delegated to exact-head CI.

## Review focus

- Wording of the new message in `codex-auth.ts` (and the locale translations).
- Whether keeping the generic message for the present-but-unreadable case is the right split.


LiuZhaoyan:
Documenting CI evidence for the two red runs on this unchanged head `f7763c38`:

**Run 1 (34932332091)** — single failure: `macos_e2e` bundle. `e2e/subagent-release-gate.spec.ts:939` (fairly schedules two upward lanes with a concurrent real user prompt) timed out after 120s waiting for the completion text, on both the initial attempt and retry #1. The same test passed cleanly on the immediately preceding PR Gate run for another PR (34931395266, no retry). No code path connects this diff to delegation scheduling.

**Run 2 (34938469679)** — root cause: `Full portable tests (Ubuntu, shard 3/3)` hit the job's `timeout-minutes: 25` and was canceled mid-run (25m15s; the same shard completed in 15m20s in run 1), so its coverage blob was never uploaded. The `Module tests and coverage` job then merged only shards 1/3 and 2/3 — 25,148 tests passed, 0 failed — and the incomplete merge dropped global line coverage to 78.82% against the 90% threshold, failing the job. Both failures are the cascade of the shard timeout on a slow runner.

All other lanes (static, unit shards 1-2, Linux runtime, Windows core/E2E, macOS journeys/presentation/regressions/delegation on run 2) pass. Re-triggered the gate; happy to re-run again if needed.

PR #2614, opened 2026-09-15
fix(specialist-marketplace): follow GitHub release redirects

### LINKED ISSUE #2612: [Bug]: Specialist Marketplace cannot install from GitHub sources because Release asset redirects are cancelled
### Preflight checklist

### What happened?

Installing any Specialist from a user-added GitHub Marketplace source fails with the generic "could not install this specialist" message. The source review, root signature and release preview all succeed; only the install step fails. Importing the identical published ZIP manually succeeds.

Expected: a verified GitHub source installs its Specialists the same way the official source does.

**Cause.** `MarketplaceService` is constructed with `fetch: netFetchStandard` (`src/main/ipc.ts`), and `fetchOne` requests every URL with `redirect: 'manual'`. For a GitHub source the only artifact URL is the Release asset (`https://github.com/<owner>/<repo>/releases/download/<tag>/<asset>`), which always answers 302 to `release-assets.githubusercontent.com`. Electron 39 `net.fetch` cannot return manual redirect responses — as the comment on `netFetchWithManualRedirect` in `src/main/skills/net-fetch.ts` already documents — so the request rejects before `assertAllowedUrl` ever sees the redirect. The official source is unaffected because its first artifact URL is the CDN, which does not redirect. The Skill Marketplace already uses `netFetchWithManualRedirect`; the Specialist Marketplace was not switched.

Reproduced outside the app with Electron 39.2.6 `net.fetch`:

| Request | Result |
| --- | --- |
| GitHub Release asset, `redirect: 'manual'` | throws `Redirect was cancelled` |
| same URL, default redirect | 200 |
| `raw.githubusercontent.com` metadata, `redirect: 'manual'` | 200 |

### Steps to reproduce

1. Add a GitHub Marketplace source, e.g. `https://github.com/mrsonord2240/openscience-specialists/tree/published` (Protocol v1, signed, artifacts published as GitHub Release assets).
2. Open any Specialist from that source and choose Install.
3. The install fails; `main.log` records only `ipc handler rejected` for `specialist:marketplace-install-prepare` with `errorCategory: error`.

### Operating system

Windows

### App version

0.30.0

```
{"level":"warn","scope":"ipc","msg":"ipc handler rejected","data":{"channel":"specialist:marketplace-install-prepare","surface":"electron","location":"local","principalKind":"human","actionOrigin":"human","durationMs":548.4457000000002,"errorCategory":"error"}}
```

mrsonord2240:
Fixes #2612

## Problem

Installing a Specialist from a user-added GitHub Marketplace source always fails with the generic install error. The source review, root signature and release preview succeed; only `specialist:marketplace-install-prepare` rejects.

`MarketplaceService` requests every URL with `redirect: 'manual'` so `assertAllowedUrl` can validate each hop before it is contacted. It is composed with `netFetchStandard`, which delegates to Electron `net.fetch`. Electron 39 `net.fetch` cannot return manual redirect responses (documented on `netFetchWithManualRedirect` in `src/main/skills/net-fetch.ts`). A GitHub source has no artifact base URL, so its only artifact URL is the Release asset, which always answers 302 to `release-assets.githubusercontent.com`. The request therefore rejects and every install from a GitHub source fails. The official source is unaffected because its first artifact URL is the CDN, which does not redirect.

Reproduced with Electron 39.2.6 `net.fetch` against a published Release asset:

| Request | Result |
| --- | --- |
| Release asset, `redirect: 'manual'` | throws `Redirect was cancelled` |
| same URL, default redirect | 200 |
| `raw.githubusercontent.com` metadata, `redirect: 'manual'` | 200 |

## Proposed change

Compose the Specialist `MarketplaceService` with `netFetchWithManualRedirect`, the GET-only adapter the Skill Marketplace already uses. It keeps the Chromium network stack (and the user's proxy) and returns each redirect to `fetchOne`, so the existing host and repository checks in `assertAllowedUrl` run before the next host is contacted.

## Scope and non-goals

- One composition change in `src/main/ipc.ts` plus a wiring assertion.
- No change to `MarketplaceService`, its redirect policy, allowed hosts, artifact verification, or the adapter.
- Non-GET requests are unaffected: the adapter delegates anything that is not a manual-redirect GET to `netFetchStandard`.

| --- | --- | --- |
| Specialist Marketplace is composed with the manual-redirect adapter (new wiring assertion; failed before the fix) | `npx vitest run src/main/application-command-wiring.test.ts` | passed |
| Marketplace redirect validation, artifact verification and install flow unchanged | `npx vitest run src/main/specialist/marketplace/service.test.ts` | passed |
| Adapter returns manual redirects and streams direct responses | `npx vitest run src/main/skills/net-fetch.test.ts` | passed (3 files, 43 tests together) |
| IPC composition owner and consumers | `npm run test:module -- ipc_surfaces` | 1972 tests passed, 0 failed; 11 suites failed to load (see below) |
| Main-process types | `npm run typecheck:node` | passed |
| Source standards | `npm run lint` | passed |

The 11 suites that failed to load all stop at `Electron failed to install correctly`: the Electron binary did not download into `node_modules/electron/dist` on this Windows machine, so suites that import `electron` cannot start. The same failure reproduces on unmodified `main` (b67ca596); it is local, not caused by this change, and CI remains authoritative for those suites.

Renderer, preload and persisted formats are untouched, so web typecheck and E2E lanes were not run locally.

All checks ran after the last material edit.

## Review focus

- The adapter only returns redirects for `redirect: 'manual'`; confirm every `MarketplaceService` request is a GET (root, signature, release descriptor, artifact, default-branch lookup).
- Uncovered risk: the fix is not exercised end to end against a live GitHub source in CI; the service's redirect handling is covered by its existing unit tests and the adapter's by `src/main/skills/net-fetch.test.ts`.


ewen-poch:
Thank you for investigating this issue and submitting the fix. Your approach is correct: the Specialist Marketplace needs the existing manual-redirect adapter so it can validate each redirect before contacting the next host.

The same production change has already been merged in #2615, together with an Electron regression test that reproduces the installation failure before the fix and passes afterward. That test is also included in the macOS and Windows CI journeys, and all selected checks passed.

Closing this PR because the fix is already covered by #2615. Thank you again for the clear diagnosis and contribution!

PR #2617, opened 2026-09-15
fix(pdf-structure): recover native figure and table layouts

ewen-poch:
## Problem

Native scientific PDFs can lose figure panels and outdented labels, assign page furniture to crops, flatten hierarchical table headers, split wrapped records, omit notes, or fail to combine continued tables. These failures were assessed across 62 supplied PDFs and 699 source pages.

## Proposed change

- Recover complete raster/vector figures, side-captioned flowcharts, exported plates, and native figure notes. Exclude evidenced running furniture, duplicate page numbers, adjacent table rules, and prose-only references.
- Recover table headers, shared statistics, section boundaries, missing records, narrative columns, and notes from native glyphs, baselines, and rules. Decode legacy symbols only when font/encoding evidence supports the mapping.
- Extend existing continuation grouping to repeated multi-level headers and explicitly continued quotation tables. Require compatible captions, consecutive pages, source geometry, complete header topology, and no unassigned text; retain source regions and notes.
- Add semantic JSONL regressions with counterexamples. Keep scientific values, source omissions, and printed typos literal.

The six development commits are consolidated into three reviewable commits: native layout recovery; grouped records and crop boundaries; mixed native layouts and notes. Rebased without conflicts onto `main` at `248c7df2`. All 226 changed paths have exactly the same content as before rebase. Most added lines are source-derived regression fixtures (~67.5k of ~76.5k added lines); no source PDFs, models, or local assessment artifacts are included.

## Scope and non-goals

- Existing extraction facade, worker result schema, renderer result contract, model/runtime dependencies, and persistence ownership remain unchanged. The renderer grouping change is a derived view over the existing cells and spans.
- Historical compatibility: no migration or compatibility shim. Existing resource fingerprinting causes normal re-extraction of stale cached results; no eager historical rewrite or alteration of existing exports.
- States/enums: no new public state or enum member.
- Persistence: no database, serialized field, cache format, path, or write-lifecycle change.

After the final rebase, run the explicitly targeted owner/contract/consumer set, not the full local portable suite:

```sh
npm test -- src/main/literature/pdf-structure src/renderer/src/pages/workspace/previews/renderers/pdf-figure-selections.test.ts src/renderer/src/pages/workspace/previews/renderers/PdfFiguresView.render.test.tsx src/shared/pdf-table-copy.test.ts --maxWorkers=4
npm run typecheck:node
npm run typecheck:web
npm run lint
```

| Changed behavior / consumer | Check | Result |
| --- | --- | --- |
| Native glyph/rule ownership, figure bounds, table grids/spans/notes -> worker decoding and cache contracts | Complete PDF owner suite in the targeted command | 131 test files / 1845 tests passed, including consumers below. |
| Derived continuation grouping -> preview rendering and table copy | `pdf-figure-selections.test.ts`, `PdfFiguresView.render.test.tsx`, `pdf-table-copy.test.ts` | Passed in the same run; includes source-derived joins and rejection cases. |
| Main/sandbox and renderer contracts | Node and web typechecks | Passed after rebase, including sandbox. |
| Changed linted source | Lint and `git diff --check` | Passed after rebase. |

Corpus evidence completed before the content-preserving rebase:

- First batch: 23 PDFs / 244 pages, 58 bounded CLI requests; actual worker-reader admission and five continuation pairs.
- Second batch: 22 PDFs / 208 pages, 50 bounded CLI requests; actual worker-reader admission and three continuation groups.
- Third batch: 17 PDFs / 247 pages, 57 bounded CLI requests; actual worker-reader admission and one continuation pair. Final output contains 55 figures and 70 table fragments, with no unassigned table text.
- Latest replay of 466 native table inputs from the first two batches: no grid/span/crop/issue regressions; only two expected note-paragraph separations. This is parser replay, not a claim that all earlier PDFs were rerun through model inference after every later change.
- All source pages were visually reviewed; changed crops and native tables were compared with source renders. Resource checksums bind the final corpus evidence to the unchanged parser source.

The repository manifest has no PDF module ID, so the targeted set follows the explicit extraction -> worker reader -> grouping/render/copy dependency map. CodeGraph was consulted, but resolves another worktree's index; current-source inspection and dynamic resource/worker contract tests supplement it. Existing engine and CLI hashing cover all `.mjs` resources. No IPC, event, build configuration, path-handling, or platform-native behavior changes require new local platform lanes. Full packaged Electron GUI and cross-platform packaging were not run. PR CI remains authoritative; independent review is pending.

## Review focus

- Native evidence and rejection conditions must prevent recovery profiles from stealing neighboring cells, notes, or prose. Check complete text ownership and row/column topology, not only zero diagnostics.
- Continuation grouping must preserve body section rows and shared cell/source regions; incompatible headers and ambiguous quotation geometry must remain separate.
- Conservative model-span warnings remain when a merge proposal conflicts with source rows/columns: five documented cases in the second batch and three in the third. They are not suppressed to improve a metric.
- One earlier source table visibly repeats three P values below its final labelled outcome. They remain unassigned because attaching them to a record would invent a relationship. Source graphic arrows remain in crops rather than becoming invented text cells.

This PR is a draft for review and CI. No merge is requested.


ewen-poch:
### Empty rule candidates

I reproduced the reported input on the reviewed head (`ad34d15f`): `recoverRepeatedUnitGrid(table, tokens, captions, [])` returns `undefined` without throwing. The `candidates[0][1]` access is inside the `Array.filter` callback, which never executes for an empty array. The resulting `footer.length < 7` check then short-circuits before either footer element is dereferenced.

The existing regression in [`table-bounded-records.test.ts`](https://github.com/aipoch/open-science/blob/ad34d15f4f0324079cea72c727f645aed9ac0dff/src/main/literature/pdf-structure/table-bounded-records.test.ts#L283) explicitly covers `repeated-unit-columns` with no rules. An additional early return would be redundant; the reported extraction failure does not occur.

### Vector wordmark exclusion

The predicates test different relationships: the first requires a confirmed running header in the same **vertical band**; the second rejects **two-dimensional overlap** with text. A header beside a wordmark satisfies both conditions, so the filter is reachable.

I replayed the final native geometry through `excludeRepeatedMarginContent`. On the affected flowchart page, the normalized rectangles are:

- Wordmark: `[0.8203125, 0.04296875, 0.921875, 0.05859375]`
- Confirmed header: `[0.1621903, 0.0409586, 0.6968218, 0.0510004]`

Their vertical intervals overlap, their horizontal intervals do not, and the wordmark is excluded as intended. Exempting header text from the overlap check is unnecessary for this case.

The targeted `table-bounded-records.test.ts` and `figure-native-side-layout.test.ts` run passed **55 tests in 2 files**. Both findings were checked against the unchanged reviewed head; no production changes are needed for these reports.

PR #2630, opened 2026-09-15
fix(notebook): restore Windows conda R and trace sandbox failures

ewen-poch:
## Problem

Windows conda R installed under `Lib/R/bin/x64` is not recognized as belonging to its conda prefix. Real R 4.4.3 consequently exits with `0xC0000135` (missing DLL), and discovery reports it unavailable. Separately, authorization waits and failed R verification lose the stage or child-process details needed to diagnose the reported sandbox failures.

## Proposed change

- Recognize the optional `x64` directory and reuse existing conda environment construction. Use shared R/Rscript resolution for x64-only external and managed conda environments across discovery, execution, readiness, provisioning, package operations and recovery. Preserve the standard path when both or neither layout exists. Revoke existing owned grants for both known executable layouts, including missing binaries.
- Add authorization and verification checkpoints using existing diagnostic operations, IDs and durations. Capture bounded, redacted child errors before cleanup or protocol conversion discards them; logging failures cannot change cleanup or the original error.

## Scope and remaining risks

No architecture, business data fields, business state enums, permission expansion, UI or historical-data migration changes. Historical compatibility preserves standard-bin interpreter identity and revokes existing current/legacy-prefix grants for both standard/x64 paths even after executable removal; the existing native receipt ownership checks remain authoritative. Persistence is limited to additional records in the existing rotating `main.log`, potentially including redacted local paths and error details.

The reporter's Windows path-API access denial and authorization IPC delay remain unconfirmed. A real AppContainer probe passed `normalizePath`, jsonlite and arithmetic with the fixed environment. Full Notebook execution could not be validated because the running application's gateway port was occupied. Diagnostic regression fixtures verify observability, not reproduction of the underlying path-API failure.

The real Windows discovery regression failed repeatedly on unfixed code with `0xC0000135`, while an independent subprocess control supplying that installation's DLL directory succeeded. The same public `rRunnable` regression passes after the fix. Portable coverage exercises both supported layouts; the real R test opts in through `OPEN_SCIENCE_TEST_R_CONDA_PREFIX`.

Diagnostic regressions failed before instrumentation and pass afterward: an actual failing subprocess writes its failure to the real log sink, pending authorization exposes its current stage, a throwing logger preserves failure and cleanup, and the real protocol probe retains child errors before returning false.

The production-change checks below ran after the final production edit; the subsequent complete-installation fixture adjustment was separately retested and typechecked:

| Affected behavior | Project-owned check | Result |
| --- | --- | --- |
| R layout detection and real launch | `npm test -- src/main/notebook/environment-discovery.test.ts src/main/notebook/windows-r-conda-discovery.integration.test.ts` | Passed in final combined run; real R enabled |
| Runtime discovery and library consumers | `npm test -- src/main/notebook/runtime-binding.discovery.test.ts src/main/notebook/external-r-library.test.ts src/main/notebook/package-listing.test.ts` | Passed in final combined run |
| Authorization, verification, execution and diagnostics | `npm test -- src/main/notebook/network-sandbox-owner.test.ts src/main/notebook/kernel-executor.test.ts src/main/notebook/runtime-workflows.test.ts src/main/notebook/runtime-diagnostics.test.ts src/main/notebook/runtime-application-commands.test.ts src/main/diagnostics/operation.test.ts` | Passed in final combined run |
| Main-process and sandbox types | `npm run typecheck:node` | Passed |
| Source conventions | `npm run lint`; Prettier and diff whitespace checks | Passed |

The expanded final targeted run used `--maxWorkers=1`: 29 files ran, 25 passed; 1312 tests passed, 128 skipped, and eleven tests failed. One runtime-service fixture represented a complete Windows installation without Rscript; it was corrected and its focused regression then passed (1 passed, 401 filtered out). The other ten existing tests in environment-lock, data-migration and migration-service failed at fixture setup because this Windows host cannot create file symlinks (EPERM). All ten fail identically with the pre-fix source files. Their CI results remain authoritative. Additional consumers covered: runtime-paths, provisioner/startup/upgrade/runtime, package-manager, runtime-service, runtime-repair-policy, runtime-relocation, reproduction-runtime, restored-environment-verification, native-lock-restoration, environment-lock, recovery-coordinator and dependency-analysis.r-paths, using `npm test -- src/main/notebook/<name>.test.ts` for those explicit files. Skipped conditional cases are not claimed as verified. Runtime consumers were included because interpreter environment selection and authorization diagnostics flow through them. No renderer checks are needed because no renderer interfaces or behavior changed; PR CI remains authoritative for other platforms. No full local suite was run, as requested by the maintainer. Independent Standards and Spec reviews found no remaining issues.

A real managed-kernel regression exposes only the real installation's x64 binaries through a temporary fixture. With the old shared-path module it reports missing standard R.exe; after the fix it returns 2 through public execute without a resolved-interpreter override. Public readiness, execution and historical-grant revocation regressions also went red-to-green. This validates host execution and adapter behavior, not native containment certification.

Historical compatibility also preserves disabled x64 R runtime IDs during data-root relocation and recognizes valid legacy x64-only prefixes before provisioning cleanup. Four public regressions failed twice before these changes (missing disabled overrides for legacy/short/named R, and deleted legacy user-package content), then passed afterward. Existing settings/ready-marker formats and cleanup ownership rules remain unchanged; no new fields or migration mechanism are introduced. The final Test Impact Set additionally includes `npm test -- src/main/notebook/managed-runtime-relocation.test.ts src/main/storage/data-migration.test.ts src/main/storage/migration-service.test.ts`, with the environment-limited cases described above.

Windows readiness now requires the Rscript executable beside the selected R.exe, and provisioning verifies that actual script entry point. Real-R and portable regressions fail on the previous implementation and pass with the fix; restoring the missing Rscript to the same real fixture is a positive control. The focused installation/path set passed (6 files, 216 tests, 4 skipped). Non-Windows verification and prefix identity are unchanged.

Compatibility decision approved by the maintainer: an existing partial Windows prefix missing its matching Rscript is now unready. If it reaches existing invalid-prefix recovery, that policy rebuilds the prefix and can remove packages installed inside it. Valid legacy installations retain their identity and packages. No new state, field, persistence format or user interface is introduced.


ewen-poch:
The x64-only discovery gap is fixed. Conda-prefix enumeration now falls back to `Lib/R/bin/x64/R.exe` when `Lib/R/bin/R.exe` is absent, for both external conda environments and app-managed environments. When both layouts exist, the original root executable remains preferred to preserve interpreter identity.

A public `defaultCandidatePaths` regression with an unactivated PATH reproduced both omissions twice before the fix and passes afterward. Root-only and both-layout cases also pass. Final validation: 11 targeted test files passed (203 tests passed, 82 skipped), including the opt-in real Windows R launch; Node/sandbox typechecks, lint and formatting passed. Independent Standards and Spec reviews found no remaining issues.

The adjustment is limited to the existing discovery source and test files and introduces no new states, persisted data, architecture or interaction design.

ewen-poch:
Shared `rBin` / `rScriptBin` resolution now selects an existing Windows x64 executable when the standard executable is absent. Discovery uses the same policy, so managed execution, readiness, provisioning, package operations and recovery consumers no longer reconstruct a different layout. Standard paths remain preferred when both layouts exist and remain the expected provisioning paths when neither exists; Linux/macOS behavior is unchanged.

Managed access revocation checks both fixed executable layouts for the existing current/legacy prefixes, even when binaries have disappeared. The native owner matches existing selected/canonical executable receipts and removes only owned permissions. This adds no receipt fields, migration, business states, or UI changes.

Public readiness, execution and revocation regressions failed before the fix and pass afterward. A real R fixture exposing only x64 binaries also fails with the old shared-path module (missing standard R.exe) and passes after the fix through the public managed kernel entry point, returning 2 without a resolved-interpreter override. This is host execution evidence, not native containment certification. Historical missing-binary grant regressions also fail before the revocation adjustment and pass afterward.

Validation after the final edits: 26 affected test files ran, with 25 files passing, 1173 tests passing and 119 skipped. One existing environment-lock test cannot create a file symlink on this Windows host (EPERM during fixture setup); the same isolated test fails identically with the pre-fix shared-path source. That environment-limited case remains for PR CI. Node/sandbox typechecks, lint, formatting and independent Standards/Spec reviews passed.

ewen-poch:
Windows managed-runtime parsing now accepts the optional x64 directory, and relocation preserves it in the destination runtime ID. Disabled overrides are retained for legacy default, short default and named R environments without transferring installation authorization or changing the stored schema.

Legacy R prefix detection now uses the shared executable-resolution policy. A materialized x64-only legacy environment remains selected under the existing marker rules, so normal provisioning preserves its user-installed packages instead of classifying the prefix as disposable residue.

Both findings were reproduced through public behavior before the fix: relocation omitted the destination false override, and provisionR deleted a package file in the legacy fixture. All four regression cases failed in two pre-fix runs and pass after the changes. No business fields, states, architecture or UI changes were introduced.

Final verification after the edits: 29 affected runtime/storage test files ran; 26 files passed, with 1310 tests passed and 128 skipped. Ten existing tests fail during file-symlink fixture creation with Windows EPERM, before exercising the implementation; all ten fail identically with the pre-fix source files. Node/sandbox typechecks, lint, formatting and independent Standards/Spec reviews passed. PR CI remains authoritative for the environment-limited cases.

ewen-poch:
The partial-installation concern is reproducible. With real Windows R binaries and required base/compiler packages present, but Rscript.exe omitted, the public verifyExecutable operation accepts R.exe. A regression expecting rejection fails consistently in two runs. The temporary fixture does not modify the installed R environment.

Requiring a complete pair in the general path resolver could also stop recognizing a damaged legacy prefix and affect preservation of its packages. I propose validating the actual Rscript launch requirement in readiness/provisioning verification while retaining the existing prefix identity and ownership rules. This reaches the provisioning-verification module outside the currently approved file set, so implementation is paused for maintainer scope approval. No source changes were pushed for this finding.

ewen-poch:
Addressed the missing-Rscript finding by verifying the matching Windows Rscript entry point and requiring that sibling at the readiness gate. Prefix identity and materialization remain unchanged, preserving recognition of legacy environments.

The real partial-x64 installation was accepted on the previous implementation (reproduced repeatedly). It now rejects with the missing Rscript path; restoring the same installation's Rscript makes the same public verification pass. Portable tests cover both layouts and reject an Rscript located only in the other layout.

The maintainer accepted the compatibility consequence: incomplete historical prefixes can enter the existing invalid-installation rebuild policy, which may remove packages within that prefix. There are no new fields, enums, persistence formats or UI changes.

Validation: the six focused files passed (216 tests, 4 skipped). The broader 29-file consumer run passed 1312 tests; a complete-installation runtime-service fixture needed its matching Rscript and its focused retest passed after correction. The remaining ten failures are previously baseline-confirmed Windows file-symlink EPERM fixture failures. Node/sandbox types, lint and formatting passed. Independent Standards and Spec reviews found no actionable findings.

ewen-poch:
Confirmed through the public provisionR operation with its existing injected installation boundary. Starting with an empty prefix, the fixture creates only Lib/R/bin/x64/R.exe and its matching Rscript. Verification still receives the cached pre-create root-bin path and fails with ENOENT. The regression failed identically in two runs on 41f014dbf.

The minimal correction is to resolve the executable again after materialization, before verification. This requires editing provisioner.ts, outside the currently approved PR file scope. The maintainer explicitly requires approval for that expansion, so the regression is retained locally and production changes are paused. No new schema, state, persistence format or UI behavior is proposed.

PR #2635, opened 2026-09-15
fix(permissions): remember web reading for a conversation

ewen-poch:
## Problem

Native web reading only offered one-call approval, so a delegated literature task repeatedly stopped while opening journal websites. Auto approve edits did not provide a bounded way to remember this permission.

## Proposed change

Offer Once or Conversation for OpenCode webfetch and Claude WebFetch bound to an observed native tool call. Restore missing Claude root permission metadata from that call. Conversation approval covers websites and delegated children of the owning conversation. Preserve the native one-call response so the application registry controls reuse and revocation. Show the scope in the card and translate it in all eight locales.

OpenCode recognition follows its [native ACP permission mapping](https://github.com/anomalyco/opencode/blob/v1.17.13/packages/opencode/src/acp/permission.ts); URL titles alone do not establish tool identity.

## Scope and non-goals

- Add capability identity `builtin:web_fetch` using existing `builtin_tool` and `session` values. No new enum values or database fields.
- Three optional transient delegation fields preserve tool identity in the approval UI. A bounded, consumed session/call-id map restores native identity and is cleared with tool/session lifecycle. This checks message consistency, not authenticity of a hostile ACP process.
- Explicit grants persist in the existing registry; no URL is persisted in the grant. No migration or conversion of historical Once approvals.
- Existing Auto policy, shell, web search and MCP permissions are unchanged. Codex response/bridge do not share this native tool contract and are excluded.

The native-call mismatch regression also failed before the correlation fix (unexpected session scope). The original frequency regression failed twice before production changes: `expected [ 'once' ] to include 'session'`. It now passes and covers another website/child, recreated registry, conversation isolation, Once, Deny and revocation.

| --- | --- | --- |
| Native classification, approval, persistence and isolation | `npm test -- src/main/acp/permission-frequency.test.ts src/main/acp/permission-policy.test.ts src/main/acp/permission-context.test.ts src/main/acp/permission-broker.test.ts src/main/acp/permission-broker-registry.test.ts src/main/permission-grants` | Passed |
| Delegated permission contracts and lifecycle | targeted `acp-execution.test.ts`, `production-composition.test.ts`, `durable-delegated-work.test.ts` | Passed |
| Approval card, presentation and settings consumers | targeted `PermissionApprovalControls` render/interaction, `permission-request-presentation`, `PermissionsPanel.render` tests | Passed |
| Eight locale catalogs | `npm test -- src/renderer/src/i18n/resources.test.ts` | Passed |
| Scope menu and Chinese screenshots | `playwright test e2e/browser/web-permission.spec.ts` using a local Edge config | 2 passed |
| Main, sandbox and renderer types; source style | `npm run typecheck`; `npm run lint` | Passed |

Final targeted Vitest run after merging current main and correcting the RPC shutdown regression: 26 files, 2402 tests passed, including `src/main/acp/runtime.test.ts` and `src/main/notebook/local-rpc-server.test.ts`. Independent Standards and Spec reviews found no issues and confirmed the final impact mapping. Delegation and settings consumers are included because the identity projection and capability catalog changed. No platform path or OS integration code changed; additional platform lanes remain CI-owned. The full local suite was not run.

Limits: ACP input was reconstructed from the report and upstream contract because the export has no original request envelope. Browser tests use the real approval component in a fixture, not a live provider or full authenticated application.

## Review focus

Native identity validation, parent conversation ownership, app-owned grant revocation, and the visible cross-website/subagent scope. This reuses the existing permission model rather than introducing a new access profile.

## RPC CI correction

The existing stop-failure test from main expected an HTTP response even after the server exhausted its 100 ms close grace period. Awaiting server closure before releasing the executor reproduced UND_ERR_SOCKET twice on unchanged production code. The test now deterministically checks forced transport shutdown while still requiring the exact stop error from turn cleanup. Early rejection handling and finally cleanup prevent unhandled rejections and leaked test resources. No production lifecycle, persistence, enum or UI change is included. The isolated shutdown matrix passed 17 tests before the final 2402-test focused run. Locale merge conflicts preserved both independent key additions.


ewen-poch:
Addressed in e857e7d2d. Native web grants now require a process-local binding to the preceding native tool_call, matched by conversation/session, call ID and framework. The binding is consumed and cleared on completion, session clear and disposal. The runtime also restores missing Claude root permission metadata from that call; initial streaming input may be empty, while final URL validation remains at grant admission.

A public-context mismatch test failed before this follow-up (unexpected session scope for an unrelated initial tool) and now passes. The original frequency regression now goes through observeToolCall + handleProviderRequest with the real persisted registry and parent conversation ownership. Final focused run: 25 files / 2287 tests passed; typecheck and lint passed; independent Standards and Spec re-review found no actionable issues.

This correlation checks consistency with the observed call. It does not authenticate a hostile ACP process, which could forge both messages. No new database fields, migrations or enum values were added; the correlation map is transient and bounded.

ewen-poch:
The reported framework type error is not reproducible on e857e7d2dd485f758dcfe6fde6d2fa66933c3e82. `observeToolCall` returns unless `framework` is one of `codex`, `opencode`, `claude-code`, or `codebuddy` before the assignment. TypeScript narrows the value through that guard. The current-head PR CI successfully ran both main and renderer typechecks: https://github.com/aipoch/open-science/actions/runs/34977427088/job/104409228625. Adding a cast would hide information without fixing a demonstrated defect, so no code change is warranted for this finding.

The actual portable-test failure is in `src/main/notebook/local-rpc-server.test.ts`: `retains a 'executeControl' stop failure 'during-close' for turn cleanup`. The blob report identifies both the failed test and an unhandled rejection with `UND_ERR_SOCKET` from `fetchLocalRpc`. CI tested merge commit 6a7ce5335c3b3073a4436522dff090ce3f44a019 (this PR merged into db5cb835bfc50f7c5e376741df5922a8f092bcec). That test is absent from this branch's original base and comes from the newer main integration. Failure: https://github.com/aipoch/open-science/actions/runs/34977427088/job/104409228818.

The exact root cause of the close/response failure still needs a focused reproduction. Its test and RPC lifecycle implementation are outside this PR's current changed files. Investigation is paused before expanding the fix scope, pending maintainer confirmation. No broad retry or unrelated implementation change has been made.

ewen-poch:
The RPC shutdown failure is fixed in d36c27013. The server intentionally force-closes pending connections after its existing 100 ms grace window. The stop-failure test incorrectly required an HTTP 500 response even when execution outlived that window.

Awaiting server closure before releasing the executor reproduced UND_ERR_SOCKET twice with production code unchanged. The corrected test deterministically covers forced transport closure and still requires `clearArtifactTurnBinding` to reject with the exact `stopError` object. It attaches a rejection handler immediately and completes cleanup even when the transport rejects; other timings retain the exact HTTP 500 payload assertion.

No production lifecycle, architecture, data model, enum, persistence or UI behavior changed. Current main was merged to resolve the reported conflicts; both independent locale-key additions were preserved. Final validation after the merge and test correction: 26 targeted files / 2402 tests passed, typecheck and lint passed. Independent Standards and Spec reviews found no issues. No full local suite or broad CI retry was used for diagnosis.

ewen-poch:
The remaining failure is the macOS delegation E2E case `parks an upward message on branch switch and resumes it after restart and restoration`. Its first attempt timed out waiting for Main's restored-branch reply; the built-in retry passed, so `--fail-on-flaky-tests` correctly failed the lane.

The captured Main log shows that the resumed prompt ran and stopped normally. That rules out assuming that the queued message was never dispatched, but the artifacts do not establish whether the reply was lost during projection or only missing from the rendered view.

The unchanged targeted test passed three times locally. Two runs with a widened renderer-startup gap and three runs with a delayed restoration reload also passed. Those diagnostic edits have been removed. No production fix, timeout increase, or retry-policy change is justified by this evidence yet. I have rerun only the failed macOS delegation job and am checking the original test under additional local scheduling pressure.

Additional validation: six more runs of the unchanged case passed with two workers requested (no retries). Overall: nine unchanged-case passes plus five temporary timing experiments. The worktree is clean. The focused macOS rerun is executing; its result is still pending, and a passing rerun will not by itself establish a root-cause fix.

PR #2722, opened 2026-09-17
fix(ci): classify pull requests from their merge base

ewen-poch:
## Problem

PR #2721 changes two connector files, but PR Gate compares the latest base tip directly with its older head. That counts 95 files, including main-only CI changes and apparent deletions, and incorrectly selects full Ubuntu and expanded desktop validation.

## Proposed change

For pull_request events, resolve the unique merge base and publish it as the existing change-comparison base. Classification, changed-file formatting and affected-test/coverage consumers use that same base. Keep a separate trusted-base output pointing to the current event base for loading the classifier and module manifest. Merge-group and manual-event comparisons retain their existing endpoints. Missing or ambiguous merge bases fail closed.

## Scope and non-goals

Only PR Gate revision resolution and its workflow contract tests change. No business logic, state enums, persistent fields, data migrations, historical-data compatibility or UI changes. CI Integrity and reviewer policy remain intact. Real PR workflow edits/deletions still trigger existing full fallbacks. This PR changes a workflow and therefore still legitimately selects full CI.

Worktree: `.worktree/ci-pr-diff-base`; branch: `fix/ci-pr-diff-base`. Local guidance and `docs/ci-execution-policy.md` are excluded.

After the final edits:
- Diverged Git history, current-base policy loading, PR deletion/rename/CI changes, queue endpoints and unrelated histories -> actual workflow shell executed in temporary Git repositories -> passed.
- Workflow, classifier, module authority, platform routing, changed formatting and CI Integrity contracts -> seven focused Vitest files -> 251 tests passed.
- Changed test/workflow -> ESLint, Prettier and git diff --check -> passed.
- Exact #2721 commits -> actual corrected revision step and authoritative classifier -> base `3bf02118`, trusted base `1736f0b2`, two changed files, selective mode and Mac smoke; no full fallback reasons.

No full local npm test was run. Full remote CI and AI Review are pending; unrelated module-registration failures previously reproduced on main are outside this two-file fix.

## Review focus

Keep change scope separate from trusted policy provenance, preserve raw merge-group endpoints, and ensure downstream affected tests use the same comparison base as Preflight.


ewen-poch:
The orphan-commit finding is a false positive on `1192fe2e`. After the `unrelated` case removes tracked files at lines 116–118, it enters the separate conditional's `else` at lines 126–129 and recreates both the source and test files. `git add .` at line 130 therefore stages two files before the commit at line 131; the index is not empty.

I reran exactly this scenario with `npx vitest run scripts/ci/pr-gate-workflow.test.ts -t 'resolves actual Git history.*unrelated'`: 1 passed, 48 unselected. It reaches and passes the assertions that revision resolution fails for unrelated histories and publishes no outputs. The complete focused set also passed all 251 tests on this same commit. No `--allow-empty` change is necessary; adding it would mask a broken fixture instead of correcting a real failure.

PR #2732, opened 2026-09-17
fix(chat): preserve transcript follow intent across layout changes

PPPGLL:
## Problem

A long conversation can stop mounting new messages while still showing a thinking indicator below older content. In the reported desktop incident, returning to the app left older replies visible even though newer replies were available; a small scroll made the missing messages appear.

The transcript window derived follow state from scroll geometry alone, so a layout-induced scroll could disable following. Distinguishing reader input also requires respecting native event timing: Chromium can begin PageUp/ArrowUp scrolling after the first animation frame. Clearing the input signal in that frame loses deliberate history navigation, allowing the next message to pull the reader back to the live tail.

### Reproduction

With 120 conversation items and the last 80 mounted:

1. While following, deliver a layout scroll slightly above the bottom and append an item. New messages should remain mounted.
2. At the physical bottom, wheel downward or press ArrowDown/PageDown/End without any viewport movement, then append item 121. The latest item should mount and the window should remain bounded.
3. In Chromium, focus a 300px viewport containing 100px rows (`scrollTop=7700`), press PageUp or ArrowUp, and wait for native movement. Append item 121. The reader should remain in history and the new item should remain outside the mounted window.

The new real-browser keyboard cases both fail on `36cf9fcc`: the viewport moves upward but following remains enabled.

## Proposed change

- Record eligible input and release following only after actual upward viewport movement.
- Retain the input signal across animation frames and zero-movement scroll events. Consume it on movement, reject it when viewport/content geometry changes, and expire unused input after 500ms. Scrollbar drags stay tracked until release; input at the physical top cannot arm upward movement.
- Ignore downward/horizontal wheel input, forward navigation, editable-control keys and bubbled nested scroll events when deciding to leave the live tail.
- Retain reading anchors for deliberate history navigation. Display current live indicators only when the mounted transcript reaches its actual end.
- Give the single-worker Windows browser suite a 600-second global budget, with matching CI contract assertions.

| Check | Result |
| --- | --- |
| Five transcript suites, two CI contract suites and E2E ownership suite | 270 passed, 3 skipped on Windows with Git Bash. Includes upward wheel/touch/key/scrollbar navigation, no-op input, nested/editable controls, delayed movement, input expiry, and viewport/content geometry changes. |
| Real-browser keyboard reproduction | PageUp and ArrowUp fail on `36cf9fcc`; both pass after the fix using Playwright native keyboard input against the production `useTranscriptWindow` hook. |
| Full Windows browser suite | 250 passed in 3.4 minutes with `--workers=1 --fail-on-flaky-tests --global-timeout=600000`. Includes 8 new native keyboard/wheel and layout cases. |
| Type checks | `npm run typecheck` passed. |
| Code style | Full repository ESLint passed. Changed files formatted with Prettier. |

## Review focus

Verify that native keyboard movement releases following even when it begins after the first animation frame, while unused input and layout changes preserve following. The browser fixture imports the production hook, uses actual Chromium scrolling, appends a new message and checks the mounted tail and reader position. Hosted required checks must also pass on the updated head before merging.


ewen-poch:
@PPPGLL Please address the following before merging this PR.

### 1. Regression: input at the bottom can stop mounting new messages

At `6870186b`, `recordUserScroll()` in `src/renderer/src/pages/workspace/use-transcript-window.ts:244-252` unconditionally clears follow intent for any wheel/touch/navigation input. A downward wheel while already at the physical bottom causes no actual scrolling, so there is no subsequent `scroll` event to restore follow state.

Reproduction:
1. Render a conversation with 120 items, following its last 80 mounted items.
2. Set the viewport to the physical bottom (`scrollHeight=10000`, `clientHeight=800`, `scrollTop=9200`).
3. Dispatch a downward wheel (`deltaY=100`). Do not synthesize a `scroll` event because the viewport did not move.
4. Append item 121.

Expected: the new item mounts, the window still contains 80 items, and the live indicator remains available.
Actual: the new item is absent and the loading indicator disappears.

I extended the existing interaction regression with this scenario: it fails on the PR and passes when both production files are restored to base `1736f0b2`. The existing five related suites all pass (187 tests), so this case needs explicit coverage.

Please distinguish actual intent to leave the bottom from no-op/downward-at-bottom input. Preserve the original layout-scroll fix and deliberate history navigation. Cover downward wheel at bottom and navigation keys that cause no movement, alongside the existing upward wheel/touch/scrollbar cases. Also check bubbled events from nested scrollable/editable controls so they do not spuriously detach the transcript.

### 2. Windows CI: suite-wide time budget, not a French Settings assertion

Failing job: https://github.com/aipoch/open-science/actions/runs/35197890702/job/105127872623

The browser suite reports `Timed out waiting 300s for the test suite to run`, with **210 passed and 32 not run**. The French Settings test was interrupted by the global timeout; its `locator.click: Test ended` is a consequence. The same job's desktop journeys (11 tests) and workspace tests (14 tests) passed.

At the time of this review, latest main (`4780fff0`) still gives this Windows browser suite 300 seconds, so merging main alone does not resolve this failure. Open PR #2729 already changes the Windows browser invocation to a single worker with a 600-second global budget and has Windows validation evidence. Prefer incorporating that fix after #2729 lands on main, then merge updated main into this branch and resolve any conflicts. If this PR must proceed earlier, coordinate the equivalent minimal CI change with #2729 rather than importing all of its unrelated fixes or weakening assertions.

Please push the code regression fix and tests, refresh this PR description with final verification, and rerun required CI on the final head. Do not treat the existing 187 passing tests or a retry of the unchanged 300-second job as sufficient evidence that these two issues are fixed.

ewen-poch:
Re-reviewed `36cf9fcc`. The previous downward-wheel/no-movement regression is fixed. I reran the five transcript suites plus the two CI contract suites: **262 passed, 3 skipped**. CI Integrity failed on npm `ECONNRESET`; I retried the failed job and it now passes.

One new blocker remains:

**[P2] Do not expire keyboard scroll intent before native scrolling starts** — `src/renderer/src/pages/workspace/use-transcript-window.ts:244-251` clears `scrollInputRef` in the first animation frame. The comment that native scroll events precede animation frames does not hold for the first frame of native keyboard scrolling.

I verified this with native PageUp in Chromium, then with a temporary browser fixture importing this PR's actual `useTranscriptWindow` hook (120 items, last 80 mounted, a 300px viewport with 100px rows). The observed order was:

```text
keydown PageUp: scrollTop=7700
requestAnimationFrame: scrollTop=7700 (input cleared)
scroll: scrollTop=7420, isFollowingEnd=true
```

The reader has actually moved upward, but `expandAtScrollEdge` sees no input and preserves following. A subsequent append therefore advances the mounted tail; in the production component `TranscriptEndSync` will also request `scrollToEnd` when the item count increases with the same mounted count. This overrides deliberate keyboard history navigation.

Please retain the input signal through the first actual native movement (while still rejecting stale/no-op input and layout-only movement), and add a real-browser keyboard regression covering PageUp/ArrowUp rather than only a synthetic keydown immediately followed by scroll. Recheck that no-op input still preserves following. No need to change the now-passing CI Integrity implementation for the transient npm failure.

PR #2767, opened 2026-09-18
fix(credential-identity): restore fresh-install startup on macOS and align packaged smoke

ewen-poch:
## Summary

The v0.31.0 Release run's packaged smoke failures (macOS arm64/x64 timeouts, Linux x64 pre-healthy exits) trace to the fail-closed credential identity system meeting real keychain/backend state for the first time; source E2E only exercises it through mocks.

**macOS fresh installs could never start.** The native probe treats a not-found lookup as uncertain when any keychain in the legacy search list is locked. macOS keeps the System keychain in every search list and it is always locked for user sessions, so definite absence was unreachable on every stock Mac: a fresh machine (no stored item for either identity) probed both, got `access-blocked`, and failed startup with `initialization-probe-access-blocked`. The exists path only accepts matches from the unlocked first database, so absence there is the matching bar — the probe now reports `not-found` whenever the first database is unlocked, and still fails closed when the first database itself is locked. Existing users keep their legacy identity (verified: legacy `exists` unchanged, new identity `not-found` on a real keychain).

**Headless launches no longer block on a modal.** A pre-ready or post-ready credential failure showed a blocking recovery dialog; on an unattended launch nobody can dismiss it, so the app hung until an external timeout. In headless mode recovery now goes to stderr and the bounded exit applies. Windowed launches keep the dialog.

**Packaged smoke launches use supported backends.** Linux `launchAndProbe` omitted `--credential-store=file`, which headless launches without a desktop Secret Service now require (the CLI smoke already passed it); every deb and AppImage launch exited before becoming healthy. The macOS startup wait also surfaces captured app output if the service never becomes healthy.

## Test plan

- `packages/credential-identity-probe-native` fixture scenarios updated for the absence policy (42 pass), plus a real-keychain check of the rebuilt probe
- new headless-recovery startup test; dialog-asserting preflight test pinned to a windowed launch
- `tsc --noEmit` node + web, ESLint, and the credential-identity/startup/smoke suites green locally


ewen-poch:
Addressed the P1 with a narrower gate than a blanket any-locked fail-closed, which would have reintroduced the fresh-install block this PR fixes. Absence is now treated as uncertain whenever any locked **user** database goes unsearched — including one that may hold a historical default-keychain account — and still returns `access-blocked` for recovery. Only the admin-managed System keychain (`/Library/Keychains/System.keychain`) is exempt, identified by its fixed path: it requires per-item authorization for every write, never holds an application identity, and macOS keeps it in every search list locked in every user session, so counting it had made definite absence unreachable on every stock Mac. Fixture scenarios now cover the stock shape (unlocked login + locked System), a locked admin database in either position, and locked user databases in both the fresh and legacy-override cases (43 scenarios pass); the safety rationale is documented at the classifier.

PR #2773, opened 2026-09-18
fix(sessions): preserve exports after main-owned task completion

wen2zhou:
## Problem

A completed CLI Task first opened in the desktop can fail conversation export with a Session revision conflict. A queued partial renderer projection enters legacy graph reconciliation before reaching the Main-owned transcript boundary, even though Main intentionally ignores those graph snapshots.

## Change

For an acknowledged Main-owned transcript, route queued snapshots to Main without legacy graph reconciliation. Main still accepts only named preferences and validated conversation commands. Deterministic regressions preserve the complete authoritative graph while applying a title preference, and continue rejecting a stale-parent command without writing any state.

Add desktop and unopened CLI Task automatic Reviewer E2E coverage: exact published Version bytes and SHA256, scoped content/trace reads, bound findings, one provider dispatch, and restart persistence. Register both cases in the regression suite.

Based directly on origin/main at `6ee880ad0` after #2758. This PR changes only five files; no engineering documents are committed.

## Final-head verification mapping

All checks below ran after the final material edit on `eb63fe140` in the new worktree. An independent root review of the implementation agent's patch confirmed the Main authority boundary, negative command guard and coverage mapping.

| Behavior | Check | Result |
| --- | --- | --- |
| Renderer persistence/flush, export UI, conversation commands, Main authority and option sanitization | Six focused Vitest files: session-persistence unit/render, ConversationExportDialog.interaction, session-conversation-command, runtime-authority, renderer-save-options | 212 passed |
| Desktop and unopened Task automatic Review, existing manual Review, CLI export/re-export/continuation | Selected cases in artifact-auto-review, subagent-model-release-gate and workspace-conversation Playwright specs | 4 passed |
| Task lifecycle, restored question/Plan/permission decisions, cancelled publication and follow-up | interaction-restart and artifact-task-lifecycle Playwright specs | 10 passed |
| Main/preload/renderer types, lint, actual Electron bundle | npm run typecheck; npm run lint; npm run build:e2e | Passed |
| Real model automatic Review and persistence | Two new DeepSeek/OpenCode Sessions, desktop and CLI Task, then normal Quit/restart and normal UI/native-save Markdown export | Both passed |

The real Reviewer read the exact Version content; expected bytes, preview bytes and SHA256 matched, checks were scope-bound, and each path had one trigger/dispatch. The Task was not opened before Review completion. After restart, the same Review/Version links persisted with idle Sessions and no activeRun/error. Main and Reviewer used existing deepseek-flash settings; no global configuration or credentials were changed. Desktop's first findings submission was correctly rejected for a missing execution-log read; it read that evidence and completed within the same dispatch. The application/controller and temporary local service exited normally.

Persistence is a shared renderer interface, so consumers and the full portable fallback are included. Restart/Electron tests cover observable persistence and export behavior. Windows remains subject to this new PR's own CI; installed upgrade/downgrade and unsupported provider paths are not certified by these checks.

## Remaining validation limits

The complete local portable suite (`OPEN_SCIENCE_TEST_MAX_WORKERS=2 npm test`) finished in 30m49s: 2,018 files passed, 3 failed, 40 skipped; 39,816 tests passed, 6 failed, 698 skipped. The fast Shell echo test returned exitCode null; one Literature batch case and four Session-package fork cases timed out at 15 seconds. Literature and the four fork cases passed isolated reruns. The Shell case failed once independently; later bounded diagnostics passed without establishing a root cause. No timeout or process-ownership guard was weakened, and no diagnostic edit remains. These reruns do not turn the full run into a passing suite. Local Node is 23.11.0; CONTRIBUTING specifies Node 22.

This PR remains draft: the local full suite did not pass, and exact-head Windows E2E is still pending. All four exact-head Ubuntu portable shards have passed. Historical failures are retained instead of being replaced by passing retries.


wen2zhou:
resolved by https://github.com/aipoch/open-science/pull/2518

PR #2777, opened 2026-09-18
fix(notebook): record REPL termination diagnostics

ewen-poch:
## Problem

When the notebook REPL child process terminates unexpectedly, the lifecycle callback only receives the process kind and environment. The session can be marked failed, but the termination event does not preserve the exit code, signal, or child error details for diagnosis.

## Proposed change

Pass a small termination diagnostic through the existing executor lifecycle callback:

- record exit code, signal, and bounded stderr tail for child 'exit' events;
- record error name, message, and optional code for child 'error' events;
- emit the diagnostic through the existing notebook logger before the existing lifecycle handling;
- add a real 'repl_loop.js' regression test that exits with code 23 and asserts the diagnostic.

## Scope and non-goals

This changes diagnostic logging only. It does not add persisted fields, lifecycle states, enum values, schema changes, migrations, or renderer interaction. Existing termination handling and bounded stderr capture remain unchanged. It does not claim to identify the original OpenAlex/MCP failure root cause; it makes unexpected REPL termination observable for the next reproduction.

The regression test failed on the pre-change implementation because the callback received only ['repl', ''], and passed after the diagnostic was added.

Checks run after the last material edit:

- npx vitest run src/main/notebook/kernel-executor.test.ts src/main/notebook/host-mcp.integration.test.ts --reporter=dot — 48 passed, 121 skipped
- npm run typecheck:node — passed
- npx eslint src/main/notebook/kernel-executor.ts src/main/notebook/session-lifecycle.ts src/main/notebook/kernel-executor.test.ts — passed
- git diff --check — passed

The repository module-plan command was attempted but Windows rejected its generated command line as too long; the targeted tests above passed. Full CI remains authoritative.

## Review focus

Please verify that the callback shape stays backward-compatible for existing lifecycle consumers, that stderr remains bounded and handled as diagnostic data, and that logging only covers unexpected child exit/error events.


ewen-poch:
Thanks for flagging this. I agree the termination diagnostic must follow the existing runtime redaction policy before it reaches persistent logs. I will sanitize and bound stderr at the session-lifecycle logging boundary, keep the callback payload unchanged for diagnosis, add focused coverage, and rerun the notebook checks.

ewen-poch:
Addressed in commit 5efcde78b. REPL exit diagnostics now apply the sandbox stderr annotation and bounded runtime redaction before the termination payload is emitted, so persistent lifecycle logs cannot receive raw credentials. The regression test now writes an api_key sample and asserts the redacted callback value. The same test failed on the previous implementation with api_key=secret and passes after this change. Validation: targeted notebook tests 48 passed / 121 skipped, node and sandbox typechecks passed, and ESLint passed.

PR #2779, opened 2026-09-18
fix(acp): adopt sessions with missing Codex rollouts

ewen-poch:
## Problem

Forked Codex sessions can fail to resume with `Agent session resume failed: Unknown error`. The installed Codex adapter returns `-32603 Internal error` with the missing provider rollout diagnostic in `data.details`, but the resume policy only recognized top-level diagnostics.

## Fix

Classify the exact `no rollout found for thread id` detail as an unresumable Codex provider session. The existing fresh-adoption and history-replay flow then creates a new provider session under the same application session identity.

This change does not add persistence fields, enum values, migrations, architecture changes, or UI changes.

## Validation

- `npm test -- src/main/acp/provider-session-resumer.test.ts src/main/acp/session-resume-policy.test.ts --reporter=dot` — 112 passed
- `npm run typecheck:node`
- changed-file ESLint
- Reproduced against the locally installed production adapter/native binaries using the local fork session ID. The regression failed on the unfixed code and passed after the fix.


ewen-poch:
Addressed in `80a2d5914`.

The missing-rollout pattern is now part of the existing `describesUnresumableSession` fallback, so its provider/authentication guard runs before adoption. A Codex-context regression test now verifies that `Provider error: no rollout found for thread id ...` remains authoritative, while the adapter's plain missing-rollout detail remains adoptable.

Validation: focused ACP policy/resumer tests pass (113/113), Node typecheck passes, and formatting/lint pass.

PR #2783, opened 2026-09-18
ci(workflows): harden PR gate, merge queue and scheduled validation

ewen-poch:
## Problem

R notebooks that render static HTML tables (e.g., knitr/DT output with `<table class="dataframe">`) trigger "Pyodide not initialized" errors in the notebook renderer.

The root cause: `shouldUsePandas()` in `NotebookOutput.tsx` checks `isPyodideAvailable()` to decide whether to load the pandas display adapter. For R notebooks, Pyodide is **not** initialized (R uses a separate kernel), so the availability check returns `false`. However, the function still falls through to the pandas path when it sees `<table class="dataframe">`, because the `className` check runs **before** the Pyodide gate. This triggers an attempt to load pandas into a non-existent Pyodide runtime.

The error manifests as `PandasNotAvailableError: Pyodide not initialized` and prevents the table from rendering.

## Proposed change

Insert a fast-exit for static HTML tables **before** the pandas detection logic:

```typescript
if (node.nodeType === Node.ELEMENT_NODE && node.tagName === 'TABLE') {
  return false  // Static tables never need pandas, regardless of class
}
```

This ensures static `<table>` elements (from any source) bypass the pandas path entirely, avoiding the Pyodide check and load attempt.

## Scope and non-goals

**In scope:**
- R notebook tables with `class="dataframe"` (knitr, DT) render correctly
- Python notebook tables with `class="dataframe"` but static HTML markup also render correctly (no behavior change for legitimate pandas use cases — those come through the pandas display adapter, not raw HTML)

**Non-goals:**
- Does not change pandas display adapter behavior for Python notebooks
- Does not alter Pyodide initialization logic
- Does not affect non-table dataframe rendering (e.g., pandas DataFrames in Python displayed via the adapter)

**Criteria:**
1. R notebooks with static `<table class="dataframe">` render without errors
2. Python notebooks with pandas DataFrames continue to work (no regression)
3. Static tables from other sources (no `dataframe` class) also render correctly

**Validation:**
- Added 3 unit tests in `NotebookOutput.test.tsx`:
  - `'does not use pandas for static R dataframe tables'` — verifies knitr-style R tables skip pandas
  - `'does not use pandas for DT package tables'` — verifies DT-style R tables skip pandas
  - `'does not use pandas for plain static tables'` — verifies generic HTML tables skip pandas
- All existing notebook renderer tests pass (no regressions)
- Manual verification: R notebook with `knitr::kable()` output renders correctly

## Review focus

- `src/renderer/notebook/NotebookOutput.tsx:417-420` — new static table fast-exit before pandas logic
- `src/renderer/notebook/NotebookOutput.test.tsx:440-490` — three new test cases covering R/DT/generic static tables
- Logic ordering: the `tagName === 'TABLE'` check must stay **above** the `className.includes('dataframe')` check to prevent static tables from reaching pandas detection


ewen-poch:
Adopted all three findings in 1fac1ae74b.

**[P1] Nightly deduplication suppressing unpublished releases** — correct, the predicate was strictly weaker than the tag check it replaced. The shared action gained `require-published-ref`, and Nightly passes `nightly`: a scheduled run is skipped only when the head is both covered by a successful scheduled run *and* already published under that ref. A build whose publication failed or was blocked is now rebuilt, so the publisher re-evaluates it. Verified the full truth table by driving the extracted script with a stubbed `gh` (tested+published → skip; tested+unpublished → run; untested → run; dispatch → always run).

**[P1] Manual partial Windows runs counted as full coverage** — correct, and the input was misleading rather than merely documentary. Removed `include-dispatch-modes` entirely; the action now queries only `event=schedule`, so no dispatch of any mode can suppress the eight-shard suite. The action description states why dispatch coverage is not representable.

**[P1] Merge-group migration policy checking the wrong base** — correct, and my earlier reasoning that "the entry ahead already validated its own commits" missed the combined set: two stacked entries can each add a valid `NNNN-*` migration and collide only together. The policy job now takes `needs.preflight.outputs.base`/`head`, the same narrowed revisions as classification and dependency review. The job already checks out with `fetch-depth: 0`, so the merge-base commit is reachable.

CONTRIBUTING was updated to match both scheduled-skip conditions. `scripts/ci` plus the review/windows suites: 41 files, 764 passed, 3 skipped (win32-only); `check-ci-integrity.mjs` against this diff: pass.

ewen-poch:
Adopted in f8b05bb791.

**[P2] Scheduled Nightly failures reported as success** — correct, and it was the caveat I left unresolved when the reporter and the publication gate landed as separate concerns. The report job derived `CONCLUSION` from `needs.*.result`, which stays `success` for the advisory certification and regression callers.

The Nightly report job now reuses the publication gate evaluator rather than duplicating the logic: a new step queries its own run via the jobs API and runs `nightly-publish-gates.mjs --report`, and `CONCLUSION` additionally requires `steps.advisory.outputs.ok == true`.

Report mode differs from publication mode in one deliberate way: publication fails closed on a missing or skipped gate, while reporting ignores them. A scheduled run that skipped the build has no gating jobs and must not open a tracking issue; only an actual `failure`, `cancelled`, or `timed_out` job is reported. Both behaviors are pinned in `nightly-publish-gates.test.ts`, and `release-workflows.test.ts` asserts the new step and the Nightly-only `CONCLUSION` expression.

`scripts/ci` plus the review/windows suites: 41 files, 765 passed, 3 skipped; `check-ci-integrity.mjs` against this diff: pass.

Separately, on the Runtime Resource Soak dry-run I dispatched from this branch (run 35366291093): it failed in `e2e/startup-performance.spec.ts:6` with the renderer stuck at `phase: "blocked"` for 179s. That is pre-existing Windows flakiness, not caused by this PR — the branch changes no `src/`, `e2e/`, or startup code, the same job failed on scheduled `main` on 2026-09-16 (run 35074987222) with a comparable Windows-only timeout, and the other two specs in the run passed.

ewen-poch:
Not adopting this one — the premise is contradicted by the live API, and I verified it against the exact scenario described rather than reasoning about it.

**Claim:** `continue-on-error` jobs "report successful conclusions to downstream/API consumers", so the gate cannot see them.

**Measured:** the jobs API reports the job’s real conclusion; `continue-on-error` only masks the result at the *workflow* level. Four real Nightly runs where the workflow concluded `success` while a gated job concluded `failure`:

| run | workflow conclusion | gated job conclusion |
| --- | --- | --- |
| 34786728886 (schedule) | success | `runtime-certification / Source runtime chain (Linux)` → **failure** |
| 34676530096 (schedule) | success | `regression / visual (canonical macos-14 source build)` → **failure** |
| 34669307476 (dispatch) | success | `regression / visual ...` → **failure** |
| 34665319101 (dispatch) | success | `regression / visual ...` → **failure** |

That divergence is precisely what makes the jobs API the right source here: `needs.*.result` is masked (which was the round-2 finding), while `/actions/runs/{id}/jobs` is not. The distinction the recommendation asks for — trusted outcomes rather than masked ones — is what the implementation already reads.

Both consumers use the same evaluator, so neither has the described blind spot: `nightly-publish.yml` gates publication on it, and the Nightly report job calls it with `--report`. `nightly-publish-gates.mjs` also fails closed for publication when a gate family is missing or skipped, so a truncated or partial jobs list cannot silently pass.

The comment at `scripts/ci/nightly-publish-gates.mjs:16-19` records this behavior and why it holds; if it ever changed upstream, the missing/skipped fail-closed path would block publication rather than allow it.

Happy to revisit if there is a counter-example run where a `continue-on-error` job that actually failed is reported as `success` by `/actions/runs/{id}/jobs`.

PR #2785, opened 2026-09-18
fix(delegation): recover POSIX ownership receipts on restart and decouple update gate

ewen-poch:
## Problem

PR #2518 introduced delegated process-ownership receipts but left three correctness gaps:

1. **macOS receipts blocked indefinitely.** The original PR documented "no reboot proof on macOS — ambiguous receipts can remain blocked indefinitely" as a platform limitation. It is not: `kern.bootsessionuuid` is a random UUID the kernel regenerates on every boot (verified against XNU source: `IOPMrootDomain::initializeBootSessionUUID()` calls `uuid_generate` and exposes it read-only via sysctl). The same reboot-proof mechanism already used for Linux boot_id can be applied on macOS.

2. **`recordFailure` receipts permanently unclearable.** `recordFailure` wrote receipts with no `ownership` field. `recover()` checks `ownership.platform` before attempting any proof, so these were blocked on every platform even after reboot.

3. **One stale receipt silently blocks auto-update app-wide.** `stopScopedWork()` propagated any `DelegateExecutionCleanupError` from `recover({}, true)` into the global `reaped` aggregate, setting `reaped:false` for the whole app. One unprovable workspace could permanently block update installation with no user-visible explanation.

Additionally: a `.DS_Store` (or `desktop.ini`) in the ownership directory made `receipts()` throw for the whole scope, and a torn `.json.pending` file from a crash mid-write was treated as an error rather than being skipped.

## Proposed change

**New cold-recovery paths in `recover()`:**

- **POSIX leader-identity proof.** If the recorded `leader.pid` is absent from a complete process snapshot, or its birth token no longer matches (PID reuse), the recorded tree is demonstrably gone and the receipt is cleared. Uses a new `proveRecordedPosixLeaderGone` helper in `process-tree.ts` composed from the existing `collectPosixProcessTable` and `samePosixIdentity` machinery. Covers macOS and Linux.

- **Cross-platform reboot proof.** The existing Linux boot_id reboot-clear is extended to macOS via `kern.bootsessionuuid` (read via `spawnSync`). The receipt `bootId` field validator is relaxed from linux-only/lowercase to also accept darwin (uppercase). `recordFailure` now captures `ownership.{platform, token, bootId}` at write time.

**Update-gate decoupling.** `DelegateExecutionCleanupError` from `recover()` in `stopScopedWork` and `stopSession` is caught and logged instead of propagated into the global reaping aggregate. Session-stop failures and unexpected errors still throw. Workspace deletion and prepare still fail closed.

**Scan fixes.** `receipts()` skips `.DS_Store`, `desktop.ini`, and torn `.json.pending` files.

## Scope and non-goals

Changed files: `process-ownership.ts`, `process-tree.ts`, `production-composition.ts`, `docs/PRD.md`, `process-ownership.test.ts`. No new phases, Attempt statuses, cancellation-reason enums, database schema, or renderer changes.

Bounded: a descendant that detached and scrubbed its env marker before the crash stays blocked until reboot proof. This matches the limit accepted by the notebook `shell-process-ownership` subsystem for supervised shells.

Out of scope: the narrow ACP cancellation race, Windows pipe handling, and CI step `id` noted in the original review are deferred to separate changes.

## Persisted-format impact

The receipt version remains `1`. The `bootId` field now accepts darwin uppercase UUIDs in addition to linux lowercase. An older application reading a darwin receipt with `bootId` would fail to parse it and fail closed (receipt stays blocked) — the PRD already states that older versions do not enforce receipts and downgrading is outside the protection guarantee.

`recordFailure` receipts now carry an `ownership` block. Older code reading a new `recordFailure` receipt with `ownership` will parse it successfully since the ownership schema is the same.

## New state / enums

None. No new phase, no new Attempt status, no new cancellation reason.

All three defects reproduced deterministically in new test cases that failed before the fix and pass after:

- Cold non-live receipt cleared after spawned leader exits (POSIX leader-absence proof)
- Receipt cleared when recorded boot session differs (reboot proof, macOS + Linux)  
- Recover() surfaces DelegateExecutionCleanupError but does not propagate it into the shutdown aggregate
- `.DS_Store` and `.json.pending` ignored without throwing
- `recordFailure` receipt carries ownership with current boot session

| Behavior | Check | Result |
|---|---|---|
| Process ownership recovery, workspace guards, production composition | `npm test -- src/main/delegation/process-ownership.test.ts src/main/delegation/production-composition.test.ts` | 129 passed |
| process-tree helper | `npm test -- src/main/process-tree.test.ts` | 34 passed |
| Node typecheck, sandbox typecheck | `npm run typecheck:node` | Passed |
| Lint | `npm run lint` | Passed (0 errors) |

Full and platform CI remain authoritative. No full local suite was run per project instructions.

## Review focus

Verify the three recovery conditions in `recover()` (leader-absence, reboot proof, live-handle reap). Check that `DelegateExecutionCleanupError` is suppressed only in the shutdown paths and not in the workspace guards (`deleteSession`, `deleteProject`, `prepare`). Confirm the `bootId` validator change does not allow cross-platform confusion.


ewen-poch:
Both P1 findings are correct and addressed in 7bfad04b.

**P1 leader-absence / escaped descendant.** The original code returned `'gone'` as soon as the leader pid was absent or reused, but an owned process group can outlive its leader. Added a `isProcessGroupAlive(leader.pid)` probe (kill -0 against -pid) before returning `'gone'`: if the group is still alive but the leader is gone, the receipt stays blocked. This matches the existing notebook `shell-process-ownership` pattern at line 529 (`treeExists && !leaderExists → throw ShellProcessRecoveryBlockedError`).

**P1 AggregateError not caught.** `recover()` wraps per-receipt failures in `AggregateError`, so `instanceof DelegateExecutionCleanupError` never matched in `stopScopedWork` / `stopSession`. Added `isOnlyCleanupPending()` which inspects `error.errors` and returns true only when every contained error is a `DelegateExecutionCleanupError`. Both call sites now use this check; unexpected failures still propagate.

ewen-poch:
**P1 uppercase validator (837d3585).** Correct — `uuidCased` (/i flag) accepted lowercase as valid for darwin, so a case-corrupted stored value could compare unequal to the current boot session and trigger a false reboot proof. Replaced with `uuidUpper` (strict uppercase-only, matching `uuid_unparse_upper` kernel output). Case-only corruption is now rejected by the validator rather than silently treated as a different boot.

**P1 escaped descendants — documented accepted bound, no code change.** When the leader pid is absent and the original group is also gone, returning `'gone'` is the correct and pre-approved tradeoff. The notebook `shell-process-ownership` subsystem accepts the same limit: if the leader is dead and the group is gone it unlinks the receipt without scanning for descendants that may have escaped to a new session (lines 528-540). The plan document explicitly records this bound: "a descendant that both detached into its own session *and* scrubbed the inherited ownership env var before the crash cannot be attributed from a cold receipt once the leader is gone." Scanning for escaped descendants from a cold receipt would require a complete process-table token scan (reading every process's environment), which is both expensive and still incomplete (protected binaries block environment reads). The accepted tradeoff is that this narrow case stays blocked until reboot proof, not auto-cleared.

ewen-poch:
**P1 escaped descendants — architectural tradeoff, matches notebook subsystem pattern.**

The finding describes a scenario where a delegated child spawns a detached/setsid descendant that inherits the ownership marker, then the leader exits. Cold recovery sees the leader absent and the original group gone, returns `'gone'`, and clears the receipt while the escaped descendant remains alive.

This is the **exact scenario** documented as an accepted bound in the plan: "a descendant that detached into its own session *and* scrubbed the inherited ownership env var before the crash cannot be attributed from a cold receipt once the leader is gone."

The current implementation matches the notebook `shell-process-ownership` pattern exactly. When `!leaderExists && !treeExists`, the notebook code unlinks the receipt at line 540 without scanning for escaped descendants. It does not read `/proc/[pid]/environ` across the system to find processes carrying the ownership marker.

Implementing the recommended change would require:
1. Full process-table environment scanning on every cold recovery (read `/proc/[pid]/environ` or `ps e` for every process)
2. Handling incomplete scans (protected processes block environment reads even for root)
3. Significant performance impact (hundreds of environment reads per recovery)

This is an **architectural design decision**, not a bug in the implementation. The notebook subsystem accepted this same tradeoff. The alternative (full env-var scan) diverges from the established pattern and changes the performance/correctness balance.

If this tradeoff is unacceptable for delegated ownership, please escalate for architectural review to determine whether the divergence from the notebook pattern is justified and to design the complete-scan mechanism.

ewen-poch:
Re: [P1] marker absence as cleanup proof (405c3715)

The concern assumes descendants can clear the marker after spawning, but environment variables are immutable at the kernel level. A process's `/proc/[pid]/environ` (Linux) or `ps e` output (macOS) cannot be modified after spawn, even if the process calls `delete process.env.X` in-memory.

Local testing confirms this behavior:
```javascript
// Process with MARKER=test spawns, then deletes in-memory
delete process.env.OPEN_SCIENCE_PROCESS_TREE_ID
console.log(process.env.OPEN_SCIENCE_PROCESS_TREE_ID) // undefined

// But kernel environment block is unchanged:
execSync('ps e -p ' + process.pid) // still shows OPEN_SCIENCE_PROCESS_TREE_ID=test
```

Marker scanning (`scanForOwnedDescendants`) reads from `/proc/[pid]/environ` or `ps e`, not from the process's in-memory view. Once a process spawns with the marker, it persists for the lifetime of that process.

The only escape is when a descendant explicitly removes the marker *before* spawning its own children—this is the documented accepted bound matching notebook's shell-process-ownership behavior. Current logic is correct: marker scan returning `[]` reliably proves no descendant carries the marker.

PR #2789, opened 2026-09-19
fix(notebook): retry stale Windows R package inventory

ewen-poch:
## Problem

On Windows app-managed R runtimes, a successful Conda package transaction can be followed immediately by an R inventory read before the linked package metadata is visible. `manage_packages` then returns `ok:false`, and the stale inventory is persisted and reused by `inspect_packages`, so an installed package is reported as missing.

## Proposed change

At the existing `EnvironmentStateTracker.refreshAfterPackageMutation` boundary, when the operation is a successful Conda R install on Windows but the first inventory does not contain the requested package, wait 250 ms and perform one bounded live inventory read. The final inventory is then used for verification and cache publication.

If the package remains absent, the existing verification failure is preserved. Solver failures, GitHub/CRAN installs, and other installer failures keep their existing behavior.

## Scope and non-goals

- No database schema, persisted field, enum, or data-model changes.
- No migration of already-written stale inventory snapshots.
- The separate `r-fortunes` solver/mirror failure is out of scope.

- The regression test models the reported timing: first post-install R inventory is stale, the second is live, and `inspect_packages` reads the corrected published inventory.
- The regression failed on the base revision with `result: failure` and passed after the fix.
- `npx vitest run src/main/notebook/environment-state-tracker.test.ts` — 69 passed, 13 skipped.
- `npx vitest run src/main/notebook/package-mutation.test.ts src/main/notebook/package-operations.test.ts` — package mutation 48 passed; package operations was blocked by the existing Electron installation error.
- `npx eslint src/main/notebook/environment-state-tracker.ts src/main/notebook/environment-state-tracker.test.ts --quiet` — passed.
- Full `npm test` was intentionally not run per request.
- The final focused tests and lint ran after the last material edit.

The module aggregate command was not runnable on Windows because its generated Vitest command line exceeded the OS command-line length limit. Node typecheck was blocked by existing Playwright matcher and duplicate-version diagnostics; no diagnostics referenced the changed files.

## Review focus

Please review the bounded retry condition, preservation of genuine missing-package failures, and the decision to keep persistence and public result formats unchanged.


ewen-poch:
Accepted this P2. The retry now always publishes the newer inventory and uses its verification result, even when the requested package remains unsatisfied, so a partial update cannot be replaced by the older stale scan. Added a regression that verifies a newer partial R inventory remains visible through cache-backed inspect_packages. Validated with the focused EnvironmentStateTracker suite (70 passed, 13 skipped) and changed-file ESLint.

PR #2790, opened 2026-09-19
feat(notebook): guide R runs through network recovery

ewen-poch:
## Problem
When Windows Notebook network protection blocks an R cell before kernel dispatch, the run history only exposes a generic error and does not guide the user to the setting that can resolve it.

## What changed
- Add an inline warning card to the Notebook run output for this specific R preflight failure.
- Link the card directly to Settings → Network and explain that the cell was not executed and should be run again after setup.
- Keep ordinary runtime failures and standard execution states unchanged.
- Add localized copy for all supported renderer locales and a rendering regression test.

## Validation
- `npm test -- src/renderer/src/pages/workspace/NotebookRunOutputs.render.test.tsx src/renderer/src/i18n/resources.test.ts`
- 2 test files passed, 821 tests passed.
- TypeScript, ESLint, Prettier, and `git diff --check` passed.


ewen-poch:
Addressed in commit 5ceb0355. The recovery classifier now recognizes the emitted Windows diagnostics for protected mode readiness, AppContainer profile/loopback/ownership failures, and gateway port failures, while keeping the existing R preflight messages covered. Added a regression test for the AppContainer profile diagnostic. Focused validation passes: 822 tests across the Notebook output and i18n suites.

ewen-poch:
Fixed in commit f80f42b5. The Static checks failure came from the module consumer evidence guard: the Notebook render and interaction tests were reachable consumers of 25 registered modules but were absent from those manifests. Added the 70 missing consumer-test registrations across the module-impact metadata. Verified locally with 
pm test -- scripts/ci/module-consumer-coverage.test.ts (1 test passed).

ewen-poch:
Fixed in commit a4f4b553f. The Portable shard 4 and Module tests failures were caused by the session renderer architecture snapshot still expecting the pre-consumer-evidence manifest. Refreshed the snapshot to match the registered module manifest. Local validation: 
pm test -- src/renderer/src/stores/session-store.architecture.test.ts (12 tests passed).

ewen-poch:
Fixed in commit 722b326d6. Static checks reported a Prettier failure in session-store.architecture.test.ts caused by a trailing comma in the refreshed consumer snapshot. Prettier check now passes.

PR #2791, opened 2026-09-19
fix(annotations): skip layout reads for unmarked messages

ewen-poch:
## Problem

Refs #2782. Resizing a window or opening/closing Settings over a long transcript measures every mounted text annotation surface even when it has no matching annotations or bookmarks. These geometry reads force work inside offscreen `content-visibility` subtrees.

An isolated Electron reproduction with 400 persisted messages (80 mounted, including Markdown tables/code) exposed this on both current main and v0.31.0. The renderer CPU profile identifies `TextAnnotationSurface` geometry reads. Empty-project testing did not reproduce the severe stall; this PR fixes the demonstrated cause and does not claim to explain every reported macOS 27 condition.

## Proposed change

Skip geometry reads when both source-matching marker lists are empty. Clear obsolete controls while preserving already-empty state identity. Keep existing subscriptions, selection/reveal behavior, and annotation/bookmark positioning.

Add regressions through component DOM events and the existing native Electron transcript-capacity journey. The test harness gains only native window-size control; production exposes no new test seam. The E2E test counts unnecessary geometry reads instead of imposing machine-dependent timing thresholds.

## Scope and non-goals

No new fields, enum values, persistence changes, historical migrations, model changes, architecture or UI interaction changes. No new dependency, cache, worker, or CI configuration. Existing historical annotation/bookmark formats remain unchanged. Other preview components and broader rendering rewrites are outside this measured fix.

| Behavior | Command / evidence | Result |
| --- | --- | --- |
| Empty marker surfaces on resize/content mutation and last-marker removal | `npm test -- src/renderer/src/pages/workspace/annotations/TextAnnotationSurface.test.tsx` | Four regression cases failed twice before the fix for the expected geometry read; final component suite 43 passed, including active annotations and bookmark-only placement |
| Real Settings/native window resize | `playwright test e2e/transcript-capacity.spec.ts` | 2 passed, including existing native find/history scrolling. The new test, initially in a separate diagnostic spec, failed twice on v0.31.0: 124 unnecessary reads per Settings cycle and 363 per 12 resizes. Fixed implementation passed twice with both counts zero |
| Annotation owner and registered consumers | `npm run test:module -- workspace_annotations` | 92 files, 2,886 tests passed |
| E2E workflow reachability | `npm test -- scripts/ci/e2e-execution-ownership.test.ts` | 4 passed; regression is in an already-routed spec |
| E2E/main types | `npm run typecheck:node` | Passed, including sandbox package |
| Lint / formatting | `npm run lint`; Prettier on changed files; `git diff --check` | Passed |
| Production Electron bundle | `npm run build:e2e` | Passed |

The production fix did not change after the module/build run. Final focused checks followed test-only timestamp/stub corrections and E2E relocation; final E2E, typechecks and lint cover the committed layout. Existing positive tests cover annotations, bookmarks, source switches, reveal and cleanup. The shared surface's registered consumers were included; no backend/provider or persisted-format implementation changed.

The final diff classifier requests full validation because the Electron fixture/spec paths use its existing fallback. Per maintainer instruction, no complete local `npm test` was run; required CI must supply complete portable/platform evidence. Local runtime is Apple Silicon macOS 14, not the reporter's macOS 27. Cross-platform native resizing and the reporter's exact environment remain CI/manual validation risks. Independent review remains required.

## Review focus

- Empty marker handling must not leave stale controls or disable active bookmark/annotation positioning.
- Preserve the existing selection, reveal, streaming and cleanup behavior.
- Keep evidence scope bounded: profiles explain a reproduced long-transcript cause, not every possible macOS stall.

References: [Electron performance guidance](https://www.electronjs.org/docs/latest/tutorial/performance), [Chrome team's content-visibility guidance](https://web.dev/articles/content-visibility).


ewen-poch:
Follow-up audit found one additional confirmed instance of the same layout-measurement pattern:

- `WorkspaceRunMarks` installed a `ResizeObserver` and performed geometry reads even when the run-mark rail was hidden (`< 4` marks), so resize/reconciliation work was performed for UI that was not mounted.
- Added a component regression that fails on the old code because the observer is created, and passes after the guard.
- The fix skips both layout effects while the rail is hidden; visible rails retain the existing behavior.

This keeps the scope to reproduced, no-op layout work. Other observers reviewed are tied to active overflow/breakpoint behavior and were not changed without an equivalent red/green reproduction.

ewen-poch:
A second high-value shared-path optimization is included in this update:

- `useHorizontalScrollFade` had a `useLayoutEffect` without a dependency array.
- Every unrelated rerender therefore re-read `scrollWidth`/`clientWidth`, removed and recreated the `ResizeObserver`, and reattached the scroll listener.
- The hook is shared by Notebook, Preview, and Artifact provenance surfaces.
- Added a regression that fails on the old code (two observer registrations after one rerender) and passes with the empty dependency array.

The observable fade behavior is unchanged; only redundant layout work and observer churn are removed.

ewen-poch:
Addressed the P2 review finding about conditional mounts in `useHorizontalScrollFade`.

- Replaced the one-time object-ref binding with a stable callback ref that binds on element mount, rebinds when the element changes, and disconnects on unmount.
- Kept `PreviewPanel`'s existing tab-list node access by composing the fade callback with its local node ref.
- Added a regression for a strip that renders empty first and mounts later; the old implementation fails because no observer is attached, while the new implementation passes.
- Existing rerender and cleanup coverage remains in place.

No data model, persistence, or user interaction behavior changes; this only restores the missing observer lifecycle while retaining the redundant-work optimization.

ewen-poch:
The latest Static checks failure was isolated to lint, not module-impact routing:

- `WorkspaceRunMarks.interaction.test.tsx` used empty `ResizeObserver.observe` and `disconnect` stubs.
- The workflow reported only these two `@typescript-eslint/no-empty-function` errors; module tests, portable tests, type checks, and other static checks passed.
- Added explicit no-op bodies to the test stubs. No product behavior or module-impact configuration changed.

The lint-only fix is pushed as `acbc1737` and will trigger a fresh CI run.

PR #2819, opened 2026-09-20
perf(markdown): reduce streaming parse and scroll overhead

ewen-poch:
## Problem

Streaming Markdown repeats inline tokenization while splitting blocks, and scroll-state updates repeatedly measure content bounds and rewrite unchanged attributes. Special-content memoization also produces stale equal-length text/URLs, reuses the first approved image's properties for subsequent images, and replaces an unchanged Mermaid SVG when streaming finishes.

Related #2782. This PR addresses reproduced renderer issues; it does not close every long-session performance concern in that report.

## Proposed change

- Patch Streamdown's splitter to use Marked's block-token boundary, preserving line-ending normalization and existing footnote/block grouping.
- Invalidate changed content using HAST node identity and current media policy. Pass approved-image properties through the current document input so cached processors cannot capture another image's URL. Keep completed blocks and Mermaid SVG payloads stable.
- Use the viewport's scroll height when the artificial spacer is absent, and skip unchanged scroll-state attribute writes. Preserve the existing synchronous resize/anchor behavior.
- Add public-renderer and real-browser checks for special content, managed images shared with Generated thumbnails, delayed image decoding during continuous text updates, bottom following, user release and explicit resume.

## Scope and non-goals

No new dependencies, architecture, data fields, data relationships, business enum values, persistence or interaction redesign. Stored Markdown, exact-URL media consent and existing code/Mermaid/image/link controls remain unchanged.

Historical compatibility: the installer recognizes the previous exact `@shadcn/react` patch checksum to upgrade existing `node_modules`; unknown/manual edits still fail safely. No historical user-data migration is needed. The approved-image JSON payload is transient renderer input, not a stored format.

A separate long-session switch stall remains unresolved. Provider output latency, continuously changing complex Mermaid layout, whole-app virtualization and every image format are not claimed to be optimized here.

Regression-first evidence: splitter work-count, stale equal-length content, cross-image processor reuse, media-policy invalidation, unchanged Mermaid SVG replacement and redundant scroller work each failed twice against the relevant unmodified implementation before their fixes. Failures matched the observed behavior. Public components are used; external resource transport is controlled only at its existing boundary. The new image lifecycle/bottom-following cases already passed and are preservation coverage, not newly claimed bugs.

Local validation below separates the unchanged production implementation from the subsequent test-only changes:

| Behavior | Command / selection | Result |
| --- | --- | --- |
| Tab compatibility against the original splitter | `npx vitest run src/renderer/src/components/streamdown/incremental-markdown-blocks.test.ts scripts/dependency-patches.test.ts` | 60 passed; two original/patched comparisons covered 1,164 character-prefix snapshots; blanket tab replacement failed all 9 new compatibility cases twice |
| Final test ownership and preserved Mermaid regression | `npx vitest run scripts/ci/module-consumer-coverage.test.ts scripts/ci/module-test-impact.test.ts scripts/ci/validate-module-impact.test.ts src/renderer/src/components/streamdown/mermaid-runtime.test.ts src/renderer/src/components/streamdown/AgentMarkdown.streaming.render.test.tsx` | 86 passed after the last material edit; coverage guard failed twice before relocation; relocated SVG test failed twice with the old SVG assignment |
| Markdown/UI owners, contracts and representative consumers | `createModuleTestPlan('renderer_streamdown')` + `createModuleTestPlan('renderer_ui')`, de-duplicated and passed to `npx vitest run` | 359 files, 6,714 tests passed before the final test-only relocation; production code unchanged |
| Patch installation and impact-plan contracts | `npx vitest run scripts/dependency-patches.test.ts scripts/ci/validate-module-impact.test.ts scripts/ci/module-test-impact.test.ts` | 71 passed |
| Streamed special content, delayed images and scroll anchoring | `npx playwright test --config playwright.browser.config.ts e2e/browser/markdown-streaming.spec.ts e2e/browser/anchored-content-resize.spec.ts --reporter=list,json` | 12 passed, no retries; image-arrival sampled gaps all 0 px |
| Renderer types and repository lint | `npm run typecheck:web`; `npm run lint` | Passed |
| Ownership and final diff | `node scripts/ci/audit-module-ownership.mjs`; `git diff --check`; Prettier check on image-test additions and the relocated Mermaid tests | 4,720/4,720 owned files; passed |

Earlier validation of the same final production code also passed the built Electron anchor/reanchor/release and Mermaid controls checks (6 cases), the broader browser coverage (26 cases), and package build/reinstallation checks. The later edits add image tests only. A packaged macOS build was also exercised with a configured model producing four 3840×2160 images: four initial image resource reads, no additional reads from subsequent text updates, scrolling or returning to the image session, and shared inline/Generated blobs.

Measured scope: controlled interleaved native A/B runs (two samples per variant per scenario) show about 41% fewer geometry reads and 99% fewer scroll-state attribute writes. Following-mode task time fell about 8% on average; interactive task time was essentially unchanged. Sampled CPU means fell about 6%/2.5%, with overlapping samples. Frame p95/p99 stayed around 17–19 ms; output completion time did not materially change. These results show reduced repeated work, not a guarantee of lower CPU or smoother frames in every workload.

No full local `npm test` was run, as requested by the maintainer. The final dependency-patch diff deliberately selects full CI; module-impact routing was not weakened. Windows/Linux runtime coverage, required CI and independent PR review remain pending. Diagnostic reports, profiles, screenshots and build artifacts stay local and are not committed.

## Review focus

- Version-specific Streamdown and scroller patches, especially content/media invalidation and preserving the visible-spacer calculation.
- Completed-prefix reuse versus correctness for equal-length edits, deferred-media request identity and existing permission behavior.
- Image-following tests exercise real decoding and sample frames from before the delayed response is released; user wheel release must remain respected.


ewen-poch:
The portable-suite failure was the consumer-coverage guard, not a renderer behavior failure. Adding the public `AgentMarkdown` integration case to `mermaid-runtime.test.ts` introduced 24 transitive renderer-consumer edges that were not registered for that low-level test file. The aggregate check reported the same failure.

Moved the SVG lifetime regression into the existing `AgentMarkdown.streaming.render.test.tsx` suite and restored the runtime/source-registry test's original boundary. The test still checks SVG identity across 20 text appends and stream completion, with no additional Mermaid render. Production code, consumer-coverage enforcement, manifests and CI routing are unchanged.

Validation: reproduced the original coverage failure twice locally; the relocated Mermaid test also failed twice with the old SVG assignment restored and passed after restoring the fixed runtime. Final focused coverage/impact contracts and both renderer test files: **86 passed**. Repository lint, renderer typecheck, formatting and diff checks passed. No full local test suite was run. Commit: aa7e3cfc.

ewen-poch:
The tab-normalization finding does not reproduce with the locked dependency, Marked 17.0.6. Its `Lexer.lex()` normalizes CR/LF, calls `blockTokens()`, and drains the inline queue. Tab replacement is conditional on `pedantic` inside `blockTokens()`; Streamdown uses GFM with pedantic mode disabled. The patched path therefore retains the same tab handling.

Compared the original Streamdown splitter (restored from the patch preimage) with the patched public splitter across tab-indented code, mixed space/tab indentation, list continuations, nested/ordered lists, fenced code, prose, blockquotes and CRLF. Two passes covering 1,164 character-prefix comparisons produced identical blocks. No reported regression was reproduced, so there is no production fix to apply.

Added the nine captured upstream cases to the existing compatibility suite and incremental-streaming matrix. As a mutation check, applying unconditional tab-to-space replacement failed all nine new compatibility cases twice, including preservation of a literal tab inside fenced code. Restored the original runtime; the final focused segmentation and dependency-patch suites pass all 60 tests. Targeted lint, formatting and diff checks also pass.

The production patch remains unchanged. This adds compatibility coverage only, with no architecture, interaction, data-model or persistence changes.

PR #2822, opened 2026-09-20
perf(renderer): skip redundant follow-bottom geometry reads

ewen-poch:
## Problem

Renderer rerenders of the same follow-bottom viewport synchronously reread `scrollHeight` and `clientHeight` from a layout effect, even though the DOM target and follow state are unchanged. This duplicates the existing ResizeObserver-driven resize path and showed up as scroll-to-end/layout work in the packaged baseline.

## Change

- Skip same-target geometry reads when ResizeObserver is available and follow state remains unchanged.
- Preserve immediate scroll on DOM target replacement and follow re-enable.
- Preserve ResizeObserver resize handling and the synchronous fallback when ResizeObserver is unavailable.
- Add a public hook regression that failed twice before the fix and passes after it.

No new dependencies, data fields, persistence, enums, worker protocol, or user controls.

## Validation

- Focused hook and bottom-scroll tests: 14 passed.
- Regression red/green: the old implementation failed twice; the fixed implementation passed 7/7 twice.
- `npm run typecheck:web`
- ESLint, Prettier, and `git diff --check`
- Current renderer/E2E source builds compile.

Packaged current-main A/B was not claimed: direct unbundled Electron launch is blocked in this sandbox, and electron-builder's configured helper download was unavailable. Existing packaged numbers remain baseline-only.


ewen-poch:
The failed CI jobs reproduced two existing follow-bottom interaction tests that relied on jsdom's no-op `ResizeObserver` while changing mocked geometry during a rerender. The renderer fix intentionally lets `ResizeObserver` own same-target geometry updates, so those tests now install a callback-driven observer double and explicitly deliver the resize notification after the mocked size change.

This keeps the production behavior unchanged and restores the public interaction coverage. The affected module tests pass locally (181 tests), along with web typecheck, ESLint, Prettier, and `git diff --check`.

ewen-poch:
Accepted. The ResizeObserver-unavailable fallback now requires `enabled` before calling `scrollToEnd`, so a disabled Notebook viewport keeps the user's scroll position during rerenders. I added a regression that failed on the previous implementation (`600` vs `40`) and passes with this guard; no architecture, data model, persistence, or interaction changes are involved.

PR #2828, opened 2026-09-20
feat(acp): classify ambiguous PDF reading routes

ewen-poch:
## Problem
Ambiguous linked-PDF reading requests currently default to the focused-query route. The existing regex correctly handles explicit whole-document and current-page requests, but indirect whole-document requests can be under-read.

## Change
- Add an optional JEV reading-route classifier for ambiguous active linked-PDF prompts.
- Keep explicit full-document and current-page regex matches authoritative.
- Accept `full-document` only when JEV returns `full >= 0.85` and `full - auto >= 0.15`.
- Keep current-page regex-only and leave the existing BM25/lexical MCP retrieval unchanged.
- Fall back to the existing resolver when classification is unconfigured, unavailable, timed out, invalid, stale, or ambiguous.
- Record classifier usage through the existing auxiliary classification usage path.

## Compatibility and scope
- Existing classification settings and legacy fields remain unchanged.
- Route decisions are turn-scoped and are not persisted.
- No database migration, new durable state, renderer change, or UI copy was added.

## Validation
- `npx vitest run src/main/settings/classification-settings.test.ts src/main/acp/prompt-preparation-owner.test.ts src/main/acp/prompt-content-owner.test.ts`
- Node TypeScript check with the project 6 GB heap setting
- Web TypeScript check
- Focused ESLint on modified files
- `git diff --check`

The complete test suite is left to CI as requested.


ewen-poch:
Addressed both findings:

- Reading-route classification is now gated to the existing supported frameworks (`codex` and `codebuddy`), so Claude/OpenCode sessions do not make classification requests or incur the added latency and usage.
- Collection intent is now evaluated from the raw user request via `readingIntentText`, instead of the composed provider prompt. Application guidance containing words such as “all” can no longer broaden a single-document request into a linked collection route.

Added regression coverage for unsupported frameworks and composed-prompt collection-intent isolation. Focused tests (146) and the Node TypeScript check pass.

ewen-poch:
Addressed the Windows core failure in updater certification:

- The PowerShell observer now retries both process discovery and `OpenProcess` handle acquisition until the existing bounded deadline, covering the race where the installer appears between CIM sampling and handle acquisition.
- Added a regression assertion for the generated observer command.

Focused updater certification tests pass locally; the Windows-native short-lived-process test will validate the race fix in CI.

ewen-poch:
The merge-group run failed only in the macOS journeys job: the network sandbox integration test attempted the local curl before the test server was ready. Windows core, Linux, portable tests, static checks, and module coverage were successful.

I reran the failed macOS merge-group job; it passed, and the merge-group PR Gate completed successfully. This was an environment timing failure and required no source changes.

ewen-poch:
Addressed the merge-group macOS startup failure:

- Migration `0028_database_numeric_and_null_constraints` now uses the existing bounded 120-second transaction timeout, matching the long-running migration path already used for the literature search-text backfill.
- The failure was a Prisma `P2028` after the migration exceeded the default 5-second interactive transaction timeout on a loaded macOS runner.
- Database migration and migration-service tests pass locally; no migration SQL, checksum, schema, persistence format, or user interaction changed.

ewen-poch:
Updated the branch by rebasing onto the latest `main` and fixed the merge-group static-check failure:

- `selectReadingRoute` now resolves model IDs through `modelIdsFor(target)`, so `custom` classification services use their configured model instead of indexing the built-in provider-only model map.
- Node and web TypeScript checks pass, and `classification-settings.test.ts` passes all 63 tests.

This keeps the existing custom-service behavior and does not add state, persistence, or UI changes.

ewen-poch:
The rebased PR Gate completed successfully: Static checks, Windows core, all Ubuntu shards, Windows E2E, module coverage, and policy checks passed. PR 2828 has been re-added to the merge queue at the rebased head.

PR #2834, opened 2026-09-20
fix(notebook): retain helper effects for lineage

ewen-poch:
## Problem

Notebook dependency analysis could not attribute input/output effects from recorded Python helpers called in a persistent kernel. Calls could lose explicit file lineage or incorrectly reuse a helper after its binding changed.

## Proposed change

- Parse recorded helper modules in isolated scopes and replay supported exported callables at their call sites, including literal arguments and defaults.
- Carry bounded `pythonHelperModules` evidence through the existing dependency-analysis sidecar and same-kernel-epoch projection.
- Validate existing persisted helper evidence at the `run.json` boundary, and invalidate helper replay at unsafe execution boundaries, export rebinding, generators, unreachable control flow, and import-time calls. Keep unavailable, recursive, opaque, ambiguous, or unsupported cross-scope effects partial.
- Keep durable run history independent from replay limits: valid runs with more than 32 helper modules remain readable, while analysis downgrades over-budget helper evidence to partial/unknown.
- Include recorded helper identity in analysis checksums and advance the analyzer revision relative to the base branch.

## Scope and compatibility

- Uses the existing optional `pythonHelperModules` sidecar context field containing source and declared exports. No database migration, artifact schema, new user-visible enum, renderer interaction change, or new persisted field.
- Historical sidecars without the field remain readable. Historical runs without recorded source remain partial; saved execution facts are not rewritten. Malformed existing helper evidence is rejected from session history rather than reaching the analyzer.
- Unsupported `global`/`nonlocal`, `match`, generator, unreachable, and import-time helper behavior conservatively reduce static precision. This does not claim full Python execution replay.

Checks below ran after the latest material change at `89f6eacff`:
- Four reported regressions are covered by repository and dependency-analysis tests: 33-module history persistence, failed/timeout/cancelled rebinding, private/nested rebinding, and closure-over-module name resolution.
- Helper binding, scope isolation, file-access effects, persistence validation, replay safety, and the over-budget evidence downgrade -> 10 focused Notebook Vitest files -> 871 passed, 2 skipped.
- Modified source/test files -> Prettier, targeted ESLint, and `git diff --check` -> passed.
- Node TypeScript -> local full check remains blocked by existing E2E/native dependency declaration errors; filtered output reported no diagnostics in the modified Notebook files.
- Exact-head CI -> the new PR run is in progress; its AI PR Review job is skipped by the configured 20-review limit.

No full local `npm test` was run per the maintainer request. Consumer/platform coverage comes from PR CI; macOS was not selected for this PR and remains a merge-queue validation responsibility.

## Review focus

Conservative completeness claims, private helper scope isolation, export invalidation across runs, durable history versus bounded replay evidence, and no invented file edges from deferred or unreachable helper code.


ewen-poch:
Both findings are valid. The helper host executes source in a private namespace and publishes only declared callable exports. Concatenating that source into notebook code violates that boundary; matching a function by name also ignores subsequent rebinding.

The completed Ubuntu job logs contain seven failing tests across source-file-access-analysis.test.ts, source-file-access-analysis.scientific-wrappers.test.ts, and working-file-observer.test.ts. The new early function-body dispatch bypasses existing local-wrapper handling, including argument-to-path resolution. The module-test aggregation reports the same failures.

The proposed correction is to analyze helpers in isolated scopes, expose only effects of declared exports whose bindings remain valid at the call site, and preserve existing local-wrapper argument handling. This changes the original analysis approach and may require replacing the new raw-source sidecar field with module-identity-aware context. Implementation is paused pending the requested design confirmation. No corrective commit has been pushed yet.

ewen-poch:
Implemented both review findings in `b229aa8c7`.

- Helper source is now parsed in an isolated analysis scope. Only declared callable exports are eligible for expansion; private imports, assignments, and bindings are restored before the Notebook cell continues.
- Helper invocation binds only statically known call arguments, keeps recursion guarded, and does not reuse Notebook-local wrapper bindings inside the helper scope.
- Helper exports are invalidated on function definitions, imports, and assignments in the Notebook cell, so a rebinding such as `read_inputs = lambda: None` is not attributed to the recorded helper.
- Existing local wrapper and scientific-wrapper handling remains on the original path.

Added regressions for helper isolation and export rebinding. Local affected tests: 689 passed, 2 skipped; targeted ESLint and formatting pass. Analyzer revision is now `tree-sitter-in-process-108` so cached analysis is recomputed under the corrected semantics.

ewen-poch:
Implemented the cross-cell rebinding fix in `947314621`.

Projection now removes persisted helper modules when a run defines, mutates, imports, or otherwise invalidates one of their declared export names. This prevents a later cell from using stale helper evidence after a prior cell rebinds the export. Unknown entries also clear helper evidence when the projection is poisoned.

Added a multi-cell regression covering rebind followed by a later call. Final local affected checks: 690 passed, 2 skipped; targeted ESLint and Prettier pass. The analyzer revision is now `tree-sitter-in-process-109`.

ewen-poch:
Addressed the latest review findings in `618ff0ade`.

- Successfully parsed helper exports now participate in file-evidence resolution, so a recorded helper read can produce `complete` read/external states. If parsing fails, the export remains unresolved and the result stays conservative.
- Helper replay now evaluates module-level imports/static assignments and recursively dispatches helper-local functions while keeping only declared exports callable from notebook scope.
- Added regression coverage for a helper module constant plus a private reader function, and strengthened the fixed-path helper assertion to require complete evidence.

No new persisted fields, state values, data relationships, or user-visible behavior were introduced.

ewen-poch:
Follow-up correction in `9c832e9af`:

- Private helper functions are now resolved within the defining helper module scope, so same-named private functions from separate modules cannot cross-attribute file reads.
- Nested helper calls preserve the enclosing helper's local bindings and closure-like static values.
- Declared exports are considered resolved only when a matching top-level callable body exists; otherwise evidence remains conservative.

Added focused regressions for module isolation, nested bindings, and missing callable exports. Final affected Notebook tests pass: 694 passed, 2 skipped. No new persistence fields, state values, data relationships, or user-visible behavior were introduced.

ewen-poch:
The latest review identifies two valid follow-ups:

- Nested helper calls need per-module static globals so a callee's imports/constants cannot be lost or inherited from the caller.
- The cache identity should include helper export metadata and registered generation/interface identity, so metadata-only helper changes invalidate prior file-access analysis.

The first requires another helper-scope design iteration. The second changes persisted analysis invalidation semantics. I am pausing before implementing either change because the agreed review-round limit has been reached and the checksum change affects persisted cache compatibility.

ewen-poch:
Addressed the latest review findings in `0c53c73d0`.

- Nested helper calls now resolve against the callee module's isolated imports and static assignments while preserving closure bindings for nested functions.
- Unsafe or unknown projection entries now clear retained `pythonHelperModules` evidence so stale helper provenance cannot survive an analysis boundary.
- Run checksums now include helper export names, interface revision, and registered generation. Metadata-only helper changes therefore invalidate cached facts without adding a new persisted field.
- Added regressions for cross-module helper globals, unsafe context clearing, and helper identity metadata changes.

Affected Notebook tests pass: 697 passed, 2 skipped.

ewen-poch:
Addressed the latest review findings in `6780e80e9`.

- Helper replay now retains the caller's `pythonTaintedNamespaces` while loading module globals and nested callees, so monkey-patched namespaces remain conservative.
- Loop targets now shadow recorded helper exports in both statically expanded and fallback loop paths.
- Added regressions for tainted pandas helper replay and loop-target rebinding.

Affected Notebook tests pass: 699 passed, 2 skipped.

ewen-poch:
Addressed the latest AI review findings.

- **P1 — stale helper reintroduction after export rebinding:** analysis now merges sticky helper metadata only with the projected file context. A helper generation that was previously observed but removed by export invalidation is not reintroduced solely because it remains present in a later run record. Added a cross-cell regression covering sticky metadata after an export rebind.
- **P2 — ordinary calls to async helpers:** async helpers are now replayed only when their call is inside an `await` expression. An ordinary call conservatively keeps file and external-state analysis partial; awaited calls retain the helper's captured file reads. Added both regression cases and explicit `await` AST conversion coverage.

Validation: 5 focused notebook test files passed (702 passed, 2 skipped); Prettier and ESLint passed. Filtered TypeScript output contains no errors in the modified notebook files.

ewen-poch:
Addressed the latest AI review findings.

- **P1 — unresolved helper aliases:** source-file access analysis now clears a helper export from `unresolvedPriorNames` only when the helper body was actually replayed. Aliases and callback-style uses remain conservative instead of being certified from the export name alone. Added an alias regression.
- **P2 — decorated helper exports:** helper replay now excludes decorated top-level and nested functions because executing only the raw body cannot account for decorator or wrapper effects. Added a decorated-export regression.

Validation: 5 focused notebook test files passed (704 passed, 2 skipped); Prettier and ESLint passed. Filtered TypeScript output contains no errors in the modified notebook files.

ewen-poch:
Addressed the latest AI review findings.

- **P1 — cross-epoch helper evidence:** prior helper identities are now compared only against runs from the current `kernelEpochId`. A helper loaded after a kernel restart can therefore be admitted again and retain its captured file dependencies.
- **P2 — nested async calls:** helper replay now marks only the direct operand of an `Await` node as awaited. Async calls made inside an awaited helper body remain conservative unless their own call is explicitly awaited.

Validation: 5 focused notebook test files passed (706 passed, 2 skipped); Prettier and ESLint passed. Filtered TypeScript output contains no errors in the modified notebook files.

ewen-poch:
Addressed the latest AI review finding.

- **P1 — opaque helper delegation:** helper replay now marks unknown non-library, non-safe callables inside helper bodies as unresolved for reads, writes, and external state. This prevents a helper such as `return transform("input.dat")` from being certified complete when the delegated callable may perform I/O. Added a regression for opaque delegation while preserving whitelisted builtins and known helper calls.

Validation: 5 focused notebook test files passed (707 passed, 2 skipped); Prettier and ESLint passed. Filtered TypeScript output contains no errors in the modified notebook files.

ewen-poch:
Addressed both P1 findings in the current helper replay scope.

- **Import-time class/decorator/default/annotation effects:** helper replay now marks unmodeled import-time constructs as partial, while preserving modeled helper body paths. This prevents class bodies and function-definition expressions from being certified complete when their I/O cannot be reconstructed.
- **Callable parameter shadowing:** replay now applies Python parameter shadowing before resolving helper calls and treats same-named callable parameters as opaque when their effects cannot be modeled. This avoids dispatching callback arguments to private helpers and omitting callback I/O.

Added focused regressions for class effects, decorators, defaults/annotations, varargs, and same-named callback parameters.

Validation: five focused notebook test files passed (715 passed, 2 skipped); Prettier, ESLint, and diff checks passed. Filtered TypeScript output contains no errors in the modified notebook files.

ewen-poch:
Addressed the P2 alias-scope finding.

- Helper replay now snapshots and restores `possibleAliases`, so conditional aliases created while replaying a helper cannot affect subsequent caller file-access analysis.
- Added a regression that exercises a helper-local conditional alias followed by caller collection mutation and file writes.

Validation: five focused notebook test files passed (716 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed both P1 findings from the latest AI review.

- **Nested helper lexical bindings:** nested helper replay now inherits the enclosing helper scope without reloading module globals. Captured locals therefore take precedence over same-named module bindings. Added regressions for scalar and collection bindings.
- **Incomplete helper evidence:** runs with incomplete helper evidence no longer reuse projected or current helper modules. Their existing dependency facts are conservatively persisted with the existing `execution-incomplete` reason, so downstream file and external-state results remain partial until helper evidence is complete.

Validation: five focused notebook test files passed (719 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed the P1 failed-run evidence finding.

- Prior helper evidence is now eligible for replay filtering only when it comes from a completed run in the same kernel epoch.
- Failed, timed-out, or cancelled runs can no longer suppress a later successful run from recording and replaying the helper source.
- Added a regression covering helper metadata recorded by a failed run followed by a successful run.

Validation: five focused notebook test files passed (720 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed the P1 recursive-helper finding.

- Re-entrant helper calls are now treated as unresolved for reads, writes, and external state. This prevents recursive calls with different arguments from being skipped while the enclosing analysis is reported complete.
- Added a regression covering a helper that recursively invokes itself with a different argument.

Validation: five focused notebook test files passed (721 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed the latest AI review findings.

- **Stale helper replays after rebinding:** helper modules are now filtered against the current run’s defined, conditionally defined, and mutated names before file context persistence. Same-cell export rebinding can no longer restore the old helper body.
- **Incomplete helper evidence:** analysis contexts now remove prior helper modules when evidence is incomplete while preserving unrelated context fields.
- **Default helper parameters:** omitted literal string defaults are bound during replay; dynamic defaults remain unresolved rather than being evaluated against the caller scope.

Validation: five focused notebook test files passed (728 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed the final AI review findings for this review batch.

- **Unknown helper imports:** imports outside the known library allowlist now conservatively mark import-time reads, writes, and external state as unresolved during helper replay.
- **Helper argument validation:** replay now rejects invalid or ambiguous calls, including extra positional arguments, unknown or duplicate keywords, positional-only conflicts, missing required parameters, and unsupported unpacking. Valid varargs and kwargs remain supported.
- Added regressions for unknown imports, invalid arity/bindings, supported variadic calls, and existing helper replay paths.

Validation: five focused notebook test files passed (747 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed the P1 final-export-binding finding.

- Helper exports are now resolved against the final top-level binding. If an exported name is rebound to an assignment, import, loop target, or other non-function binding, replay is skipped and the analysis remains partial.
- Repeated function definitions still resolve to the last function definition.
- Added a regression test for an exported helper rebound to another callable.

Validation: the five focused notebook test files passed (748 passed, 2 skipped); Prettier, ESLint, diff checks, and the modified-file TypeScript check passed.

ewen-poch:
Addressed both P1 findings from this review.

- Top-level `ClassDef` bindings now invalidate an earlier exported helper function, so replay occurs only when the final export binding is a supported function definition.
- Nested helper definitions are replayed only when they are direct, unambiguous definitions in the enclosing helper body. Conditional or otherwise nested redefinitions are treated as unresolved, keeping file and external-state analysis partial instead of selecting a raw AST winner.
- Added regressions for class rebinding and conditional nested helper definitions.

Validation: the five focused notebook test files passed (750 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed the P1 conditional-rebinding finding.

- Export binding analysis now walks module control-flow statements and treats assignments, imports, deletes, loop targets, `with` bindings, exception names, class definitions, and nested function definitions as possible rebinding of exported names.
- An exported helper is replayed only when the final module binding is proven to be its supported top-level function. Conditional or ambiguous rebinding keeps the analysis partial.
- Added regressions covering `if`, `try`, `finally`, `while`, `with`, `match`, loops, imports, deletes, decorated definitions, and valid final-binding cases.

Validation: the five focused notebook test files passed (767 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed. The full repository TypeScript command still reports unrelated existing E2E and native-module errors outside the modified notebook files.

ewen-poch:
Addressed the independent worktree review findings.

- Exception aliases are now preserved in the converted AST and invalidate a helper export before later calls.
- Match capture bindings conservatively invalidate helper replay when their exact stores cannot be reconstructed.
- Helpers containing `global` or `nonlocal` rebinding are kept partial instead of replaying stale closure/module bindings.
- Added regressions for exception captures, match captures, module globals, closure rebinding, and ordinary exception handling.

Validation: five focused notebook test files passed (774 passed, 2 skipped); Prettier, ESLint, diff checks, and the filtered notebook TypeScript check passed.

ewen-poch:
Addressed the independent worktree review findings:

- Persisted helper evidence is now validated at the `run.json` boundary. Helper modules require bounded source, identity, exports, dependencies, and a matching SHA-256 digest; helper evidence status is restricted to the persisted union. Malformed records are rejected from session history instead of reaching dependency analysis and throwing.
- Helper replay now stays partial when a callable contains `yield`/`yield from` or unreachable statements after a terminating control-flow statement, so lazy or dead code cannot create concrete file edges.
- Top-level calls in a helper module are treated as import-time opaque effects and are no longer attributed to a later exported-call replay.

Validation: repository tests passed (38 passed); five focused Notebook Vitest files passed (777 passed, 2 skipped); Prettier, ESLint, `git diff --check`, and the filtered Notebook TypeScript check passed.

ewen-poch:
Addressed the four reported issues in `89f6eacff`:

- **Valid history with more than 32 helper modules:** removed the helper-count/size replay limits from durable `run.json` validation, so valid execution history remains readable after reopen. The limits now apply only when replaying helper evidence; over-budget evidence is excluded from analysis and produces partial/unknown facts.
- **Failed-run helper rebinding:** dispatched failed, timed-out, and cancelled runs no longer make sticky helper metadata authoritative by themselves. The next run only reuses evidence when the cached analysis retained the same helper binding and remains available; otherwise the helper is re-analyzed and stale paths cannot be restored.
- **Private and nested function rebinding:** helper binding collection now tracks final bindings inside module and callable scopes, including private and nested functions, assignments, imports, loop targets, and deletes. Rebound functions are invalidated instead of replaying their old bodies.
- **Closure precedence:** when replaying a helper call, the captured lexical scope is retained before module-level helper functions, matching Python name lookup and preventing a module function from overriding a same-named closure binding.

Added regression coverage for each case plus the 33-module persistence and replay-budget boundary. Focused Notebook tests pass (`871 passed, 2 skipped`); formatting, targeted ESLint, and diff checks pass. The full local type check still reports existing E2E/native dependency declaration errors, with no diagnostics in the modified Notebook files.

ewen-poch:
Aligned the analyzer cache revision with the requested value in `95a38a884`:

- `NOTEBOOK_ANALYZER_REVISION` is now `tree-sitter-in-process-107`.
- This reuses analysis entries carrying revision `107`; no persisted schema, state, or user-facing behavior changed.
- Prettier and `git diff --check` pass for the change.

PR #2841, opened 2026-09-20
perf(annotations): pause streaming layout observers

ewen-poch:
## Summary

Streaming transcript messages currently keep a `MutationObserver`, `ResizeObserver`, and window resize listener active for every `TextAnnotationSurface`, even when no draft selection can be retargeted. Markdown commits can therefore deliver redundant observer callbacks and marker geometry reads while the text tree is being replaced.

This change pauses those listeners during streaming when there is no live draft selection, and guards marker measurement against stale callbacks. The observers are restored when streaming ends; a live draft selection still retains the mutation observer so its range can follow streamed text.

## Validation

- Added regression coverage proving streaming surfaces create no mutation or resize observers, then restore both after streaming.
- Added coverage proving streaming resize notifications do not read marker geometry.
- `TextAnnotationSurface.test.tsx`: 51 passed.
- Targeted ESLint and Prettier checks: passed.

No data model, persistence, protocol, architecture, or user interaction changes.


ewen-poch:
Addressed. Streaming now filters annotation and bookmark marker controls against the current matching source IDs, so removing an annotation during a stream cannot leave its old edit marker or bookmark marker mounted. The geometry pass remains deferred while streaming, and the existing final measurement restores the marker positions when streaming ends.

Added a regression test that fails before this change and passes after it: render an annotated surface, remove the annotation while `isAnimating`, and assert that the marker is gone. The annotation test file now passes 52 tests.

ewen-poch:
Addressed. Existing annotation and bookmark controls are now hidden for the duration of streaming, so deferred geometry cannot leave a clickable control at a stale coordinate. The normal reconcile and measurement pass runs when streaming ends and restores the controls.

Added a regression test that fails before this change and passes after it: an annotated surface shows its marker, hides it while `isAnimating`, and restores it after streaming. The annotation test file now passes 53 tests.

PR #2843, opened 2026-09-20
fix(renderer): auto-hide idle scrollbars

ewen-poch:
## Problem

Persistent workspace scrollbars compete with conversation and preview content, especially in the main transcript.

## Proposed change

- Apply transparent idle native scrollbars to the session sidebar and tool-preview scroll container, with a thin, theme-aware thumb on hover or focus.
- Retain faint non-transcript overflow indicators on touch and hybrid devices (`hover: none` or `any-pointer: coarse`), including while a mouse is primary.
- Keep the transcript thumb transparent, including on no-hover devices, while retaining native scrolling and existing navigation buttons.
- Keep the shared Radix hover scrollbar mounted so keyboard focus can reveal it. Restore pointer events while focused and preserve overflow detection and explicit `type="always"` behavior.

## Scope and non-goals

Renderer styles, class wiring, and the ConversationPanel ResizeObserver test double. No new application state, enum values, persistence, historical-data compatibility, IPC, or backend changes. This does not restyle third-party iframe contents or every file-preview renderer.

After the latest test-only edit (3669514):

- ConversationPanel cleanup and workspace wiring -> focused Vitest run of `ConversationPanel.interaction.test.tsx` and `workspace-components.test.tsx` -> 196/196 passed, including all eight previously failing CI cases.
- Test-file Prettier, ESLint, and `git diff --check` -> passed.

After the final production-code edit (5f18f32):

- Workspace integration -> `vitest run src/renderer/src/pages/workspace/workspace-components.test.tsx` -> 30/30 passed.
- Changed CSS/TSX formatting and focused ESLint -> passed.
- Focus/blur visibility, pointer-event restoration, End-key scrolling, and transparent transcript in desktop and touch Chromium -> passed using a temporary Playwright script against the project modal-scrollbars fixture, with its `type="always"` replaced in memory to exercise default hover mode.

Earlier validation: `npm run typecheck:web` passed before the final CSS-only touch/hybrid adjustments. `npm run test:e2e:browser -- e2e/browser/modal-scrollbars.spec.ts` passed 13/13 before the review corrections. PreviewPanel and WorkspaceSidebar rendering suites passed. Two message-scroller unit failures also reproduced on the unchanged local main checkout; portable tests passed on an earlier PR head. A subsequent CI run exposed eight ConversationPanel cleanup failures caused by an incomplete ResizeObserver test double; 3669514 corrects that mock. On the current head 7f84ad0, portable CI shards 2, 3, and 4 passed; shard 1 and other checks were still running at handoff.

The final hybrid CSS branch also passed the browser checks with a test-only media-predicate override. Simultaneous fine/coarse pointer hardware was not emulated; physical hybrid-device validation remains uncovered.

No complete local test suite was run, as requested. Browser testing covers shared Radix consumers and native CSS in Chromium; final Windows/platform CI and visual inspection remain relevant for native scrollbar rendering. No main-process checks were selected because this patch changes no main-process behavior.

## Review focus

Verify the sidebar/tool-preview hover and focus treatment, persistent invisibility of the transcript thumb, and the shared Radix scrollbar's keyboard-only focus path. The source-string assertions guard wiring; the temporary browser check provides behavioral evidence but is not a committed regression test.

Latest AI review: **mergeable**, with no actionable findings on 7f84ad0 ([review](https://github.com/aipoch/open-science/pull/2843#issuecomment-5750961654)). The last commit only clarifies Radix overflow ownership in comments; it does not change runtime behavior.


ewen-poch:
Addressed the P2 finding about keyboard focus and the Radix scrollbar.

The custom scrollbar is a sibling of the viewport, so `:focus-within` on the root did not override its hidden state. The shared scrollbar CSS now explicitly restores opacity and pointer events for a hidden Radix scrollbar while its scroll area contains focus. Added a workspace regression assertion and reran the focused workspace test and Web typecheck; both pass.

ewen-poch:
Fixed the no-hover cascade issue in 7758f4e: the touch fallback now excludes `.scrollbar-message`, preserving the requested invisible transcript scrollbar.

Also completed the keyboard-focus fix: Radix unmounts its hover scrollbar while idle, so the CSS override alone was insufficient. The existing `ScrollBar` now uses `forceMount`, allowing the focus rule to act on a mounted node while Radix retains its overflow detection and explicit `type="always"` behavior. No new state or persistence was introduced.

Verified in Chromium using the existing modal-scrollbars fixture with its default hover mode (test-only in-memory override): desktop and touch contexts both pass focus reveal, pointer-event restoration, End-key scrolling, blur hiding, and transparent transcript checks. The workspace suite passes 30/30; formatting and focused ESLint pass. Web typecheck also passed for the touch CSS correction.

ewen-poch:
Addressed touch visibility for Radix scrollbars in d14f73d. The no-hover rule now targets the custom hidden scrollbar as well as native scrollbars. It only reveals a bar containing Radix's thumb, so overflow detection remains with Radix; the transcript exception is preserved.

Chromium checks passed in desktop and touch contexts: focus visibility, pointer events, End-key scrolling, the intended idle appearance, disappearance after content shrinks below the viewport, and transparent transcript scrollbars. The workspace suite passes 30/30; formatting and diff checks pass. This correction changes only the existing CSS file and introduces no new state or persistence.

ewen-poch:
The hybrid-device observation is valid: `hover` describes the primary input, while `any-pointer: coarse` detects an available coarse pointer, not the currently used input (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/any-pointer).

Applying that fallback to every touch-capable laptop would keep non-transcript scrollbar indicators visible even during mouse use. That changes the agreed hover/focus-only desktop presentation. Tracking the active input would add state and interaction policy beyond this CSS correction.

Leaving this point pending the maintainer's design decision rather than changing that behavior implicitly. The transcript remains hidden in all cases; the existing pure-touch fallback and keyboard-focus corrections are implemented and browser-checked.

ewen-poch:
Implemented the hybrid touch fallback in 5f18f32 following the maintainer's request to continue and complete the improvement. The media condition now matches either `hover: none` or `any-pointer: coarse`. Touch-capable laptops therefore retain faint overflow indicators even when a mouse is primary. The transcript exclusion remains unchanged; no active-input tracking, new application state, or persistence was added.

Validation after the change: workspace tests 30/30, Prettier, focused ESLint, and diff checks passed. Chromium checks passed for desktop and pure-touch contexts, covering focus/blur visibility, pointer events, End-key scrolling, no-overflow removal, and transparent transcript scrollbars. The compiled hybrid CSS branch passed the same checks using a test-only media-predicate override. Chromium did not support emulating simultaneous fine/coarse pointers, so this last check is controlled CSS-branch coverage, not physical hybrid-device validation.

ewen-poch:
I could not reproduce the phantom-thumb finding and am retaining the current implementation rather than introducing a second overflow state owner.

In the installed `@radix-ui/react-scroll-area@1.2.13` implementation:

- `ScrollAreaScrollbarHover` destructures `forceMount` and uses it only for its outer Presence; it does not forward it to the nested `ScrollAreaScrollbarAuto`.
- That nested Auto component still measures viewport overflow and gates the rendered scrollbar on that result.
- `ScrollAreaThumb` has its own separate `forceMount` prop. Our wrapper does not set it, so thumb presence still depends on Radix's `hasThumb`.

Chromium checks pass for both initially non-overflowing content (including focus and hover) and overflowing content that subsequently shrinks: no scrollbar remains in the default hover ScrollArea. These checks passed in desktop and pure-touch contexts and with the controlled hybrid CSS branch. The hybrid check remains a media-predicate override, not physical hybrid hardware emulation.

No code changes for this finding. The configured five-review limit has been reached, so no further revision/review cycle is being started. The automated verdict is still `needs changes`; this response records the source and behavioral evidence for maintainer review.

ewen-poch:
The new overflow finding repeats the same premise as the previous review. `forceMount` on `ScrollAreaScrollbar` does not force-mount `ScrollAreaThumb`: in the installed Radix 1.2.13 hover implementation, the outer Presence is forced, the nested Auto scrollbar still gates itself on measured overflow, and the thumb has its own separate Presence controlled by `hasThumb`. Our wrapper does not pass `forceMount` to the thumb. Existing Chromium checks cover initially short content and content shrinking below the viewport, including focus and hover, with no phantom scrollbar. I am not adding duplicate application overflow state for this finding.

Separately, the current portable-test failure is actionable. CI shard 2 reports `TypeError: resizeObserver.unobserve is not a function` in `ConversationPanel.interaction.test.tsx`. A focused local run reproduces the same eight failures (158 other tests pass): that file's ResizeObserver double defines `observe` and `disconnect`, but omits `unobserve`, which Radix now calls during cleanup. The proposed fix is to complete that test double, with no production behavior changes. This file is outside the existing PR diff, so implementation is paused pending the maintainer's explicit approval to extend the file scope.

ewen-poch:
Completed the ResizeObserver test double in `ConversationPanel.interaction.test.tsx` by adding the missing `unobserve()` cleanup method in 3669514, following maintainer approval to include this test file. Production code and the agreed scrollbar behavior are unchanged; no state, persistence, or compatibility changes are involved.

Validation after this edit: the ConversationPanel interaction suite and workspace component suite pass together (196/196 tests, including the eight previously failing cases). Focused ESLint, Prettier, and `git diff --check` also pass. The full portable and platform suites remain delegated to PR CI.

ewen-poch:
The overflow finding relies on conflating the scrollbar's `forceMount` prop with the thumb's independent `forceMount` prop. The latter is not set by our wrapper. In Radix 1.2.13, `ScrollAreaScrollbarHover` consumes its own `forceMount` for the outer Presence without forwarding it to the nested Auto component; `ScrollAreaThumb` separately uses `forceMount || scrollbarContext.hasThumb`.

Added comments in the existing ScrollArea and CSS files in 7f84ad0 to make that dependency explicit for future maintenance. This is documentation only; the verified no-overflow behavior is unchanged and no second overflow state owner was added. Prettier, focused ESLint, and diff checks pass. The existing Chromium checks cover both initially fitting content and overflow removal after content shrinks, with no phantom thumb on desktop, touch, or the controlled hybrid CSS branch.

The previously failing portable-test shard 2 now passes on 3669514 after the approved test-double fix; the local related suites pass 196/196.

PR #2853, opened 2026-09-21
feat(pdf): add persistent annotations and document notebooks

ewen-poch:
## Summary

Add persistent PDF notes and annotations so users can read, mark, organize and find evidence without losing it when switching views or conversations. Notes belong to a managed file version: Library attachments share their notebook across references, projects and sessions, while project uploads/artifacts share it across sessions in the owning project.

- Support text styles, area marks, page/document notes, comments, colors, global Tags, undo/redo and native PDF annotation import. Export a separate annotated PDF or Markdown/CSV notes without changing source bytes.
- Reuse one document provider across the PDF overlay, notebook and responsive notes sidebar. Keep narrow toolbars/tabs compact, filters collapsed by default, Tag controls stable, and annotation editing aligned with the selection toolbar.
- Include notes and quotes under Library in global search. Selecting a result shows a detail pane with comments and quoted text separated; its source action opens the PDF directly. Settings Tags also open the PDF modal without navigating away or opening bibliographic details first.

## Persistence and performance

- One migration, `0043_pdf_annotations`, creates annotations and native-import receipts; Tags reuse the existing global tables. Receipts prevent deleted imported marks from returning. Released main databases follow the normal migration path; obsolete branch-only development schemas are intentionally unsupported.
- Validate managed source/version ownership in main, save annotation/Tag changes transactionally, reject stale writes with compare-and-swap tokens, and synchronize open views through document events. Session IDs record provenance rather than partitioning the notebook.
- Parse native annotations and export PDFs in workers, page list IPC in batches of 100, progressively mount notebook cards, index marks by page, and debounce notebook search by 250 ms with IME support.
- Global note search remains SQLite substring search; the provider still loads the full document in bounded pages and progressive rendering is not virtualization.

## Validation

- Focused database, migration, PDF service/repository, renderer annotation, notebook, global-search and i18n suites: **1,483 passed, 3 skipped**.
- `npm run typecheck:node`, `npm run typecheck:web`, `npm run db:schema:check`, ESLint on the final follow-up changes, formatting and `git diff --check`: passed.
- Production Electron build and PDF annotation end-to-end journey: validated native import, persisted edits, undo provenance, export, selection/editor positioning, search detail preview and direct source navigation.
- Earlier notebook-capacity Electron coverage exercised progressive rendering and searching beyond the first rendered batch.

The broader PDF module run previously exposed five unrelated Markdown-rendering test failures, reproduced on clean main. The relevant focused suites pass; this PR does not claim a clean full-repository test run. Performance observations are local measurements, not cross-platform frame-rate guarantees.


ewen-poch:
- **Nullable scope CHECK (P1):** No schema change is needed. SQLite rejects a CHECK only when its expression evaluates to zero; NULL is accepted. The existing migration regression test inserts both Library annotations with NULL project/session fields and project annotations with a NULL session successfully. Invalid scope combinations are still rejected by the separate scope constraint. Re-ran the actual migration test successfully.
- **Import cancellation (P2):** Passed the combined abort signal into native annotation persistence, checked it before and after the write, and rethrew cancellation instead of swallowing it as enrichment failure. Also check cancellation after the final item read. Added regressions for cancellation during attachment, annotation persistence, and a failing annotation write. These verify cancellation is reported, completion is not reported, staging is released, and an already-attached immutable PDF is retained. A completed database transaction is not rolled back by a later cancellation.

Validation: 12 importer/migration tests passed. No new state, persistence fields, or schema changes.

ewen-poch:
- **Project annotation preview (P1):** The selected-session lookup is not the only source path. `PdfPreviewRendererContent` falls back to the document-scoped provider's source, populated from the saved annotations for the exact file/version. Added regression coverage that mounts this provider and reveals a saved project annotation with no selected session and with another project's session selected. Both scenarios reveal the annotation successfully without calling the session-bound resolver or changing session selection, so no production navigation change is needed.
- **Region text search (P2):** Updated the shared SQL quote expression to use `selector.text` for region annotations alongside `selector.exact` for text annotations. The existing nonempty filter, match predicate, count and relevance scoring all reuse that expression. Added coverage for note-less region results, case-insensitive matching, count-only search and exact-match ranking ahead of prefix matches. No new fields, indexes, state or persistence are introduced.

Validation: 127 catalog tests and 93 PDF preview tests passed. The existing area-selection test now waits for the closing menu to restore focus before simulating the next pointer gesture, avoiding a test-only timing race.

ewen-poch:
- **Literature annotation ownership (P1):** The approved model shares a Library attachment-version notebook across every project that references that version. Project-filtered search therefore uses the paper's project membership for these shared notes; imposing project ownership on canonical Library notes would contradict that design. Inspection did uncover a legacy mixed scope that still permits a literature source with a project-owned annotation.
- **Project-only literature creation (P2):** This concerns the same mixed scope. Extending its sessionless write path would preserve a second ownership model that the current Library UI does not use. The proposed direction is to retain file-version sharing and close the legacy mixed-scope contract, rather than introduce project-isolated Library notebooks.

Implementation is paused pending the author's confirmation because resolving this ambiguity changes the accepted scope contract. No schema, ownership or navigation changes have been made for these findings.

ewen-poch:
- Library annotation ownership: Library notebooks remain shared by immutable attachment version across projects, as confirmed by the author. The request contract, service boundary and existing unreleased migration now reject project-owned Library annotations. Removed the legacy session-dependent resolver instead of introducing a second ownership model. Added regression coverage for mixed-scope rejection (with and without a session), database enforcement, and shared search results in two projects linked to the same paper.
- Project-only Library creation: Such requests are now invalid at the boundary. Canonical Library writes continue to use the attachment authority without a Project or Session; project upload/artifact notes continue to share by managed file version within their owning project.
- Rebased onto `f586c070` from `main`. Resolved additive CI/test-list conflicts by retaining both the background-delivery/cancellation coverage and the PDF coverage. The annotation schema still uses one migration; no extra table, column or UI behavior was introduced.

ewen-poch:
The document-note reveal finding is not reproducible. `PdfPreviewContent` already subscribes to reveal preparation, matches the immutable source and switches to the Notes tab for `document-note`. Once active, `PdfNotebookView` registers its reveal listener, scrolls/focuses the card and acknowledges the pending request.

Added a real Electron regression to the existing PDF journey: create a document note, find it through global search, open its source modal, assert the Notes tab is selected and the exact card receives focus, and assert no location error is shown. The test passes with the current production navigation code. No alternate navigation state or special-case shortcut was added.

The user-requested compact-header alignment and persistent sidebar toggle changes are also included. Renderer/i18n tests (916), web typecheck, lint, build and the full isolated Electron PDF journey pass.

ewen-poch:
- Aligned the annotation color indicator, page/kind labels, tag trigger and source/edit/delete actions on the same vertical center in compact note headers.
- Kept the notes-sidebar toggle visible in the Original PDF toolbar even below the side-by-side width budget. In narrow layouts it remains keyboard discoverable, exposes its disabled state and explains that the window must be wider; the existing full Notes tab remains available. Restoring sufficient width preserves the notebook and draft.
- Validation: 916 renderer/i18n tests, web typecheck, lint and production build passed. The isolated Electron PDF journey passed, including a real geometry assertion (action-center spread no greater than 1 CSS pixel), persistent narrow-width toggle and draft preservation.

ewen-poch:
Sidebar availability now uses the full preview/modal container width. Expanding the left navigation no longer reduces that eligibility budget, hides the open notes sidebar or disables its toggle. Both sidebars can coexist, and notes can be closed and reopened while navigation stays expanded.

Validation: 93 PDF renderer tests, web typecheck, lint and build pass. The real Electron journey also verifies simultaneous sidebars, closing/reopening notes with navigation open, compact-header alignment and global-search document-note reveal.

PR #2878, opened 2026-09-21
fix(regressions): stabilize source regression flows

ewen-poch:
## Problem

The scheduled Source Regression run in https://github.com/aipoch/open-science/actions/runs/35592133585 surfaced three independent root causes:

1. **Settings keyboard focus restoration:** opening Settings makes the workspace inert before `SettingsPage` can reliably identify the opener. A later or repeated autofocus callback can then capture a Settings control or a detached replacement instead of the original trigger. This occurs when a completed message run has rebuilt workspace controls around the Settings transition. If left unfixed, Escape can close Settings without restoring focus to the invoking control, and the accessibility reporter fails the run.
2. **PDF annotation tag layout:** at the 660×900 fixture viewport, 2× zoom, and seven selected tags, the selected-tag strip gains vertical height when horizontal overflow appears. The Save button moves by six pixels. If left unfixed, the dialog visibly jumps and the layout regression assertion remains unstable.
3. **Side-chat readiness:** reply text can render before the session has reached its terminal idle state. The test could click the branch-send menu while the run was still `Thinking`, producing a retry-pass that `--fail-on-flaky-tests` correctly treats as a failure. If left unfixed, scheduled regression runs can fail despite the eventual UI being correct.

The aggregate enforcement failures, skipped tests, and workflow timeout were downstream consequences, so this change does not raise timeouts or weaken the reporters.

Refs #2875.

## Proposed change

- Capture the focused Settings opener synchronously in the navigation action, before `isSettingsOpen` changes. Store it in a module-local, one-shot DOM reference and consume it from `SettingsPage`; preserve the first connected target across repeated autofocus callbacks.
- Keep the selected-tag strip at the existing control height while retaining horizontal scrolling.
- Make the accessibility and side-chat journeys wait for the real idle control before navigating or opening the send menu.
- Remove the artificial delayed-focus mutation from the accessibility test so the test exercises the production interaction directly.

These fixes are deliberately narrow. Capturing the opener before hiding/inerting its surface is the standard dialog-focus pattern, and waiting for the idle affordance uses the UI's existing readiness contract rather than adding test-only state.

## Scope and non-goals

- No database or persistent-storage changes.
- No new data fields, schema changes, migrations, or data-model/relationship changes.
- No new enum or persisted status values.
- No historical-data compatibility path is required.
- No copy or localization changes.
- No increase to global or per-workflow timeouts.
- No attempt to redesign Settings navigation, the tag model, or side-chat state management.

All checks below ran after the last material edit to the area they cover.

- Settings focus capture and architecture boundary:
  - `renderer_state` module: 376 files / 8,224 tests passed.
  - `settings_application_presentation` module: 244 files / 5,628 tests passed.
  - The first presentation-module run hit one unrelated `SubagentReleaseSurfaces` timing failure; the isolated case passed three consecutive reruns, and the complete module rerun passed.
- Focus interaction and accessibility reporting:
  - Accessibility Electron suite: 11/11 passed with `--fail-on-flaky-tests`.
  - Reporter result: passed, 21 axe scans, zero UI findings.
- Tag layout:
  - `ResourceTagControls.render.test.tsx`: 24/24 passed.
  - `pdf-annotations.spec.ts`: 1/1 passed.
- Side-chat readiness:
  - `side-chat-admission.spec.ts --repeat-each=3`: 6/6 passed.
- Static/build checks:
  - ESLint on all changed files passed.
  - Node and Web TypeScript checks passed.
  - Electron production build passed (existing Vite/CSS warnings only).
  - `git diff --check` passed.

Per request, the complete local `npm test` suite was not run; PR CI remains the authoritative full validation.

## Review focus

- Confirm the opener is captured only on the closed-to-open transition and consumed once.
- Confirm repeated Settings autofocus events cannot overwrite a still-connected return target.
- Confirm the fixed tag-strip height preserves horizontal scrolling without clipping chips.
- Confirm the E2E waits represent user-visible idle readiness rather than masking product failures.


ewen-poch:
Addressed the tag overflow clipping concern. The selected-tag strip still scrolls horizontally, but now hides the native scrollbar with cross-platform rules so Windows/Linux scrollbar gutters cannot reduce the 32px badge viewport. Added a render regression test with seven selected tags covering the overflow case and verifying the scrollbar rules.

ewen-poch:
Addressed the focus-restoration findings. Settings navigation now resolves a focused Radix menu item to its connected menu trigger before the menu unmounts, so closing Settings returns focus to the stable opener for menu-based entry points. The accessibility journey also retains the delayed-focus pass, and the settings navigation tests cover one-shot opener capture plus transient menu-item resolution. No data, persistence, schema, or interaction model changes are involved.

